PR #16, opened 2025-12-11
add simple LLVM semantics

tobiasgrosser:
Add key definitions for the LLVM Int and Byte types. We only add a few operations to keep the PR small and ensure that we are aligned on the data-structure definitions.

### DIFF
--- a/Veir/Dialects/LLVM/Byte/Basic.lean
+++ b/Veir/Dialects/LLVM/Byte/Basic.lean
@@ -8,8 +8,15 @@ namespace Byte
 
 public section
 
+/--
+The `Byte` type can have any bitwidth `w`. It carries a two's complement integer
+value of width `w` and a per-bit poison mask modeling bitwise delayed undefined
+behavior.
+-/
 structure Byte (w : Nat) where
+  /-- A two's complement integer value of width `w`. -/
   val : BitVec w
+  /-- A per-bit poison mask of width `w`. -/
   poison : BitVec w
 
 def xor {w : Nat} (x y : Byte w) : Byte w :=
@@ -23,6 +30,10 @@ def and {w : Nat} (x y : Byte w) : Byte w :=
 
 open Int
 
+/-- Convert from `Byte` and `Int`.
+  A byte where no bit is poison is equal to an integer value.
+  If any bit is poison, the integer type is also poison.
+-/
 def toInt {w : Nat} (x : Byte w) : Int w :=
   if x.poison = 0 then
     .val x.val
--- a/Veir/Dialects/LLVM/Int/Basic.lean
+++ b/Veir/Dialects/LLVM/Int/Basic.lean
@@ -4,8 +4,15 @@ namespace Veir.Dialects.LLVM.Int
 
 public section
 
+/--
+The `Int` type can have any bitwidth `w`. It is either a two's complement
+integer value of width `w` or a poison value indicating delayed undefined
+bahavior.
+-/
 inductive Int (w : Nat) where
+/-- A two's complement integer value of width `w`. -/
 | val : BitVec w → Int w
+/-- A poison value indicating delayed undefined behavior. -/
 | poison : Int w
 deriving DecidableEq
 


Veir/Dialects/LLVM/Byte/Basic.lean
```diff
@@ -0,0 +1,41 @@
...
+
+public import Veir.Dialects.LLVM.Int.Basic
+
+namespace Veir.Dialects.LLVM
+
+namespace Byte
+
+public section
+
+structure Byte (w : Nat) where
```
math-fehr, line 16: Is that the new type that will be added to LLVM?
Shouldn't it be only 8 bits and not be parametric?

Veir/Dialects/LLVM/Byte/Basic.lean
```diff
@@ -0,0 +1,41 @@
...
+  val : BitVec w
+  poison : BitVec w
+deriving DecidableEq
```
math-fehr: This should be changed to a custom implementation where bits that are poison are considered equals.

Veir/Dialects/LLVM/Byte/Basic.lean
```diff
@@ -0,0 +1,41 @@
...
+def and {w : Nat} (x y : Byte w) : Byte w :=
+  ⟨x.val &&& y.val, x.poison ||| y.poison⟩
+
+open Int
+
+def toInt {w : Nat} (x : Byte w) : Int w :=
+  if x.poison = 0 then
+    .val x.val
+  else
+    .poison
```
math-fehr, line 41: This looks weird to me. I would pass the value the integer should be in case the byte is poison. Here you are relying on the internals of the byte implementation which would prevent some reasoning.

Veir/Dialects/LLVM/Int/Basic.lean
```diff
@@ -0,0 +1,26 @@
+module
+
```
math-fehr, line 2: Why do we have both Int and Byte that looks very similar? I don't understand what Byte is supposed to represent in LLVM compared to Int here.

math-fehr (approved):
I left 2-3 comments on the code, but otherwise I think it's fine.

I'm just very not sure if we should put this code in these folders, I think these folders should be more for the dialect definitions, and then having separate folders for the semantics. We can always change that later though.

Veir/Dialects/LLVM/Byte/Basic.lean
tobiasgrosser: Yes, this is an upcoming LLVM type needed to replace `undef` on the memory side of things.

https://blog.llvm.org/posts/2025-08-29-gsoc-byte-type/

Here is a potential comment.

> A byte type of any bitwidth `w` carries a two's complement integer value and a per-bit poison mask modeling bitwise delayed undefined behavior.

Veir/Dialects/LLVM/Byte/Basic.lean
tobiasgrosser: Good catch. I dropped this for now.

Veir/Dialects/LLVM/Int/Basic.lean
tobiasgrosser: Byte carries bitwise poison, vs Int carries one poison for everything. Byte is needed to load memory into registers at arbitrary offsets without collapsing poison. I am adding it now as I want to explore interactions across these two types.

I hope the additional documentation helps.

Veir/Dialects/LLVM/Byte/Basic.lean
tobiasgrosser: The integer is an inductive type. This function says:

> A byte where no bit is poison is equal to an integer value. If any bit is poison, the integer type is also poison.

But maybe I don't get what you mean. Can you clarify this or even suggest code?

Veir/Dialects/LLVM/Byte/Basic.lean
tobiasgrosser: I added relevant comments.

Veir/Dialects/LLVM/Byte/Basic.lean
math-fehr: Nevermind I just misunderstood the code!

tobiasgrosser:
I am happy to move things around later.

PR #20, opened 2025-12-14
Add GetSet lemmas for linkBetweenWithParent

math-fehr:
Also add the GetSet lemmas for `linkBetween` and `setParentWithCheck`.

These are the last GetSet lemmas missing for Operation in ˋLinkedList`, the other functions where unused and removed.

### DIFF
--- a/Mlir/Rewriter/LinkedList/GetSet.lean
+++ b/Mlir/Rewriter/LinkedList/GetSet.lean
@@ -396,8 +396,8 @@ theorem OpOperandPtr.get!_OpOperandPtr_insertIntoCurrent {opOperand : OpOperandP
         simp only [get!_OpOperandPtr_setBack, Ne.symm heq, ↓reduceIte, get!_ValuePtr_setFirstUse]
         simp only [get!_OpOperandPtr_setNextUse, h, ↓reduceIte]
         simp only [get!_OpOperandPtr_setBack]
-        simp only [ite_eq_right_iff] -- Why does grind needs these here to work?
-        grind
+        --grind    -- Why does 'grind' not work here?
+        simp only [h, ↓reduceIte]
 
 @[simp, grind =]
 theorem OperationPtr.getNumSuccessors!_OpOperandPtr_insertIntoCurrent {operation : OperationPtr} :


Mlir/Rewriter/WellFormed/Rewriter/Operation.lean
```diff
@@ -34,8 +34,7 @@ theorem OperationPtr.get!_linkBetweenWithParent (op : OperationPtr)
       { op.get! ctx with prev := self }
     else
       op.get! ctx := by
-  unfold linkBetweenWithParent at heq
-  grind (splits := 20)
+  sorry
```
tobiasgrosser, line 37: Why does this proof do not go through anymore?

Mlir/Rewriter/LinkedList/GetSet.lean
```diff
@@ -476,234 +476,683 @@ theorem OpOperandPtrPtr.get!_OpOperandPtr_insertIntoCurrent {opOperandPtr : OpOp
...
+/- OperationPtr.linkBetween -/
+section linkBetween
+attribute [local grind] OperationPtr.linkBetween
 
 @[simp, grind =]
-theorem OperationPtr.getNumResults_OpOperandPtr_setBack :
-    OperationPtr.getNumResults op (OpOperandPtr.setBack operandPtr ctx h₁ v) hop =
-    OperationPtr.getNumResults op ctx (by grind) := by
+theorem IRContext.topLevelOp_OperationPtr_linkBetween :
+    (op'.linkBetween ctx prev next selfIn prevIn nextIn).topLevelOp =
```
tobiasgrosser, line 485: why o you use `op'` even tough no `op` is used?

Mlir/Rewriter/LinkedList/GetSet.lean
```diff
@@ -476,234 +476,683 @@ theorem OpOperandPtrPtr.get!_OpOperandPtr_insertIntoCurrent {opOperandPtr : OpOp
...
+    newCtx.topLevelOp = ctx.topLevelOp := by
   grind
 
-@[simp, grind =>]
-theorem BlockArgumentPtr.get!_OpOperandPtr_linkBetweenWithParent (res : BlockArgumentPtr)  (op : OperationPtr) (h₁ : op.InBounds ctx)
-    (heq : op.linkBetweenWithParent ctx prevOp nextOp parent hp h₁ h₂ h₃ = some newCtx) :
-    res.get! newCtx = res.get! ctx := by
-  unfold OperationPtr.linkBetweenWithParent at heq
+grind_pattern IRContext.topLevelOp_OperationPtr_setParentWithCheck =>
+  op'.setParentWithCheck ctx newParent selfIn, some newCtx, newCtx.topLevelOp
```
tobiasgrosser, line 667: Why does this PR start to use `grind_pattern`? What do they do?

Mlir/Rewriter/LinkedList/GetSet.lean
math-fehr: In all the getset lemmas we use `'` on the pointers that get modified. It just makes this easier to copy paste so we are sure there is no name clashing.

Mlir/Rewriter/LinkedList/GetSet.lean
math-fehr: Here `grind .` is an annotation that requires to have both the getter of `ctx` and `newCtx` in the context to trigger the rewrite. What we want here is that the rewrite gets triggered only when we only see the patterns specified by `grind_pattern`, which is only `newCtx` and the given hypothesis about `newCtx`.

Mlir/Rewriter/WellFormed/Rewriter/Operation.lean
math-fehr: This is a getter that only talk about `Operation.get`. We are removing those in favor of getters for each field, which seems to make this faster for grind. So this doesn't go through because grind doesn't `case` the `Operation` struct. But this proof will likely be removed when I do my pass on this file, so I sorried it instead of fixing it to not waste time

ineol: I don't think this is correct? It looks to me that it only talks about `get!`, but maybe it's replaced by a series of lemmas about the components of `Operation` in the `GetSet` file?

Mlir/Rewriter/LinkedList/GetSet.lean
ineol: To give a bit more context, this is a bit more tricky that most get-set lemmas because the function returns an option so it's not an equality lemma. We should investigate if we could give more precise patterns to the lemmas about the `allocEmpty` functions

Mlir/Rewriter/WellFormed/Rewriter/Operation.lean
math-fehr: Yes sorry about `get!`

math-fehr: We did replace it with all the different components, I actually haven't tried to remove the lemma to see what happens because I haven't touched this file yet. I'll try this after lunch

math-fehr: Yeah we need to add the lemmas about `insertOp?` first before doing that, so it's better to sorry this now because this is going to be removed soon!

Mlir/Rewriter/LinkedList/GetSet.lean
tobiasgrosser: I see. Cool.

PR #21, opened 2025-12-15
Small cleanups

ineol:
- add module files which import the submodules
- remove `satisfies`
- remove some unsused lemmas about `USize`
- make some definitions irreducible in `LinkedList.Basic`


Mlir/PatternRewriter/Basic.lean
```diff
@@ -76,6 +76,12 @@ def Worklist.remove (worklist: Worklist) (op: OperationPtr) : Worklist :=
   | none =>
     worklist
 
+-- TODO: remove this lemma and/or move it somewhere reasonable
+@[grind →]
+theorem OperationPtr.inBounds_of_mem_operations_keys (ctx : IRContext) :
+    (op ∈ ctx.operations.keys) → op.InBounds ctx := by
+  grind [OperationPtr.InBounds]
```
math-fehr: Can we make this a local grind or is it needed elsewhere?

math-fehr: Also, why is it needed here?

Mlir/Rewriter/LinkedList/GetSet.lean
```diff
@@ -676,7 +676,7 @@ theorem OpOperandPtr.get_OpOperandPtr_linkBetween (res : OpOperandPtr) (h : res.
 theorem OperationPtr.getParent_OperationPtr_linkBetween :
     (OperationPtr.get op (OperationPtr.linkBetween op' ctx prevOp nextOp h₁ h₂ h₃) hop).parent =
       (OperationPtr.get op ctx (by grind)).parent := by
-  unfold OperationPtr.linkBetween
+  unfold OperationPtr.linkBetween at *
```
math-fehr: I'm removing this lemma in my PR, can you remove the change so we don't have a conflict here?

Mlir/Rewriter/LinkedList/Basic.lean
```diff
@@ -56,6 +57,7 @@ theorem OpOperandPtr.insertIntoCurrent_inBounds (ptr : GenericPtr) :
   Add previous and next links to `newOp`, linking it after `self`.
   In particular, this does not update any parent pointers.
 -/
+@[irreducible]
```
math-fehr: Can you remove the changes in `linkNextOp` and `linkPrevOp` since I'm removing them anyway because we don't use them at all?

Mlir/Rewriter/LinkedList/Basic.lean
```diff
@@ -171,6 +176,7 @@ theorem OperationPtr.setParentWithCheck_inBounds (ptr : GenericPtr)
   Add previous and next links to `newBlock`, linking it after `self`.
   In particular, this does not update any parent pointers.
 -/
+@[irreducible]
```
math-fehr, line 119: Same here

Mlir/Rewriter/LinkedList/Basic.lean
```diff
@@ -239,6 +245,7 @@ theorem OperationPtr.linkBetweenWithParent_fieldsInBounds (hx : ctx.FieldsInBoun
   Add previous and next links to `newOp`, linking it before `self`.
   In particular, this does not update any parent pointers.
 -/
+@[irreducible]
```
math-fehr, line 189: Same here

Mlir/PatternRewriter/Basic.lean
ineol: In that case it's because some function iterates over all the operations to populate the work list.

ineol: I put the two lemmas as local grind

Mlir/Rewriter/LinkedList/Basic.lean
ineol: We use them in `Rewriter.Rewriter` no?

Mlir/Rewriter/LinkedList/Basic.lean
math-fehr: Oh I haven't removed the block one my bad

PR #25, opened 2025-12-16
Merge Builder and Rewriter files

math-fehr:
This does not yet merge the files in `Rewriter.WellFormed` folder to not make this PR too big.


math-fehr:
This currently has a bug in `Mlir/Rewriter/RewriterGetSetInBounds.lean`, where `grind` does not work anymore for reasons I really don't understand (the proof was just moved, it does not lose any context). 
It seems the bug comes from a match not being able to be split.
I'm looking into it now.

tobiasgrosser:
LGTM.

ineol:
Thank you Mathieu!

PR #48, opened 2025-12-21
Add well-formedness lemma for Rewriter.insertOp?

math-fehr:
This finishes the proof using the lemmas we defined for `insertIntoCurrent`

### DIFF
--- a/Veir/Rewriter/WellFormed/Rewriter/Operation.lean
+++ b/Veir/Rewriter/WellFormed/Rewriter/Operation.lean
@@ -8,7 +8,6 @@ namespace Veir
 
 section InsertOp
 
-unseal Rewriter.insertOp? in
 theorem Rewriter.insertOp?_WellFormed (ctx : IRContext) (hctx : ctx.WellFormed)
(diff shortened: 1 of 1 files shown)


Veir/Rewriter/WellFormed/Rewriter/Operation.lean
```diff
@@ -6,260 +6,14 @@ import Veir.Rewriter.InsertPoint
...
 namespace Veir
 
-@[grind]
-def InsertPoint.Wf (insertionPoint : InsertPoint) (ctx : IRContext) (newOp : OperationPtr) :=
-  match insertionPoint with
-  | before op => newOp ≠ op ∧ newOp ≠ (op.get! ctx).prev ∧ op ≠ (op.get! ctx).prev
-  | atEnd bl => newOp ≠ (bl.get! ctx).lastOp
+section InsertOp
 
 unseal Rewriter.insertOp? in
```
tobiasgrosser: Is this still needed?

math-fehr: Good catch, I removed it

PR #53, opened 2025-12-23
Add `OperationPtr.dealloc`

math-fehr:
This abstraction was missing and is useful to talk about InBounds and Get/Set lemmas involving deallocation (for instance in eraseOp).

### DIFF
--- a/Veir/IR/Basic.lean
+++ b/Veir/IR/Basic.lean
@@ -473,6 +473,8 @@ def OperationPtr.allocEmpty (ctx : IRContext) (opType : Nat) : Option (IRContext
   let ctx := newOpPtr.set ctx operation
   (ctx, newOpPtr)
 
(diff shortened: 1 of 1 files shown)


Veir/IR/Basic.lean
```diff
@@ -473,6 +473,10 @@ def OperationPtr.allocEmpty (ctx : IRContext) (opType : Nat) : Option (IRContext
   let ctx := newOpPtr.set ctx operation
   (ctx, newOpPtr)
 
+set_option linter.unusedVariables false in
+def OperationPtr.dealloc (op : OperationPtr) (ctx : IRContext)
+    (inBounds: op.InBounds ctx := by grind) : IRContext :=
```
tobiasgrosser, line 480: Maybe add a comment to the PR message on why we need an unused variable here. This is not clear to me.

tobiasgrosser (approved):

PR #54, opened 2025-12-23
Add hasUses definitions and theorems

math-fehr:
This is going to be useful to reason about eraseOp, as an operation needs to have no uses to be erased while preserving well-formedness.


tobiasgrosser (approved):
Nice.

PR #60, opened 2025-12-28
Make OperationPtr, BlockPtr, and RegionPtr opaque types

math-fehr:
This improves by 20% our proof times.
Note that this breaks further the proofs in `createOp`, but this definition should be improved anyway.

### DIFF
--- a/Veir/Benchmarks.lean
+++ b/Veir/Benchmarks.lean
@@ -446,17 +446,17 @@ builtin.module() {
 info: builtin.module() {
   ^2:
     %3 = arith.constant 42 ⏎
-    %25 = arith.addi(%3, %3)
-    %28 = arith.addi(%25, %25)
-    %27 = arith.addi(%28, %28)
-    %31 = arith.addi(%27, %27)
-    %29 = arith.addi(%31, %31)
-    %30 = arith.addi(%29, %29)
-    %33 = arith.addi(%30, %30)
+    %34 = arith.addi(%3, %3)
+    %33 = arith.addi(%34, %34)
     %32 = arith.addi(%33, %33)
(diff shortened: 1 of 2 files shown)


Veir/Benchmarks.lean
```diff
@@ -443,21 +443,20 @@ builtin.module() {
...
-    %25 = arith.addi(%26, %26)
-    test.test(%25)
+    %31 = arith.addi(%27, %27)
+    %29 = arith.addi(%31, %31)
+    %30 = arith.addi(%29, %29)
+    %33 = arith.addi(%30, %30)
+    %32 = arith.addi(%33, %33)
+    %34 = arith.addi(%32, %32)
+    %26 = arith.addi(%34, %34)
+    test.test(%26)
```
tobiasgrosser: Is this related? Otherwise, maybe a separate PR?

math-fehr: So the reason this changed is because the hashmap is now having the opaque keys rather than Nats (from what I understand). This means that the ordering in which we traverse the IR when we do pattern rewrites changed (because we still haven't implemented the "actual" one), which is why this happens.
I think it's not worth trying to fix this here unless it has a performance implication, since we will fix this ordering later on

tobiasgrosser: That's fine. A 1-sentence explanation in the PR message will avoid confusion when people later on look at this commit.

For example: 

> We update a VEIR benchmark, as the changed key type in our hashmaps results in a change in rewrite order on our IR. We will resolve this in a future commit when we finalize the rewriter.

math-fehr: This is fixed by using the same Hash as `Nat`

PR #68, opened 2025-12-30
Use `grind =>` instead of `grind .` for allocEmpty

math-fehr:
Using `grind =>` means that `h -> x = y` will be used by grind if `h` and `x` are in the context. Previously, with `grind .`, this lemma was added in the context if both `h`, `x`, and `y` were in the context.

For `allocEmpty`, this means that upon seeing `allocEmpty ctx = some (ctx', op')` and `ptr.get! ctx'`, the value of `ptr.get! ctx` will be added in the grind context. Previously, it was required for `ptr.get! ctx` to appear to be added in the context, which was preventing some `grind` to succeed in future PRs.

### DIFF
--- a/Veir/IR/GetSet.lean
+++ b/Veir/IR/GetSet.lean
@@ -54,7 +54,7 @@ theorem IRContext.topLevelOp_OperationPtr_allocEmpty (heq : OperationPtr.allocEm
     ctx'.topLevelOp = ctx.topLevelOp := by
   grind
 
-@[simp, grind .]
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
The PR is code-wise trivial. A 2-3 sentence explanation, why this change is useful would be useful documentation for the future. Also, does this change have a performance impact?

math-fehr:
added the explanation, and I checked and there is no performance impact.

PR #85, opened 2026-01-08
Add a first implementation of an MLIR lexer

math-fehr:
The lexer does not rely on any parsing/lexing library, and instead follows the MLIR/xDSL implementation directly.

For now, it only supports lexing bare identifiers and whitespace skipping.


tobiasgrosser (approved):

.github/workflows/lean_action_ci.yml
```diff
@@ -15,3 +15,4 @@ jobs:
       - uses: leanprover/lean-action@v1
         with:
           build-args: "--iofail"
+          test: true
```
tobiasgrosser: Add a newline?

PR #89, opened 2026-01-09
Lex `/` and comments starting with `//`.

math-fehr:


Veir/Parser/Lexer.lean
```diff
@@ -191,6 +194,19 @@ partial def lexBareIdentifier (state : LexerState) (tokStart : Nat) : Token × L
   let newState := { state with pos := pos }
   (newState.formToken TokenKind.BareIdent tokStart, newState)
 
+def skipComments (state : LexerState) : LexerState :=
+  if h: state.isEof then
```
tobiasgrosser, line 198:
```suggestion
  if state.isEof then
```

No need to name this condition, no?

PR #92, opened 2026-01-09
Add lexing for @-identifiers

math-fehr:
@-identifiers can either be followed by a string literal or a bare identifier.

### DIFF
--- a/Veir/Parser/Lexer.lean
+++ b/Veir/Parser/Lexer.lean
@@ -256,12 +256,12 @@ def lexAtIdentifier (state : LexerState) (tokStart : Nat) : Except String (Token
   else
     let c := state.input[state.pos]'(by grind [LexerState.isEof])
     if UInt8.isAlphaOrUnderscore c then
-      let (token, state') := lexBareIdentifier state tokStart
-      return (LexerState.formToken state' TokenKind.AtIdent tokStart, state')
+      let (token, state) := lexBareIdentifier state tokStart
+      return (LexerState.formToken state TokenKind.AtIdent tokStart, state)
     else if c == '"'.toUInt8 then
       let newState := { state with pos := state.pos + 1 }
-      let (token, state') ← lexStringLiteral newState tokStart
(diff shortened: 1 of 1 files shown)


Veir/Parser/Lexer.lean
```diff
@@ -244,6 +244,27 @@ termination_by state.input.size - state.pos
...
+  The first character `@` is expected to have already been parsed.
+-/
+def lexAtIdentifier (state : LexerState) (tokStart : Nat) : Except String (Token × LexerState) := do
+  if h: state.isEof then
+    .error "expected identifier or string literal after '@'"
+  else
+    let c := state.input[state.pos]'(by grind [LexerState.isEof])
+    if UInt8.isAlphaOrUnderscore c then
+      let (token, state') := lexBareIdentifier state tokStart
+      return (LexerState.formToken state' TokenKind.AtIdent tokStart, state')
```
ineol: `formToken` is a strange name? why not `mkToken` or `createToken`?

Veir/Parser/Lexer.lean
```diff
@@ -244,6 +244,27 @@ termination_by state.input.size - state.pos
...
+    else if c == '"'.toUInt8 then
+      let newState := { state with pos := state.pos + 1 }
+      let (token, state') ← lexStringLiteral newState tokStart
```
ineol: How about always using `state` as the variable name to make sure it's always used linearly?

Veir/Parser/Lexer.lean
math-fehr: That's the same name as MLIR, I can change it to either if you want, though I'd like to do that on a PR on top of these if possible.

Veir/Parser/Lexer.lean
math-fehr: Oh yes my bad

Veir/Parser/Lexer.lean
math-fehr: I put a PR on top of these ones where I change it to `mkToken`

PR #99, opened 2026-01-10
Modify createOp to take as input the regions

math-fehr:
Previously, we passed the number of regions to create. It makes more sense to pass existing regions instead.

### DIFF
--- a/Veir/Rewriter/Basic.lean
+++ b/Veir/Rewriter/Basic.lean
@@ -333,30 +333,29 @@ def Rewriter.createRegion (ctx: IRContext) : Option (IRContext × RegionPtr) :=
   RegionPtr.allocEmpty ctx
 
 @[irreducible]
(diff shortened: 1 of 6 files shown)


Veir/Rewriter/Basic.lean
```diff
@@ -524,28 +524,17 @@ def IRContext.create : Option (IRContext × OperationPtr) :=
     grind [Rewriter.createEmptyOp, OperationPtr.allocEmpty, Operation.empty, OperationPtr.set, RegionPtr.InBounds]
```
math-fehr, line 524: This change breaks a bit these proofs, but I feel like we should fix these proofs later when we want to make this sorry-free.

ineol: I agree, we need a good strategy for this function

PR #105, opened 2026-01-13
Add parsing of keywords.

math-fehr:
Parsing a keyword means that we are parsing a bare identifier that we expect to be a specific identifier.

### DIFF
--- a/Veir/Parser/Parser.lean
+++ b/Veir/Parser/Parser.lean
@@ -55,7 +55,7 @@ private def consumeToken : m Token := do
   Otherwise, return none.
 -/
 private def parseOptionalToken (tokType : TokenKind) : m (Option Token) := do
-  if (←get).currentToken.kind == tokType then
+  if (←get).currentToken.kind = tokType then
     some <$> consumeToken
   else
     return none
@@ -159,7 +159,7 @@ def parseOptionalKeyword (keyword : ByteArray) : m Bool := do
   match ← peekToken with
   | {kind := .bareIdent, slice := slice : Token} =>
     let ident := slice.of ((← get).input)
-    if ident == keyword then
+    if ident = keyword then
       let _ ← consumeToken
       return true
     else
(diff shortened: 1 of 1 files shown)


Veir/Parser/Parser.lean
```diff
@@ -150,23 +150,45 @@ def parseIdentifier (errorMsg : String := "identifier expected") : m ByteArray :
   | none => throw errorMsg
 
 /--
-  Parse a boolean with grammar rule `true | false`, if present.
-  If the next token is a boolean, consume it and return its value.
+  Parse optionally a specific keyword.
+  The given keyword should be parseable as an identifier.
+  If the next token is an identifier matching the given keyword, consume it and return it.
```
tobiasgrosser, line 155:
```suggestion
  If the next token is an identifier, match the given keyword, consume it, and return it.
```

?

tobiasgrosser (approved):

Veir/Parser/Parser.lean
math-fehr: It is not completely the same right?
Here we only consume it if we match it, I feel that writing `, match the given keyword` doesn't make it obvious that we don't return an error if we have an identifier that doesn't match the keyword?

tobiasgrosser: I got it.

Veir/Parser/Parser.lean
```diff
@@ -150,23 +150,45 @@ def parseIdentifier (errorMsg : String := "identifier expected") : m ByteArray :
...
   Otherwise, return none.
 -/
-def parseOptionalBoolean : m (Option Bool) := do
+def parseOptionalKeyword (keyword : ByteArray) : m Bool := do
   match ← peekToken with
-  | {kind := .bareIdent, slice := slice} =>
+  | {kind := .bareIdent, slice := slice : Token} =>
     let ident := slice.of ((← get).input)
-    if ident == "true".toByteArray then
+    if ident == keyword then
```
ineol: Why not `ident = keyword`?

Veir/Parser/Parser.lean
```diff
@@ -150,23 +150,45 @@ def parseIdentifier (errorMsg : String := "identifier expected") : m ByteArray :
...
+  | _ => return false
+
+/--
+  Parse a specific keyword.
+  The given keyword should be parseable as an identifier.
+  If the next token is an identifier matching the given keyword, consume it and return it.
+  Otherwise, return an error with the given message.
+-/
+def parseKeyword (keyword : ByteArray) (errorMsg : String := s!"expected keyword '{String.fromUTF8! keyword}'") : m Unit := do
+  match ← parseOptionalKeyword keyword with
```
ineol: `match` on a boolean is a bit strange?

ineol (approved):

PR #108, opened 2026-01-14
Add parsing of string literals

math-fehr:
Currently, the parser does not handle escape sequences in string literals.

### DIFF
--- a/Veir/Parser/Parser.lean
+++ b/Veir/Parser/Parser.lean
@@ -174,18 +174,11 @@ def parseOptionalKeyword (keyword : ByteArray) : M Bool := do
   If the next token is an identifier matching the given keyword, consume it and return it.
   Otherwise, return an error with the given message.
 -/
-<<<<<<< HEAD
-def parseKeyword (keyword : ByteArray) (errorMsg : String := s!"expected keyword '{String.fromUTF8! keyword}'") : m Unit := do
+def parseKeyword (keyword : ByteArray) (errorMsg : String := s!"expected keyword '{String.fromUTF8! keyword}'") : M Unit := do
   if ← parseOptionalKeyword keyword then
     return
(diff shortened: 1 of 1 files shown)


Veir/Parser/Parser.lean
```diff
@@ -179,6 +179,31 @@ def parseKeyword (keyword : ByteArray) (errorMsg : String := s!"expected keyword
...
+    match String.fromUTF8? slice with
+    | some str => return some str
+    | none => throw "internal error: failed converting string literal"
+  | none => return none
+
+/--
+  Parse a string literal.
+  Raise an error if the next token is not a string literal.
+-/
+def parseStringLiteral (errorMsg : String := "string literal expected") : m String := do
```
tobiasgrosser: I am not sure what this `m` is doing here? Can you explain this to me (just that I learn?)

tobiasgrosser (approved):

Veir/Parser/Parser.lean
math-fehr: That's something that Leo teached me so we can have "inheritance" in our parser structure.

Early in the file we have `variable [Monad m] [MonadExcept String m] [MonadStateOf ParserState m]`
So here, `m` is any monad that has both an except part and a `ParserState` state component.

So we can call this function on a monad that has a state "bigger" than `ParserState`, and this will only modify `ParserState`.
You can see an example in `AttrParser` where we have a monad that has both `ParserState` and `AttrParserState`.

tobiasgrosser: Maybe there is a more descriptive name then `m`?

ineol: It's the Lean convention as far as I can see. Maybe calling in `M` would be even better since its range is `Type`

PR #109, opened 2026-01-14
Parse simple MLIR operations, regions, and blocks.

math-fehr:
Currently, the limitations are:
* Operations have no operands, results, attributes, successors, or properties.
* Blocks have no arguments.
* Regions have always a single block.
* Operations have at most one region.

We should also do the following:
* Create a `!` variant of our IR rewrites so that we don't have to provide proofs or `(by sorry)` in our code
* Have a monad for our `Rewriter`, so that we don't have to `<- get` and `<- set` our context manually.

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -29,6 +29,26 @@ structure MlirParserState where
 abbrev MlirParserM := StateT MlirParserState (EStateM String ParserState)
 
 /--
+  Execute the action with the given initial state.
+  Returns the result along with the final state, or an error message.
+-/
+def MlirParserM.run (self : MlirParserM α)
+  (mlirParserState : MlirParserState) (parserState: ParserState) : Except String (α × MlirParserState × ParserState) :=
+  match (StateT.run self mlirParserState).run parserState with
+  | .ok (a, mlirParserState) parserState => .ok (a, mlirParserState, parserState)
+  | .error err _ => .error err
+
+/--
+  Execute the action with the given initial state.
+  Returns the result or an error message.
+-/
+def MlirParserM.run' (self : MlirParserM α)
+  (mlirParserState : MlirParserState) (parserState: ParserState) : Except String α :=
+  match self.run mlirParserState parserState with
+  | .ok (a, _, _) => .ok a
+  | .error err => .error err
+
+/--
   Get the current IR context that is stored in the parser state.
 -/
 def getContext : MlirParserM IRContext := do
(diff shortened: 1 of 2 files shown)


Veir/Parser/MlirParser.lean
```diff
@@ -0,0 +1,119 @@
...
+def operationNameToOpId (name : String) : Nat :=
+  match name with
+  | "builtin.module" => 0
+  | "arith.constant" => 1
+  | "arith.addi" => 2
+  | "return" => 3
+  | "arith.muli" => 4
+  | "arith.andi" => 5
+  | "test.test" => 99
+  | _ => 42
```
tobiasgrosser, line 24: Is there a reason you do not put the unregistered name to `0`? That seems more canonical.

Veir/Parser/MlirParser.lean
```diff
@@ -0,0 +1,119 @@
...
+/--
+  Parse a region.
+-/
+partial def parseRegion : MlirParserM RegionPtr := do
+  parsePunctuation "{"
+  let ctx := ← getContext
+  match Rewriter.createRegion ctx with
+  | some (ctx, region) =>
+    setContext ctx
+    let _ ← parseBlock (BlockInsertPoint.atEnd region)
```
tobiasgrosser, line 114:
```suggestion
    parseBlock (BlockInsertPoint.atEnd region)
```

math-fehr: I'll try this tomorrow again, but since `parseBlock` returns something, I think Lean wasn't happy about it.

Veir/Parser/MlirParser.lean
math-fehr: I think it's because I added this later, I can move this to `0` if you want, in any case we will change this soon so we can add any operation we want.

Veir/Parser/MlirParser.lean
tobiasgrosser: i tried. it indeed does not work.

Veir/Parser/MlirParser.lean
```diff
@@ -0,0 +1,119 @@
...
+    return #[]
+  let region ← parseRegion
+  parsePunctuation ")"
+  return #[region]
+
+/--
+  Parse an operation, if present, and insert it at the given insert point.
+-/
+partial def parseOptionalOp (ip : Option InsertPoint) : MlirParserM (Option OperationPtr) := do
+  match ← parseOptionalStringLiteral with
```
ineol: `let some .. | return none`

Veir/Parser/MlirParser.lean
```diff
@@ -0,0 +1,119 @@
...
+    parsePunctuation ")"
+    let regions ← parseOpRegions
+    parsePunctuation ":"
+    parsePunctuation "("
+    parsePunctuation ")"
+    parsePunctuation "->"
+    parsePunctuation "("
+    parsePunctuation ")"
+    let opId := operationNameToOpId opName
+    match Rewriter.createOp (← getContext) opId 0 #[] regions 0 ip (by grind) (by sorry) (by sorry) (by sorry) with
```
ineol: also

Veir/Parser/MlirParser.lean
```diff
@@ -0,0 +1,119 @@
...
+    | some (ctx, op) =>
+      setContext ctx
+      return op
+    | none => throw "internal error: failed to create operation"
+
+/--
+  Parse an operation.
+-/
+partial def parseOp (ip : Option InsertPoint) : MlirParserM OperationPtr := do
```
ineol, line 101: also

PR #120, opened 2026-01-16
Parse operation results in MLIR format

math-fehr:
We do not handle yet the parsing of result tuples, and do not set the result types, as no types exist yet.

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -76,12 +76,10 @@ def parseOpResult : MlirParserM Slice := do
   either as a list of values followed by '=', or nothing.
 -/
 def parseOpResults : MlirParserM (Array Slice) := do
-  if (← peekToken).kind = .percentIdent then
-    let results ← parseList parseOpResult
-    parsePunctuation "=" "'=' expected after operation results"
-    return results
-  else
-    return #[]
+  let .percentIdent := (← peekToken).kind | return #[]
+  let results ← parseList parseOpResult
+  parsePunctuation "=" "'=' expected after operation results"
+  return results
 
 /--
   Parse a type, if present.


Veir/Parser/MlirParser.lean
```diff
@@ -62,6 +64,57 @@ def getContext : MlirParserM IRContext := do
...
+def parseOpResult : MlirParserM Slice := do
+  let nameToken ← parseToken .percentIdent "operation result expected"
+  return { nameToken.slice with start := nameToken.slice.start + 1 } -- skip % character
+
+/--
+  Parse the results before an operation definition,
+  either as a list of values followed by '=', or nothing.
+-/
+def parseOpResults : MlirParserM (Array Slice) := do
+  if (← peekToken).kind = .percentIdent then
```
ineol: Fallible let?

Veir/Parser/MlirParser.lean
```diff
@@ -75,18 +128,16 @@ partial def parseOpRegions : MlirParserM (Array RegionPtr) := do
...
-  parsePunctuation ")"
-  parsePunctuation "->"
-  parsePunctuation "("
-  parsePunctuation ")"
+  let (inputTypes, outputTypes) ← parseOperationType
+  if outputTypes.size ≠ results.size then
+    throw s!"operation '{opName}' declares {outputTypes.size} result types, but {results.size} result names were provided"
   let opId := operationNameToOpId opName
-  let some (ctx, op) := Rewriter.createOp (← getContext) opId 0 #[] regions 0 ip (by grind) (by sorry) (by sorry) (by sorry)
+  let some (ctx, op) := Rewriter.createOp (← getContext) opId results.size #[] regions 0 ip (by grind) (by sorry) (by sorry) (by sorry)
```
ineol, line 138: I'm curious, how difficult are the `sorry`s?

ineol (approved):

Veir/Parser/MlirParser.lean
math-fehr: I think they would be doable if we propagate the right information.
I guess the `ctx.FieldsInBounds` should always be correct, and otherwise we just need 1-2 lemmas and that should work.
So I can try to make this work later if you want!

PR #121, opened 2026-01-17
Make MlirType a String

math-fehr:
While this is not the end-goal with MlirType, it allows us to use it in the parser and the printer, and allows us to catch some bugs in the codebase that relied on the fact that all types are equal.


Veir/IR/GetSet.lean
```diff
@@ -750,12 +751,20 @@ theorem ValuePtr.getFirstUse!_OperationPtr_setResults {value : ValuePtr} :
...
+      if result.op = operation' then
+        newResults[result.index]!.type
+      else
+        value.getType! ctx
+    | _ =>
+      value.getType! ctx := by
   grind
 
+
 @[grind =]
```
tobiasgrosser, line 767:
```suggestion

@[grind =]
```
Remove unnecessary newline edition.

PR #127, opened 2026-01-18
Add simple veir-opt tool

math-fehr:
`veir-opt` now parses then prints MLIR files.
This is not the full tool yet, just a small script.
Most of this code is going to be refactored once we implement passes, read files from stdin, etc...

### DIFF
--- a/lakefile.toml
+++ b/lakefile.toml
@@ -1,6 +1,6 @@
 name = "VeIR"
 version = "0.1.0"
-defaultTargets = ["veir-opt"]
+defaultTargets = ["veir-opt", "run-benchmarks", "Veir"]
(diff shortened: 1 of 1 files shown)


math-fehr:
Good point, I put `RunBenchmarks` back as an executable.
@tobiasgrosser @ineol, do you know how to add these to the CI? Are they automatically added?

tobiasgrosser:
> Good point, I put `RunBenchmarks` back as an executable. @tobiasgrosser @ineol, do you know how to add these to the CI? Are they automatically added?

You need to add them to the `defaultTargets` in `lakefile.yaml`. I did this and now all is compiled by default. Thank you for this nice addition!

PR #128, opened 2026-01-20
Generalize IRContext.WellFormed by allowing missing uses

math-fehr:
In the proofs related to `insertIntoCurrent`, we need to be in a state where the IR is well formed except for some uses.

### DIFF
--- a/Veir/IR/WellFormed.lean
+++ b/Veir/IR/WellFormed.lean
@@ -611,9 +611,9 @@ structure IRContext.WellFormed (ctx : IRContext)
   (missingSuccessorUses : Std.ExtHashSet BlockOperandPtr := ∅) : Prop where
   inBounds : ctx.FieldsInBounds
   valueDefUseChains (valuePtr : ValuePtr) (valuePtrInBounds : valuePtr.InBounds ctx) :
-    ∃ array, ValuePtr.DefUse valuePtr ctx array (missingOperandUses.filter (λ use => (use.get! ctx).value = valuePtr))
+    ∃ array, ValuePtr.DefUse valuePtr ctx array (missingOperandUses.filter (fun use => (use.get! ctx).value = valuePtr))
   blockDefUseChains (blockPtr : BlockPtr) (blockPtrInBounds : blockPtr.InBounds ctx) :
-    ∃ array, BlockPtr.DefUse blockPtr ctx array (missingSuccessorUses.filter (λ use => (use.get! ctx).value = blockPtr))
+    ∃ array, BlockPtr.DefUse blockPtr ctx array (missingSuccessorUses.filter (fun use => (use.get! ctx).value = blockPtr))
   opChain (blockPtr : BlockPtr) (blockPtrInBounds : blockPtr.InBounds ctx) :
     ∃ array, BlockPtr.OpChain blockPtr ctx array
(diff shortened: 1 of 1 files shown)


Veir/IR/WellFormed.lean
```diff
@@ -790,10 +792,16 @@ theorem BlockPtr.operationList.mem :
 noncomputable def ValuePtr.defUseArray (value : ValuePtr) (ctx : IRContext) (hctx : ctx.WellFormed) (hvalue : value.InBounds ctx) : Array OpOperandPtr :=
   (hctx.valueDefUseChains value hvalue).choose
 
+theorem test :
+    ValuePtr.DefUse value ctx array →
+    array = array' →
+    ValuePtr.DefUse value ctx array' := by
+  grind
```
tobiasgrosser: Why adding a theorem named `test`?

tobiasgrosser (approved):
LGTM, just one comment.

luisacicolini (approved):

Veir/IR/WellFormed.lean
```diff
@@ -606,12 +606,14 @@ structure Region.WellFormed (region : Region) (ctx : IRContext) (regionPtr : Reg
...
   parent_op {op} (heq : region.parent = some op) : ∃ i, i < op.getNumRegions! ctx → op.getRegion! ctx i = regionPtr
 
-structure IRContext.WellFormed (ctx : IRContext) : Prop where
+structure IRContext.WellFormed (ctx : IRContext)
+  (missingOperandUses : Std.ExtHashSet OpOperandPtr := ∅)
+  (missingSuccessorUses : Std.ExtHashSet BlockOperandPtr := ∅) : Prop where
   inBounds : ctx.FieldsInBounds
   valueDefUseChains (valuePtr : ValuePtr) (valuePtrInBounds : valuePtr.InBounds ctx) :
-    ∃ array, ValuePtr.DefUse valuePtr ctx array
+    ∃ array, ValuePtr.DefUse valuePtr ctx array (missingOperandUses.filter (λ use => (use.get! ctx).value = valuePtr))
```
ineol: Can you write `((\cdot.get! ctx).value = valuePtr)` instead?

Veir/ForLean.lean
```diff
@@ -115,6 +117,11 @@ theorem Nat.eq_iff_forall_lessthan :
 
 deriving instance DecidableEq for Except
 
+@[simp, grind =]
```
ineol, line 120: Is it really necessary??

ineol (approved):

Veir/IR/WellFormed.lean
math-fehr: My bad, this was an experiment I forgot to remove

Veir/ForLean.lean
math-fehr: So https://leanprover-community.github.io/mathlib4_docs/Std/Data/ExtHashSet/Lemmas.html#Std.ExtHashSet.filter_eq_empty_iff is the nearest theorem, otherwise there is no existing theorem like this.
`grind =` is necessary for some proofs when I checked.

Veir/IR/WellFormed.lean
math-fehr: This doesn't work for me, because `(\cdot.get! ctx).value = valuePtr` makes the inner parenthesis a function, so `.value` doesn't work after that.

PR #132, opened 2026-01-20
Prove well-formedness of Rewriter.detachOperands

math-fehr:


Veir/Rewriter/WellFormed/Rewriter/Operation.lean
```diff
@@ -222,6 +222,43 @@ theorem Rewriter.detachOpIfAttached_WellFormed (ctx : IRContext) (wf : ctx.WellF
 
 end detachOpIfAttached
 
+theorem Rewriter.detachOperands.loop_wellFormed
```
luisacicolini, line 225: given that you have a lot of `Rewriter` lemmas it might be worth using a namespace?

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
```diff
@@ -222,6 +222,43 @@ theorem Rewriter.detachOpIfAttached_WellFormed (ctx : IRContext) (wf : ctx.WellF
...
+theorem Rewriter.detachOperands_wellFormed
+    (wf : IRContext.WellFormed ctx missingOperands missingSuccessors)
+    (hMissingOperands : ∀ i, (OpOperandPtr.mk op i) ∉ missingOperands) :
+    (Rewriter.detachOperands ctx op hCtx hOp).WellFormed
+      (missingOperands.insertMany ((0...op.getNumOperands! ctx).toList.map (fun i => OpOperandPtr.mk op i)))
+      missingSuccessors := by
+  simp only [Rewriter.detachOperands]
+  split
+  case isTrue h =>
+    rw [←OperationPtr.getNumOperands!_eq_getNumOperands (hin := by grind)] at h
```
luisacicolini:
```suggestion
    rw [← OperationPtr.getNumOperands!_eq_getNumOperands (hin := by grind)] at h
```

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
```diff
@@ -222,6 +222,43 @@ theorem Rewriter.detachOpIfAttached_WellFormed (ctx : IRContext) (wf : ctx.WellF
 
 end detachOpIfAttached
 
+theorem Rewriter.detachOperands.loop_wellFormed
+    (wf : IRContext.WellFormed ctx missingOperands missingSuccessors)
+    (hMissingOperands : ∀ i, i <= index → (OpOperandPtr.mk op i) ∉ missingOperands) :
+    (Rewriter.detachOperands.loop ctx op index hCtx hOp hIndex).WellFormed
```
luisacicolini, line 228:
```suggestion
      (Rewriter.detachOperands.loop ctx op index hCtx hOp hIndex).WellFormed
```
for consistency with the rest of the lemma (afaict)

ineol: Why spacing here?

Veir/ForLean.lean
```diff
@@ -140,6 +140,11 @@ theorem Std.ExtHashSet.filter_insert_eq_of_false_eq {α : Type} [Hashable α] [B
   intro h
   ext; grind
 
+theorem Std.ExtHashSet.insertMany_list_insert_comm [BEq α] [EquivBEq α] [Hashable α] [LawfulHashable α] [BEq α] [LawfulBEq α]
```
ineol, line 143: Does it rely on `l` being a list?

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
ineol: In general we've been bad at using `namespace` I think

math-fehr: What kind of namespaces are you suggesting?
My understanding is that we should keep the namespace of either the function, or the prop. Are you suggesting having a different namespace for these proofs?
I'm open to anything that would make this nicer (though in a different refactoring PR)!

Veir/ForLean.lean
math-fehr: We talked about it in person, we'll keep this for now as adapting it for a generic `ForInt` type seems annoying and not worth it.

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
math-fehr: Oh I see what you mean actually.
Well here this is only 3 lemmas, so I'm not sure it's worth it for this?

luisacicolini: I took a better look at the repo organization and this would be my suggestion with namespace (for a refactoring PR, though): 
* in the `Basic` file (e.g. `Rewriter/Basic`) also declare a `namespace` (afaict this is also common in lean4)
* the namespace should be opened and closed when you have multiple lemmas about `Rewriter` - this I am not sure "how many lemmas" are enough to oppen and close the namespace, but my understanding is that whenever you have more than one theorem having that in a namespace makes things a bit easier, because if you need any other namespace-related lemmas you have them loaded there

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
luisacicolini: for consistency with the body of the theorem statement which is indented more (e.g. line below)

luisacicolini: (I could not find a good ref for indentation of lemmas with multi-line hypotheses in lean4, so anything works as long as we're consistent)

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
math-fehr: Okay that makes sense!
I'll probably try to make some refactoring PRs on the "core" files, then we can discuss everything there (thanks a lot for these comments btw)

PR #133, opened 2026-01-21
Dominance Analysis

axelcool1234:
This implements CFG dominance analysis on top of the dataflow framework: #264 

This is based off of the Cooper-Harvey-Kennedy algorithm.

### DIFF
--- a/Veir/Analysis/DataFlow/DominanceAnalysis.lean
+++ b/Veir/Analysis/DataFlow/DominanceAnalysis.lean
@@ -1,0 +1,375 @@
+module
+
+public import Veir.Analysis.DataFlowFramework
+public import Std.Data.HashSet
+public import Std.Data.HashMap
+
+public section
+
+namespace Veir
+
+open Std (HashMap HashSet)
+
+structure DominatorState extends AnalysisStateHeader where
+  iDom : Option BlockPtr := none
+deriving TypeName
+
+instance : AnalysisState DominatorState where
+  typeNameInst := inferInstance
+  mkState := fun anchor =>
+    { anchor := anchor
+      dependents := #[]
+      iDom := none }
+  onUpdate (state : DominatorState) (dfCtx : DataFlowContext)
+      (_irCtx : IRContext OpCode) : DataFlowContext :=
+    { dfCtx with workList := state.enqueueDependents dfCtx.workList }
+  toHeader := DominatorState.toAnalysisStateHeader
+
+/-- Hack to cache the post ordering of a region's blocks so we don't
+have to recompute it. Stored in the entry block of each region.
+-/
+structure RegionMetadataState extends AnalysisStateHeader where
+  postOrderIndex : HashMap BlockPtr Nat := {}
+deriving TypeName
+
+instance : AnalysisState RegionMetadataState where
+  typeNameInst := inferInstance
+  mkState := fun anchor =>
+    { anchor := anchor
+      dependents := #[]
+      postOrderIndex := {} }
+  onUpdate (_state : RegionMetadataState) (dfCtx : DataFlowContext)
+      (_irCtx : IRContext OpCode) : DataFlowContext :=
+    dfCtx
+  toHeader := RegionMetadataState.toAnalysisStateHeader
+
+namespace DominatorState
+
+def get? (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option DominatorState :=
+  dfCtx.getState? (.ProgramPoint (ProgramPoint.beforeBlock block irCtx))
+
+/--
+Get the immediate dominator for a block. 
+-/
+def getIDom? (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  match get? block dfCtx irCtx with
+  | some state =>
+    state.iDom
+  | none =>
+    none
+
+/--
+Walk up the dominator tree via iDoms to gather the dominators for a block. 
+-/
+def getDoms? (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option (HashSet BlockPtr) :=
+  match get? block dfCtx irCtx with
+  | none =>
+    none
+  | some state =>
+    Id.run do
+      let mut doms : HashSet BlockPtr := (∅ : HashSet BlockPtr).insert block
+      let mut current := state.iDom
+      while current.isSome do
+        let dom := current.get!
+        if dom == block then
+          return some doms
+        doms := doms.insert dom
+        let next := getIDom? dom dfCtx irCtx
+        if next == some dom then
+          return some doms
+        current := next
+      some doms
+
+end DominatorState
+
+namespace RegionMetadataState
+
+def get? (dfCtx : DataFlowContext) (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Option RegionMetadataState :=
+  match (region.get! irCtx).firstBlock with
+  | some entry =>
+    dfCtx.getState? (.ProgramPoint (ProgramPoint.beforeBlock entry irCtx))
+  | none =>
+    none
+
+end RegionMetadataState
+
+namespace DominanceAnalysis
+
+def id : Lean.Name :=
+  Lean.Name.mkSimple "DominanceAnalysis"
+
+/--
+Get the lattice anchor of a given block. 
+-/
+private def blockEntryAnchor (block : BlockPtr) (irCtx : IRContext OpCode) : LatticeAnchor :=
+  .ProgramPoint (ProgramPoint.beforeBlock block irCtx)
+
+/--
+Get the successors of a given block. 
+-/
+private def successorsInRegion
+    (block : BlockPtr)
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr :=
+  match (block.get! irCtx).lastOp with
+  | none =>
+    #[]
+  | some term =>
+    Id.run do
+      let mut succs := #[]
+      for i in [0:term.getNumSuccessors! irCtx] do
+        let succ := term.getSuccessor! irCtx i
+        if (succ.get! irCtx).parent == some region then
+          succs := succs.push succ
+      succs
+
+/--
+The returned array is CFG in postorder, and the map assigns 
+each block a postorder index used by `intersect`.
+-/
+private def collectPostOrder
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr × HashMap BlockPtr Nat :=
+  match (region.get! irCtx).firstBlock with
+  | none =>
+    (#[], {})
+  | some entry =>
+    Id.run do
+      let mut worklist : Array (BlockPtr × Bool) := #[(entry, false)]
+      let mut seen : HashSet BlockPtr := ∅
+      let mut postOrder : Array BlockPtr := #[]
+      let mut postOrderIndex : HashMap BlockPtr Nat := {}
+
+      while !worklist.isEmpty do
+        let (block, visited) := worklist.back!
+        worklist := worklist.pop
+
+        if visited then
+          let index := postOrder.size
+          postOrder := postOrder.push block
+          postOrderIndex := postOrderIndex.insert block index
+        else if seen.contains block then
+          continue
+        else
+          seen := seen.insert block
+          worklist := worklist.push (block, true)
+
+          for succ in successorsInRegion block region irCtx do
+            if !seen.contains succ then
+              worklist := worklist.push (succ, false)
+      (postOrder, postOrderIndex)
+
+/- Initialize the dominators in post order and enqueue them in 
+reverse post order.
+-/
+private def initializeRegion
+    (region : RegionPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  match (region.get! irCtx).firstBlock with
+  | none =>
+    dfCtx
+  | some entry =>
+    Id.run do
+      -- Calculate the post order of the blocks and cache it.
+      let (postOrder, postOrderIndex) := collectPostOrder region irCtx
+      let reversePostOrder := postOrder.reverse
+      let metadata : RegionMetadataState :=
+        { anchor := blockEntryAnchor entry irCtx
+          dependents := #[]
+          postOrderIndex := postOrderIndex }
+      let mut dfCtx := DataFlowContext.setState dfCtx (blockEntryAnchor entry irCtx) (fun _ => metadata)
+
+      -- Initialize the blocks in post order.
+      for block in postOrder do
+        let mut dependents := #[]
+        for succ in successorsInRegion block region irCtx do
+          dependents := dependents.push (ProgramPoint.beforeBlock succ irCtx, id)
+        let state : DominatorState :=
+          { anchor := blockEntryAnchor block irCtx
+            dependents := dependents
+            iDom := if block == entry then some entry else none }
+        dfCtx := DataFlowContext.setState dfCtx (blockEntryAnchor block irCtx) (fun _ => state)
+
+      -- Enqueue the blocks in reverse post order.
+      for block in reversePostOrder do
+        match blockEntryAnchor block irCtx with
+        | .ProgramPoint point =>
+          dfCtx := { dfCtx with workList := dfCtx.workList.enqueue (point, id) }
+        | _ =>
+          unreachable!
+
+      dfCtx
+
+/--
+Recursively initialize the analysis on nested regions.
+-/
+partial def initializeRecursively
+    (op : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  let mut dfCtx := dfCtx
+
+  for i in [0:op.getNumRegions! irCtx] do
+    let region := op.getRegion! irCtx i
+    dfCtx := initializeRegion region dfCtx irCtx
+
+    let regionData := region.get! irCtx
+
+    let mut maybeBlock := regionData.firstBlock
+    while hBlock : maybeBlock.isSome do
+      let block := maybeBlock.get hBlock
+      let mut maybeOp := (block.get! irCtx).firstOp
+      while hOp : maybeOp.isSome do
+        let nestedOp := maybeOp.get hOp
+        dfCtx := initializeRecursively nestedOp dfCtx irCtx
+        maybeOp := (nestedOp.get! irCtx).next
+      maybeBlock := (block.get! irCtx).next
+
+  dfCtx
+
+def init
+    (top : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  initializeRecursively top dfCtx irCtx
+
+
+/--
+Find the nearest common dominator of `block1` and `block2`.
+
+On each step, the cursor with the smaller postorder index is moved upward
+until both cursors coincide. The block where they meet is the nearest common
+dominator of the two input blocks.
+-/
+private def intersect
+    (block1 block2 : BlockPtr)
+    (postOrderIndex : HashMap BlockPtr Nat)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : BlockPtr := Id.run do
+  let mut finger1 := block1
+  let mut finger2 := block2
+  while finger1 != finger2 do
+    while postOrderIndex[finger1]! < postOrderIndex[finger2]! do
+      finger1 := (DominatorState.getIDom? finger1 dfCtx irCtx).get!
+    while postOrderIndex[finger2]! < postOrderIndex[finger1]! do
+      finger2 := (DominatorState.getIDom? finger2 dfCtx irCtx).get!
+  finger1
+
+/--
+Compute the next immediate dominator candidate for `block`.
+
+The entry block dominates itself. For every other block, we scan its predecessors, pick the first one
+whose dominator state has already been computed, and then repeatedly `intersect` that candidate with each
+other processed predecessor. Predecessors whose `iDom` is still unknown are skipped until a later solver
+iteration.
+-/
+private def computeImmediateDominator
+    (block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  let region := ((block.get! irCtx).parent).get!
+  let entry := ((region.get! irCtx).firstBlock).get!
+  match RegionMetadataState.get? dfCtx region irCtx with
+  | none =>
+    none
+  | some metadata =>
+    if block == entry then
+      some entry
+    else
+      Id.run do
+        let mut maybePredUse := (block.get! irCtx).firstUse
+        let mut newIDom : Option BlockPtr := none
+
+        while hUse : maybePredUse.isSome do
+          let predUse := maybePredUse.get hUse
+          let predUseStruct := predUse.get! irCtx
+          maybePredUse := predUseStruct.nextUse
+          let predOp := predUseStruct.owner
+          let some predBlock := (predOp.get! irCtx).parent
+            | continue
+          if (predBlock.get! irCtx).parent != some region then
+            continue
+          let some _ := DominatorState.getIDom? predBlock dfCtx irCtx
+            | continue
+          newIDom := match newIDom with
+            | none => some predBlock
+            | some current => some (intersect predBlock current metadata.postOrderIndex dfCtx irCtx)
+
+        newIDom
+
+def visit
+    (point : ProgramPoint)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  if !point.isBlockStart irCtx then
+    dfCtx
+  else
+    let block := (point.getBlock?).get!
+    match computeImmediateDominator block dfCtx irCtx with
+    | none =>
+      dfCtx
+    | some newIDom =>
+      let anchor := blockEntryAnchor block irCtx
+      DataFlowContext.setStateAndUpdate dfCtx anchor (State := DominatorState) (fun oldState =>
+        ({ oldState with iDom := some newIDom }, some newIDom != oldState.iDom)) irCtx
+
+/-- Get a block's immediate dominator. -/
+def immediateDominator?
+    (block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  match DominatorState.getIDom? block dfCtx irCtx with
+  | some iDom =>
+    if iDom == block then none else some iDom
+  | none =>
+    none
+
+/-- Recurse up the dominator tree to see if `dominator` 
+is hit. If it is, then `dominator` dominates `block`.
+-/
+def dominates
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  if dominator == block then
+    (DominatorState.get? block dfCtx irCtx).isSome
+  else
+    match DominatorState.get? block dfCtx irCtx with
+    | none =>
+      false
+    | some state =>
+      Id.run do
+        let mut current := state.iDom
+        while current.isSome do
+          let candidate := current.get!
+          if candidate == dominator then
+            return true
+          let next := DominatorState.getIDom? candidate dfCtx irCtx
+          if next == some candidate then
+            return false
+          current := next
+        false
+
+/-- Same thing as `dominates` except a blcok dominating itself
+doesn't count.
+-/
+def strictlyDominates
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  dominator != block && dominates dominator block dfCtx irCtx
+
+end DominanceAnalysis
+
+def DominanceAnalysis : DataFlowAnalysis :=
+  DataFlowAnalysis.new
+    DominanceAnalysis.id
+    DominanceAnalysis.init
+    DominanceAnalysis.visit
+
+end Veir
--- a/Veir/Analysis/DataFlow/DominanceAnalysis.lean
+++ b/Veir/Analysis/DataFlow/DominanceAnalysis.lean
@@ -1,0 +1,310 @@
+module
+
+public import Veir.Analysis.DataFlowFramework
+public import Std.Data.HashSet
+public import Std.Data.HashMap
+
+public section
+
+namespace Veir
+
+open Std (HashMap HashSet)
+
+/-!
+# Dominance analysis
+
+This module implements immediate dominator analysis using the Cooper Harvey
+Kennedy algorithm described in their paper "A Simple, Fast Dominance Algorithm."
+
+Like the algorithm in that paper, we initialize the entry block to dominate
+itself, process reachable blocks in reverse postorder, and iteratively refine
+each block's immediate dominator by intersecting the dominator chains of its
+already processed predecessors. The `intersect` helper uses cached postorder
+indices of two blocks as pointers into their dominator chains. The lower
+ranked pointer is moved upward until both pointers meet at their nearest
+common dominator. `computeImmediateDominator` implements the
+paper's update step by choosing the first predecessor whose immediate
+dominator is already known as an initial candidate, then repeatedly
+intersects that candidate with the other predecessors whose immediate
+dominator is already known. The resulting candidate is the current
+immediate dominator estimate for the block. As more predecessor dominator
+facts become available, the worklist revisits the block and recomputes that
+estimate. Each recomputation either preserves the estimate or moves it upward
+in the dominator tree (note that this is monotonic!), and the process repeats
+until the facts reach a fixpoint.
+
+In VeIR, dominator facts are attached to block entry `InsertPoint`s. A separate
+region metadata fact stores the postorder numbering needed by `intersect`, and
+the ordinary dataflow worklist is used to revisit dependent successors until the
+immediate dominator facts reach a fixpoint.
+-/
+
+namespace BlockPtr
+
+def getDominatorFact? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option DominatorFact :=
+  dfCtx.getFact? .dominator (.InsertPoint (InsertPoint.atStart! block irCtx))
+
+def getIDom? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  block.getDominatorFact? dfCtx irCtx >>= (·.iDom)
+
+def getDoms? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option (HashSet BlockPtr) := do
+  let fact <- block.getDominatorFact? dfCtx irCtx
+  let mut doms := (∅ : HashSet BlockPtr).insert block
+  let mut current := fact.iDom
+  while let some dom := current do
+    if dom = block then
+      return doms
+    doms := doms.insert dom
+    let next := dom.getIDom? dfCtx irCtx
+    if next = some dom then
+      return doms
+    current := next
+  doms
+
+end BlockPtr
+
+namespace RegionPtr
+
+def getRegionMetadataFact? [FactSpec .regionMetadata] (region : RegionPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option RegionMetadataFact :=
+  (region.get! irCtx).firstBlock >>= 
+    dfCtx.getFact? .regionMetadata ∘ (InsertPoint.atStart! · irCtx)
+
+end RegionPtr
+
+namespace DominatorFact
+
+def mkDefault (anchor : LatticeAnchor) : DominatorFact :=
+  { anchor := anchor
+    dependents := #[]
+    payload := { iDom := none } }
+
+def propagate (fact : DominatorFact) (dfCtx : DataFlowContext)
+    (_irCtx : IRContext OpCode) : DataFlowContext :=
+  { dfCtx with workList := fact.enqueueDependents dfCtx.workList }
+
+instance : FactSpec .dominator where
+  mkDefault := DominatorFact.mkDefault
+  propagate := DominatorFact.propagate
+
+end DominatorFact
+
+namespace RegionMetadataFact
+
+def mkDefault (anchor : LatticeAnchor) : RegionMetadataFact :=
+  { anchor := anchor
+    dependents := #[]
+    payload := { postOrderIndex := {} } }
+
+def propagate (_fact : RegionMetadataFact) (dfCtx : DataFlowContext)
+    (_irCtx : IRContext OpCode) : DataFlowContext :=
+  dfCtx
+
+instance : FactSpec .regionMetadata where
+  mkDefault := RegionMetadataFact.mkDefault
+  propagate := RegionMetadataFact.propagate
+
+end RegionMetadataFact
+
+namespace DominanceAnalysis
+
+def kind : AnalysisKind :=
+  .dominance
+
+/--
+The returned array is CFG in postorder, and the map assigns each block a
+postorder index used by `intersect`.
+-/
+private def collectPostOrder
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr × HashMap BlockPtr Nat := Id.run do
+  let mut postOrder : Array BlockPtr := #[]
+  let mut postOrderIndex : HashMap BlockPtr Nat := {}
+  if let some entry := (region.get! irCtx).firstBlock then
+    let mut worklist : Array (BlockPtr × Bool) := #[(entry, false)]
+    let mut seen : HashSet BlockPtr := ∅
+
+    while !worklist.isEmpty do
+      let (block, visited) := worklist.back!
+      worklist := worklist.pop
+
+      if visited then
+        postOrder := postOrder.push block
+        postOrderIndex := postOrderIndex.insert block postOrder.size
+      else if seen.contains block then
+        continue
+      else
+        seen := seen.insert block
+        worklist := worklist.push (block, true)
+
+        if let some terminator := (block.get! irCtx).lastOp then
+          for succ in terminator.getSuccessors! irCtx do
+            if !seen.contains succ then
+              worklist := worklist.push (succ, false)
+  (postOrder, postOrderIndex)
+
+/-- Initialize the dominators in post order and enqueue them in reverse post order. -/
+private def initializeRegion
+    (region : RegionPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  let mut dfCtx := dfCtx
+  if let some entry := (region.get! irCtx).firstBlock then
+    let (postOrder, postOrderIndex) := collectPostOrder region irCtx
+    let reversePostOrder := postOrder.reverse
+    dfCtx :=
+      dfCtx.modifyFact .regionMetadata (InsertPoint.atStart! entry irCtx) fun fact =>
+        fact.setPostOrderIndex postOrderIndex
+
+    for block in postOrder do
+      let mut dependents := #[]
+      if let some terminator := (block.get! irCtx).lastOp then
+        for succ in terminator.getSuccessors! irCtx do
+          dependents := dependents.push (InsertPoint.atStart! succ irCtx, kind)
+      dfCtx := dfCtx.modifyFact .dominator (InsertPoint.atStart! block irCtx) fun fact =>
+        (fact.setDependents dependents).setIDom
+          (if block = entry then some entry else none)
+
+    for block in reversePostOrder do
+      dfCtx := dfCtx.enqueue (InsertPoint.atStart! block irCtx, kind)
+  dfCtx
+
+/-- Recursively initialize the analysis on nested regions. -/
+partial def initializeRecursively
+    (op : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  let mut dfCtx := dfCtx
+
+  for region in (op.get! irCtx).regions do
+    dfCtx := initializeRegion region dfCtx irCtx
+
+    let mut currentBlock := (region.get! irCtx).firstBlock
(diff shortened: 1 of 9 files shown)


Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,218 @@
...
+namespace Veir
+public section
+
+structure DomTreeNodePtr where
+  region: RegionPtr
+  index: Nat
+deriving Inhabited, Repr, DecidableEq, Hashable
+
+structure DomTreeNode where
+  -- The block in question
```
ineol: You can use `/-- -/` to write doc comments.

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,218 @@
...
+  sibling : Option DomTreeNodePtr
+  deriving Inhabited, Repr
+
+abbrev DomTree := Array DomTreeNode
+abbrev DomContext := HashMap RegionPtr DomTree 
+
+instance : Repr DomContext where
+  reprPrec ctx prec := reprPrec (ctx.toList) prec
+
+def DomTreeNode.new (block : BlockPtr) (iDom: Option DomTreeNodePtr) : DomTreeNode :=
```
ineol:
```suggestion
def DomTreeNode.new (block : BlockPtr) (iDom : Option DomTreeNodePtr) : DomTreeNode :=
```

ineol: Also in lots of other places :)

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,218 @@
...
+def DomContext.updateNode! (ptr: DomTreeNodePtr) (f : DomTreeNode -> DomTreeNode) (ctx: DomContext) : DomContext :=
+  let tree := (ptr.getDomTree! ctx)
+  let tree' := tree.set! ptr.index (f (ptr.get! ctx))
+  ctx.insert ptr.region tree'
+
+def DomTreeNodePtr.addChild! (parent child: DomTreeNodePtr) (ctx: DomContext) : DomContext := 
+  let parentNode := (parent.get! ctx) 
+
+  if (child.getSibling! ctx).isSome then 
+    panic! "cannot add child that already has siblings"
```
ineol: Would it be better to return `none`. @math-fehr do you have an opinion?

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,218 @@
...
+  else if !(parent.isLeaf! ctx) && 
+          (parent.getLastChildSibling! ctx).isSome then 
+    panic! "sibling of last child must be none" 
+  else
+    match parentNode.lastChild with
+    | none => (ctx.updateNode! parent) (fun n => { n with firstChild := some child, lastChild := some child })
```
ineol: In Lean a trailing function parameter does not need parens. So I think you can remove the parens around `fun n => ...`

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,218 @@
...
+    | some last => 
+      let ctx := (ctx.updateNode! last) (fun n => { n with sibling := some child })
+      (ctx.updateNode! parent) (fun parentNode => { parentNode with lastChild := some child })
+
+def DomTreeNodePtr.removeChild! (parent child: DomTreeNodePtr) (ctx: DomContext) : DomContext := Id.run do
+  let childSibling := (child.getSibling! ctx)
+  let parentLast := (parent.getLastChild! ctx)
+  let parentFirst := (parent.getFirstChild! ctx)
```
ineol: The parens are not needed

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,218 @@
...
+
+  -- Check invariants
+  if !(parent.isLeaf! ctx) && 
+     (parent.getLastChildSibling! ctx).isSome then 
+    panic! "sibling of last child must be none" 
+  if childSibling.isNone ≠ (parentLast == some child) then
+      panic! "parent's last child is not the same as the last sibling"  
+  
+  -- Special case: child being removed is first child in sibling list
+  if parentFirst == child then
```
ineol: Usually, it's more canonical to use `=` instead of `==` in Lean

ineol (commented):
Thank you!

Maybe we should consider splitting the dominance tree off of the PR to keep it manageable?

Veir/IR/Dominance.lean
axelcool1234: Done! Thanks for the suggestion.

Veir/IR/Dominance.lean
axelcool1234: Done! I had no idea this was allowed! Thank you.

Veir/IR/Dominance.lean
axelcool1234: Thanks

Veir/IR/Dominance.lean
axelcool1234: Done. What are some of the benefits of giving `if ... then ... else ...` a `Prop` instead of a `Bool`?

ineol: I think it's more idiomatic, and if you use `==` you then need to use a lemma that says that `x == y` implies `x = y` when it comes to do reasoning, though it's mostly done automatically when using `simp` or `grind`.

Veir/IR/Dominance.lean
math-fehr: Discussed it with Leo, and that should be fine because this is never supposed to be executed (only if there is a bug in the algorithm).

axelcool1234: That was the intention, yeah. Ideally this version of the tree won't be used long term (replaced by a verified version).

Veir/IR/Dominance.lean
ineol: @axelcool1234 please switch to doc comments, also for the other fields.

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+  sibling : Option DomTreeNodePtr
+  deriving Inhabited, Repr
+
+abbrev DomTree := Array DomTreeNode
+abbrev DomContext := HashMap RegionPtr DomTree 
+
+instance : Repr DomContext where
+  reprPrec ctx prec := reprPrec (ctx.toList) prec
+
+def DomTreeNode.new (block : BlockPtr) (iDom : Option DomTreeNodePtr) : DomTreeNode :=
```
ineol: Since it's a pure function I would call it `empty` or somehting similar

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+  block := block
+  iDom := iDom 
+  firstChild := none
+  lastChild := none
+  sibling := none
+}
+
+def RegionPtr.getDomTree! (ptr : RegionPtr) (ctx : DomContext) : DomTree := ctx[ptr]!
+
+def RegionPtr.newDomTreeNode! (ptr : RegionPtr) (block : BlockPtr) (ctx : DomContext) : DomContext :=
```
ineol: Consider using `HashMap.modify` instead to make sure values are used linearly. I'm afraid this will have to copy the tree associated to `ptr`.

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+def DomTreeNodePtr.getFirstChild! (ptr : DomTreeNodePtr) (ctx : DomContext) : Option DomTreeNodePtr :=
+  (ptr.get! ctx).firstChild
+
+def DomTreeNodePtr.getLastChild! (ptr : DomTreeNodePtr) (ctx : DomContext) : Option DomTreeNodePtr :=
+  (ptr.get! ctx).lastChild
+
+def DomTreeNodePtr.getSibling! (ptr : DomTreeNodePtr) (ctx : DomContext) : Option DomTreeNodePtr :=
+  (ptr.get! ctx).sibling
+
+def DomTreeNodePtr.getLastChildSibling! (ptr : DomTreeNodePtr) (ctx : DomContext) : Option DomTreeNodePtr :=
```
ineol: I'm confused. According to the invariant it should always return `none`?

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+  (ptr.getLastChild! ctx).get!.getSibling! ctx
+
+def DomTreeNodePtr.isLeaf! (ptr : DomTreeNodePtr) (ctx : DomContext) : Bool :=
+  (ptr.getFirstChild! ctx).isNone
+
+def DomContext.updateNode! (ptr : DomTreeNodePtr) (f : DomTreeNode -> DomTreeNode) (ctx : DomContext) : DomContext :=
```
ineol: Also use `modify` here? also consider calling the function `modifyNode` if it's similar to the standard library's `update` functions

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+  
+  -- Special case: child being removed is first child in sibling list
+  if parentFirst = child then
+    if parentFirst = parentLast then
+      ctx.updateNode! parent fun n => { n with firstChild := none, lastChild := none }
+    else
+      let ctx := ctx.updateNode! parent fun n => { n with firstChild := childSibling }
+      ctx.updateNode! child fun n => { n with sibling := none } 
+  else -- Iterate
+    let prev := parentFirst
```
ineol: This definition seems useless since it's not mutable?

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+      if iDom = newIDom then 
+        return ctx
+      else
+        let ctx := iDom.removeChild! ptr ctx
+        let ctx := DomContext.updateNode! ptr (fun n => { n with iDom := some newIDom }) ctx
+        newIDom.addChild! ptr ctx
+
+-- Uses the Cooper Harvey Kennedy algorithm
+def RegionPtr.computeDomTree! (ptr : RegionPtr) (domCtx : DomContext) (irCtx : IRContext OpInfo) : DomContext := Id.run do
+  let intersect (block1 : BlockPtr) (block2 : BlockPtr) (idx : HashMap BlockPtr DomTreeNodePtr) (domCtx : DomContext) : BlockPtr := Id.run do
```
ineol:
```suggestion
  let intersect (block1 block2 : BlockPtr) (idx : HashMap BlockPtr DomTreeNodePtr) (domCtx : DomContext) : BlockPtr := Id.run do 
```
you can combine args with the same type.

ineol: why is it an inner function?

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+    let mut finger1 := idx[block1]!
+    let mut finger2 := idx[block2]!
+    while finger1 != finger2 do
+      while finger1.index < finger2.index do
+        finger1 := (finger1.getIDom! domCtx).get! 
+      while finger2.index < finger1.index do  
+        finger2 := (finger2.getIDom! domCtx).get! 
+    return (finger1.getBlock! domCtx)
+
+  -- Postorder traversal of blocks, insert into DomTree (which is just an array!)
```
ineol: Would it make sense to split this into a function and document what it does? this function is really large

Veir/IR/Dominance.lean
```diff
@@ -0,0 +1,220 @@
...
+
+namespace Veir
+public section
+
+variable {OpInfo : Type} [HasOpInfo OpInfo]
+
+structure DomTreeNodePtr where
+  region : RegionPtr
+  index : Nat
+deriving Inhabited, Repr, DecidableEq, Hashable
```
math-fehr: What does the `index` correspond to here?

regehr:
@axelcool1234 do you want to rename UnitTest/DataFlowFramework.lean to be something like UnitTest/DataFlowFramework/Dominance.lean?

or wait, it looks like there's some common code... but maybe take the new dominance tests and put them into a separate file?

axelcool1234:
> @axelcool1234 do you want to rename UnitTest/DataFlowFramework.lean to be something like UnitTest/DataFlowFramework/Dominance.lean?
> 
> or wait, it looks like there's some common code... but maybe take the new dominance tests and put them into a separate file?

Sounds good, I can have a folder for tests. I'll have a common file each analysis test file imports from. I was in fact just throwing everything on top of this one file which can easily become infeasible.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+
+def getDominatorFact? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option DominatorFact :=
+  dfCtx.getFact? .dominator (.InsertPoint (InsertPoint.atStart! block irCtx))
+
+def getIDom? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  match block.getDominatorFact? dfCtx irCtx with
+  | some fact => fact.iDom
+  | none => none
```
ineol:
```suggestion
  block.getDominatorFact? dfCtx irCtx |>.map (·.iDom)
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+      while current.isSome do
+        let dom := current.get!
+        if dom == block then
+          return some doms
+        doms := doms.insert dom
+        let next := dom.getIDom? dfCtx irCtx
+        if next == some dom then
+          return some doms
+        current := next
+      some doms
```
ineol:
```suggestion
def getDoms? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
    (irCtx : IRContext OpCode) : Option (HashSet BlockPtr) := do
  let fact ← block.getDominatorFact? dfCtx irCtx
  let mut doms : HashSet BlockPtr := (∅ : HashSet BlockPtr).insert block
  let mut current := fact.iDom
  while current.isSome do
    let dom := current.get!
    if dom = block then
      return some doms
    doms := doms.insert dom
    let next := dom.getIDom? dfCtx irCtx
    if next = some dom then
      return some doms
    current := next
  pure doms
```
In Lean it's more idiomatic to compare for equality with `=` rather than `==` unless you have a good reason :)

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+
+instance : FactSpec .dominator where
+  mkDefault := DominatorFact.mkDefault
+  propagate := DominatorFact.propagate
+
+/-- Get the immediate dominator for a block. -/
+def getIDom? (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  block.getIDom? dfCtx irCtx
+
```
ineol: Why is this necessary? isn't it basically equivalent to
```lean
def getIDom? := BlockPtr.getIDom?
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+/-- Walk up the dominator tree via iDoms to gather the dominators for a block. -/
+def getDoms? (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option (HashSet BlockPtr) :=
+  block.getDoms? dfCtx irCtx
```
ineol: same question as above

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+
+namespace RegionPtr
+
+def getRegionMetadataFact? [FactSpec .regionMetadata] (region : RegionPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option RegionMetadataFact :=
+  match (region.get! irCtx).firstBlock with
+  | some entry =>
+    dfCtx.getFact? .regionMetadata (.InsertPoint (InsertPoint.atStart! entry irCtx))
+  | none =>
+    none
```
ineol:
```suggestion
def getRegionMetadataFact? [FactSpec .regionMetadata] (region : RegionPtr) (dfCtx : DataFlowContext)
    (irCtx : IRContext OpCode) : Option RegionMetadataFact := do
  let entry ← (region.get! irCtx).firstBlock
  dfCtx.getFact? .regionMetadata (.InsertPoint (InsertPoint.atStart! entry irCtx))
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+  .InsertPoint (InsertPoint.atStart! block irCtx)
+
+/-- Get the successors of a given block. -/
+private def successorsInRegion
+    (block : BlockPtr)
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr :=
+  match (block.get! irCtx).lastOp with
+  | none =>
+    #[]
```
ineol:
```suggestion
  | none => #[]
```
of use `if let`?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+  | some term =>
+    Id.run do
```
ineol:
```suggestion
  | some term => Id.run do
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+      let mut succs := #[]
+      for i in [0:term.getNumSuccessors! irCtx] do
+        let succ := term.getSuccessor! irCtx i
+        if (succ.get! irCtx).parent == some region then
+          succs := succs.push succ
```
ineol:
```suggestion
        if (succ.get! irCtx).parent = some region then
          succs := succs.push succ
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+/--
+The returned array is CFG in postorder, and the map assigns each block a
+postorder index used by `intersect`.
+-/
+private def collectPostOrder
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr × HashMap BlockPtr Nat :=
+  match (region.get! irCtx).firstBlock with
+  | none =>
+    (#[], {})
```
ineol:
```suggestion
  | none => (#[], {})
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+  | some entry =>
+    Id.run do
```
ineol:
```suggestion
  | some entry => Id.run do
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+      let mut postOrderIndex : HashMap BlockPtr Nat := {}
+
+      while !worklist.isEmpty do
+        let (block, visited) := worklist.back!
+        worklist := worklist.pop
+
+        if visited then
+          let index := postOrder.size
+          postOrder := postOrder.push block
+          postOrderIndex := postOrderIndex.insert block index
```
ineol:
```suggestion
          postOrder := postOrder.push block
          postOrderIndex := postOrderIndex.insert block postOrder.size
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+        else
+          seen := seen.insert block
+          worklist := worklist.push (block, true)
+
+          for succ in successorsInRegion block region irCtx do
+            if !seen.contains succ then
+              worklist := worklist.push (succ, false)
+      (postOrder, postOrderIndex)
+
+/- Initialize the dominators in post order and enqueue them in reverse post order. -/
```
ineol: If it's a doc comment, use `/-- -/` instead :)

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+private def initializeRegion
+    (region : RegionPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  match (region.get! irCtx).firstBlock with
+  | none =>
+    dfCtx
+  | some entry =>
+    Id.run do
```
ineol:
```suggestion
  | none => dfCtx
  | some entry => Id.run do
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+      let (postOrder, postOrderIndex) := collectPostOrder region irCtx
+      let reversePostOrder := postOrder.reverse
+      let mut dfCtx :=
+        dfCtx.modifyFact .regionMetadata (blockEntryAnchor entry irCtx) (fun fact =>
+          fact.setPostOrderIndex postOrderIndex)
```
ineol:
```suggestion
      let mut dfCtx :=
        dfCtx.modifyFact .regionMetadata (blockEntryAnchor entry irCtx) fun fact =>
          fact.setPostOrderIndex postOrderIndex
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+private def computeImmediateDominator
+    (block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  let region := ((block.get! irCtx).parent).get!
+  let entry := ((region.get! irCtx).firstBlock).get!
+  match region.getRegionMetadataFact? dfCtx irCtx with
+  | none =>
+    none
+  | some metadata =>
```
ineol: you can replace by a monadic bind (see above)

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+    if block == entry then
```
ineol:
```suggestion
    if block = entry then
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+      some entry
+    else
+      Id.run do
```
ineol: This would not be necessary if the whole function is in the `Option` monad

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+        let mut newIDom : Option BlockPtr := none
+
+        while hUse : maybePredUse.isSome do
+          let predUse := maybePredUse.get hUse
+          let predUseStruct := predUse.get! irCtx
+          maybePredUse := predUseStruct.nextUse
+          let predOp := predUseStruct.owner
+          let some predBlock := (predOp.get! irCtx).parent
+            | continue
+          if (predBlock.get! irCtx).parent != some region then
```
ineol: Use `\neq` instead

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+        (fact.setIDom (some newIDom), some newIDom != fact.iDom)) irCtx
+
+/-- Get a block's immediate dominator. -/
+def immediateDominator?
+    (block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  match block.getIDom? dfCtx irCtx with
+  | some iDom => if iDom == block then none else some iDom
+  | none => none
```
ineol: monadic bind would look a bit nicer I think

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+
+/-- Recurse up the dominator tree to see if `dominator` dominates `block`. -/
+def dominates
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  if dominator == block then
```
ineol:
```suggestion
  if dominator = block then
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+    (block.getDominatorFact? dfCtx irCtx).isSome
+  else
+    match block.getDominatorFact? dfCtx irCtx with
+    | none => false
+    | some fact =>
+      Id.run do
```
ineol:
```suggestion
    | some fact => Id.run do
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+        let mut current := fact.iDom
+        while current.isSome do
+          let candidate := current.get!
+          if candidate == dominator then
+            return true
+          let next := candidate.getIDom? dfCtx irCtx
+          if next == some candidate then
+            return false
```
ineol: `s/==/=/`

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,353 @@
...
+          current := next
+        false
+
+/-- Same thing as `dominates` except a block dominating itself doesn't count. -/
+def strictlyDominates
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  dominator != block && dominates dominator block dfCtx irCtx
```
ineol: \neq

ineol (commented):
Thank you for rebasing!

I made a superficial review of the style, I haven't looked at the correctness yet.

Could you write a high level explanation of the algorithm/abstract domains etc at the beginning of `DominanceAnalysis.lean`?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
axelcool1234: So this doesn't work, because `(·.iDom)` is an error since `getDominatorFact?` returns `Option DominatorFact`. Should I change the `match` to something else though?

axelcool1234: For example, should I use `do` and do something like:
```
let fact <- block.getDominatorFact? dfCtx irCtx
pure fact.iDom
```

axelcool1234: I figured out why this didn't work actually. `Fact.iDom` returns an `Option BlockPtr`, which means this function, which has a return type of `Option BlockPtr`, doesn't match the `Option.map` call return `Option (Option BlockPtr)`. I need to flatten the `Option`s. So I need to use `bind` instead.

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -67,22 +52,40 @@ deriving BEq, Hashable
...
+/--
+The immediate dominator fact attached to a block entry.
+-/
+structure DominatorPayload where
+  iDom : Option BlockPtr := none
+
+/--
+Hack to cache the post ordering of a region's blocks.
+
+Stored in the entry block of each region.
```
math-fehr: I would remove the "Hack" part here to be honest.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+
+def getDominatorFact? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option DominatorFact :=
+  dfCtx.getFact? .dominator (.InsertPoint (InsertPoint.atStart! block irCtx))
+
+def getIDom? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  block.getDominatorFact? dfCtx irCtx >>= (·.iDom)
+
+def getDoms? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
```
math-fehr, line 74: This function is not used anywhere, can you add a test for it?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+    dfCtx
+  else
+    let block := (point.block! irCtx).get!
+    computeImmediateDominator block dfCtx irCtx |> (Option.elim · dfCtx fun newIDom =>
+      let anchor := blockEntryAnchor block irCtx
+      dfCtx.modifyFactAndPropagate .dominator anchor (fun fact =>
+       (fact.setIDom (some newIDom), some newIDom ≠ fact.iDom)) irCtx)
+
+/-- Get a block's immediate dominator. -/
+def immediateDominator?
```
math-fehr: This function is not used anywhere, can you add a test for it?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+            return false
+          current := next
+        false))
+
+/-- Same thing as `dominates` except a block dominating itself doesn't count. -/
+def strictlyDominates
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  dominator ≠ block && dominates dominator block dfCtx irCtx
```
math-fehr, line 330: These two functions are not used anywhere, can you add a test for them?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+  let mut doms := (∅ : HashSet BlockPtr).insert block
+  let mut current := fact.iDom
+  while current.isSome do
+    let dom := current.get!
+    if dom = block then
+      return doms
+    doms := doms.insert dom
+    let next := dom.getIDom? dfCtx irCtx
+    if next = some dom then
+      return doms
```
math-fehr, line 83: This can be simplified by:
```
let some next := dom.getIDom? dfCtx irCtx | return doms
```

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+      (Id.run do
+        if dominator = block then 
+          return true
+        let mut current := ·.iDom
+        while let some candidate := current do
+          if candidate = dominator then
+            return true
+          let next := candidate.getIDom? dfCtx irCtx
+          if next = some candidate then
+            return false
```
math-fehr, line 321: Same here, you can use `let some next := ... | return false`

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+end RegionMetadataFact
+
+namespace DominanceAnalysis
+
+def kind : AnalysisKind :=
+  .dominance
+
+/-- Get the lattice anchor of a given block. -/
+private def blockEntryAnchor (block : BlockPtr) (irCtx : IRContext OpCode) : LatticeAnchor :=
+  .InsertPoint (InsertPoint.atStart! block irCtx)
```
math-fehr: Instead of returning `LatticeAnchor`, can you instead return the `InsertPoint` directly, and use coercions to go from `InsertPoint` to `LatticeAnchor`? That would prevent you from using `unreachable!` in `initializeRegion`.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+  doms
+
+end BlockPtr
+
+namespace RegionPtr
+
+def getRegionMetadataFact? [FactSpec .regionMetadata] (region : RegionPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option RegionMetadataFact :=
+  (region.get! irCtx).firstBlock >>= 
+    dfCtx.getFact? .regionMetadata ∘ (.InsertPoint <| InsertPoint.atStart! · irCtx)
```
math-fehr: Should you use `blockEntryAnchor` here directly?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+private def successorsInRegion
+    (block : BlockPtr)
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr := Id.run do
+  let mut succs := #[]
+  if let some term := (block.get! irCtx).lastOp then
+    for i in [0:term.getNumSuccessors! irCtx] do
+      let succ := term.getSuccessor! irCtx i
+      if (succ.get! irCtx).parent = some region then
+        succs := succs.push succ
```
math-fehr: You can assume this `if` to always be `True`. Successors of verified programs should always point to blocks in the same region.
In general, this function should probably moved to the `BlockPtr` namespace, in `IR/Basic.lean`.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+      dfCtx := dfCtx.modifyFact .dominator (blockEntryAnchor block irCtx) fun fact =>
+        (fact.setDependents dependents).setIDom
+          (if block = entry then some entry else none)
+
+    for block in reversePostOrder do
+      match blockEntryAnchor block irCtx with
+      | .InsertPoint point =>
+        dfCtx := dfCtx.enqueue (point, kind)
+      | _ =>
+        unreachable!
```
math-fehr: This is why I'm suggesting to change `blockEntryAnchor`

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+
+  while hUse : maybePredUse.isSome do
+    let predUse := maybePredUse.get hUse
+    let predUseStruct := predUse.get! irCtx
+    maybePredUse := predUseStruct.nextUse
+    let predOp := predUseStruct.owner
+    let some predBlock := (predOp.get! irCtx).parent
+      | continue
+    if (predBlock.get! irCtx).parent ≠ some region then
+      continue
```
math-fehr: You can remove these check as well, this should never be `True` in a verified context. If the context is not verified, then computing dominance doesn't make sense anyway.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+      (intersect predBlock · metadata.postOrderIndex dfCtx irCtx)
+
+  newIDom
+
+def visit
+    (point : InsertPoint)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  if point.prev! irCtx ≠ none then
+    dfCtx
```
math-fehr, line 296: Would it be an error if this branch is taken? If so, can you add a comment so we remember this?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+    if (predBlock.get! irCtx).parent ≠ some region then
+      continue
+    let some _ := predBlock.getIDom? dfCtx irCtx
+      | continue
+    newIDom := newIDom.elim predBlock 
+      (intersect predBlock · metadata.postOrderIndex dfCtx irCtx)
+
+  newIDom
+
+def visit
```
math-fehr, line 290: Can you add documentation to explain what this does?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  if point.prev! irCtx ≠ none then
+    dfCtx
+  else
+    let block := (point.block! irCtx).get!
+    computeImmediateDominator block dfCtx irCtx |> (Option.elim · dfCtx fun newIDom =>
+      let anchor := blockEntryAnchor block irCtx
+      dfCtx.modifyFactAndPropagate .dominator anchor (fun fact =>
+       (fact.setIDom (some newIDom), some newIDom ≠ fact.iDom)) irCtx)
```
math-fehr: Instead of using `Option.elim`, can you just `match` on the result of `computerImmediateDominator`? I feel this makes it much more readable

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,334 @@
...
+
+/-- Get a block's immediate dominator. -/
```
math-fehr: Can you document why this is different than `getIDom?`

math-fehr (commented):
Nice! I added some comments, but the design seems good otherwise.

In the tests, can you add one where there is a block that is "disconnected" from any other block. So a block that doesn't have any predecessor. Similarly, a group of blocks that are not dominated by the entry block, because there is no path to them.

Otherwise, I think we should only have the "analysis" part in `DominanceAnalysis`, and move some of the useful API outside of it in a `IR/Dominance.lean` file. So you would still need to use the dominance analysis, but the public API functions would be stored there to clearly separate what is useful for computing the analysis, versus what is useful for querying which block dominates which block.

Finally, your notion of dominance does not take into account dominance between regions. Do you know if MLIR uses any other data structure to compute that, and if so, can these be added on top of these ones in the future, or are they incompatible?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
axelcool1234: I actually think this would be wrong. I'm checking if the current block, assigned to `dom` via `current.get!` is equal to `next`. I should remove `some` from `if next = some dom` though because Lean will coerce, I believe.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
axelcool1234: Ideally a verified fixpoint loop would never lead to `DominanceAnalysis`'s `visit` call to be done on anything but the entry to a block. I'll write a comment.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+    (irCtx : IRContext OpCode) : Option DominatorFact :=
+  dfCtx.getFact? .dominator (.InsertPoint (InsertPoint.atStart! block irCtx))
+
+def getIDom? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  block.getDominatorFact? dfCtx irCtx >>= (·.iDom)
+
+def getDoms? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option (HashSet BlockPtr) := do
+  let fact <- block.getDominatorFact? dfCtx irCtx
```
ineol: Please use unicode left arrow (also below)

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+  let mut doms := (∅ : HashSet BlockPtr).insert block
+  let mut current := fact.iDom
+  while let some dom := current do
+    if dom = block then
+      return doms
```
ineol: can this happen after the 1st iteration?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
ineol: Is it only meant for testing / debugging?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+/--
+The returned array is CFG in postorder, and the map assigns each block a
+postorder index used by `intersect`.
+-/
+private def collectPostOrder
+    (region : RegionPtr)
+    (irCtx : IRContext OpCode) : Array BlockPtr × HashMap BlockPtr Nat := Id.run do
+  let mut postOrder : Array BlockPtr := #[]
+  let mut postOrderIndex : HashMap BlockPtr Nat := {}
+  if let some entry := (region.get! irCtx).firstBlock then
```
ineol: a fallible let would be nicer I think, so that the whole function is not indented

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+    let mut worklist : Array (BlockPtr × Bool) := #[(entry, false)]
```
ineol: I would call it a stack, since the order is relevant to the algorithm

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+
+    for block in postOrder do
+      let mut dependents := #[]
+      if let some terminator := (block.get! irCtx).lastOp then
+        for succ in terminator.getSuccessors! irCtx do
+          dependents := dependents.push (InsertPoint.atStart! succ irCtx, kind)
+      dfCtx := dfCtx.modifyFact .dominator (InsertPoint.atStart! block irCtx) fun fact =>
+        (fact.setDependents dependents).setIDom
+          (if block = entry then some entry else none)
+
```
ineol: Does the order of initialisation matter? if not it can be combined with the next loop

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+pick the first one whose dominator fact has already been computed, and then
+repeatedly `intersect` that candidate with each other processed predecessor.
+-/
+private def computeImmediateDominator
+    (block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr := do
+  let region := ((block.get! irCtx).parent).get!
+  let entry := ((region.get! irCtx).firstBlock).get!
+  let metadata <- region.getRegionMetadataFact? dfCtx irCtx
```
ineol: unicode

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+  while let some predUse := currentPredUse do
+    let predUseStruct := predUse.get! irCtx
+    currentPredUse := predUseStruct.nextUse
+    let predOp := predUseStruct.owner
+    let some predBlock := (predOp.get! irCtx).parent
+      | continue
+    let some _ := predBlock.getIDom? dfCtx irCtx
+      | continue
+    newIDom := newIDom.elim predBlock 
+      (intersect predBlock · metadata.postOrderIndex dfCtx irCtx)
```
ineol: I don't think we should use `Option.elim` in programs, it's a bit confusing... can you rewrite with monadic code?

ineol (commented):
It's starting to look good!

Veir/Analysis/DataFlow/DominanceAnalysis.lean
axelcool1234: It'll be used internally by a useful API that Mathieu said should be written in IR/Dominance.lean

axelcool1234: Actually I'll probably move that function to `IR/Dominance.lean` tonight.

regehr: at present there are some calls to this function from `UnitTest/DataFlowFramework/Dominance.lean` so I think this is fine?

I don't have an actual use case for this function in mind, but it seems like something that someone might want at some point, and it's less than 15 lines of code, so I'd suggest that we keep it

Veir/Analysis/DataFlow/DominanceAnalysis.lean
regehr: meanwhile claude believes it can be replaced by this
```
current := (dom.getIDom? dfCtx irCtx).filter (· ≠ dom)
```

axelcool1234:
Thanks @regehr. I'll cover everything else within the next couple hours.

Veir/Analysis/DataFlow/DominanceAnalysis.lean
math-fehr: Sorry, my bad, let's keep it as it is (I find this more readable than the claude version).

axelcool1234: > Sorry, my bad, let's keep it as it is (I find this more readable than the claude version).

No worries, the tests didn't use it at first and I changed that. Also I think the claude version is wrong too

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+def getDominatorFact? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option DominatorFact :=
+  dfCtx.getFact? .dominator (.InsertPoint (InsertPoint.atStart! block irCtx))
+
+def getIDom? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option BlockPtr :=
+  block.getDominatorFact? dfCtx irCtx >>= (·.iDom)
+
+def getDoms? [FactSpec .dominator] (block : BlockPtr) (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Option (HashSet BlockPtr) := do
```
math-fehr, line 75: Can you document these functions?

Veir/Analysis/DataFlow/DominanceAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+    if next = some dom then
+      return doms
+    current := next
+  doms
+
+end BlockPtr
+
+namespace RegionPtr
+
+def getRegionMetadataFact? [FactSpec .regionMetadata] (region : RegionPtr) (dfCtx : DataFlowContext)
```
math-fehr, line 97: Same here

math-fehr (commented):
Thanks! I think there is just a few functions that need more documentation, but otherwise that's good to me!

math-fehr (approved):

PR #134, opened 2026-01-21
Add `BlockPtr.setArguments` and `BlockPtr.pushArgument`

math-fehr:
Both methods were missing. Their `GetSet` lemmas have been added as well.

### DIFF
--- a/Veir/IR/Basic.lean
+++ b/Veir/IR/Basic.lean
@@ -778,6 +778,15 @@ theorem BlockPtr.getArgument_block {block : BlockPtr} {index : Nat} :
     (BlockPtr.getArgument block index).block = block := by
   grind [getArgument]
 
+def BlockPtr.setArguments (block: BlockPtr) (ctx: IRContext)
+    (newArguments: Array BlockArgument) (inBounds: block.InBounds ctx := by grind) : IRContext :=
+  let oldBlock := block.get ctx (by grind)
+  block.set ctx { oldBlock with arguments := newArguments }
+
+def BlockPtr.pushArgument (block : BlockPtr) (ctx : IRContext) (result : BlockArgument)
+      (inBounds : block.InBounds ctx := by grind) :=
+    block.setArguments ctx ((block.get ctx).arguments.push result)
+
 /-
  BlockArgumentPtr accessors
 -/
@@ -1254,7 +1263,7 @@ macro "setup_grind_with_get_set_definitions" : command => `(
   attribute [local grind] BlockArgumentPtr.get! BlockArgumentPtr.setFirstUse BlockArgumentPtr.set BlockArgumentPtr.setType BlockArgumentPtr.setLoc
   attribute [local grind] OperationPtr.setOperands OperationPtr.setResults OperationPtr.pushResult OperationPtr.setRegions OperationPtr.setProperties  OperationPtr.pushOperand OperationPtr.allocEmpty OperationPtr.dealloc OperationPtr.setNextOp OperationPtr.setPrevOp OperationPtr.setParent OperationPtr.getNumResults! OperationPtr.getNumOperands! OperationPtr.getNumRegions! OperationPtr.getRegion! OperationPtr.getNumSuccessors! OperationPtr.set
   attribute [local grind] Operation.empty
-  attribute [local grind] BlockPtr.get! BlockPtr.setParent BlockPtr.setFirstUse BlockPtr.setFirstOp BlockPtr.setLastOp BlockPtr.setNextBlock BlockPtr.setPrevBlock BlockPtr.allocEmpty Block.empty BlockPtr.getNumArguments! BlockPtr.set
+  attribute [local grind] BlockPtr.get! BlockPtr.setParent BlockPtr.setFirstUse BlockPtr.setFirstOp BlockPtr.setLastOp BlockPtr.setNextBlock BlockPtr.setPrevBlock BlockPtr.allocEmpty Block.empty BlockPtr.getNumArguments! BlockPtr.set BlockPtr.setArguments BlockPtr.pushArgument
   attribute [local grind =] Option.maybe_def
   attribute [local grind] OpOperandPtr.get! BlockOperandPtr.get! OpResultPtr.get! BlockArgumentPtr.get! OperationPtr.get!
   attribute [local grind] BlockOperandPtr.setBack BlockOperandPtr.setNextUse BlockOperandPtr.setOwner BlockOperandPtr.setValue BlockOperandPtr.set
--- a/Veir/IR/GetSet.lean
+++ b/Veir/IR/GetSet.lean
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
     opOperandPtr.get! (BlockOperandPtr.setValue operand' ctx newValue hoperand') =
     opOperandPtr.get! ctx := by
   grind
+
+/- BlockPtr.setArguments -/
+
+@[simp, grind =]
+theorem IRContext.topLevelOp_BlockPtr_setArguments :
+    (BlockPtr.setArguments block' ctx newArguments hblock').topLevelOp =
+    ctx.topLevelOp := by
+  grind
+
+@[simp, grind =]
+theorem BlockPtr.firstUse!_BlockPtr_setArguments {block : BlockPtr} :
+    (block.get! (BlockPtr.setArguments block' ctx newArguments hblock')).firstUse =
+    (block.get! ctx).firstUse := by
+  grind
+
+@[simp, grind =]
+theorem BlockPtr.prev!_BlockPtr_setArguments {block : BlockPtr} :
+    (block.get! (BlockPtr.setArguments block' ctx newArguments hblock')).prev =
+    (block.get! ctx).prev := by
+  grind
+
+@[simp, grind =]
+theorem BlockPtr.next!_BlockPtr_setArguments {block : BlockPtr} :
+    (block.get! (BlockPtr.setArguments block' ctx newArguments hblock')).next =
+    (block.get! ctx).next := by
+  grind
+
+@[simp, grind =]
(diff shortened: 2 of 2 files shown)


Veir/IR/Basic.lean
```diff
@@ -778,6 +778,15 @@ theorem BlockPtr.getArgument_block {block : BlockPtr} {index : Nat} :
     (BlockPtr.getArgument block index).block = block := by
   grind [getArgument]
 
+def BlockPtr.setArguments (block: BlockPtr) (ctx: IRContext)
```
luisacicolini, line 781: I have the same concern with a namespace for `BlockPtr` - but that should be refactored separately from this PR

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
     opOperandPtr.get! (BlockOperandPtr.setValue operand' ctx newValue hoperand') =
     opOperandPtr.get! ctx := by
   grind
+
+/- BlockPtr.setArguments -/
```
luisacicolini, line 5977:
```suggestion
/-! ### BlockPtr.setArguments -/
```
This would be more canonical (although I see that the file has a different style, so feel free to ignore)

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+
+@[simp, grind =]
+theorem IRContext.topLevelOp_BlockPtr_setArguments :
```
luisacicolini, line 5980: here again I am wondering if there's a better way to handle namespaces and group lemmas differently? I would be happy to go through the repo organization together and see where things live

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+    blockOperandPtr.get! ctx := by
+  grind
+
+@[grind =]
+theorem BlockPtr.getNumArguments!_BlockPtr_setArguments {block : BlockPtr} :
+    block.getNumArguments! (BlockPtr.setArguments block' ctx newArguments hblock') =
+    if block = block' then
+      newArguments.size
+    else
+      block.getNumArguments! ctx := by
```
luisacicolini, line 6081:
```suggestion
    if block = block' then newArguments.size else block.getNumArguments! ctx := by
```

luisacicolini: I find this more readable, but feel free to ignore

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+  grind
+
+@[grind =]
+theorem BlockArgumentPtr.get!_BlockPtr_setArguments {blockArg : BlockArgumentPtr} :
+    blockArg.get! (BlockPtr.setArguments block' ctx newArguments hblock') =
+    if blockArg.block = block' then
+      newArguments[blockArg.index]!
+    else
+      blockArg.get! ctx := by
```
luisacicolini, line 6090:
```suggestion
    blockArg.get! (BlockPtr.setArguments block' ctx newArguments hblock') =
    if blockArg.block = block' then newArguments[blockArg.index]! else blockArg.get! ctx := by
```
same as above

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+
+@[grind =]
+theorem ValuePtr.getFirstUse!_BlockPtr_setArguments {value : ValuePtr} :
+    value.getFirstUse! (BlockPtr.setArguments block' ctx newArguments hblock') =
+    match value with
+    | .blockArgument arg =>
+      if arg.block = block' then
+        newArguments[arg.index]!.firstUse
+      else
+        value.getFirstUse! ctx
```
luisacicolini, line 6107:
```suggestion
      if arg.block = block' then newArguments[arg.index]!.firstUse else value.getFirstUse! ctx
```
same as above

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+@[grind =]
+theorem ValuePtr.getType!_BlockPtr_setArguments {value : ValuePtr} :
+    value.getType! (BlockPtr.setArguments block' ctx newArguments hblock') =
+    match value with
+    | .blockArgument arg =>
+      if arg.block = block' then
+        newArguments[arg.index]!.type
+      else
+        value.getType! ctx
+    | .opResult _ =>
```
luisacicolini, line 6121: I'm not a huge fan of a `match` statement in the statement - is there a reason why we want this? As an alternative, I would considering splitting the lemma into two, passing an hypothesis over the constructor of value (this is similar to what we do with definitions using induction on naturals: we have a lemma for `_succ` and a lemma for `_zero`).

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+  grind
+
+@[grind =]
+theorem OpOperandPtrPtr.get!_BlockPtr_setArguments {opOperandPtr : OpOperandPtrPtr} :
+    opOperandPtr.get! (BlockPtr.setArguments block' ctx newArguments hblock') =
+    match opOperandPtr with
+    | .valueFirstUse (.blockArgument arg) =>
+      if arg.block = block' then
+        newArguments[arg.index]!.firstUse
+      else
```
luisacicolini, line 6132: same concern as above

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+        opOperandPtr.get! ctx
+    | _ =>
+      opOperandPtr.get! ctx := by
+  grind
+
+/- OperationPtr.pushOperand -/
```
luisacicolini, line 6138:
```suggestion
/-! ### OperationPtr.pushOperand -/
```
same style concern, feel free to ignore

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+    blockOperandPtr.get! ctx := by
+  grind
+
+@[grind =]
+theorem BlockPtr.getNumArguments!_BlockPtr_pushArgument {block : BlockPtr} {hop} :
+    block.getNumArguments! (BlockPtr.pushArgument op ctx newOperand hop) =
+    if block = op then
+      block.getNumArguments! ctx + 1
+    else
+      block.getNumArguments! ctx := by
```
luisacicolini, line 6248:
```suggestion
    if block = op then block.getNumArguments! ctx + 1 else block.getNumArguments! ctx := by
```

Veir/IR/GetSet.lean
```diff
@@ -5973,3 +5973,322 @@ theorem OpOperandPtrPtr.get!_BlockOperandPtr_setValue {opOperandPtr : OpOperandP
...
+@[grind =]
+theorem BlockArgumentPtr.get!_BlockPtr_pushArgument {blockArg : BlockArgumentPtr} :
+    blockArg.get! (BlockPtr.pushArgument block' ctx newArgument hblock') =
+    if blockArg.block = block' then
+      if blockArg.index = (block'.getNumArguments! ctx) then
+        newArgument
+      else
+        blockArg.get! ctx
+    else
+      blockArg.get! ctx := by
```
luisacicolini, line 6260: this seems not super readable to me, I would split it and pass hypotheses (also depends on whether you plan to expose this to users? if it's a theorem you only use rarely in some complex infra it's fine)

ineol: These lemmas (and the one with `match` above) are designed to be used by grind, so we want the LHS to be without hypothesis to be applicable by `grind`. This is supposed to mirror the hashmap lemmas in Lean, where the general lemma with a conditional on the RHS are annotated with grind, and the lemmas with hypothesis are annotated with `simp`. Maybe there are better ways, but that's the thinking behind it.

luisacicolini: ah got it, that makes sense!

Veir/IR/GetSet.lean
tobiasgrosser: I would vote for the multi-line style, as it stands out more when scanning for `if then else` that should not have a `simp` tag.

Veir/IR/GetSet.lean
math-fehr: I'm gonna make a refactoring PR for most files I think, so that we start getting consistent everywhere!

Veir/IR/GetSet.lean
math-fehr: Yes let's do that!
Btw, the reasonning in having the lemmas that way is that it is super easy for me to add new methods.
The file is 6k lines of lean, but adding a new method is essentially copy-pasting one section and then changing the method name in there (and fixing the proofs that break).
Having everything grouped would mean that adding a new method is annoying, and we might miss some methods (which happened in the past)

Veir/IR/GetSet.lean
math-fehr: The main reason we want to keep this lemma is that this is heavily helping `grind`.
When grind sees the `lhs`, it creates the `rhs`, and then `cases` on the match/if.
If we had only the two lemmas, then `grind` wouldn't know to `cases` here.

PR #145, opened 2026-01-24
Add missing invariant in Operation.WellFormed

math-fehr:
When an operation has `none` as a parent, we should ensure that both its previous and next operations are `none` as well. This is to ensure that `prev` and `next` are always in bounds when deleting an operation. The alternative solution would be to not require `prev` and `next` to be in bounds when `parent` is in bounds, but that seems annoying to use.

To make this work, I had to make the following changes:
* Add more arguments to `Operation.WellFormed_changed`, and update the proofs that were using it.
* Update `detachOp` to also set `prev` and `next` to `none`, and thus update the get/set of `detachOp`

### DIFF
--- a/Veir/Rewriter/Basic.lean
+++ b/Veir/Rewriter/Basic.lean
@@ -18,7 +18,7 @@ def Rewriter.insertOp? (ctx: IRContext) (newOp: OperationPtr) (insertionPoint: I
     newOp.linkBetweenWithParent ctx prev next parent (by grind) (by grind) (by grind) (by grind)
 
 /--
-  Set the parent, previous and next operation pointers of an operation to `none`.
+  Set the parent, previous, and next operation pointers of an operation to `none`.
   This method should not be used from outside the rewriter, and is only used to
   make proofs easier for grind.
 -/
--- a/Veir/Rewriter/LinkedList/WellFormed.lean
+++ b/Veir/Rewriter/LinkedList/WellFormed.lean
@@ -688,8 +688,8 @@ theorem RegionPtr.blockChain_OperationPtr_linkBetweenWithParent
 
 theorem Operation.wellFormed_OperationPtr_linkBetweenWithParent
     (ctxInBounds: IRContext.FieldsInBounds ctx)
-    (prevOpParent : prevOp.maybe₁ (fun prev => (prev.get! ctx).parent = some parentBlock) )
-    (nextOpParent : nextOp.maybe₁ (fun next => (next.get! ctx).parent = some parentBlock) )
+    (prevOpParent : prevOp.maybe₁ (fun prev => (prev.get! ctx).parent = some parentBlock))
+    (nextOpParent : nextOp.maybe₁ (fun next => (next.get! ctx).parent = some parentBlock))
     (hctx : op.linkBetweenWithParent ctx prevOp nextOp parentBlock selfIn prevIn nextIn parentIn = some newCtx) :
     (OperationPtr.get! opPtr ctx).WellFormed ctx opPtr opInBounds →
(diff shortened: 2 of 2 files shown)


Veir/Rewriter/LinkedList/WellFormed.lean
```diff
@@ -688,11 +688,18 @@ theorem RegionPtr.blockChain_OperationPtr_linkBetweenWithParent
...
+    (nextOpParent : nextOp.maybe₁ (fun next => (next.get! ctx).parent = some parentBlock) )
     (hctx : op.linkBetweenWithParent ctx prevOp nextOp parentBlock selfIn prevIn nextIn parentIn = some newCtx) :
     (OperationPtr.get! opPtr ctx).WellFormed ctx opPtr opInBounds →
     (OperationPtr.get! opPtr newCtx).WellFormed newCtx opPtr (by grind) := by
-  intros
-  apply Operation.WellFormed_unchanged (ctx := ctx) <;> grind
+  intro wf
+  constructor
+  case region_parent =>
+    -- TODO: why does grind does not work here and require this simp?
```
tobiasgrosser, line 699:
```suggestion
    -- TODO: why does grind not work here and require this simp?
```

Veir/Rewriter/WellFormed/Rewriter/Operation.lean
```diff
@@ -159,8 +163,15 @@ theorem Rewriter.detachOp_WellFormed (ctx : IRContext) (wf : ctx.WellFormed)
   case operations =>
     intros op' hop'
     have : op'.InBounds ctx := by grind
-    have ⟨ha, hb, hc, hd, he, hf, hg⟩ := h₆ op' this
-    apply Operation.WellFormed_unchanged (ctx := ctx) <;> grind
+    have ⟨ha, hb, hc, hd, he, hf, hg, hh⟩ := h₆ op' this
+    constructor
+    case region_parent =>
+      -- TODO: why does grind does not work here and require this simp?
```
tobiasgrosser, line 169:
```suggestion
      -- TODO: why does grind not work here and require this simp?
```

Veir/Rewriter/Basic.lean
```diff
@@ -17,11 +17,34 @@ def Rewriter.insertOp? (ctx: IRContext) (newOp: OperationPtr) (insertionPoint: I
     let next := insertionPoint.next
     newOp.linkBetweenWithParent ctx prev next parent (by grind) (by grind) (by grind) (by grind)
 
+/--
+  Set the parent, previous and next operation pointers of an operation to `none`.
```
tobiasgrosser:
```suggestion
  Set the parent, previous, and next operation pointers of an operation to `none`.
```

Veir/Rewriter/LinkedList/WellFormed.lean
```diff
@@ -688,11 +688,18 @@ theorem RegionPtr.blockChain_OperationPtr_linkBetweenWithParent
 
 theorem Operation.wellFormed_OperationPtr_linkBetweenWithParent
     (ctxInBounds: IRContext.FieldsInBounds ctx)
+    (prevOpParent : prevOp.maybe₁ (fun prev => (prev.get! ctx).parent = some parentBlock) )
+    (nextOpParent : nextOp.maybe₁ (fun next => (next.get! ctx).parent = some parentBlock) )
```
luisacicolini:
```suggestion
    (prevOpParent : prevOp.maybe₁ (fun prev => (prev.get! ctx).parent = some parentBlock))
    (nextOpParent : nextOp.maybe₁ (fun next => (next.get! ctx).parent = some parentBlock))
```

PR #148, opened 2026-01-26
Add well-formedness preservation of `dealloc`

math-fehr:
This follows nicely from the `FieldsInBounds` preservation lemma.

### DIFF
--- a/Veir/IR/DeallocLemmas.lean
+++ b/Veir/IR/DeallocLemmas.lean
@@ -231,4 +231,151 @@ theorem IRContext.fieldsInBounds_OperationPtr_dealloc {ctx : IRContext} {inBound
     · grind
     · grind [IRContext.WellFormed, Operation.WellFormed]
 
+theorem Operation.wellFormed_OperationPtr_dealloc
+    (wf : ctx.WellFormed (Std.ExtHashSet.fromOperands ctx op) (Std.ExtHashSet.fromSuccessors ctx op))
+    (inBounds' : OperationPtr.InBounds opPtr' ctx)
+    (huses : ¬ op.hasUses ctx)
+    (hparent : (op.get! ctx).parent = none)
+    (hregions : op.getNumRegions! ctx = 0)
+    (htoplevelOp : ctx.topLevelOp ≠ op)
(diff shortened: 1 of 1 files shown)


Veir/IR/DeallocLemmas.lean
```diff
@@ -227,4 +227,151 @@ theorem IRContext.fieldsInBounds_OperationPtr_dealloc {ctx : IRContext} {inBound
     · grind
     · grind [IRContext.WellFormed, Operation.WellFormed]
 
+theorem Operation.wellFormed_OperationPtr_dealloc
```
luisacicolini, line 234: one thing that I just noticed: why `OperationPtr` with capital `O`?

Veir/IR/DeallocLemmas.lean
```diff
@@ -227,4 +227,151 @@ theorem IRContext.fieldsInBounds_OperationPtr_dealloc {ctx : IRContext} {inBound
...
+    (wf : ctx.WellFormed (Std.ExtHashSet.fromOperands ctx op) (Std.ExtHashSet.fromSuccessors ctx op))
```
luisacicolini, line 235: does any of the subsequent types rely on `wf`? If not, does it really need to be explicit?

luisacicolini: and the same question for all the theorems below

Veir/IR/DeallocLemmas.lean
ineol: We started like this and it became the predominant style :( cf `Veir/IR/GetSet.lean`. We should batch change it at some point.

Veir/IR/DeallocLemmas.lean
ineol: What do you mean?

luisacicolini: I would put `{wf : ...}` unless the subsequent types rely on it

ineol: I would do the opposite, because if `wf` is never used in the other hyps Lean won't be able to infer it, no?

luisacicolini: my reasoning is: if that's the case you don't need to specify anything about `wf` as any object of that type will work, hence implicit (?) but I trust your intuition better!

PR #163, opened 2026-02-04
Implement OpCodes

math-fehr:
This commit introduces the `OpCode` inductive type to represent operation types in VeIR. Instead
of storing an integer for each operation type, we
instead add to this inductive type a constructor
for each registered operation.

This is likely not the final design, but would allow us to move forward with other parts of the codebase.

### DIFF
--- a/Veir/OpCode.lean
+++ b/Veir/OpCode.lean
@@ -21,42 +21,42 @@ public section
   Each OpCode corresponds to a specific operation.
 -/
 inductive OpCode
-  | unregistered
-  | arith.constant
-  | arith.addi
-  | arith.subi
-  | arith.muli
-  | arith.andi
-  | builtin.module
-  | test.test
+  | arith_constant
+  | arith_addi
+  | arith_subi
+  | arith_muli
+  | arith_andi
+  | builtin_unregistered
+  | builtin_module
+  | test_test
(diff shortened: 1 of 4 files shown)


Veir/OpCode.lean
```diff
@@ -0,0 +1,62 @@
...
+-/
+inductive OpCode
+  | unregistered
+  | arith.constant
+  | arith.addi
+  | arith.subi
+  | arith.muli
+  | arith.andi
+  | builtin.module
+  | test.test
```
math-fehr: How should we write these names?
I used `dialect.name` to make this easy to read, but this prevents Lean from accepting `.arith.constant`, and requires me to add `OpCode` first.
Do you have any suggestions for it?

tobiasgrosser: What about CamelCase? arithMuli, testTest, arithConstant, ...?

luisacicolini: fwiw we had a similar problem with the riscv ISA and called them the same name without dot (e.g. slli.uw became slliuw)

tobiasgrosser: Alternatively, we can use `arith_add`.

math-fehr: I think `arith_add` is probably better than `arithAdd` or `arithadd` yeah

PR #165, opened 2026-02-04
Add first verification infrastructure

math-fehr:
The goal of verification on an IR is to ensure that an IR always satisfy some invariants.
Here, I add a simple `verify` function on `IRContext` that checks that the number of operands/results,regions, is the expected one.

This will likely change in the future whenever we will have a more modular way of defining dialects, but at least it will help us moving forward.


Veir/ForLean.lean
```diff
@@ -150,6 +151,17 @@ theorem Std.ExtHashSet.insertMany_empty_eq_ofList [BEq α] [EquivBEq α] [Hashab
     (∅ : Std.ExtHashSet α).insertMany l = Std.ExtHashSet.ofList l := by
   ext; grind
 
+/--
+  A version of `Std.HashMap.keys.for` that also pass to the given function a proof
+  that the key exists in the map.
+  It is currently not as efficient as it could be, as we check dynamically the the key is
```
tobiasgrosser:
```suggestion
  It is currently not as efficient as it could be, as we check dynamically that the key is
```

Veir/IR/Basic.lean
```diff
@@ -1264,6 +1265,13 @@ theorem hasUses!_eq_false_iff_hasUses!_opResult_eq_false {op : OperationPtr}
 
 end OperationPtr
 
+/--
+  Run a function on all operations in the context.
+  In particular, the function provides a proof that the operation pointer is in bounds.
+-/
+def IRContext.forOpsDepM (ctx : IRContext) {m : Type w → Type w'} [Monad m] (p : ∀ (op : OperationPtr), op.InBounds ctx → m PUnit) : m PUnit :=
```
luisacicolini: line too long?

Veir/ForLean.lean
```diff
@@ -150,6 +151,17 @@ theorem Std.ExtHashSet.insertMany_empty_eq_ofList [BEq α] [EquivBEq α] [Hashab
     (∅ : Std.ExtHashSet α).insertMany l = Std.ExtHashSet.ofList l := by
   ext; grind
 
+/--
+  A version of `Std.HashMap.keys.for` that also pass to the given function a proof
```
```suggestion
  A version of `Std.HashMap.keys.for` that also passes to the given function a proof
```

Veir/ForLean.lean
```diff
@@ -150,6 +151,17 @@ theorem Std.ExtHashSet.insertMany_empty_eq_ofList [BEq α] [EquivBEq α] [Hashab
...
+  that the key exists in the map.
+  It is currently not as efficient as it could be, as we check dynamically the the key is
+  indeed in the map (although it should always be the case). This could be improved by
+  modifying the `Std.HashMap` implementation to pass such proof statically.
+-/
+def Std.HashMap.forKeysDepM [BEq α] [Hashable α] {m : Type w → Type w'} [Monad m]
+    (b : Std.HashMap α β) (f : ∀ (a : α), a ∈ b → m PUnit) : m PUnit :=
+  b.forM (fun k v => do if h: k ∈ b then f k (by grind))
```
```suggestion
  b.forM (fun k v => do if h : k ∈ b then f k (by grind))
```

Test/test-verifier.mlir
```diff
@@ -0,0 +1,6 @@
+// RUN: veir-opt %s 2>&1 | filecheck %s
+
+// CHECK: Error verifying input program: Expected 1 region
+"builtin.module"() ({
+
+}, {}) : () -> ()
```
tobiasgrosser: File not ending with newline.

Veir/Verifier.lean
```diff
@@ -0,0 +1,95 @@
...
+  | .test_test => do
+    if op.getNumOperands ctx opIn ≠ 1 then
+      throw "Expected 1 operand"
+    if op.getNumResults ctx opIn ≠ 1 then
+      throw "Expected 1 result"
+    if op.getNumRegions ctx opIn ≠ 0 then
+      throw "Expected 0 regions"
+    if op.getNumSuccessors ctx opIn ≠ 0 then
+      throw "Expected 0 successors"
+    pure ()
```
tobiasgrosser, line 87: There seems to be a lot of replication here. Eventually, we probably want to factor this into a per-op verify function or a per-op function that returns the number of expected operands/results,...

I would also expect a proof to be generated eventually. For now this seems a good step to gain test coverage.

Issue #170, opened 2026-02-05
LLVM `func` parsing


luisacicolini:
An example of a program I would like to be able to parse: 
```mlir
"builtin.module"() ({
  "func.func"() <{function_type = (i64, i32) -> i32, sym_name = "main"}> ({
  ^bb0(%arg0: i64, %arg1: i32):
    %0 = "llvm.sext"(%arg1) : (i32) -> i64
    %1 = "llvm.srem"(%arg0, %0) : (i64, i64) -> i64
    %2 = "llvm.trunc"(%1) <{overflowFlags = 0 : i32}> : (i64) -> i32
    "func.return"(%2) : (i32) -> ()
  }) : () -> ()
}) : () -> ()
```

tobiasgrosser:
@luisacicolini, is this meanwhile resolved?

luisacicolini:
it is only parsed correctly if we pass the flag as `<{nuw, nsw}>` (not `overflowFlags =`...). If we are fine with that it is resolved. On the LLVM-RISCV side, we should make sure to only generate flags in that format, and it should be doable.

PR #173, opened 2026-02-06
Add simple interpreter

math-fehr:
This is a first implementation of the interpreter, the design might change in the future, but at least we should be able to test it now.

The interpreter maintains a mapping from IR values (`ValuePtr`) to runtime values (`UInt64`). Each supported operation reads its operands from this mapping and writes its results back into it.

The interpreter walks the linked list of operations in a block. It continues until a `func.return` is encountered, at which point the returned values are collected and propagated to the caller. (This also adds `func.return` as needed for #170).

As we do not parse constants yet, we only test it with constant 0 (the default on arith.constant).

### DIFF
--- a/Test/Interpreter/test-veir-interpret.mlir
+++ b/Test/Interpreter/test-veir-interpret.mlir
@@ -1,9 +1,10 @@
 // RUN: veir-interpret %s | filecheck %s
 
-// CHECK: Program output: #[0]
 "builtin.module"() ({
   %lhs = "arith.constant"() : () -> i32
   %rhs = "arith.constant"() : () -> i32
   %x = "arith.addi"(%lhs, %rhs) : (i32, i32) -> i32
   "func.return"(%x) : (i32) -> ()
 }) : () -> ()
+
+// CHECK: Program output: #[0]
--- a/Veir/Interpreter.lean
+++ b/Veir/Interpreter.lean
@@ -35,9 +35,9 @@ structure InterpreterState where
 /--
   Set the value of a variable.
 -/
-def InterpreterState.setVar (state : InterpreterState) (var : ValuePtr) (val : RuntimeValue)
-    : InterpreterState :=
-  { state with variables := state.variables.insert var val }
+def InterpreterState.setVar (state : InterpreterState) (var : ValuePtr) (val : RuntimeValue) :
+    InterpreterState :=
+  {state with variables := state.variables.insert var val}
 
 /--
   Get the value of a variable, if the variable exists.
@@ -62,8 +62,8 @@ inductive ControlFlowAction where
   | continue
 
 /--
-  Interprets a single operation given the runtime values of its operands.
-  Returns the result runtime values and a control flow action indicating how
+  Interpret a single operation given the runtime values of its operands.
+  Return the result runtime values and a control flow action indicating how
   to continue the interpretation.
   If any error occurs during interpretation (e.g., unknown operation, missing variable),
   returns `none`.
@@ -84,11 +84,11 @@ def interpretOp' (ctx : IRContext) (opPtr : OperationPtr) (operands: Array Runti
   | _ => none
 
 /--
-  Interprets a single operation given the current interpreter state.
-  Returns an updated interpreter state and a control flow action indicating how
+  Interpret a single operation given the current interpreter state.
+  Return an updated interpreter state and a control flow action indicating how
   to continue the interpretation.
   If any error occurs during interpretation (e.g., unknown operation, missing variable),
-  returns `none`.
+  return `none`.
 -/
 def interpretOp (ctx : IRContext) (opPtr : OperationPtr) (state : InterpreterState)
     (opPtrInBounds : opPtr.InBounds ctx := by grind)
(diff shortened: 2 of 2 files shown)


Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+
+  The interpreter walks the linked list of operations in a block. It continues
+  until a `func.return` is encountered, at which point the returned values are
+  collected and propagated to the caller.
+-/
+
+namespace Veir
+
+/--
+  The representation of a vaule in the interpreter.
```
tobiasgrosser, line 24:
```suggestion
  The representation of a value in the interpreter.
```

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+  Returns the values returned by the block, if any.
+  Returns `none` if any errors occur during interpretation.
+-/
+def interpretBlock (ctx : IRContext) (blockPtr : BlockPtr) (state : InterpreterState) (blockInBounds : blockPtr.InBounds ctx := by grind) (wf : ctx.WellFormed := by grind) : Option (Array RuntimeValue) := do
+  let block := blockPtr.get ctx (by grind)
+  rlet firstOp ← (blockPtr.get ctx).firstOp
+  interpretOpList ctx firstOp state
+
+/--
+  Interprets a bulitin.module operation.
```
tobiasgrosser:
```suggestion
  Interprets a builtin.module operation.
```

Test/Interpreter/test-veir-interpret.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// CHECK: Program output: #[0]
```
tobiasgrosser: I would expect `CHECK` lines to follow after the program output.

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+  It includes a mapping from IR values to their runtime values.
+-/
+structure InterpreterState where
+  variables : Std.HashMap ValuePtr RuntimeValue
+
+/--
+  Set the value of a variable.
+-/
+def InterpreterState.setVar (state : InterpreterState) (var : ValuePtr) (val : RuntimeValue)
+    : InterpreterState :=
```
luisacicolini: nit: is there a reason why the first `:` is on this line? it feels to me like it should be on the previous one? (and same below)

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+  { state with variables := state.variables.insert var val }
```
luisacicolini:
```suggestion
  {state with variables := state.variables.insert var val}
```

(I know we discussed this spacing already and don't remember what the conclusion was, feel free to ignore if it's consistently spaced)

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+    return (#[op.properties], .continue)
+  | .arith_addi => do
+    let #[lhs, rhs] := operands | none
+    return (#[lhs + rhs], .continue)
+  | .func_return => do
+    return (#[], .return operands)
+  | _ => none
+
+/--
+  Interprets a single operation given the current interpreter state.
```
luisacicolini:
```suggestion
  Interpret a single operation given the current interpreter state.
```
for consistency with the comments above

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+  Returns an updated interpreter state and a control flow action indicating how
```
luisacicolini:
```suggestion
  Return an updated interpreter state and a control flow action indicating how
```

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+  to continue the interpretation.
+  If any error occurs during interpretation (e.g., unknown operation, missing variable),
+  returns `none`.
```
luisacicolini:
```suggestion
  return `none`.
```

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+    let value := resultValues[idx]!
+    s.setVar result value) newState
+  return (newState, action)
+
+/--
+  Interprets a list of operations, starting from the given operation pointer.
+  Returns an array of values corresponding to the values returned by the block, if any.
+  Continues to interpret operations until a `return` control flow action is encountered,
+  or the end of the block is reached.
+  Returns `none` if any errors occur during interpretation.
```
luisacicolini:
```suggestion
  Interpret a list of operations, starting from the given operation pointer.
  Return an array of values corresponding to the values returned by the block, if any.
  Continue to interpret operations until a `return` control flow action is encountered,
  or the end of the block is reached.
  Return `none` if any errors occur during interpretation.
```

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+    rlet next ← (op.get ctx).next
+    interpretOpList ctx next state
+  | .return results =>
+    return results
+partial_fixpoint
+
+/--
+  Interprets a block of operations, starting from the first operation in the block.
+  Returns the values returned by the block, if any.
+  Returns `none` if any errors occur during interpretation.
```
luisacicolini:
```suggestion
  Interpret a block of operations, starting from the first operation in the block.
  Return the values returned by the block, if any.
  Return `none` if any errors occur during interpretation.
```

Veir/Interpreter.lean
luisacicolini:
```suggestion
  Interpret a builtin.module operation.
```

Veir/Interpreter.lean
```diff
@@ -0,0 +1,151 @@
...
+  let block := blockPtr.get ctx (by grind)
+  rlet firstOp ← (blockPtr.get ctx).firstOp
+  interpretOpList ctx firstOp state
+
+/--
+  Interprets a bulitin.module operation.
+  This is done by interpreting the first block of the first region of the operation.
+  Reurns the values returned by the block.
+  If any errors occur during interpretation, returns `none`.
+-/
```
luisacicolini, line 141:
```suggestion
  This is done by interpreting the first block of the first region of the operation.
  Returns the values returned by the block.
  If any errors occur during interpretation, returns `none`.
-/
```

Veir/OpCode.lean
```diff
@@ -28,6 +28,7 @@ inductive OpCode
   | arith_andi
   | builtin_unregistered
   | builtin_module
+  | func_return
```
tobiasgrosser: This was now added in #174.

Veir/Interpreter.lean
math-fehr: I hate it and I'll change to the Lean style ;)

PR #176, opened 2026-02-07
Add `Attribute` and `TypeAttr` inductive types

math-fehr:
These two types are used to represent attributes and types in the IR, respectively. This is not the final design, as they should be extensible and not require to recompile the whole codebase whenever a new attribute or type is added. However, this is a first step to at least be able to represent attributes and types in the IR.

Similar to xDSl, I merged Attribute and Type into a single type, so that we can write functions that can both operate on attributes and types without having to duplicate code.

Currently, `TypeAttr` is defined as `{attr : attr.isType}`.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -16,7 +16,10 @@ module
   completely disjoint in MLIR). The reason for this lack of separation in VeIR is
   that merging both concepts into a single `Attribute` type allows to define
   functions that can work with both types and attributes without needing to define
-  separate functions for each case.
+  separate functions for each case. For instance, `mlir::AttrTypeWalker` can be
+  defined for both `TypeAttr` and `Attribute` without needing to define separate
+  walkers for each case. Similarly, `mlir::TypeAttr` is not needed, as we can
+  store any `TypeAttr` as an `Attribute`.
 -/
 
 namespace Veir


Veir/IR/Attribute.lean
```diff
@@ -0,0 +1,90 @@
...
+  A data structure that represents compile-time information in the IR.
+  Attributes are used either as type annotations for SSA values, or
+  as extra information stored in operations.
+-/
+inductive Attribute
+| integerType (bitwidth : Nat)
+/--
+  An attribute from an unknown dialect.
+-/
+| unregistered (value : String) (isType : Bool)
```
```suggestion
/- An attribute from an unknown dialect. -/
| unregistered (value : String) (isType : Bool)
```

This breaks the flow less. Or even `-- An attribute` at the end of the `unregistered` line.

tobiasgrosser (approved):
Can we use a coercion, instead of the `toType` maybe? I guess, @ineol's input would be useful here as well beyond the usual review from @luisacicolini.

math-fehr:
I just wanted to try something, since anyway we should probably wait Monday for merging this.
Now, we create for each attribute a different structure (which will help us having methods for them). So when we create an `IntegerType`, we can directly have a coercion to both a TypeAttr or an Attribute

Veir/IR/Attribute.lean
```diff
@@ -0,0 +1,119 @@
...
+
+  The `TypeAttr` definition is a subtype of `Attribute` that carries the additional
+  invariant that the attribute is a valid type annotation.
+
+  `TypeAttr` corresponds to `mlir::Type`, and `Attribute`s that are not
+  `TypeAttr`s correspond to an `mlir::Attribute` (as attributes and types are
+  completely disjoint in MLIR). The reason for this lack of separation in VeIR is
+  that merging both concepts into a single `Attribute` type allows to define
+  functions that can work with both types and attributes without needing to define
+  separate functions for each case.
```
luisacicolini: nit: maybe an example of two mlir functions working with a type and an attribute (and how they live in veIR) would clarify?

Veir/IR/Attribute.lean
```diff
@@ -0,0 +1,119 @@
...
+-/
+
+namespace Veir
+public section
+
+/--
+  The definition of `!builtin.integer`.
+  It is an integer type with a given bitwidth.
```
luisacicolini:
```suggestion
  A `!builtin.integer` is an integer type with a given bitwidth.
```

luisacicolini: (I find this style of docstring slightly clearer but feel free to ignore)

Veir/IR/Attribute.lean
```diff
@@ -0,0 +1,119 @@
...
+-/
+def isType (attr : Attribute) : Bool :=
+  match attr with
+  | .integerType _ => true
+  | .unregisteredAttr attr => attr.isType
+
+@[simp, grind =]
+theorem isType_integerType type : (integerType type).isType = true := by rfl
+@[simp, grind =]
+theorem isType_unregistered unregistered : (unregisteredAttr unregistered).isType = unregistered.isType := by rfl
```
luisacicolini: this line feels a bit too long?

Veir/Rewriter/Basic.lean
```diff
@@ -380,7 +380,7 @@ theorem Rewriter.initOpRegions_inBounds_mono (ptr : GenericPtr) :
     ptr.InBounds ctx → ptr.InBounds (initOpRegions ctx opPtr regions n opPtrInBounds hregions hctx hn) := by
   induction n generalizing ctx <;> simp only [initOpRegions] <;> grind
 
-def Rewriter.initOpResults (ctx: IRContext) (opPtr: OperationPtr) (resultTypes: Array MlirType) (index: Nat := 0) (hop : opPtr.InBounds ctx)
+def Rewriter.initOpResults (ctx: IRContext) (opPtr: OperationPtr) (resultTypes: Array TypeAttr) (index: Nat := 0) (hop : opPtr.InBounds ctx)
```
luisacicolini: line seems long

Veir/PatternRewriter/Basic.lean
```diff
@@ -142,7 +142,7 @@ theorem addUsersInWorklist_same_ctx :
   simp [addUsersInWorklist]
 
 def createOp (rewriter: PatternRewriter) (opType: OpCode)
-    (resultTypes: Array MlirType) (operands: Array ValuePtr) (blockOperands: Array BlockPtr) (regions: Array RegionPtr) (properties: UInt64)
+    (resultTypes: Array TypeAttr) (operands: Array ValuePtr) (blockOperands: Array BlockPtr) (regions: Array RegionPtr) (properties: UInt64)
```
luisacicolini: line seems long

PR #178, opened 2026-02-07
Add IntBv, a bit-blastable LLVM type

tobiasgrosser:
This type allows us to call `bv_decide` right on the `IntBv` type.

### DIFF
--- a/Veir/Dialects/LLVM/IntBv/Basic.lean
+++ b/Veir/Dialects/LLVM/IntBv/Basic.lean
@@ -5,9 +5,7 @@ namespace Veir.Dialects.LLVM.IntBv
 public section
 
 /--
-The `IntBv` type can have any bitwidth `w`. It is either a two's complement
-integer value of width `w` or a poison value indicating delayed undefined
-behavior.
+The `IntBv` type is either a two's complement integer value of width `w` or a poison value indicating delayed undefined behavior.
 -/
 structure IntBv (w : Nat) where
   /-- A two's complement integer value of width `w`. -/
@@ -18,7 +16,7 @@ structure IntBv (w : Nat) where
   h : poison → (val = 0) := by simp
 deriving DecidableEq
 
-def IntBv.mkPoison {w : Nat} : IntBv w :=
+def IntBv.mkPoison (w : Nat) : IntBv w :=
   ⟨0, true, by simp⟩
 
 def IntBv.mkVal {w : Nat} (val : BitVec w): IntBv w :=
--- a/Veir/Dialects/LLVM/IntBv/Lemmas.lean
+++ b/Veir/Dialects/LLVM/IntBv/Lemmas.lean
@@ -36,7 +36,7 @@ theorem ext_iff {w : Nat} (x y : IntBv w) :
     x = y ↔ (x.val = y.val ∧ x.poison = y.poison) := by
   rw [@IntBv.mk.injEq]
 
-/- # add -/
+/-! # add -/
 
 @[bv_normalize]
 theorem add_eq {w : Nat} (x y : IntBv w) :
@@ -63,7 +63,7 @@ theorem add_comm {w : Nat} (x y : IntBv w) : x + y = y + x := by
   simp only [HAdd.hAdd, Add.add, add]
   cases x.poison <;> cases y.poison <;> simp [BitVec.add_comm]
 
-/- # mul -/
+/-! # mul -/
 
 @[simp, bv_normalize]
 theorem poison_mul {w : Nat} (x : IntBv w) : IntBv.mkPoison * x = IntBv.mkPoison := by


Veir/Dialects/LLVM/IntBv/Basic.lean
```diff
@@ -0,0 +1,45 @@
+module
```
math-fehr, line 1: Should this file just replace `Int/Basic.lean`? It seems this is exactly the same structure, just with a different encoding?

Veir/Dialects/LLVM/IntBv/Basic.lean
```diff
@@ -0,0 +1,45 @@
...
+integer value of width `w` or a poison value indicating delayed undefined
+behavior.
+-/
+structure IntBv (w : Nat) where
+  /-- A two's complement integer value of width `w`. -/
+  val : BitVec w
+  /-- A poison flag for `w`. -/
+  poison : Bool
+  /-- Invariant: If this value is poison the `val` is zero. -/
+  h : poison → (val == 0) := by simp
```
math-fehr:
```suggestion
  h : poison → (val = 0) := by simp
```
This is the usual style from what I understand

Veir/Dialects/LLVM/IntBv/Basic.lean
```diff
@@ -0,0 +1,45 @@
...
+deriving DecidableEq
+
+def IntBv.mkPoison {w : Nat} : IntBv w :=
```
math-fehr:
```suggestion
def IntBv.mkPoison (w : Nat) : IntBv w :=
```
I think @ineol  or @luisacicolini  might know what is best here, but I feel that it should be explicit here, as you can only deduce it from the return type, not from other arguments.

Veir/Dialects/LLVM/IntBv/Examples.lean
```diff
@@ -0,0 +1,17 @@
+module
+
+import Std.Tactic.BVDecide
+import Veir.Dialects.LLVM.IntBv.Basic
+import all Veir.Dialects.LLVM.IntBv.Lemmas
+
+namespace Veir.Dialects.LLVM.IntBv
+
+example (w : Nat) (x y : IntBv w) : x + y = y + x := by
+  grind
```
math-fehr, line 10: Why does this grind works btw? Is it unfolding `add` by default here?

Veir/Dialects/LLVM/IntBv/Lemmas.lean
```diff
@@ -0,0 +1,82 @@
...
+@[simp, bv_normalize]
+theorem add_poison {w : Nat} (x : IntBv w) : x + IntBv.mkPoison = IntBv.mkPoison := by
+  simp [HAdd.hAdd, Add.add, add]
+
+@[grind =]
+theorem add_comm {w : Nat} (x y : IntBv w) : x + y = y + x := by
+  simp only [HAdd.hAdd, Add.add, add]
+  cases x.poison <;> cases y.poison <;> simp [BitVec.add_comm]
+
+/- # mul -/
```
math-fehr:
```suggestion
/-! # mul -/
```

Veir/Dialects/LLVM/IntBv/Lemmas.lean
```diff
@@ -0,0 +1,82 @@
...
+@[simp, bv_normalize]
+theorem poison_mkPoison {w : Nat} : (IntBv.mkPoison (w := w)).poison = true := by
+  simp only [IntBv.mkPoison]
+
+@[bv_normalize]
+theorem ext_iff {w : Nat} (x y : IntBv w) :
+    x = y ↔ (x.val = y.val ∧ x.poison = y.poison) := by
+  rw [@IntBv.mk.injEq]
+
+/- # add -/
```
math-fehr:
```suggestion
/-! # add -/
```

Veir/Dialects/LLVM/IntBv/Basic.lean
```diff
@@ -0,0 +1,45 @@
+module
+
+namespace Veir.Dialects.LLVM.IntBv
+
+public section
+
+/--
+The `IntBv` type can have any bitwidth `w`. It is either a two's complement
+integer value of width `w` or a poison value indicating delayed undefined
+behavior.
```
math-fehr: I would just remove the first sentence here?

tobiasgrosser:
```suggestion
The `IntBv` type is either a two's complement integer value of width `w` or a poison value indicating delayed undefined behavior.
```

Veir/Dialects/LLVM/IntBv/Basic.lean
luisacicolini: would a `Data/` folder be useful? All the data structures used by dialects (which should not anyways be that many) could live there. I find it a little bit confusing to have the data types with the semantics of the operations

luisacicolini: I also don't understand what';s the difference between `Int` and `IntBv`?
And a design question: why do we want to have a bool for poison? I find the `Option` more intuitive, especially because I suppose in a lot of cases we won't be able to specify a priori what the value of `poison` is (?)

luisacicolini: Another alternative that comes to mind is creating a type that contains all integers (bitvecs) + a special element poison - have we considered this? What are the advantages of this def?

Veir/Dialects/LLVM/IntBv/Examples.lean
luisacicolini: no clue, please let me know if you find out!

PR #179, opened 2026-02-08
Add LocalRewritePattern to represent simple rewrite patterns

math-fehr:
LocalRewritePatterns represent rewrites that match a single rooted DAG, and then replace matched root with a list of new operations. This is what is used by InstCombine in LLVM, and should be easier to
prove these patterns correct compared to the more general PatternRewriter.

Note that this currently does not delete unused operations. so the constant folding example still keep the old constants.

### DIFF
--- a/Veir/Benchmarks.lean
+++ b/Veir/Benchmarks.lean
@@ -72,7 +72,7 @@ def addIConstantFoldingLocal (ctx: IRContext) (op: OperationPtr) :
 
   -- Sum both constant values
   let newVal := lhsOpStruct.properties + rhsOpStruct.properties
-  let (ctx, newOp) ← Rewriter.createOp ctx .arith_constant #["i32"] #[] #[] #[] newVal none sorry sorry sorry sorry sorry
+  let (ctx, newOp) ← Rewriter.createOp ctx .arith_constant #[IntegerType.mk 32] #[] #[] #[] newVal none sorry sorry sorry sorry sorry
   return (ctx, some (#[newOp], #[newOp.getResult 0]))
 
 def addIZeroFolding (rewriter: PatternRewriter) (op: OperationPtr) : Option PatternRewriter := do


Veir/Benchmarks.lean
```diff
@@ -45,6 +45,35 @@ def addIConstantFolding (rewriter: PatternRewriter) (op: OperationPtr) : Option
     rewriter ← rewriter.eraseOp rhsOp sorry
   return rewriter
 
+def addIConstantFoldingLocal (ctx: IRContext) (op: OperationPtr) :
+    Option (IRContext × List OperationPtr × List ValuePtr) := do
```
tobiasgrosser: Should this not take an `InBounds` proof?

math-fehr: Well, that's what we need to decide at some point.
I want to take some time to clean up the IR/Basic API, and after that I'll take care of this one.
I think we want both a version with InBounds and one without. Right now I kept the consistency with the non-local version

Veir/PatternRewriter/Basic.lean
```diff
@@ -208,6 +208,28 @@ def replaceValue (rewriter: PatternRewriter) (oldVal newVal: ValuePtr)
 end PatternRewriter
 
 abbrev RewritePattern := PatternRewriter → OperationPtr → Option (PatternRewriter)
+/--
```
luisacicolini, line 213:
```suggestion

/--
```

Veir/Benchmarks.lean
```diff
@@ -45,6 +45,35 @@ def addIConstantFolding (rewriter: PatternRewriter) (op: OperationPtr) : Option
...
+  -- Check that the operation is an arith.addi operation
```
luisacicolini:
```suggestion
  -- Check that the operation is an `arith.addi` operation
```

Veir/PatternRewriter/Basic.lean
```diff
@@ -208,6 +208,28 @@ def replaceValue (rewriter: PatternRewriter) (oldVal newVal: ValuePtr)
...
+  A local rewrite that can only replace a matched operation with a list of new operations.
+  The pattern returns, if successful, a list of new operations to insert and a list of values to
+  replace the old results with.
+-/
+abbrev LocalRewritePattern := IRContext → OperationPtr → Option (IRContext × List OperationPtr × List ValuePtr)
```
ineol: What happens if the rewrite does not apply? it returns `none`? but then we lose linearity of the context because the function below keeps a copy of the context for that case, right?

ineol: Also, do we want a list or an array or new operations?

Veir/Benchmarks.lean
```diff
@@ -45,6 +45,35 @@ def addIConstantFolding (rewriter: PatternRewriter) (op: OperationPtr) : Option
...
+  if op.getOpType ctx sorry ≠ .arith_addi then
+    none
+
+  -- Get the lhs and check that it is a constant
+  let lhsValuePtr := op.getOperand ctx 0 (by sorry) (by sorry)
+  let lhsOp ← match lhsValuePtr with
+  | ValuePtr.opResult lhsOpResultPtr => some lhsOpResultPtr.op
+  | _ => none
+  let lhsOpStruct := lhsOp.get ctx (by sorry)
+  if lhsOpStruct.opType ≠ .arith_constant then
```
ineol: Shouldn't it be `arithConstant`?

Veir/PatternRewriter/Basic.lean
math-fehr: Ah fuck you're right, I'm changing this!
I'll also use arrays here

math-fehr: Should be fixed!

Veir/Benchmarks.lean
math-fehr: So what we talked about last time is that I wanted to have `arith.constant`, or `arith_constant`, so that it is a bit more readable than `arithConstant` imo. Because we want `arith` to be some kind of namespace?
`arith.constant` wouldn't work and was annoying, so I tried `arithConstant`. If you want we can try to change it to something else, but I really felt like `arithConstant` was weird there

ineol: Makes sense!

PR #182, opened 2026-02-09
Refactor ordering of definitions in Attribute.lean (NFC)

math-fehr:
Moving all definitions together is necessary to define mutual definitions later on for other attributes.


tobiasgrosser (approved):
LGTM.

PR #183, opened 2026-02-10
Make parseOptionalDelimitedList return an Option

math-fehr:
This should have never returned an empty array when the left delimiter was missing.


tobiasgrosser (approved):
LGTM

PR #191, opened 2026-02-12
Add parseOptionalIdentifierOrStringLiteral

math-fehr:
This function is often used in the MLIR textual format, where identifiers might need to be escaped, so a string literal is also accepted.

### DIFF
--- a/UnitTest/AttrParser.lean
+++ b/UnitTest/AttrParser.lean
@@ -35,28 +35,6 @@ def testAttr (s : String) : Except String Attribute := do
   parseAttribute.run' AttrParserState.mk parser
 
 /--
-  Run parseOptionalAttr on the given input string.
--/
-def testOptionalAttr (s : String) : Except String (Option Attribute) :=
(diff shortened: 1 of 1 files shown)


Veir/Parser/Parser.lean
```diff
@@ -207,6 +207,27 @@ def parseStringLiteral (errorMsg : String := "string literal expected") : M Stri
...
+  match ← parseOptionalIdentifier with
+  | some ident => return ident
+  | none =>
+    match ← parseOptionalStringLiteral with
+    | some str => return some str.toByteArray
+    | none => return none
+
+/--
+  Parses either an identifier or a string literal.
+  Raise an error if the next token is neither an identifier nor a string literal.
```
luisacicolini:
```suggestion
  Raises an error if the next token is neither an identifier nor a string literal.
```

luisacicolini (approved):

ineol (approved):

PR #192, opened 2026-02-13
Add support for operation properties parsing

math-fehr:
We do not yet do anything with the properties, as we need to add support for them in the IR first.

### DIFF
--- a/UnitTest/AttrParser.lean
+++ b/UnitTest/AttrParser.lean
@@ -35,28 +35,6 @@ def testAttr (s : String) : Except String Attribute := do
   parseAttribute.run' AttrParserState.mk parser
 
 /--
-  Run parseOptionalAttr on the given input string.
--/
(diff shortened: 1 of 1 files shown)


Veir/Parser/MlirParser.lean
```diff
@@ -246,6 +246,20 @@ def parseTypedValue : MlirParserM (ByteArray × TypeAttr) := do
...
 
+/--
+  Parse the properties of an operation.
+  Currently, these properties are not stored in the IR, but we still need to parse them to be able
+  to parse valid MLIR syntax.
+-/
+def parseOpProperties : MlirParserM Unit := do
+  if not (← parseOptionalPunctuation "<") then
+    return ()
+  match AttrParser.parseAttributeDictionary.run AttrParserState.mk (← getThe ParserState) with
```
ineol, line 257: use monadic syntax?

ineol (approved):

PR #194, opened 2026-02-13
Use LLVM.Int in the interpreter instead of Nat

math-fehr:
This will allow us to use the actual LLVM semantics, since Int can represent poison values.

### DIFF
--- a/Veir/Interpreter.lean
+++ b/Veir/Interpreter.lean
@@ -5,9 +5,7 @@ import Veir.IR.WellFormed
 import Veir.PatternRewriter.Basic
 import Veir.Dialects.LLVM.Int.Basic
 
-open Veir.Dialects.LLVM.Int
-abbrev IntBv := Veir.Dialects.LLVM.Int.Int
-
+open Veir.Dialects
 /-!
   # Veir Interpreter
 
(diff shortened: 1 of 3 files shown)


Veir/Interpreter.lean
```diff
@@ -74,11 +84,16 @@ def interpretOp' (ctx : IRContext) (opPtr : OperationPtr) (operands: Array Runti
...
+    let res ← op.results[0]?
+    let .integerType bw := res.type.val
+      | none
+    return (#[.int bw.bitwidth (.val (BitVec.ofNat bw.bitwidth value.toNat))], .continue)
   | .arith_addi => do
-    let #[lhs, rhs] := operands | none
-    return (#[lhs + rhs], .continue)
+    let #[.int bw lhs, .int bw' rhs] := operands | none
+    if h: bw' ≠ bw then none else
+    let rhs := Dialects.LLVM.Int.cast rhs (by simp at h; exact h)
```
math-fehr: This is a bit ugly, but since we are overloading "Int", we need to proved the entire name here? If someone has a better idea to make this nicer please let me know!

tobiasgrosser: Does `rhs.cast` not work?

luisacicolini: I think that should work, if it doesn't it might be because of namespaces (and if you create a namespace `Int` or `llvmInt` or whatever that should then work afaik

tobiasgrosser (approved):

luisacicolini (approved):

PR #196, opened 2026-02-14
Remove IRContext.topLevelOp

ineol:
The `topLevelOp` field was problematic, because the empty context was `FieldsInBounds`.

--------

@math-fehr The worklist for the pattern rewriter was initialized with all ops except `topLevel` and now it's all ops. Should it be given an explicit top level op? or should it test for the existence of parents when inserting in the worklist?

The grind patterns for the `allocEmpty` functions are not optimal (also for Operation and Block probably). I'll write the hand written ones tomorrow.

### DIFF
--- a/Veir/IR/GetSet.lean
+++ b/Veir/IR/GetSet.lean
@@ -47,7 +47,6 @@ setup_grind_with_get_set_definitions
 
 /- OperationPtr.allocEmpty -/
 
-
 @[simp, grind =>]
 theorem BlockPtr.get!_OperationPtr_allocEmpty {block : BlockPtr}
     (heq : OperationPtr.allocEmpty ctx ty = some (ctx', op')) :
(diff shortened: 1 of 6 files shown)


Veir/IR/GetSet.lean
```diff
@@ -4016,15 +3852,112 @@ theorem OpOperandPtrPtr.get!_OpResultPtr_setFirstUse {opOperandPtr : OpOperandPt
...
+theorem ValuePtr.getFirstUse!_RegionPtr_allocEmpty {value : ValuePtr}
+    (heq : RegionPtr.allocEmpty ctx = some (ctx', rg')) :
+    value.getFirstUse! ctx' = value.getFirstUse! ctx := by
+  grind
+
+-- @[simp] --
+-- theorem ValuePtr.getType!_RegionPtr_allocEmpty {value : ValuePtr}
+--     -- (_ : OperationPtr.allocEmpty ctx = some (ctx', rg')) : :
+--     value.getType! ctx' = value.getType! ctx := by
+--   grind
```
math-fehr: What is the problem with that theorem?

Veir/Rewriter/Basic.lean
```diff
@@ -595,45 +610,17 @@ def IRContext.create : Option (IRContext × OperationPtr) :=
...
-  -- Note: We inline part of the definition of `createOp` because the above
-  -- `ctx` does not satisfy `ctx.FieldsInBounds` because `topLevelOp` is an
-  -- invalid pointer.
+
+  have : ctx.FieldsInBounds := by
+    constructor
+    · grind [OperationPtr.InBounds]
+    · grind [BlockPtr.InBounds]
+    · grind [RegionPtr.InBounds]
+
```
math-fehr: I guess we should now maybe separate `IRContext.create` and `IRContext.createWithModule` somehow?

Veir/IR/GetSet.lean
ineol: It was commented out for the other `allocEmpty`s, I've uncommented all of them (I think they were weird because the types were `Unit` so it was tautological.

PR #199, opened 2026-02-15
Use `module` in more files

ineol:
----

The file DeallocLemmas unfolds definitions

### DIFF
--- a/Veir/IR/WellFormed.lean
+++ b/Veir/IR/WellFormed.lean
@@ -1,7 +1,14 @@
-import Veir.IR.Basic
+module
+
+public import Veir.IR.Basic
+public import Veir.IR.Fields
+import Veir.IR.GetSet
+import Veir.IR.InBounds
 import Veir.IR.Grind
 import Veir.ForLean
 import Std.Data.ExtHashSet
+
+public section
 
 namespace Veir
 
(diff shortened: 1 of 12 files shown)


Veir/Prelude.lean
```diff
@@ -22,7 +22,7 @@ macro:50 "rlet" pat:term "←" expr:term "in" rest:term : term =>
       | none => none
       | some $pat => $rest)
 
-def Option.maybe (p : α → β → Prop) (x : Option α) (y : β) : Prop :=
+@[expose] def Option.maybe (p : α → β → Prop) (x : Option α) (y : β) : Prop :=
   ∀ z, x = some z → p z y
```
math-fehr, line 26: Should we instead rely on `Option.maybe_def`?

Veir/IR/WellFormed.lean
```diff
@@ -855,7 +862,7 @@ theorem BlockPtr.OpChain_next_ne
 theorem ValuePtr.DefUse.hasUses_iff
     (hWF : ValuePtr.DefUse value ctx array missingUses) :
     value.hasUses ctx ↔ array ≠ #[] := by
-  grind [DefUse, hasUses]
+  grind [DefUse, hasUses_def] -- :(
```
math-fehr: Why is this sad btw?

ineol: Back then I thought there were missing lemmas, I removed it!

PR #201, opened 2026-02-16
Add `StringAttr` attribute

math-fehr:
It just contains a string element, represented as a byte array.
We will add filecheck tests for printing string attributes when properties are properly supported.

### DIFF
--- a/Veir/Parser/AttrParser.lean
+++ b/Veir/Parser/AttrParser.lean
@@ -70,6 +70,25 @@ def parseOptionalIntegerAttr : AttrParserM (Option IntegerAttr) := do
   let integerType ← parseIntegerType "integer type expected after ':' in integer attribute"
   return some (IntegerAttr.mk value integerType)
 
+/--
+  Parse a string attribute, if present.
+  Its syntax is a string literal enclosed in double quotes, e.g., `"foo"`.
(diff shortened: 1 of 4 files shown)


Veir/Parser/AttrParser.lean
```diff
@@ -70,6 +70,25 @@ def parseOptionalIntegerAttr : AttrParserM (Option IntegerAttr) := do
   let integerType ← parseIntegerType "integer type expected after ':' in integer attribute"
   return some (IntegerAttr.mk value integerType)
 
+/--
+  Parse a string attribute, if present.
+  Its syntax is a string literal enclosed in double quotes, e.g., `"foo"`.
+-/
+def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
```
luisacicolini, line 77: I think Lean usually uses `?` in the naming of this type of theorems, in case you want to consider that? (e.g. `parseStringAttr?`)?

math-fehr: That's a very good point, I haven't considered it for these parser methods, but maybe we should have a PR to move everything to it!
I'll make sure to keep that in mind and probably write a PR that fixes that all at once, thanks!

PR #204, opened 2026-02-17
Print properties in a generic way

math-fehr:
Properties were hardcoded for arith.constant, now
they should work for any operation


tobiasgrosser (approved):
Nice!

PR #205, opened 2026-02-17
Add parsePrefixedKeyword

math-fehr:
This utility function is used to parse identifiers prefixed with a sigil, such as `@`, `#`, `!`, `^`, or `%`.

### DIFF
--- a/Veir/Parser/Lexer.lean
+++ b/Veir/Parser/Lexer.lean
@@ -115,7 +115,7 @@ deriving Inhabited, Repr, DecidableEq
 namespace TokenKind
 
 /--
-  Checks if the token kind either a hash, percent, caret, or exclamation identifier.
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):

Veir/Parser/Lexer.lean
```diff
@@ -112,7 +112,29 @@ inductive TokenKind
   | fileMetadataEnd
 deriving Inhabited, Repr, DecidableEq
 
-instance TokenKind.toString : ToString TokenKind where
+namespace TokenKind
+
+/--
+  Checks if the token kind either a hash, percent, caret, or exclamation identifier.
```
luisacicolini:
```suggestion
  Checks if the token kind is a hash, percent, caret, or exclamation identifier.
```

ineol: also `@`?

PR #209, opened 2026-02-18
Add Unit attributes

math-fehr:
Unit attributes contain no information, and are usually used as markers. They have a special syntax for dictionary entries.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -56,7 +56,7 @@ instance : Repr StringAttr where
   reprPrec attr _ := "StringAttr.mk " ++ repr (String.fromUTF8! attr.value)
 
 /--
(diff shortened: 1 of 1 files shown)


Veir/IR/Attribute.lean
```diff
@@ -55,6 +55,12 @@ deriving Inhabited, DecidableEq, Hashable
 instance : Repr StringAttr where
   reprPrec attr _ := "StringAttr.mk " ++ repr (String.fromUTF8! attr.value)
 
+/--
+  A unit attribute that carries no information.
```
```suggestion
  A unit attribute that carries no information, but the information that it exists.
```

luisacicolini: that it exists?

luisacicolini (approved):

ineol (approved):

PR #213, opened 2026-02-18
Improved error messages on parser failures

AlexanderViand:
Adds better error messages on parser failures, for example:

```
Error: Error parsing operation: attribute expected
at line 8, column 52 (byte 422) near token exclamationIdent '!felt.type'
      "struct.member"() <{sym_name = "val", type = !felt.type}> {llzk.pub} : () -> ()
``` 

Also, vibe-coded by AI and zero testing, but @math-fehr  wanted to have this here

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -207,11 +207,13 @@ def resolveOperand (operand : UnresolvedOperand) (expectedType : TypeAttr) : Mli
   Parse a type, if present.
 -/
 def parseOptionalType : MlirParserM (Option TypeAttr) := do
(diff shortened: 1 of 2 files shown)


math-fehr:
I'll take care of this PR, but I'll actually change most of the code.
I think we want to create a new `ParserError` structure and return this instead of a `String` on all our parser errors.
While this works, it's only a bandage solution and I prefer writing the real thing first.

tobiasgrosser:
@math-fehr pushed plenty of improvements here. Given that this PR won't apply cleanly and a manual implementation route was chosen, I close this PR. Please raise specific issues, if this there are further error messages we should emit.

I close this for now to keep our PR stack clean.

PR #217, opened 2026-02-19
Add attribute dictionary to operations

math-fehr:
Each operation has an attribute dictionary that is a `DictionaryAttr`.
This changes slightly the syntax we used to have.


Veir/IR/Basic.lean
```diff
@@ -576,6 +575,20 @@ theorem pushOperand!_eq_pushOperand {op : OperationPtr} (inBounds: op.InBounds c
     op.pushOperand! ctx operands = op.pushOperand ctx operands inBounds := by
   grind [pushOperand, pushOperand!]
 
+def setAttributes (op: OperationPtr) (ctx: IRContext) (newAttrs: DictionaryAttr)
+    (inBounds: op.InBounds ctx := by grind) : IRContext :=
+  let oldOp := op.get ctx (by grind)
```
```suggestion
  let oldOp := op.get ctx
```
should work?

PR #218, opened 2026-02-19
Add ArrayAttr

math-fehr:
This is an attribute that contains a sequence of other attributes.


Veir/IR/Attribute.lean
```diff
@@ -171,6 +184,17 @@ decreasing_by
   · have := @FunctionType.sizeOf_elems_outputs
     grind
 
+def ArrayAttr.decEq (arr1 arr2 : ArrayAttr) : Decidable (arr1 = arr2) :=
+  let value1 := arr1.value
+  let value2 := arr2.value
+  match Array.instDecidabelEq' value1 value2 (fun x y _ _ => Attribute.decEq x y) with
```
ineol:
```suggestion
  match Array.instDecidabelEq' value1 value2 (fun x y _ _ => x.decEq y) with
```

PR #221, opened 2026-02-19
chore: make documentation into docstrings in IR/Basic.lean

alexkeizer:
I noticed some documentation in `Veir/IR/Basic.lean` that use `/- -/` type comments, but which would be better as proper docstrings (`/-- -/`), so they show up in hover and such, this PR addresses that.


ineol (approved):
Thank you!

PR #222, opened 2026-02-19
Verified parser

GZGavinZhao:
Depends on #383, #384.

PR #226, opened 2026-02-21
Implement fieldInBounds preservation for setArguments and pushArguments

math-fehr:
This should simplify a lot the proof required in #222

### DIFF
--- a/Veir/IR/Fields.lean
+++ b/Veir/IR/Fields.lean
@@ -739,6 +739,27 @@ theorem BlockPtr.allocEmpty_fieldsInBounds (heq : allocEmpty ctx = some (ctx', p
     ctx.FieldsInBounds → ctx'.FieldsInBounds := by
   prove_fieldsInBounds
 
+attribute [local grind →] Array.getElem_mem in
+attribute [local grind] Block.empty in
+attribute [local grind =] BlockArgumentPtr.inBounds_def in
+attribute [local grind =>] BlockPtr.allocEmpty_genericPtr_iff' in
+@[grind .]
+theorem BlockPtr.setArguments_fieldsInBounds
+    (hIncreaseSize : block.getNumArguments! ctx ≤ newArguments.size)
+    (hp : ∀ arg ∈ newArguments, arg.FieldsInBounds ctx) :
+    ctx.FieldsInBounds → (setArguments block ctx newArguments h).FieldsInBounds := by
+  prove_fieldsInBounds
+
+attribute [local grind →] Array.getElem_mem in
+attribute [local grind] Block.empty in
+attribute [local grind =] BlockArgumentPtr.inBounds_def in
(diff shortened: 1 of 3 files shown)


Veir/IR/Fields.lean
```diff
@@ -800,6 +800,59 @@ theorem BlockPtr.allocEmpty_fieldsInBounds (heq : allocEmpty ctx = some (ctx', p
...
+    · grind
+    · rintro opr hopr heq
+      constructor <;> grind
+  · intros bl hbl
+    constructor
+    case arguments_inBounds =>
+      intros blockArg blockArgIn blockArgBlock
+      constructor <;> grind [Array.getElem_mem, BlockArgumentPtr.inBounds_def]
+    all_goals grind [=> BlockPtr.allocEmpty_genericPtr_iff', Block.empty]
+  · grind
```
tobiasgrosser: It might be interesting to note in the commit message why these proofs are rather long.

tobiasgrosser (approved):
LGTM

Veir/IR/Fields.lean
math-fehr: Oh, all the proofs in that file are this long, it's just that we prove something with 20-ish cases, and we manually unfold all of them.

luisacicolini: I don't exactly see where all the cases come from, but I am wondering if there is anything you could replace with `cases` or `rcases`? That might make the stratey more evident

luisacicolini (approved):

Veir/IR/Fields.lean
math-fehr: So all of these come from nested `structures`, do you mean using `rcases` so that we have only a single call? Does `rcases` work when you need to `intro` in some cases before recursively casing?

luisacicolini: rcases just splits an inductive type into its constructors, so you can intro first and then rcases on a value of a certain type, and it will be split into the constructors

luisacicolini: yeah, I think all the `constructor` make it a bit hard to read (but readability is also not necessarily sth you need, the proof proves)

ineol: Also, this should easily follow from the previous lemma if we have a `_def` lemma for `pushArgument` no?

ineol (approved):

Veir/IR/Fields.lean
math-fehr: We don't expose `pushArgument`, and I'm not sure we should.
Because `pushArgument` is something like `setArguments ((op.get ctx).arguments.push ...)`, but I would argue we do not want to mention anything that touch the `arguments` field dierctly (and instead should rely on `genNumArguments` and `getArgument`.

luisacicolini:
I'm a fan of the tactic! LGTM :)

PR #237, opened 2026-02-22
Add more bang definitions

math-fehr:
Add all bang definitions in LinkedList.Basic, and add the missing theorems from IR.Basic.


luisacicolini (approved):
lgtm

PR #235, opened 2026-02-21
Add Generic OpInfo to IRContext and Operation

tobiasgrosser:
We introduce a new typeclass `OpInfo` which augments an inductive carrying the `OpCode`s with information about which `Properties` are carried by a given `OpCode`. These changes make most of the `IR` datastructures generic over the specific set of `OpCodes` we are using. `OpCode.lean` still leaks into the IR datastructures via the `Properties.lean` include. In subsequent PRs, we will eliminate this dependency. When finished, one can add operations without recompiling our core data structures.

The main changes in this PR are in `OpInfo.lean`, as well as in `Operation` and `IRContext`. Otherwise, we mostly funnel the `OpInfo` type through the different files.

We do not yet generalize all of our Parser/Printer/Interpreter infrastructure but hardcode them to `OpCode`. Even the rewriters are not as generic as they could be. We leave this to future PRs to keep this PR focused.

PR #238, opened 2026-02-22
InsertPoint: avoid the use of all-implicit theorems

tobiasgrosser:
Theorems that are not equational rewrites, but inequalities or properties should take their arguments as explicit arguments. Forcing all implicit arguments to be explicit at the call site will cause unrelated breakage, e.g., in the upcoming PR #235. To make the upcoming PR less noisy, we remove uses of `@` at call sites in `InsertPoint`.

### DIFF
--- a/Veir/Rewriter/InsertPoint.lean
+++ b/Veir/Rewriter/InsertPoint.lean
@@ -207,7 +207,7 @@ theorem InsertPoint.idxIn.getElem? :
     simp only [next_before_eq]
     apply Array.getElem?_idxOf
     suffices _ : op ∈ blockPtr.operationList ctx ctxWf blockInBounds by grind
-    have := @BlockPtr.operationListWF ctx blockPtr blockInBounds ctxWf
+    have := BlockPtr.operationListWF ctx blockPtr blockInBounds ctxWf
     have := this.allOpsInChain
     grind
   case atEnd bl =>


Veir/IR/WellFormed.lean
```diff
@@ -791,7 +791,8 @@ theorem Region.WellFormed_unchanged
 noncomputable def BlockPtr.operationList (block : BlockPtr) (ctx : IRContext) (hctx : ctx.WellFormed) (hblock : block.InBounds ctx) : Array OperationPtr :=
   (hctx.opChain block hblock).choose
 
-theorem BlockPtr.operationListWF {hctx : IRContext.WellFormed ctx} :
+theorem BlockPtr.operationListWF (ctx : IRContext) (block : BlockPtr) (hblock : block.InBounds ctx)
+  (hctx : ctx.WellFormed) :
```
math-fehr, line 795: In these kind of theorems, should we have `ctx` and `block` as impliict or explicit, since they are present in `hblock` and `hctx`?

tobiasgrosser: I guess this depends. If `hblock` is always given as a value, we can use `ctx` and `block` implicitly. If `hblock` may be derived using `grind`, we probably want them to be explicit, no?

math-fehr: Yeah that's probably true, we should find a good rule at some point!

tobiasgrosser: OK, I merge this as is for now.

PR #250, opened 2026-02-26
Add nsw and nuw flags to arith.addi, arith.muli, llvm.add, and llvm.mul operations.

math-fehr:
Also, support the flags in the interpreter, and test them for the `llvm` dialect.


Test/Arith/identity.mlir
```diff
@@ -33,33 +39,39 @@
 }) : () -> ()
 
 // CHECK:      "builtin.module"() ({
-// CHECK-NEXT: ^4():
```
math-fehr, line 36: I updated the test to use `%{{.*}}` instead of using the ssa value name.
This is more future-proof

PR #254, opened 2026-03-01
Make InBounds predicate decidable

math-fehr:


luisacicolini (approved):

Veir/IR/Basic.lean
```diff
@@ -1888,6 +1936,9 @@ variable {ctx : IRContext OpInfo}
 @[simp, grind =, grind =_] theorem iff_value (ptr : ValuePtr) : (value ptr).InBounds ctx ↔ ptr.InBounds ctx := by grind [InBounds]
 @[simp, grind =, grind =_] theorem iff_opOperandPtr (ptr : OpOperandPtrPtr) : (opOperandPtr ptr).InBounds ctx ↔ ptr.InBounds ctx := by grind [InBounds]
 
+instance : Decidable (InBounds ptr ctx) := by
```
ineol, line 1950: non terminal simp.

Veir/IR/Basic.lean
```diff
@@ -289,6 +289,11 @@ namespace OperationPtr
 def InBounds (op : OperationPtr) (ctx : IRContext OpInfo) : Prop :=
   op ∈ ctx.operations
 
+def inBounds_def : InBounds op ctx ↔ op ∈ ctx.operations := by rfl
+
+instance : Decidable (InBounds op ctx) := by
+  simp [inBounds_def]; infer_instance
```
```suggestion
@[no_expose]
instance : Decidable (InBounds op ctx) := by
  unfold InBounds; infer_instance
```

it feels a bit better, because I don't think it's a good idea to use `simp` to define a runtime definition. The `@[no_expose]` is required so that the body of the instance be in the private scope and has access to the definition of `InBounds`

Veir/IR/Basic.lean
math-fehr: What's the best way to do this?
I don't really want to have to list the 10 different simp lemmas that this is using (the ones right on top).
Is it fine if I use `try infer_instance` at the end?

math-fehr: Is this okay now?

Veir/IR/Basic.lean
math-fehr: Yeah I used it for all instances, so should be good now

Veir/IR/Basic.lean
```diff
@@ -291,8 +291,9 @@ def InBounds (op : OperationPtr) (ctx : IRContext OpInfo) : Prop :=
 
 def inBounds_def : InBounds op ctx ↔ op ∈ ctx.operations := by rfl
 
+@[no_expose]
 instance : Decidable (InBounds op ctx) := by
-  simp [inBounds_def]; infer_instance
+  simp only [inBounds_def]; infer_instance
```
ineol: you can just `unfold` instead of `simp` with the `no_expose`

math-fehr: Should be done now!

PR #256, opened 2026-03-01
Add getOperands! definition and GetSet lemmas

math-fehr:
`getOperands!` is a quite useful function to have, especially when proving things about rewrites and semantics.

### DIFF
--- a/Veir/IR/Basic.lean
+++ b/Veir/IR/Basic.lean
@@ -204,6 +204,14 @@ theorem default_results_eq :
 
 end Operation
 
+namespace OpOperand
+
+theorem default_value_eq :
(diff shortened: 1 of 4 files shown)


Veir/IR/Basic.lean
```diff
@@ -204,6 +204,14 @@ theorem default_results_eq :
 
 end Operation
 
+namespace OpOperand
+
+theorem default_value_eq :
```
tobiasgrosser, line 209: Should this be `simp`?

math-fehr: We didn't mark `simp` any of the `default` so that we only rely on them explicitely. At least this was the logic we had

PR #263, opened 2026-03-03
Refactor interpretOp

math-fehr:
This refactors the following:
* Use getOperandValues and setResultValues to simplify `interpretOp` implementation, making it easier to formalise
* Make `interpretOp'` not depend from the context anymore, and move the operation getters to `interpretOp` instead. This also helps for formalization and automation


tobiasgrosser:
Nice!

PR #264, opened 2026-03-03
Dataflow Analysis Framework

axelcool1234:
The following is a dataflow analysis framework based off the C++ MLIR implementation.

An analysis I've implemented on top of the framework is SCCP: https://github.com/opencompl/veir/pull/456

### DIFF
--- a/UnitTest/DataFlowFramework.lean
+++ b/UnitTest/DataFlowFramework.lean
@@ -1,0 +1,77 @@
+import Veir.Analysis.DataFlowFramework 
+import Veir.Parser.MlirParser
+
+open Std (HashMap)
+open Veir
+open Veir.Parser
+
+namespace UnitTest.DataFlowFramework
+
+abbrev MismatchReport := Array String
+
+def renderReport (report : MismatchReport) : String :=
+  if report.isEmpty then
+    "ok"
+  else
+    "mismatches:\n" ++ String.intercalate "\n" report.toList
+
+def parseTopLevelOp (s : String) : Except String (OperationPtr × MlirParserState) := do
+  let some (ctx, _) := IRContext.create OpCode
+    | throw "internal error: failed to create IR context"
+  let parserState ← ParserState.fromInput (s.toByteArray)
+  let (op, mlirState, _) ← (parseOp none).run (MlirParserState.fromContext ctx) parserState
+  pure (op, mlirState)
+
+def labeledBlockNames (mlir : String) : Array String := Id.run do
+  let mut names := #[]
+  for line in mlir.splitOn "\n" do
+    let trimmed := (line.trimAsciiStart).toString
+    if trimmed.startsWith "^" then
+      let name := ((trimmed.drop 1).takeWhile fun c => c != ':' && c != '(').toString
+      names := names.push name
+  names
+
+partial def collectBlocksDepthFirst
+    (op : OperationPtr)
+    (irCtx : IRContext OpCode)
+    (acc : Array BlockPtr := #[]) : Array BlockPtr := Id.run do
+  let mut acc := acc
+  for i in [0:op.getNumRegions! irCtx] do
+    let region := (op.getRegion! irCtx i).get! irCtx
+    let mut maybeBlock := region.firstBlock
+    while h : maybeBlock.isSome do
+      let block := maybeBlock.get h
+      acc := acc.push block
+      let mut maybeOp := (block.get! irCtx).firstOp
+      while hOp : maybeOp.isSome do
+        let nestedOp := maybeOp.get hOp
+        acc := collectBlocksDepthFirst nestedOp irCtx acc
+        maybeOp := (nestedOp.get! irCtx).next
+      maybeBlock := (block.get! irCtx).next
+  acc
+
+def namedBlocks?
+    (top : OperationPtr)
+    (irCtx : IRContext OpCode)
+    (mlir : String) : Option (HashMap String BlockPtr) :=
+  let names := labeledBlockNames mlir
+  let blocks := collectBlocksDepthFirst top irCtx
+  if names.size != blocks.size then
+    none
+  else
+    some <| Id.run do
+      let mut blockMap : HashMap String BlockPtr := HashMap.emptyWithCapacity names.size
+      for i in [0:names.size] do
+        blockMap := blockMap.insert names[i]! blocks[i]!
+      blockMap
+
+def runWithAnalyses
+    (mlir : String)
+    (analyses : Array DataFlowAnalysis)
+    (check : OperationPtr -> DataFlowContext -> MlirParserState -> MismatchReport) : String :=
+  match parseTopLevelOp mlir with
+  | .error err =>
+    s!"parse failed: {err}"
+  | .ok (top, parserState) =>
+    let dfCtx := fixpointSolve top analyses parserState.ctx
+    renderReport (check top dfCtx parserState)
--- a/UnitTest/DataFlowFramework.lean
+++ b/UnitTest/DataFlowFramework.lean
@@ -7,22 +7,40 @@ open Veir.Parser
 
 namespace UnitTest.DataFlowFramework
 
+/--
+Human readable mismatches collected by a dataflow test.
+-/
 abbrev MismatchReport := Array String
 
+/--
+Render a mismatch report in the format expected by the `#guard_msgs` tests below.
+-/
 def renderReport (report : MismatchReport) : String :=
   if report.isEmpty then
     "ok"
   else
     "mismatches:\n" ++ String.intercalate "\n" report.toList
 
+/--
+Parse one top level MLIR operation together with the parser state that owns its IR context.
+-/
 def parseTopLevelOp (s : String) : Except String (OperationPtr × MlirParserState) := do
-  let some (ctx, _) := IRContext.create OpCode
+  let some (ctx, _) := WfIRContext.create OpCode
     | throw "internal error: failed to create IR context"
-  let parserState ← ParserState.fromInput (s.toByteArray)
-  let (op, mlirState, _) ← (parseOp none).run (MlirParserState.fromContext ctx) parserState
-  pure (op, mlirState)
+  let parserState ←
+    match ParserState.fromInput s.toByteArray with
+    | .ok parserState => pure parserState
+    | .error err => throw (toString err)
+  match (parseOp none).run (MlirParserState.fromContext ctx) parserState with
+  | .ok (op, mlirState, _) =>
+    pure (op, mlirState)
+  | .error err =>
+    throw (toString err)
 
-def labeledBlockNames (mlir : String) : Array String := Id.run do
+/--
+Recover textual block labels such as `^bb0` from an MLIR snippet string.
+-/
+def blockLabelsFromMlir (mlir : String) : Array String := Id.run do
   let mut names := #[]
   for line in mlir.splitOn "\n" do
     let trimmed := (line.trimAsciiStart).toString
@@ -31,7 +49,40 @@ def labeledBlockNames (mlir : String) : Array String := Id.run do
       names := names.push name
   names
 
-partial def collectBlocksDepthFirst
+/--
+Extract one textual SSA name from a definition fragment such as `%x : i32` or `%x`.
+-/
+def parseSsaName? (fragment : String) : Option String :=
+  let trimmed := (fragment.trimAsciiStart).toString
+  if trimmed.startsWith "%" then
+    let name := ((trimmed.drop 1).takeWhile fun c => c != ':' && c != ' ' && c != '\t').toString
+    if name.isEmpty then none else some name
+  else
+    none
+
+/--
+Recover textual SSA names from block arguments and operation result definitions.
+-/
+def definedSsaNamesFromMlir (mlir : String) : Array String := Id.run do
+  let mut names := #[]
+  for line in mlir.splitOn "\n" do
+    let trimmed := (line.trimAsciiStart).toString
+    if trimmed.startsWith "^" then
+      let args := (((trimmed.splitOn "(").drop 1 |>.headD "").takeWhile fun c => c != ')').toString
+      for arg in args.splitOn "," do
+        if let some name := parseSsaName? arg then
+          names := names.push name
+    else if trimmed.startsWith "%" then
+      let lhs := (trimmed.splitOn "=").headD ""
+      for result in lhs.splitOn "," do
+        if let some name := parseSsaName? result then
+          names := names.push name
+  names
+
+/--
+Collect all blocks reachable by recursively traversing nested regions in source order.
+-/
+partial def collectBlocksInSourceOrder
     (op : OperationPtr)
     (irCtx : IRContext OpCode)
     (acc : Array BlockPtr := #[]) : Array BlockPtr := Id.run do
@@ -45,26 +96,72 @@ partial def collectBlocksDepthFirst
       let mut maybeOp := (block.get! irCtx).firstOp
       while hOp : maybeOp.isSome do
         let nestedOp := maybeOp.get hOp
-        acc := collectBlocksDepthFirst nestedOp irCtx acc
+        acc := collectBlocksInSourceOrder nestedOp irCtx acc
         maybeOp := (nestedOp.get! irCtx).next
       maybeBlock := (block.get! irCtx).next
   acc
 
-def namedBlocks?
+/--
+Collect all SSA values in source order.
+This includes the operation's results, followed by block arguments and nested
+operation results inside each region.
+-/
+partial def collectDefinedValuesInSourceOrder
     (top : OperationPtr)
     (irCtx : IRContext OpCode)
-    (mlir : String) : Option (HashMap String BlockPtr) :=
-  let names := labeledBlockNames mlir
-  let blocks := collectBlocksDepthFirst top irCtx
-  if names.size != blocks.size then
+    (acc : Array ValuePtr := #[]) : Array ValuePtr := Id.run do
+  let mut acc := acc
+  for i in [0:top.getNumResults! irCtx] do
+    acc := acc.push (ValuePtr.opResult (top.getResult i))
+  for i in [0:top.getNumRegions! irCtx] do
+    let region := (top.getRegion! irCtx i).get! irCtx
+    let mut maybeBlock := region.firstBlock
+    while h : maybeBlock.isSome do
+      let block := maybeBlock.get h
+      for j in [0:block.getNumArguments! irCtx] do
+        acc := acc.push (ValuePtr.blockArgument (block.getArgument j))
+      let mut maybeOp := (block.get! irCtx).firstOp
+      while hOp : maybeOp.isSome do
+        let nestedOp := maybeOp.get hOp
+        acc := collectDefinedValuesInSourceOrder nestedOp irCtx acc
+        maybeOp := (nestedOp.get! irCtx).next
+      maybeBlock := (block.get! irCtx).next
+  acc
+
+/--
+Recovered name maps built by pairing MLIR source names with IR nodes in source order.
+-/
+structure RecoveredNames where
+  blocks : HashMap String BlockPtr
+  values : HashMap String ValuePtr
+
+/--
+Recover block and SSA value maps by pairing MLIR source names with IR traversal order.
+-/
+def recoverNames?
+    (top : OperationPtr)
+    (irCtx : IRContext OpCode)
+    (mlir : String) : Option RecoveredNames :=
+  let blockLabels := blockLabelsFromMlir mlir
+  let blocks := collectBlocksInSourceOrder top irCtx
+  let valueNames := definedSsaNamesFromMlir mlir
+  let values := collectDefinedValuesInSourceOrder top irCtx
+  if blockLabels.size != blocks.size || valueNames.size != values.size then
     none
   else
     some <| Id.run do
-      let mut blockMap : HashMap String BlockPtr := HashMap.emptyWithCapacity names.size
-      for i in [0:names.size] do
-        blockMap := blockMap.insert names[i]! blocks[i]!
-      blockMap
+      let mut blockMap : HashMap String BlockPtr := HashMap.emptyWithCapacity blockLabels.size
+      for i in [0:blockLabels.size] do
+        blockMap := blockMap.insert blockLabels[i]! blocks[i]!
+      let mut valueMap : HashMap String ValuePtr := HashMap.emptyWithCapacity valueNames.size
+      for i in [0:valueNames.size] do
+        valueMap := valueMap.insert valueNames[i]! values[i]!
+      { blocks := blockMap, values := valueMap }
 
+/--
+Parse an MLIR snippet string, run the requested dataflow analyses to a fixpoint, and
+render any test mismatches produced by `check`.
+-/
 def runWithAnalyses
     (mlir : String)
     (analyses : Array DataFlowAnalysis)
@@ -73,5 +170,8 @@ def runWithAnalyses
   | .error err =>
     s!"parse failed: {err}"
   | .ok (top, parserState) =>
-    let dfCtx := fixpointSolve top analyses parserState.ctx
-    renderReport (check top dfCtx parserState)
+    match fixpointSolve top analyses parserState.ctx with
+    | some dfCtx =>
+      renderReport (check top dfCtx parserState)
+    | none =>
+      "analysis did not converge"
--- a/Veir/Analysis.lean
+++ b/Veir/Analysis.lean
@@ -1,0 +1,5 @@
+module
+
+public import Veir.Analysis.DataFlowFramework
+
+public section
--- a/Veir/Analysis.lean
+++ b/Veir/Analysis.lean
@@ -2,4 +2,3 @@ module
 
 public import Veir.Analysis.DataFlowFramework
 
-public section
--- a/Veir/Analysis/DataFlow/Domains.lean
+++ b/Veir/Analysis/DataFlow/Domains.lean
@@ -1,0 +1,17 @@
+module
+
+public section
+
+namespace Veir
+
+/--
+An algebraic definition of a lattice. In particular, no partial order is defined
+on the domain.
+-/
+class LatticeElement (Domain : Type) extends BEq Domain where
+  bottom : Domain
+  top : Domain
+  join : Domain -> Domain -> Domain
+  meet : Domain -> Domain -> Domain
+
+end Veir
--- a/Veir/Analysis/DataFlow/Facts.lean
+++ b/Veir/Analysis/DataFlow/Facts.lean
@@ -1,0 +1,126 @@
+module
+
+public import Std.Data.HashMap
+public import Init.Data.Queue
+public import Veir.Analysis.DataFlow.Domains
+public import Veir.IR.Basic
+public import Veir.GlobalOpInfo
+public import Veir.Rewriter.InsertPoint
+
+open Std (HashMap Queue)
+
+public section
+
+namespace Veir
+
+/-!
+# Facts in the dataflow framework
+
+In the mathematics of monotone dataflow frameworks, an analysis assigns abstract
+states (each from some lattice) to program points and repeatedly applies monotone
+transfer functions until a fixpoint is reached. A fixpoint `f` is such that 
+`f(x) = x`. In other words, when nothing changes.
+
+This file defines the objects used to store those abstract states. A
+`Fact` is one piece of analysis information and a `LatticeAnchor` says 
+where in the IR that information lives (such as an SSA value or a CFG 
+edge).
+-/
+
+namespace InsertPoint
+
+/--
+Return whether the insertion point is at the beginning of a block.
+-/
+def isBlockStart (point : InsertPoint) (irCtx : IRContext OpCode) : Bool :=
+  match point with
+  | .before op =>
+    match (op.get! irCtx).parent with
+    | some block => (block.get! irCtx).firstOp == some op
+    | none => false
+  | .atEnd block => (block.get! irCtx).firstOp == none
+
+end InsertPoint
+
+/--
+A directed control flow edge between two blocks.
+-/
+structure CFGEdge where
+  source : BlockPtr
+  target : BlockPtr
+deriving BEq, Hashable
+
+/--
+The control flow graph positions and SSA values where dataflow facts are attached.
+-/
+inductive LatticeAnchor
+  | InsertPoint (point : InsertPoint)
+  | ValuePtr (value : ValuePtr)
+  | CFGEdge (edge : CFGEdge)
+deriving BEq, Hashable
+
+/-!
+# Analyses and facts
+-/
+
+/--
+Tags to match on for different `DataFlowAnalysis` types.
+-/
+inductive AnalysisKind where
+deriving BEq, Hashable, Repr, DecidableEq
+
+/--
+Tags to match on for different fact types.
+-/
+inductive FactKind where
+deriving BEq, ReflBEq, LawfulBEq, Hashable, Repr, DecidableEq
+
+abbrev WorkItem := InsertPoint × AnalysisKind
+abbrev WorkList := Queue WorkItem
+
+/--
+The fact specific data stored for each fact kind.
+-/
+@[expose] def FactPayload : FactKind -> Type
+  | kind => nomatch kind
+
+/--
+A dataflow fact stored by the framework.
+
+Each fact carries a lattice anchor (some location in the program this fact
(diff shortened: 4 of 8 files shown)


Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+public import Veir.IR.Basic
+public import Veir.GlobalOpInfo
+
+open Std(HashMap Queue)
+
+public section
+
+namespace Veir
+
+inductive ProgramPoint where
```
math-fehr: Instead of defining `ProgramPoint`, can you use `InsertPoint`?
It should be equivalent, and almost all methods are already available for it.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+instance : Hashable CFGEdge :=
+  inferInstanceAs (Hashable (BlockPtr × BlockPtr))
+
+namespace CFGEdge
+
+def getFrom (edge : CFGEdge) : BlockPtr :=
+  edge.1
+
+def getTo (edge : CFGEdge) : BlockPtr :=
+  edge.2
```
math-fehr: If you are anyway defining a new type, it's probably better to use a `structure`. That way, you have for free `getFrom` and `getTo` (which you should rename to `from` and `to`, and you can use `deriving BEq, Hashable`

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+
+end CFGEdge
+
+inductive LatticeAnchor
```
math-fehr: Can you add documentation to that inductive?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+  | ProgramPoint (point : ProgramPoint)
+  | ValuePtr (value : ValuePtr)
+  | CFGEdge (edge : CFGEdge)
+deriving BEq, Hashable
+
+class LatticeElement (Domain : Type) extends BEq Domain where
+  typeNameInst : TypeName Domain
```
math-fehr: What's the difference between `extends TypeName Domain` and asking it as parameter here?
Also, I'm not sure I would require this typeclass to be existing here, this should be probably done elsewhere.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+  top : Domain
+  join : Domain -> Domain -> Domain
+  meet : Domain -> Domain -> Domain
+
+attribute [instance] Veir.LatticeElement.toBEq
+
+@[reducible] instance [LatticeElement Value] : TypeName Value :=
+  LatticeElement.typeNameInst
+
+-- =============================== DataFlowAnalysis ============================== --
```
math-fehr: Instead of having these comments, can you use something like `/-! ## DataFlowAnalysis -/`?
`/-!` delimits markdown documentation.

Same below

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+-- `Dynamic` is ALWAYS `DataFlowContext`; panic if this invariant is broken.
+structure DataFlowAnalysis where
+  private mk ::
+  id : Lean.Name
+  initDyn : OperationPtr -> Dynamic -> IRContext OpCode -> Dynamic
+  visitDyn : ProgramPoint -> Dynamic -> IRContext OpCode -> Dynamic
```
math-fehr: I think it's a bit broken having `Dynamic` instead of `DataFlowContext`. Also, the comment is likely wrong, as I don't think you can panic if the invariant is broken.
Is there a way you can make this a mutual `structure` with `DataFlowContext`? If you have trouble doing this (because Lean has to prove that it terminates), let me know!

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+-- =============================================================================== --
+
+-- ================================== WorkList =================================== --
+-- A queued item stores a program point and the analysis to run.
+@[expose] def WorkItem := ProgramPoint × Lean.Name
+@[expose] def WorkList := Queue WorkItem
```
math-fehr: It's also possible to use `abbrev` instead of `def`, which fully expose the type. Not sure what's best (Maybe @ineol you have some opinions on this?).

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+-- =============================================================================== --
+
+-- ============================= ErasedAnalysisState ============================= --
+-- `valueDyn` stores the concrete state object. `onUpdateDyn` expects `(valueDyn, dfCtxDyn)`.
+structure ErasedAnalysisState where
+  private mk ::
+  valueDyn : Dynamic
+  -- State -> DataFlowContext -> IRContext -> DataFlowContext
+  onUpdateDyn : Dynamic -> Dynamic -> IRContext OpCode -> Dynamic
```
math-fehr: Can you add this comment on the fields directly, with `/-- -/` format?
Also, what is the comment in the middle of the structure?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+-- =============================================================================== --
+
+-- =============================== DataFlowContext =============================== --
+structure DataFlowContext where
+  /-- Registered analyses for the current solver run, keyed by analysis id. -/
+  analyses : HashMap Lean.Name DataFlowAnalysis
```
math-fehr: @ineol, @luisacicolini, is having a `Lean.Name` overkill here or is it the standard practice? Compared to have a `String`.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+
+instance : Inhabited DataFlowContext where
+  default := DataFlowContext.empty
+
+private unsafe def asDataFlowContextImpl (ctxDyn : Dynamic) : DataFlowContext :=
+  unsafeCast ctxDyn
+
+private unsafe def asDynamicImpl (ctx : DataFlowContext) : Dynamic :=
+  unsafeCast ctx
+
```
math-fehr: I think we should avoid this at all costs.
Let's figure out how to encode the data structures correctly in Lean, and hopefully this will go away.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+  { id := id
+    initDyn := fun top dfCtxDyn irCtx =>
+      asDynamic (init top (asDataFlowContext dfCtxDyn) irCtx)
+    visitDyn := fun point dfCtxDyn irCtx =>
+      asDynamic (visit point (asDataFlowContext dfCtxDyn) irCtx)
+  }
+
+end DataFlowAnalysis
+
+structure AnalysisStateHeader where
```
math-fehr: Can you add documentation for this structure and the following classes/definitions? Otherwise it is a bit hard to understand what is happening here.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,351 @@
...
+  if changed then
+    AnalysisState.onUpdate state dfCtx irCtx
+  else
+    dfCtx
+
+end DataFlowContext
+-- =============================================================================== --
+
+-- =============================== Fixpoint Solver =============================== --
+partial def run (dfCtx : DataFlowContext) (irCtx : IRContext OpCode) : DataFlowContext :=
```
math-fehr: I think it is better to return an `Option` here (where `None` represents an infinite loop).

That way, we will be able to actually write theorems about it.

UnitTest/DataFlowFramework.lean
```diff
@@ -0,0 +1,77 @@
+import Veir.Analysis.DataFlowFramework 
+import Veir.Parser.MlirParser
```
math-fehr, line 2: It would be nice to add some documentation here, to explain what these functions are used for.

UnitTest/DataFlowFramework.lean
```diff
@@ -0,0 +1,77 @@
...
+
```
math-fehr, line 3: Do you think you could a very very dumb and easy analysis to test the framework here? Something like having an `Int` lattice, which associates `1` to each operation?

math-fehr (commented):
Sorry for the delay!
I think what would help a lot reviewing this is adding more documentation on the later part of the `Veir/Analysis/DataFlowFramework.lean` file.
Maybe adding a markdown documentation (`/-! -/`) at the beginning of the function to explain the relationship between the different structures?

Otherwise, the main issue I see right now is the structures that are using your `Dynamic` types. Do you think you can see a way to define them with a `mutual` somehow? If you don't, we can try to look into it.

Veir/Analysis/DataFlowFramework.lean
ineol: It depends what we want. expose is for the module system, so that other modules can unfold the definition. `abbrev` says that this definition is transparent and should be always unfolded.

Veir/Analysis/DataFlowFramework.lean
ineol: I think the one difference is that if you have it as fields you also have to add instances 
```lean
instance : LatticeElement D -> BEq D := fun l => l.instBEq
...
```

Veir/Analysis/DataFlowFramework.lean
ineol: I don't think we should use `Lean.Name` in non meta programming settings (indeed the `Lean` namespace is for Lean's  internal data structures). It's a basically a list of Strings s a String would also be more efficient for our use cases I think.

axelcool1234:
Thank you for the review comments. I'll try to get back to them as soon as I can (my laptop decided to have some hardware issues so it might be a second, but I _will_ get to them). I apologize for the poor documentation (I'm still getting used to collaborating with others - didn't have much of that at this scale in my undergrad unfortunately). I should've done an additional once over my PRs and ensure I've commented on everything for better clarity. I'll make sure to do that as well.

I will also reconsider the structure of the framework as a whole now that I have a strong understanding of how the original MLIR version that this is mirroring works. I have a couple of ideas of some restructuring that should eliminate the strange hacks/workarounds and make this simpler. Once I get my laptop up and working I will work on a proposal for this via an additional commit.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: Good idea. Ideally this'll go away entirely once a verified version is written as we can prove it terminates with monotonicity proofs or something.

UnitTest/DataFlowFramework.lean
axelcool1234: Good idea. I'll work on this.

UnitTest/DataFlowFramework.lean
axelcool1234: Yep, I'll make sure to document these.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: Yeah you're right. I'll reconsider my approach on the structure of the framework. I have a stronger sense of how dataflow analysis is achieved in MLIR having mirrored it to Lean. I believe I now have the confidence to make a very drastic turn in the structure to make it more Lean-idiomatic and so that we can avoid this hack entirely. I'll work on this soon as an additional commit. It might or might not involve having every analysis in an inductive type. Ideally it wouldn't, but if I truly can't figure out a way to avoid that then that'll be the solution as we discussed on Slack.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: Yeah I stuck with Lean.Name simply because that's why Dynamic was using under the hood. I'll switch to String.

axelcool1234:
@math-fehr @ineol I have pushed the changes I've mentioned in our last meeting! This should be cleaner I hope. 

I think I need to think about how to organize the `Analysis/` folder - I currently have this strange split between `DataFlowFramework.lean` and `Facts.lean` (which is a misleading name for the file as it's doing more than just fact related stuff) to prevent import cycles.

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+public section
+
+namespace Veir
+
+/-!
+# Control flow positions
+-/
+
+inductive ProgramPoint where
+  | inBlock (block : BlockPtr) (nextOp : Option OperationPtr)
```
math-fehr: I guess you should use `InsertPoint` here, as this is exactly the kind of interface you want (either before an operation, or the end of a block)?

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+  | standaloneOp (op : OperationPtr)
```
math-fehr: Is `standaloneOp` ever needed? Is that something that MLIR has as well?
Usually, the toplevel operation is a `module`, so we should never really have a standalone operation?

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+deriving BEq, Hashable
+
+namespace ProgramPoint
+
+/--
+Create the program point at the beginning of a block.
+-/
+def beforeBlock (block : BlockPtr) (irCtx : IRContext OpCode) : ProgramPoint :=
```
math-fehr: I would probably rename this `blockStart`?
And below renaming `afterBlock` to `blockEnd`.

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+
+/-!
+# Analyses and facts
+-/
+
+inductive AnalysisKind where
+deriving BEq, Hashable, Repr, DecidableEq
+
+inductive FactKind where
+deriving BEq, ReflBEq, LawfulBEq, Hashable, Repr, DecidableEq
```
math-fehr, line 76: Can you add documentation of what these two inductives are going to be?

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+
+abbrev WorkItem := ProgramPoint × AnalysisKind
+abbrev WorkList := Queue WorkItem
+
+/--
+Shared information carried by every dataflow fact.
+-/
+structure FactHeader where
+  anchor : LatticeAnchor
+  dependents : Array WorkItem := #[]
```
math-fehr, line 97: What is this information supposed to be? Can you document this? Is it the location where the fact is, and the locations that needs to be updated upon computation of the fact?
i.e., document what it means, in addition to how it is used.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+public section
+
+namespace Veir
+
+/-!
+# Dataflow framework 
+-/
+
+/--
+The mutable solver state (product lattice and work list)
```
math-fehr:
```suggestion
The solver state containing all dataflow facts and the worklist of location to update.
```

Veir/Analysis/DataFlow/Facts.lean
math-fehr: I also feel that `FactHeader` is not that descriptive, do you have an idea for a better name?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+
+/--
+Specification for fact types.
+
+This provides the fact kind, the proof that the fact is the payload of
+that kind, the default value, and the propagation hook triggered when that fact 
+changes.
+-/
+class FactSpec (α : Type) extends HasFactKind α where
+  mkDefault : LatticeAnchor -> α
```
math-fehr:
```suggestion
  mkTop : LatticeAnchor -> α
```
If I understand correctly, this should always be the top value, or am I misunderstanding what is happening?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+    workList := .empty }
+
+/--
+Specification for fact types.
+
+This provides the fact kind, the proof that the fact is the payload of
+that kind, the default value, and the propagation hook triggered when that fact 
+changes.
+-/
+class FactSpec (α : Type) extends HasFactKind α where
```
math-fehr: What is alpha supposed to be here? Maybe you sholud change the name of it?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+  kind : AnalysisKind
+  init : OperationPtr -> DataFlowContext -> IRContext OpCode -> DataFlowContext
+  visit : ProgramPoint -> DataFlowContext -> IRContext OpCode -> DataFlowContext
+
+namespace DataFlowContext
+
+/--
+Read the fact of type `α` stored at `anchor`, if any. 
+-/
+def getFact? [HasFactKind α]
```
math-fehr: I think you should add an explicit `alpha` argument here, otherwise there is no way to call that function without writing `getFact? (alpha = ...)`.
Also, is there a reason `HasFactKind` is a class? Would it be bad to pass it as an argument explicitely, like a structure?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+    (ctx : DataFlowContext) (anchor : LatticeAnchor) : Option α := do
+  let facts ← ctx.lattice.get? anchor
+  let payload <- DHashMap.get? facts (HasFactKind.kind (α := α))
+  pure <| HasFactKind.ofPayload payload
+
+/--
+Read the fact of type `α` at `anchor`, creating the default fact if it is absent.
```
math-fehr: Can you document that the context is unchanged, just making sure that reader won't think that this will insert that default fact in the context.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+def getFact? [HasFactKind α]
+    (ctx : DataFlowContext) (anchor : LatticeAnchor) : Option α := do
+  let facts ← ctx.lattice.get? anchor
+  let payload <- DHashMap.get? facts (HasFactKind.kind (α := α))
+  pure <| HasFactKind.ofPayload payload
+
+/--
+Read the fact of type `α` at `anchor`, creating the default fact if it is absent. 
+-/
+def getOrMkFact [FactSpec α]
```
math-fehr: I would rename this `getFactD`, this is the convention in Lean.
Maybe this should be renamed `getD` and `get?` for the previous function, but I don't feel strongly about it.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,149 @@
...
+    (ctx : DataFlowContext) (anchor : LatticeAnchor) : α :=
+  match ctx.getFact? (α := α) anchor with
+  | some fact => fact
+  | none => FactSpec.mkDefault (α := α) anchor
+
+/--
+Overwrite the stored fact for `anchor`. 
+-/
+private def writeFact [FactSpec α]
```
math-fehr: Since you named the other functions `get...`, should you name this one `set...`?

Veir/Analysis.lean
```diff
@@ -0,0 +1,5 @@
+module
+
+public import Veir.Analysis.DataFlowFramework
+
+public section
```
math-fehr:
```suggestion
```

math-fehr (commented):
Nice! It looks much much better now!
I just added comments mostly about documentation and naming!

I guess to me the biggest question is wether or not some `class` should be `structure` instead, like `HasFactKind` (which I'm not 100% sure I understand what it does). Other than that, the structure looks good now to me!

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+/--
+An abstract domain equipped with the lattice operations used by the framework.
+-/
+class LatticeElement (Domain : Type) extends BEq Domain where
+  bottom : Domain
+  top : Domain
+  join : Domain -> Domain -> Domain
+  meet : Domain -> Domain -> Domain
+
+attribute [instance] Veir.LatticeElement.toBEq
```
ineol:
```suggestion
attribute [instance] LatticeElement.toBEq
```

ineol: Are you sure it's needed? I would have thought the extends already does that

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,175 @@
...
+/--
+Payload type stored for each fact kind.
+-/
+@[expose] def FactPayload : FactKind -> Type
+  | kind => nomatch kind
+
+/--
+Identifies which `FactKind` a fact type inhabits.
+-/
+class HasFactKind (α : Type) where
```
ineol: Can you explain the purpose of this class?

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: `standaloneOp` is used in MLIR, but if our version never has an operation pointer outside of a block then I could simplify it.

axelcool1234: Here's the relevant code:
https://github.com/llvm/llvm-project/blob/38c36f9c8269661bae13bfbe006d78622897e645/mlir/include/mlir/Analysis/DataFlowFramework.h#L140-L147

Can an operation pointer ever lack a parent block?

Veir/Analysis/DataFlowFramework.lean
axelcool1234: My analyses actually start at bottom, not top. The name `mkDefault` is ambiguous because you can make an analysis with top/meet instead of bottom/join since they are dual.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: It's the custom fact type you are implementing a "Fact Specification" (`FactSpec`) for. I just renamed alpha to `FactType` and updated the comment.

axelcool1234: Okay I broke a bunch of stuff. I'm gonna clean this up (everything uses `FactType` instead of alpha) along with merging `HasFactKind` into `FactSpec` and making `FactType` be an explicit argument in methods.

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: `FactHeader` is the information every single dataflow fact needs to have. A lattice fact always associates itself with some sort of location in the program (thus the `anchor : LatticeAnchor`), as well as other lattice facts that need to be updated if this lattice fact is updated (`dependents`). I will make sure to explain this in the comment. I am not sure of a better name than `FactHeader` though. I'll think about it. Maybe `FactBase`? Feels just as ambiguous.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: usually `getD` requires you provide a specific default. However, this always calls our `mkDefault` method. Is it still appropriate to call it `getFactD`?

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: It was unnecessary, I made it a part of `FactSpec` now.

Veir/Analysis/DataFlow/Domains.lean
```diff
@@ -0,0 +1,18 @@
...
+
+public section
+
+namespace Veir
+
+/--
+An abstract domain equipped with the lattice operations used by the framework.
+
+The solver only relies on equality and the four distinguished lattice operations below.
+-/
```
math-fehr, line 10:
```suggestion
/--
An algebraic definition of a lattice. In particular, no partial order is defined
on the domain.
-/
```

I prefer not talking about how the solver or the framework uses this here.

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -0,0 +1,140 @@
...
+-/
+
+/--
+Tags to match on for different `DataFlowAnalysis` types.
+-/
+inductive AnalysisKind where
+deriving BEq, Hashable, Repr, DecidableEq
+
+/--
+Tags to match on for different fact types.
```
math-fehr, line 73: Maybe explain what is a fact here? I don't think most people would know what a `fact` is supposed to be at lesat.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,167 @@
...
+/--
+Implement this class to register a custom type to be recognized as
+a fact type by the dataflow framework.
+`mkDefault`: Default state a fact starts in. Typically either bottom or top.
+`propagate`: Hook that's called when the fact changes state. Typically used
+             to enqueue a fact's dependents because it changed.
+-/
+class FactSpec (kind : FactKind) where
+  mkDefault : LatticeAnchor -> Fact kind
+  propagate : Fact kind -> DataFlowContext -> IRContext OpCode -> DataFlowContext
```
math-fehr, line 42: Can you add documentatino here on what this function is supposed to do? More importantly, what is `Fact kind` here?
Is there any case where `propagate` would do anything else than adding its content to the dataflow context?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,167 @@
...
+
+/--
+Implement this class to register a custom type to be recognized as
+a fact type by the dataflow framework.
+`mkDefault`: Default state a fact starts in. Typically either bottom or top.
+`propagate`: Hook that's called when the fact changes state. Typically used
+             to enqueue a fact's dependents because it changed.
+-/
+class FactSpec (kind : FactKind) where
+  mkDefault : LatticeAnchor -> Fact kind
```
math-fehr, line 37: I believe this should probably also include an `IRContext`? Or maybe I'm wrong there?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,167 @@
...
+  FactSpec.mkDefault (kind := kind) anchor
+
+/--
+Run the fact kind's propagation hook.
+-/
+def propagate [FactSpec kind]
+    (fact : Fact kind)
+    (ctx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext :=
+  FactSpec.propagate (kind := kind) fact ctx irCtx
```
math-fehr, line 59: Why is `kind := kind` needed here?

math-fehr: And actually why do you need this function instead of just calling `FactSpec.propagate` directly?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,167 @@
...
+
+end Fact
+
+/--
+A single transfer problem scheduled by the fixpoint solver.
+-/
+structure DataFlowAnalysis where
```
math-fehr, line 66: Can you add documentation for each of these fields?
For instance, in `init`, what is supposed to be passed? The module?
For `visit`, why do we pass an `InsertPoint`, shouldn't it be a `LatticeAnchor`?

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,167 @@
...
+  let facts := (ctx.lattice.getD anchor ∅).insert kind fact
+  { ctx with lattice := ctx.lattice.insert anchor facts }
+
+/--
+Apply an update with `f` to the fact of kind `kind` stored at `anchor`. 
+-/
+def modifyFact (kind : FactKind) [FactSpec kind]
+    (ctx : DataFlowContext) (anchor : LatticeAnchor) (f : Fact kind -> Fact kind) : DataFlowContext :=
+  let current := ctx.getOrMkFact kind anchor
+  ctx.setFact kind anchor (f current)
```
math-fehr, line 122: Should this made private? Is there a reason we should ever use it?

math-fehr (commented):
Okay that looks much much better again, nice!
I added some comments for documentation (mostly places where I feel it is harder to understand what's going on).
I'll wait that @ineol reviews it (probably early next week), and to me that should be probably good to go after that!

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: I just wrote a top level comment for `Facts.lean`.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: `propagate` for `DominanceAnalysis` simply enqueues the fact's dependents. However, for SCCP, it enqueues the fact's dependents and then enqueues "subscribed analyses", an abstraction used by MLIR's dataflow framework. We can discuss whether or not this abstraction even makes sense with our version. There may be a possible simplification here, but I think we should discuss this in the SCCP PR.

axelcool1234: Additionally, `Fact kind` is documented for the `Fact` and `FactKind` type.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: `mkDefault` is more like an internal method that is then used by `DataFlowContext` methods. So you'd call `DataFlowContext.getOrMkFact` which would insert the result of `mkDefault` into the dataflow context.

It doesn't need `IRContext` either, because it doesn't need it. There's no checks going on to see if `LatticeAnchor` is valid, it just assumes the `InsertPoint`, `ValuePtr`, or `CFGEdge` that the `LatticeAnchor` is made up of is valid and creates a `Fact` with it as its `anchor`.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: An implementation of a custom `Fact` type will not need to take `kind` as a parameter because we do `(kind := kind)` here. 

When we are handling `Fact` objects, such as if we had a `Fact` object `f`, we can call `f.propagate dfCtx irCtx` directly instead of `FactSpect.propagate (kind := kind) f dfCtx irCtx`. It's purely for more comfortability and readability when writing new analyses.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: I'm writing the documentation, thanks for pointing this out.

For `visit`, this requires an explanation. With how the dataflow framework is currently designed, the worklist of the fixpoint loop schedules _points in the program_ (`InsertPoint`) for the transfer function to execute on rather than _fact location_ (`LatticeAnchor`) for the transfer function to execute on. So when a control flow edge fact turns live for dead code analysis, the block start of the destination block of that edge is enqueued into the worklist. Then when that is dequeued, we continue with the dead code analysis of finding successor blocks and determining if those would now turn live because this block has turned live. We can discuss whether this distinction is worthwhile if you'd like.

Veir/Analysis/DataFlowFramework.lean
axelcool1234: There are some modifications that can be made to a custom `Fact` type that does not necessitate calling the `propagate` hook. For example, the subscribed analyses abstraction I brought up in another comment. We should discuss this further soon.

Veir/Analysis/DataFlowFramework.lean
```diff
@@ -0,0 +1,172 @@
...
+  mkDefault : LatticeAnchor -> Fact kind
+  propagate : Fact kind -> DataFlowContext -> IRContext OpCode -> DataFlowContext
+
+namespace Fact
+
+/--
+Construct the default fact for a given lattice anchor. 
+-/
+def mkDefault (kind : FactKind) [FactSpec kind] (anchor : LatticeAnchor) : Fact kind :=
+  FactSpec.mkDefault (kind := kind) anchor
```
math-fehr, line 50:
```suggestion
def mkDefault (kind : FactKind) [FactSpec kind] (anchor : LatticeAnchor) : Fact kind :=
  FactSpec.mkDefault kind anchor
```
Does this works?

axelcool1234:
```
Application type mismatch: The argument
  kind
has type
  FactKind
but is expected to have type
  LatticeAnchor
in the application
  FactSpec.mkDefault kind
```

Seems no. Refer to this for why I have this:
https://github.com/opencompl/veir/pull/264#discussion_r3263356734

PR #267, opened 2026-03-03
add RISCV immediate arithmetic ops

luisacicolini:
This PR adds a first batch of the RISCV immediate arithmetic operations.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -172,19 +172,16 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
     let imm := BitVec.ofNat 20 properties.value.value.toNat
     return (#[.reg (RISCV.auipc imm op)], .continue)
   | .riscv_addi => do
-    let value := opPtr.getProperties! ctx .riscv_addi
     let #[.reg op] := operands | none
-    let imm := BitVec.ofNat 12 value.value.value.toNat
+    let imm := BitVec.ofNat 12 properties.value.value.toNat
     return (#[.reg (RISCV.addi imm op)], .continue)
   | .riscv_slti => do
-    let value := opPtr.getProperties! ctx .riscv_slti
     let #[.reg op] := operands | none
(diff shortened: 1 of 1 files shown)


Veir/Data/RISCV/Reg/Basic.lean
```diff
@@ -39,6 +39,33 @@ def lui (imm : BitVec 20) : Reg :=
 def auipc (imm : BitVec 20) (pc : BitVec 64) : BitVec 64 :=
   BitVec.add (BitVec.signExtend 64 (BitVec.append imm (0x0 : BitVec 12))) pc
 
+/--
+  Adds the sign-extended 12-bit immediate to register rs1. Arithmetic overflow is ignored and the
+  result is simply the low XLEN bits of the result. ADDI rd, rs1, 0 is used to implement the MV
```
math-fehr:
```suggestion
  result is simply the low 64 bits of the result. ADDI rd, rs1, 0 is used to implement the MV
```

Veir/Data/RISCV/Reg/Basic.lean
```diff
@@ -63,3 +90,31 @@ def ori (imm : BitVec 12) (rs1_val : BitVec 64) : BitVec 64 :=
...
+/--
+  Performs logical right shift on the value in register rs1 by the shift amount held in the lower 5
+  bits of the immediate In RV64, bit-25 is used to shamt[5].
+-/
+def srli (shamt : BitVec 6) (rs1_val : BitVec 64) : BitVec 64 := rs1_val >>> shamt
+
+/--
+  Performs arithmetic right shift on the value in register rs1 by the shift amount held in the
+  lower 5 bits of the immediate In RV64, bit-25 is used to shamt[5].
+-/
```
math-fehr: I don't think you should talk about `bit-25` here, since it only make sense in the RISC-V assembly encoding context?
Here, my understanding is that we use the 6 bits because we are encoding RV64, so I just would not mention these parts.

luisacicolini: addressed

PR #271, opened 2026-03-04
fix: correct parsing of negative immediates

luisacicolini:
This PR fixes the parsing of negative immediates for RISCV operations. 
The current parsing is consistent with [Lean-MLIR](https://github.com/opencompl/lean-mlir/blob/814a365c00dc641c9647bec4369e7b3bd882372f/SSA/Projects/RISCV64/Base.lean#L491). 

In particular, for `slli`, `srli`, `srai`, we coerce negative attributes to their corresponding (positive) natural, as the Sail model in Lean [does](https://github.com/rems-project/lean-sail/blob/1768d356cc4ec3f0385c94d539b2c785fa064f97/Sail/Sail.lean#L172).

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -209,7 +209,7 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
     return (#[.reg (RISCV.srli imm op)], .continue)
   | .riscv_srai => do
     let #[.reg op] := operands | none
-    let imm := BitVec.toNat (BitVec.ofInt 6 properties.value.value)
+    let imm := BitVec.ofInt 6 properties.value.value
     return (#[.reg (RISCV.srai imm op)], .continue)
   | _ => none
 


tobiasgrosser (approved):
Nice. Thank you!

Let's merge this, that we can get your other PR in.

Veir/Interpreter/Basic.lean
```diff
@@ -158,68 +158,58 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
...
     return (#[.reg (RISCV.xori imm op)], .continue)
   | .riscv_addiw => do
     let #[.reg op] := operands | none
-    let imm := BitVec.ofNat 12 properties.value.value.toNat
+    let imm := BitVec.ofInt 12 properties.value.value
     return (#[.reg (RISCV.addiw imm op)], .continue)
   | .riscv_slli => do
     let #[.reg op] := operands | none
-    let imm := BitVec.ofNat 6 properties.value.value.toNat
+    let imm := BitVec.toNat (BitVec.ofInt 6 properties.value.value)
```
tobiasgrosser:
```suggestion
    let imm := BitVec.ofInt 6 properties.value.value
```

Veir/Interpreter/Basic.lean
```diff
@@ -158,68 +158,58 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
...
     return (#[.reg (RISCV.slli imm op)], .continue)
   | .riscv_srli => do
     let #[.reg op] := operands | none
-    let imm := BitVec.ofNat 6 properties.value.value.toNat
+    let imm := BitVec.toNat (BitVec.ofInt 6 properties.value.value)
```
tobiasgrosser:
```suggestion
    let imm := BitVec.ofInt 6 properties.value.value
```

Veir/Interpreter/Basic.lean
```diff
@@ -158,68 +158,58 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
...
     return (#[.reg (RISCV.srli imm op)], .continue)
   | .riscv_srai => do
     let #[.reg op] := operands | none
-    let imm := BitVec.ofNat 6 properties.value.value.toNat
+    let imm := BitVec.toNat (BitVec.ofInt 6 properties.value.value)
```
tobiasgrosser:
```suggestion
    let imm := BitVec.ofInt 6 properties.value.value
```

Veir/Interpreter/Basic.lean
tobiasgrosser: The `BitVec.toNat` is not needed, AFAIU. Why do. you think otherwise.

luisacicolini: Because for `slli`, `srli`, `srai` we need this to be a natural. and if we use `ofNat` we will get `0`, which differs from what happens in the sail model, which is our standard (e.g. [here](https://github.com/rems-project/lean-sail/blob/1768d356cc4ec3f0385c94d539b2c785fa064f97/Sail/Sail.lean#L172))

luisacicolini: In this way I believe our semanitcs are consistent with Sail.

tobiasgrosser: Did you try my changes? They seem to pass all your tests. I am not saying you should use `ofNat`. The change to use `toNat` is fine, but there is no need for another `toNat` conversion.

PR #273, opened 2026-03-04
feat: RISCV `w` instructions 

luisacicolini:
This PR adds the `w` instructions for the RISCV64 Base extension. This completes our coverage of its arithmetic fragment.


math-fehr (approved):
Looks good to me, but before merging can you add tests where the 32 lowest bits overflow?

PR #282, opened 2026-03-05
feat: interpreter support for arith.[addi|subi|muli|divui|divsi|remui|remsi|shl|shrui|shrsi]

tobiasgrosser:
.. in interpreter and tests.

We currently reuse the LLVM semantics, but this may be split in the future. Similarly, we adapted the test cases from the LLVM dialect with the only difference being that we renamed the operations.

### DIFF
--- a/Test/Interpreter/Arith/subi_unsigned_overflow.mlir
+++ b/Test/Interpreter/Arith/subi_unsigned_overflow.mlir
@@ -1,13 +1,13 @@
 // RUN: veir-interpret %s | filecheck %s
 
 "builtin.module"() ({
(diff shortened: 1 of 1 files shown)


Test/Interpreter/Arith/subi_unsigned_overflow.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: veir-interpret %s | filecheck %s
+
```
math-fehr, line 2: This file has the wrong name

tobiasgrosser: Fixed. Thank you for spotting this.

PR #283, opened 2026-03-06
fix `Reg` type

luisacicolini:
This PR introduces a proper definition for the `Reg` type, which is un/wrapped in the RISCV semantics.

### DIFF
--- a/Veir/Data/RISCV/Reg/Basic.lean
+++ b/Veir/Data/RISCV/Reg/Basic.lean
@@ -117,14 +117,12 @@ def slli (shamt : BitVec 6) (rs1 : Reg) : Reg :=
 def srli (shamt : BitVec 6) (rs1 : Reg) : Reg :=
   ⟨rs1.val >>> shamt⟩
 
-
 /--
   Performs arithmetic right shift on the value in register rs1 by the shift amount held in the
   lower 6 bits of the immediate in RV64.
 -/
 def srai (shamt : BitVec 6) (rs1 : Reg) : Reg :=
   ⟨BitVec.signExtend 64 (BitVec.sshiftRight' rs1.val shamt)⟩
-
 
 /--
   Adds the registers rs1 and rs2 and stores the result in rd. Arithmetic overflow is ignored and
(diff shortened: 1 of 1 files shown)


Veir/Data/RISCV/Reg/Basic.lean
```diff
@@ -5,10 +5,14 @@ namespace Veir.Data.RISCV
 public section
 
 /--
-The `reg` type is a wrapper around a `reg` representing the content of a register.
-
+The `Reg` type is a `BitVec 64` representing the content of a register.
 -/
-abbrev Reg := BitVec 64
+inductive Reg where
+  | val : BitVec 64 → Reg
```
math-fehr: Should this be moved to a structure instead of a single-case inductive?

Veir/Data/RISCV/Reg/Basic.lean
```diff
@@ -22,332 +26,445 @@ abbrev Reg := BitVec 64
...
   result is simply the low 64 bits of the result. ADDI rd, rs1, 0 is used to implement the MV
   rd, rs1 assembler pseudo-instruction.
 -/
-def addi (imm : BitVec 12) (rs1_val : BitVec 64) : BitVec 64 :=
-  let immext : BitVec 64 := (BitVec.signExtend 64 imm) ;
-  BitVec.add rs1_val immext
+def addi (imm : BitVec 12) (rs1 : Reg) : Reg :=
+  let immext : BitVec 64 := (BitVec.signExtend 64 imm)
+  let ⟨rs1_val⟩ := rs1
+  ⟨BitVec.add rs1_val immext⟩
```
math-fehr: If you are using a structure, you can probably expose something like `val` so you could write `<BitVec.add rs1.val immext>` instead, which should make the code a bit more concise?

Veir/Data/RISCV/Reg/Basic.lean
luisacicolini: addressed

Veir/Data/RISCV/Reg/Basic.lean
luisacicolini: done

PR #284, opened 2026-03-06
feat: add llvm.[and|or|xor] and arith.[andi|ori|xori]

tobiasgrosser:


math-fehr:
@copilot Can you address @math-fehr comment? Just change `llvm.` operations to `arith.` operations in the `Arith/` tests added in this pr.

PR #290, opened 2026-03-08
Ensure InlineBlock.lean is compiled in the CI and fix it

math-fehr:
This file was never checked by the CI, so it got a bit out of sync.


Veir/Rewriter/InlineBlock.lean
```diff
@@ -283,24 +288,26 @@ theorem BlockPtr.operationList_Rewriter_inlineBlock_from
...
     l.extract idx ub := by
   sorry
 
+set_option warn.sorry false in
 theorem Array.append_singleton_erase_of_getElem?_zero [BEq α] {l : Array α} :
     l[0]? = some x →
     #[x] ++ (l.erase x) = l := by
   sorry
 
-
```
tobiasgrosser, line 303: This is an unrelated change.

PR #291, opened 2026-03-09
Proof for `ArrayInv.popNones`

AlexanderViand:
Removes the sorry for `ArrayInv.popNones`, as a byproduct of more explorations in applying Claude Code (with Lean4 LSP MPC and the Lean4 skill).

This PR intentionally comes to you in two commits, to show the difference between Sonnet 4.6 (first commit) and Opus 4.6 (second commit, asked to clean up helper situation). It's not quite an apples-to-apples comparison since there's a chance Sonnet would have also found the simpler version with enough time, but it didn't seem to be getting anywhere anymore no matter how long it tried.

### DIFF
--- a/Veir/PatternRewriter/Basic.lean
+++ b/Veir/PatternRewriter/Basic.lean
@@ -22,12 +22,8 @@ abbrev ArrayInv (stack : Array (Option α)) (index : HashMap α Nat) :=
 
 omit [DecidableEq α] in
 @[grind .]
-theorem ArrayInv.popNones (h : ArrayInv (α := α) stack index) : ArrayInv (stack.popWhile (· = none)) index := by
-  intro op i hi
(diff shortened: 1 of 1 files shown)


Veir/ForLean.lean
```diff
@@ -2,6 +2,7 @@ module
 
 public import Std.Data.ExtHashSet
 public import Std.Data.HashMap
+import all Init.Data.Array.Basic -- unfold [Array.popWhile] in Array.getElem?_popWhile_of_false needs to see Array.popWhile.eq_def
```
AlexanderViand: Not sure if this is good Lean-ing or not. Alternatively, it's not too hard to restate and proof Array.popWhile.eq_def as a private helper here, but it seems a bit redundant/not very pretty

PR #324, opened 2026-03-23
Add invariant that nextID is free

ineol:
It makes allocating a new operation/block/region infallible.

@math-fehr what do you think?


math-fehr:
The reason I didn't make allocating faillible is because whenever we are going to put this into a bytearray, we will have the invariant that the array size is less than USize (or less), and then we have to fail during allocations here I believe?

I'm fine in removing this restriction now, but note that this is very much likely something we will have to add back for buffed.

tobiasgrosser:
@ineol, is this still something we want to do? Otherwise, can we close this issue?

PR #346, opened 2026-03-28
Add standalone IRNode containment module

snarang181:
## Summary
  - Adds `Veir/IR/Containment.lean` with a generic IRNode inductive type and contains! function with supporting lemmas

  ## Test plan                                                                                                          
  - [x] `lake build` passes                                                                            
  - [x] `lake test` passes                                                                     
  - [x] `uv run lit Test/` passes


snarang181:
cc @math-fehr (Sorry about the explicit ping, but can't request review as of now)

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
+/-
+  Containment relationships for the IR tree.
+
+  The IR tree hierarchy is: Operation → Region → Block → Operation → ...
+  Parent pointers go upward: Block.parent → Region, Region.parent → Operation, Operation.parent → Block.
```
math-fehr:
```suggestion
```
This should be unneeded as this is a description of the base data structures.

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
...
+
+  This file defines an inductive `IRNode` type representing operations, blocks, or regions,
+  and a `contains` function that determines whether one node is an ancestor of another by
```
math-fehr:
```suggestion
  and a `contains!` function that determines whether one node is an ancestor of another by
```

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
...
+
+namespace IRNode
+
+/-- Get the parent of an IR node, if it exists. -/
+def parent! (node : IRNode) (ctx : IRContext OpInfo) : Option IRNode :=
+  match node with
+  | .op ptr => (ptr.get! ctx).parent |>.map (.block ·)
+  | .block ptr => (ptr.get! ctx).parent |>.map (.region ·)
+  | .region ptr => (ptr.get! ctx).parent |>.map (.op ·)
+
```
math-fehr, line 41: Can you also add the `parent` version that takes the proof of `InBounds node`.

For that, you should also add `IRNode.InBounds`, and the related lemmas (for each constructor).

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
...
+def contains! (ancestor descendant : IRNode) (ctx : IRContext OpInfo)
+    (fuel : Nat := ctx.nextID) : Bool :=
+  match fuel with
+  | 0 => true  -- conservative: assume containment if we can't determine
+  | n + 1 =>
+    match descendant.parent! ctx with
+    | none => false
+    | some parentNode =>
+      if parentNode == ancestor then true
+      else contains! ancestor parentNode ctx n
```
math-fehr, line 72: Instead of return a `Bool`, can you return an `Option`?
That way, we do not need to be conservative on wether or not all the fuel was used, and having the proof `contains! ancestor descendant ctx = some ...` will make reasonning simpler overall.

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
...
+/--
+  Check whether `ancestor` contains `descendant` in the IR tree by walking
+  up from `descendant` through parent pointers.
+
+  Uses fuel-based recursion for termination. The fuel parameter defaults to
+  `ctx.nextID`, which is an upper bound on the number of nodes in the context.
+  Returns `true` if `ancestor` is found on the path from `descendant` to the root,
+  or if fuel is exhausted (conservative approximation).
+-/
+def contains! (ancestor descendant : IRNode) (ctx : IRContext OpInfo)
```
math-fehr, line 63: Can you also add a `contains` version and a proof of equivalence?

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
...
+    match descendant.parent! ctx with
+    | none => false
+    | some parentNode =>
+      if parentNode == ancestor then true
+      else contains! ancestor parentNode ctx n
+
+/-- `ancestor.contains! descendant` with zero fuel is always true (conservative). -/
+theorem contains!_fuel_zero (ancestor descendant : IRNode) (ctx : IRContext OpInfo) :
+    contains! ancestor descendant ctx 0 = true := by
+  rfl
```
math-fehr: I don't think this theorem should be needed (especially if we return options now)

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,70 @@
...
+    (fuel : Nat := ctx.nextID) : Bool :=
+  match fuel with
+  | 0 => true  -- conservative: assume containment if we can't determine
+  | n + 1 =>
+    match descendant.parent! ctx with
+    | none => false
+    | some parentNode =>
+      if parentNode == ancestor then true
+      else contains! ancestor parentNode ctx n
+
```
math-fehr, line 73: I would also add the basic lemmas about "unfolding" `contains!`.
Like `y.parent = some yParent -> contains! x y fuel = some z -> contains! x yParent (fuel - 1) = some y` if that makes sense

math-fehr (commented):
Thanks! I left some comments, mostly about the design, but otherwise I think that's good!

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,109 @@
+import Veir.IR.Basic
```
math-fehr, line 1: Can you make this file a module?

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,109 @@
...
+def InBounds (node : IRNode) (ctx : IRContext OpInfo) : Prop :=
+  match node with
+  | .op ptr => ptr.InBounds ctx
+  | .block ptr => ptr.InBounds ctx
+  | .region ptr => ptr.InBounds ctx
+
+instance : Decidable (InBounds node (ctx : IRContext OpInfo)) := by
+  unfold InBounds; cases node <;> infer_instance
+
+theorem InBounds.op {ptr : OperationPtr} (h : ptr.InBounds (ctx : IRContext OpInfo)) :
```
math-fehr, line 26: @ineol, I always forget, should the name be this one or `inBounds_op`?

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-- Get the parent of an IR node (safe version with InBounds proof). -/
+def parent (node : IRNode) (ctx : IRContext OpInfo) (h : node.InBounds ctx := by grind) : Option IRNode :=
+  match node, h with
+  | .op ptr, h => (ptr.get ctx h).parent |>.map (.block ·)
+  | .block ptr, h => (ptr.get ctx h).parent |>.map (.region ·)
+  | .region ptr, h => (ptr.get ctx h).parent |>.map (.op ·)
+
+/-- `parent!` agrees with `parent` when the node is in bounds. -/
+theorem parent!_eq_parent (node : IRNode) (ctx : IRContext OpInfo) (h : node.InBounds ctx) :
```
math-fehr, line 50: Can you add this to the `eq_bang` grind set?

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,109 @@
...
+
+variable {OpInfo : Type} [HasOpInfo OpInfo]
+
+/-- An IR node is an operation, block, or region. -/
+inductive IRNode where
+  | op (ptr : OperationPtr)
+  | block (ptr : BlockPtr)
+  | region (ptr : RegionPtr)
+deriving DecidableEq, Repr
+
```
math-fehr, line 13: Can you add a coercion from `OperationPtr`, `BlockPtr`, and `RegionPtr` to `IRNode`.
That way, it should be much easier to write most code using this

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,109 @@
...
+/--
+  Check whether `ancestor` contains `descendant` in the IR tree by walking
+  up from `descendant` through parent pointers.
+
+  Uses fuel-based recursion for termination. The fuel parameter defaults to
+  `ctx.nextID`, which is an upper bound on the number of nodes in the context.
+  Returns `some true` if `ancestor` is found, `some false` if the root is reached
+  without finding it, or `none` if fuel is exhausted.
+-/
```
math-fehr, line 62:
```suggestion
  without finding it, or `none` if fuel is exhausted.
  Node that the fuel should never be exhausted unless there is a cycle
  where an operation contains itself.
-/
```
Just noting this as we now have a proof that `ctx.nextID` is greater than all pointers in our context.

Veir/IR/Containment.lean
```diff
@@ -0,0 +1,109 @@
...
+  match fuel with
+  | 0 => none
+  | n + 1 =>
+    match descendant.parent ctx h with
+    | none => some false
+    | some parentNode =>
+      if parentNode == ancestor then some true
+      else if hb : parentNode.InBounds ctx then
+        contains ancestor parentNode ctx hb n
+      else none
```
math-fehr, line 107: Can you move this next to `contains!`, and add the proof of equivalence?

math-fehr (commented):
Nice! I just have some small comments, but otherwise this looks good!

tobiasgrosser:
This is almost ready to merge, AFAIU. @snarang181, any plans to finish this PR?

math-fehr:
Superseded by #1100

PR #383, opened 2026-04-09
Rewriter.createBlock and Rewriter.createOp proofs for parser

GZGavinZhao:
Required by #222.

### DIFF
--- a/Veir/Rewriter/Basic.lean
+++ b/Veir/Rewriter/Basic.lean
@@ -230,6 +230,20 @@ def Rewriter.insertBlock? (ctx: IRContext OpInfo) (newBlock: BlockPtr)
     let next := insertionPoint.next
     newBlock.linkBetweenWithParent ctx prev next parent (by grind) (by grind) (by grind) (by grind)
 
+@[grind .]
+theorem Rewriter.insertBlock?_inBounds (ptr : GenericPtr)
+    (heq : insertBlock? ctx newBlock ip h₁ h₂ h₃ = some newCtx) :
+    ptr.InBounds ctx ↔ ptr.InBounds newCtx := by
+  simp only [insertBlock?] at heq
+  grind
+
+@[grind .]
+theorem Rewriter.insertBlock?_fieldsInBounds_mono
+    (heq : insertBlock? ctx newBlock ip h₁ h₂ h₃ = some newCtx) :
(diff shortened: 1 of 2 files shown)


ineol (commented):
LGTM!

But it may interact with #381 since the unseal may stop working...

Veir/Rewriter/GetSetInBounds.lean
```diff
@@ -1993,6 +1993,37 @@ theorem OpOperandPtrPtr.get!_detachBlockOperands {opOperandPtr : OpOperandPtrPtr
...
+theorem Rewriter.createBlock_fieldsInBounds_mono
+    (heq : createBlock ctx ip hctx hip = some ⟨newCtx, newPtr⟩) :
+    ctx.FieldsInBounds → newCtx.FieldsInBounds := by
+  simp only [createBlock] at heq
+  split at heq
+  · grind
+  · split at heq
+    · simp [Option.bind] at heq
+      split at heq <;> grind
+    · grind
```
math-fehr: Can you just move these ones to `Rewriter.Basic` as well (right after their definition), then I can merge this!

GZGavinZhao: Moved. A consequence is that `Rewriter.insertBlock?_mono` and related theorems must also be moved to `Rewriter.Basic`, which I think is actually desireable because now `Rewriter.insertBlock?_XXX` theorems are also right after their definition.

Veir/Rewriter/Basic.lean
```diff
@@ -230,6 +230,20 @@ def Rewriter.insertBlock? (ctx: IRContext OpInfo) (newBlock: BlockPtr)
     let next := insertionPoint.next
     newBlock.linkBetweenWithParent ctx prev next parent (by grind) (by grind) (by grind) (by grind)
 
+@[grind .]
+theorem Rewriter.insertBlock?_inBounds (ptr : GenericPtr)
```
GZGavinZhao, line 234: Am I understanding the naming correctly? `inBounds_mono` is for `→` implication proofs, while `inBounds` is for `↔` if-and-only-if proofs

math-fehr: Yes I think so!
I want to do a pass at some point to be sure we are consistent, but I believe this is the naming convention we should follow!

math-fehr (approved):
Nice!

PR #384, opened 2026-04-09
IRContext.default and IRContext.create proofs for parser

GZGavinZhao:
Depends on #383.

Required by #222.

### DIFF
--- a/Veir/IR/Fields.lean
+++ b/Veir/IR/Fields.lean
@@ -77,6 +77,19 @@ structure IRContext.FieldsInBounds (ctx : IRContext OpInfo) : Prop where
 
 attribute [local grind =] Option.maybe_def
 
+section default 
+
+@[grind .]
+theorem IRContext.default_FieldsInBounds : IRContext.FieldsInBounds (default : IRContext OpInfo) := by
+  stop
+  -- Invalid simp theorem `instInhabitedIRContext.default`: Expected a definition with an exposed body
+  -- TODO(gzgz): is this due to `module` weirdness? I can't figure it out.
+  simp [default, instInhabitedIRContext.default]
+  grind [IRContext.FieldsInBounds, OperationPtr.InBounds, BlockPtr.InBounds, RegionPtr.InBounds]
+
+
+end default
+
 section get
 
 /-
--- a/Veir/IR/Fields.lean
+++ b/Veir/IR/Fields.lean
@@ -77,16 +77,13 @@ structure IRContext.FieldsInBounds (ctx : IRContext OpInfo) : Prop where
 
 attribute [local grind =] Option.maybe_def
 
-section default 
+section default
 
 @[grind .]
 theorem IRContext.default_FieldsInBounds : IRContext.FieldsInBounds (default : IRContext OpInfo) := by
-  stop
-  -- Invalid simp theorem `instInhabitedIRContext.default`: Expected a definition with an exposed body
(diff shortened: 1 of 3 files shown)


GZGavinZhao:
@ineol @math-fehr would appreciate some help on definition visibility and file organization.

Veir/Rewriter/GetSetInBounds.lean
```diff
@@ -4384,3 +4384,10 @@ theorem OperationPtr.getNumOperands_iff_replaceValue?
     OperationPtr.getNumOperands op ctx' h_op =
     OperationPtr.getNumOperands op ctx (by grind) := by
   grind [OpOperandPtr.inBounds_if_operand_size_eq]
+
+-- TODO(gzgz): is this the correct place to put this theorem?
```
tobiasgrosser: Please drop this comment from the diff and add a comment to the PR instead.

tobiasgrosser (commented):
The PR currently does not compile.

Veir/Rewriter/GetSetInBounds.lean
```diff
@@ -4384,3 +4415,9 @@ theorem OperationPtr.getNumOperands_iff_replaceValue?
     OperationPtr.getNumOperands op ctx' h_op =
     OperationPtr.getNumOperands op ctx (by grind) := by
   grind [OpOperandPtr.inBounds_if_operand_size_eq]
+
+@[grind →]
+theorem IRContext.create_fieldsInBounds {op: OperationPtr} (h : IRContext.create OpInfo = some (ctx, op)) :
```
GZGavinZhao, line 4370: @math-fehr suggested to put this theorem right after the definition of `IRContext.create` (in `Rewriter.Basic`), but the proof requires `Rewriter.createBlock_fieldsInBounds_mono` which is in the current file (`Rewriter.GetSetInBounds`).

math-fehr: Oh my bad, then sorry let's move it back to where it was, I forgot this relied on this proof!

Veir/IR/Fields.lean
```diff
@@ -77,6 +77,19 @@ structure IRContext.FieldsInBounds (ctx : IRContext OpInfo) : Prop where
 
 attribute [local grind =] Option.maybe_def
 
+section default 
+
+@[grind .]
+theorem IRContext.default_FieldsInBounds : IRContext.FieldsInBounds (default : IRContext OpInfo) := by
+  stop
+  -- Invalid simp theorem `instInhabitedIRContext.default`: Expected a definition with an exposed body
+  -- TODO(gzgz): is this due to `module` weirdness? I can't figure it out.
```
GZGavinZhao: @math-fehr @ineol Would like some help with this proof when you get a chance. The exact same proof works in `MlirParser` but not here...

math-fehr: I pushed a commit to fix it!
When you use modules in Lean, you can't unfold definitions that are not marked `@[expose]`.
Funnily enough, you cannot expose `instInhabitedIRContext.default` because you can only expose at definition time (which is done automatically so you cannot here).
So I added a theorem that exposes it.
So we use the manual `*_def` lemmas that rewrite things into their definitions.

tobiasgrosser (approved):
This LGTM. Let's see what @math-fehr says.

math-fehr (approved):

PR #386, opened 2026-04-09
instruction selection: `llvm.shl`, `llvm.lshr`

luisacicolini:
Add the instruction selection lowerig patterns for `llvm.shl`, `llvm.lshr`


Test/Passes/InstructionSelection/RISCV64/lshr.mlir
```diff
@@ -0,0 +1,34 @@
+// RUN: veir-opt %s -p=isel-riscv64 | filecheck %s
+
+"builtin.module"() ({
+    "func.func"() ({
```
math-fehr, line 4: Can you have multiple functions or just have everything in the same block?
The problem with having multiple blocks is that at some point we will add terminators and that will break this test!

Test/Passes/InstructionSelection/RISCV64/shl.mlir
```diff
@@ -0,0 +1,34 @@
+// RUN: veir-opt %s -p=isel-riscv64 | filecheck %s
+
+"builtin.module"() ({
+    "func.func"() ({
+    ^bb0(%a: i64, %b: i64):
```
math-fehr, line 5: Same here

math-fehr (approved):
All good to me, I would just change the test so it won't break in the future!

Test/Passes/InstructionSelection/RISCV64/lshr.mlir
luisacicolini: addressed

Test/Passes/InstructionSelection/RISCV64/shl.mlir
luisacicolini: addressed

luisacicolini:
addressed the changes, thank you @math-fehr!

PR #392, opened 2026-04-10
factor propertiesOf into dialect-specific functions

tobiasgrosser:
This is a first step towards making a our dialect definitions more modular.

### DIFF
--- a/Veir/IR/Simp.lean
+++ b/Veir/IR/Simp.lean
@@ -10,5 +10,5 @@ public section
 /- Simp set converting safe operations to their unsafe (!) counterparts. -/
 register_simp_attr eq_bang
 
-/- Simp set for folding away `Dialect`.opProperties calls -/
+/- Simp set for folding away `Dialect`.propertiesOf calls -/
 register_simp_attr properties_of
--- a/Veir/IR/Simp.lean
+++ b/Veir/IR/Simp.lean
@@ -9,3 +9,6 @@ public section
 
 /- Simp set converting safe operations to their unsafe (!) counterparts. -/
 register_simp_attr eq_bang
+
+/- Simp set for folding away `Dialect`.opProperties calls -/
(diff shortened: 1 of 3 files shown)


Veir/Properties.lean
```diff
@@ -162,98 +162,135 @@ def ModArithConstantProperties.fromAttrDict (attrDict : Std.HashMap ByteArray At
...
+instance : HasDialectOpInfo OpCode where
   propertiesOf := propertiesOf
   propertiesHash := by
-    unfold propertiesOf
+    simp
     intros opCode
     cases opCode <;> (try simp) <;> (rename_i op; cases op <;> infer_instance)
   propertiesDefault := by
-    unfold propertiesOf
+    simp only [propertiesOf, Arith.propertiesOf, Llvm.propertiesOf, Riscv.propertiesOf, Mod_Arith.propertiesOf]
```
tobiasgrosser: I had hoped to use `HasDialectOpInfo` to avoid `Arith.propertiesOf`, but I found no way to make it work. Ideas welcome.

math-fehr: Can you maybe create a simp set?

tobiasgrosser: Added. This now looks a lot nicer.

tobiasgrosser:
I addressed all comments. A final round of reviews would be appreciated.

Veir/IR/Simp.lean
```diff
@@ -9,3 +9,6 @@ public section
 
 /- Simp set converting safe operations to their unsafe (!) counterparts. -/
 register_simp_attr eq_bang
+
+/- Simp set for folding away `Dialect`.opProperties calls -/
```
tobiasgrosser:
```suggestion
/- Simp set for folding away `Dialect`.propertiesOf calls -/
```

math-fehr (approved):

PR #395, opened 2026-04-10
Parse `cf.br` op

nchappe:
This adds the `cf.br` op to the parser and the verifier.
Note that I don't verify that the number of arguments is consistent with the destination block. This is not a local invariant, so it seemed out of scope for the current verifier.


tobiasgrosser (approved):
Perfect. Let's wait for @luisacicolini's feedback and then get this in.

luisacicolini (approved):
LGTM! Thanks a lot :)

PR #398, opened 2026-04-11
instruction selection: `llvm.select`

luisacicolini:
Adds the instructions selection pattern (with branching) for `llvm.select` operation. 
We ignore the `ld` and `sd` operations for the time being and focus on the construction of the correct blocks. 
We also update the verifier to check the width of the predicate of the operation (`i1`) 
We implement the RISCV assembly branching operation `bnez` following #391, without introducing its semantics.


tobiasgrosser:
This seems to be outdated. Let me close this. Feel free to reopen, if there is more to be done, @luisacicolini.

PR #402, opened 2026-04-13
Merge benchmarks and docs in a single CI workflow

snarang181:
## Current Issue                                                                                                                                                                                                                                                                  
  - opencompl.github.io/veir/dev/bench/ returns 404 because the docs deployment (deploy-pages@v4) overwrites the benchmark pages that  benchmark-action pushes to gh-pages   

## Summary of PR

- Merges benchmarks.yml and build_docs.yml into a single ci.yml with 3 jobs: benchmark, build-docs, and deploy
-  The deploy job checks out gh-pages (with fresh benchmark data), overlays docs on top preserving dev/, and pushes once so both docs and benchmarks are served from the same branch  
-  After merge to main, opencompl.github.io/veir/ serves docs   and  opencompl.github.io/veir/dev/bench/ serves benchmark charts


snarang181:
benchmark-action tries to post a comment on the PR with benchmark results, which requires pull_requests=write permission. But GitHub Actions gives read-only tokens to PRs from forks so the workflow is failing. This is not related to the change.

tobiasgrosser (approved):

PR #413, opened 2026-04-13
[ci] Fix noise in benchmarks 

snarang181:
Each benchmark ran once which produced highly variable results.

  ## Fix 
  1. .github/scripts/run_benchmarks.sh — Run each benchmark 5 times and report the median
  2. .github/workflows/ci.yml: 
    - Pass 5 as the iterations argument: run_benchmarks.sh 1000 100 5
    - Increase threshold to a very large value, effectively disabling alerts.

### DIFF
--- a/.github/scripts/run_benchmarks.sh
+++ b/.github/scripts/run_benchmarks.sh
@@ -1,13 +1,16 @@
 #!/usr/bin/env bash
 # Runs all benchmarks and outputs results in github-action-benchmark JSON format.
-# Usage: ./run_benchmarks.sh [count] [pc]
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
This build is failing again. Let's leave the alert threshold unchanged for now and let the CI gather data for a couple of days. We can then see what a good threshold is.

snarang181:
> This build is failing again. Let's leave the alert threshold unchanged for now and let the CI gather data for a couple of days. We can then see what a good threshold is.

@tobiasgrosser, I think it is a permissions issue while running from a fork PR. The benchmarks itself run fine, I think.

PR #414, opened 2026-04-13
refinement relation from `LLVM.Int` to `RISCV.Reg`

luisacicolini:
We add the definition of the refinement relation from `LLVM.Int` to `RISCV.Reg`. This is the core relation we want to reason about when proving the correctness of instruction selection patterns. We say an integer `i : LLVM.Int` is refined by a register `r : RISCV.Reg` if the behaviours allowed by the register are a subset smaller or equal to the integer's behaviour.

Formally, an integer `i : LLVM.Int` is refined by a register `r : RISCV.Reg` iff: 
* if `i` is a concrete value `.val v`, then the register `r` must contain `v` (i.e., `r.val = v`), and the two behave identically.
* if `i` is a `.poison`, any value can refine it and the relation always holds, as the register displays a specific behaviour among the different ones `poison` can have.

### DIFF
--- a/Veir/Data/Refinement.lean
+++ b/Veir/Data/Refinement.lean
@@ -1,0 +1,19 @@
+module
+
+public import Veir.Data.LLVM.Int.Basic
+public import Veir.Data.RISCV.Reg.Basic
+
+public section
+
+/-!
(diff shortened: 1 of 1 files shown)


math-fehr (approved):
All good to me, I think after we should try to see what happens when we want to have a non-64 bits refinement!

PR #415, opened 2026-04-13
llvm: `constant` semantics

luisacicolini:
We add explicit, declarative semantics of `llvm.constant` operation. This is useful to reason about the correctness of rewrites, e.g. in the context of instruction selection. The semantics are unchanged in practice, but wrapped in their own deifnition just like every other operation.

### DIFF
--- a/Veir/Data/LLVM/Int/Basic.lean
+++ b/Veir/Data/LLVM/Int/Basic.lean
@@ -66,6 +66,12 @@ instance {w : Nat} : ToString (Int w) where
     | .poison => "poison"
 
 /--
+  We define the semantics of a constant value
+-/
+def constant {w : Nat} (v : BitVec w) : Int w := Id.run do
+  val v
+
+/--
 The ‘add’ instruction returns the sum of its two operands.
 
 If the sum has unsigned overflow, the result returned is the mathematical result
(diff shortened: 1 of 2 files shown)


Veir/Data/LLVM/Int/Basic.lean
```diff
@@ -65,6 +65,13 @@ instance {w : Nat} : ToString (Int w) where
     | .val v => toString v
     | .poison => "poison"
 
+/--
+  We define the semantics of a `constant` operation.
+  The result of this operation is never poison.
+-/
+def constant (w : Nat) (v : _root_.Int) : Int w := Id.run do
+  val (BitVec.ofInt w v)
```
math-fehr:
```suggestion
def constant (w : Nat) (v : _root_.Int) : Int w := val (BitVec.ofInt w v)
```
There should be no need for `Id.run` since we are not using a monad here?

math-fehr (approved):
All good to me, just one comment.
Feel free to merge it whenever you fix the comment, if the comment is fine for you

Veir/Data/LLVM/Int/Basic.lean
luisacicolini: indeed! that was a copy-paste mistake

PR #416, opened 2026-04-13
instruction selection: `constant`, `add` refinement proof

luisacicolini:
We prove the correctness of the lowering pattern for `constant` and `add`. If we agree on this format and organization I will write more proofs and push them. Note that the proof for `constant` does not deal with `.poison`, as `llvm.constant` is never poison. 

I think we have two options to consider wrt. lemmas' formulation: 
* wrapping the semantics of the `unrealized_conversion_cast` (i.e.,`toReg : i64 -> !reg`) directly in the lemma and proving the refinement `i64` to `!reg`. 
* wrapping the semantics of the `unrealized_conversion_cast` *in both directions* (i.e.,`toReg : i64 -> !reg` and `toInt : !reg -> i64`). This would be more consistent with what the rewriter actually does, and the refinement goes from `i64` to `i64`. 

I am pushing both option and look forward to gathering opinions :)

Also, the tactic is the usual:
* unwrap semantics
* split + get rid of monad
* `bv_decide`

If we agree on these steps we can wrap it into an actual tactic, as we did in `lean-MLIR` (for a subsequent PR).

### DIFF
--- a/Veir/Data/Refinement.lean
+++ b/Veir/Data/Refinement.lean
@@ -9,23 +9,13 @@ public section
   This file defines the refinement relation from the `LLVM.Int` to `RISCV.Reg` type.
 -/
 
-open Classical in
-
-/--
-  `LLVM.Int` is refined by `RISCV.Reg`.
--/
-def IntIsRefinedByReg [DecidableEq Prop] (i : Veir.Data.LLVM.Int 64) (r : Veir.Data.RISCV.Reg) : Prop :=
-  match i with
-  | .val v => r.val = v
-  | .poison => true
-
 /--
   `LLVM.Int` is refined by `LLVM.Int`.
   `i'` refines `i` if its behaviour are a subset of the behaviours allowed by `i`.
   In particular, any concrete `i'` refines a poison `i`, but a poison `i'` does *not* refine
   any `i`.
 -/
-def IntIsRefinedByInt [DecidableEq Prop] (i i' : Veir.Data.LLVM.Int 64) : Prop :=
+def isRefinedBy (i i' : Veir.Data.LLVM.Int 64) : Prop :=
   match i with
   | .val v =>
     match i' with
--- a/Veir/Data/Refinement.lean
+++ b/Veir/Data/Refinement.lean
@@ -18,3 +18,17 @@ def IntIsRefinedByReg [DecidableEq Prop] (i : Veir.Data.LLVM.Int 64) (r : Veir.D
   match i with
   | .val v => r.val = v
   | .poison => true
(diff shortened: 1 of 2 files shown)


Veir/Data/Refinement.lean
```diff
@@ -17,3 +18,17 @@ def IntIsRefinedByReg [DecidableEq Prop] (i : Veir.Data.LLVM.Int 64) (r : Veir.D
...
   | .val v => r.val = v
   | .poison => true
+
+/--
+  `LLVM.Int` is refined by `LLVM.Int`.
+  `i'` refines `i` if its behaviour are a subset of the behaviours allowed by `i`.
+  In particular, any concrete `i'` refines a poison `i`, but a poison `i'` does *not* refine
+  any `i`.
+-/
+def IntIsRefinedByInt [DecidableEq Prop] (i i' : Veir.Data.LLVM.Int 64) : Prop :=
```
ineol:
```suggestion
def IntIsRefinedByInt (i i' : Veir.Data.LLVM.Int 64) : Prop :=
```

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -9,13 +10,50 @@ import Std.Tactic.BVDecide
 -/
 
 open Veir
+open Classical
```
ineol: You can also remove this if you remove the DecidableEq for Prop above

Veir/Data/Refinement.lean
luisacicolini: done

Veir/Passes/InstructionSelection/Proofs.lean
luisacicolini: removed

luisacicolini:
after chatting with @math-fehr we resolved to only use refinement from integers to integers, adhering to the way rewrite patterns actually work.

math-fehr (approved):

nchappe:
I'm a bit late, but wouldn't the LLVM-only `isRefinedBy` definition make sense for any bitwidth instead of just 64?
It can probably be useful beyond the RISC-V lowering, e.g. for LLVM IR optimizations.

luisacicolini:
Thanks for looking into this @nchappe :)
Yes, we will at some point expand the definition for different widths. At the moment we do not need to do this, because the instruction selection only supports `i64` - however, this requires some more reasoning about the correctness of zero/signextensions. I am not yet sure how this should work, so I started with this patch: if you have any suggestions I'm happy to talk about it more :)

PR #418, opened 2026-04-14
Interpret `cf.br`

nchappe:
Blocks to interpret are iterated over using a `partial_fixpoint`,
until a `func.return` terminator is eventually reached (if any).

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -56,13 +56,6 @@ def InterpreterState.setVar (state : InterpreterState) (var : ValuePtr) (val : R
   {state with variables := state.variables.insert var val}
 
 /--
-  Set the value of a block argument.
--/
-def InterpreterState.setBlockArgument (state : InterpreterState) (arg : BlockArgument) (val : RuntimeValue) :
-    InterpreterState :=
-  state.setVar (ValuePtr.blockArgument {block := arg.owner, index := arg.index}) val
-
-/--
   Get the value of a variable, if the variable exists.
 -/
 def InterpreterState.getVar? (state : InterpreterState) (var : ValuePtr)
@@ -99,14 +92,13 @@ def InterpreterState.setResultValues (state : InterpreterState) (ctx : IRContext
 -/
 def InterpreterState.setArgumentValues (state : InterpreterState) (ctx : IRContext OpInfo)
     (block : BlockPtr) (values : Array RuntimeValue) : InterpreterState :=
-  let args := (block.get! ctx).arguments
   let rec loop (state : InterpreterState) (i : Nat) :=
     match i with
     | 0 => state
     | i + 1 =>
-      let arg := args[i]!
+      let arg := block.getArgument i
       let value := values[i]!
-      let newState := state.setBlockArgument arg value
+      let newState := state.setVar arg value
       loop newState i
   loop state (block.getNumArguments! ctx)
 
@@ -134,7 +126,7 @@ inductive ControlFlowAction where
   returns `none`.
 -/
 def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
-    (resultTypes : Array TypeAttr) (operands : Array RuntimeValue) (blockOperands : Array BlockOperand)
+    (resultTypes : Array TypeAttr) (operands : Array RuntimeValue) (blockOperands : Array BlockPtr)
     : Option ((Array RuntimeValue) × Option ControlFlowAction) :=
   match opType with
   | .arith .constant => do
@@ -329,7 +321,7 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
     return (#[], some (.return operands))
   | .cf .br => do
     let #[dest] := blockOperands | none
-    return (#[], some (.branch operands dest.value))
+    return (#[], some (.branch operands dest))
   /- Bitblastable semantics of RISC-V assembly instructions. -/
   | .riscv .li => do
     let imm := BitVec.ofInt 64 properties.value.value
@@ -662,7 +654,7 @@ def interpretOp (ctx : IRContext OpCode) (op : OperationPtr) (state : Interprete
     : Option (InterpreterState × Option ControlFlowAction) := do
   let operands ← state.getOperandValues ctx op
   let opType := op.getOpType! ctx
-  let (resultValues, action) ← interpretOp' opType (op.getProperties! ctx opType) ((op.get! ctx).results.map (·.type)) operands (op.get! ctx).blockOperands
+  let (resultValues, action) ← interpretOp' opType (op.getProperties! ctx opType) ((op.get! ctx).results.map (·.type)) operands ((op.get! ctx).blockOperands.map (·.value))
   let newState := state.setResultValues ctx op resultValues
   return (newState, action)
 
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -117,539 +117,537 @@ def InterpreterState.empty : InterpreterState :=
   { variables := Std.HashMap.emptyWithCapacity 8 }
 
 /--
-  How the control flow should proceed after interpreting an operation.
+  How the control flow should proceed after interpreting a terminator.
   - `return` indicates that the current block should return with the given values.
   - `branch` indicates that the interpreter should jump to another block
-  - `continue` indicates that the interpreter should continue to the next operation in the block.
 -/
 inductive ControlFlowAction where
   | return (vals : Array RuntimeValue)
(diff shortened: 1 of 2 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -87,541 +87,562 @@ def InterpreterState.setResultValues (state : InterpreterState) (ctx : IRContext
...
-  How the control flow should proceed after interpreting an operation.
+  How the control flow should proceed after interpreting a terminator.
   - `return` indicates that the current block should return with the given values.
-  - `continue` indicates that the interpreter should continue to the next operation in the block.
+  - `branch` indicates that the interpreter should jump to another block
 -/
 inductive ControlFlowAction where
   | return (vals : Array RuntimeValue)
-  | continue
+  | branch (vals : Array RuntimeValue) (dest : BlockPtr)
```
ineol, line 118: What's the rationale for removing the `continue` control flow action?

Veir/Interpreter/Basic.lean
```diff
@@ -87,541 +87,562 @@ def InterpreterState.setResultValues (state : InterpreterState) (ctx : IRContext
...
+    let value := values[i]!
+    let newState := state.setVar (ValuePtr.blockArgument {block := arg.owner, index := arg.index}) value
+    InterpreterState.setArgumentValues_loop newState ctx args values i
+
+/--
+  Set the values of block arguments.
+-/
+def InterpreterState.setArgumentValues (state : InterpreterState) (ctx : IRContext OpInfo)
+    (block : BlockPtr) (values : Array RuntimeValue) : InterpreterState :=
+  InterpreterState.setArgumentValues_loop state ctx (block.get! ctx).arguments values (block.getNumArguments! ctx)
```
ineol: Can you put the `..._loop` function in a where block inside the `setArgumentValues` instead and just call it `loop` or `go` or something similar?

math-fehr: If you do that, is it actually possible to verify it? I've been doing the `..._loop` in the codebase so far for that reason, let me know if I should change it as well in another PR.

Veir/Interpreter/Basic.lean
nchappe: At the end of a block, the `continue` action does not make sense, so I just removed it altogether. `interpretOp` now returns an `Option ControlFlowAction` so that `none` covers the former `continue` case in the middle of a block. I put this part in a separate commit on purpose, I can always revert it.

Veir/Interpreter/Basic.lean
```diff
@@ -633,43 +654,58 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
...
   let block := blockPtr.get ctx (by grind)
   rlet firstOp ← (blockPtr.get ctx).firstOp
   interpretOpList ctx firstOp state
 
+/--
+  Interpret a region, starting from the given block.
+  Return the values eventually returned, if any.
+  Return `none` if any errors occur during interpretation.
+-/
+def interpretRegion (ctx : IRContext OpCode) (blockPtr : BlockPtr) (state : InterpreterState) (blockInBounds : blockPtr.InBounds ctx := by grind) (wf : ctx.WellFormed := by grind) : Option (Array RuntimeValue) := do
```
math-fehr: Do you think we could have a different value whenever we have a failure than when we have an infinite loop?
Currently, if we return `none`, we cannot reason if this is because of an infinite loop or because there was an error while interpreting.

Veir/Interpreter/Basic.lean
math-fehr: I think the cons are:
* We might have non-continue `ControlFlowAction` in the middle of block at some point, this is something that is being upstreamed soon to MLIR
* Performance wise, `Option ControlFlowAction` has one more pointer indirection.

The pros are:
* It's "easier" to check that `ControlFlowAction` is not a continue for the terminator, because you can write `some action` and that directly removes that case.


I'm not sure which one is better, since they are both semantically equivalent I'm not feeling strongly for any so far, so both are fine for me.

Veir/Interpreter/Basic.lean
nchappe: I feel like this will be solved by the use of coinductive semantics for Veir: an infinite loop would be an infinite itree/ctree/whatever, whereas an error would be a finite tree returning none.

Veir/Interpreter/Basic.lean
nchappe: I mimicked `setResultValues` when writing this function but another option would be to write it in a functional style.

ineol: We just talked with Mathieu and we agree the `where` clause should be good.

nchappe:
I rebased and cleaned up the PR, including a `let rec` in `setArgumentValues`.

ineol (approved):
Thank you!

Veir/Interpreter/Basic.lean
```diff
@@ -55,6 +55,13 @@ def InterpreterState.setVar (state : InterpreterState) (var : ValuePtr) (val : R
     InterpreterState :=
   {state with variables := state.variables.insert var val}
 
+/--
+  Set the value of a block argument.
+-/
+def InterpreterState.setBlockArgument (state : InterpreterState) (arg : BlockArgument) (val : RuntimeValue) :
+    InterpreterState :=
+  state.setVar (ValuePtr.blockArgument {block := arg.owner, index := arg.index}) val
```
math-fehr: `BlockArgument` should never be used in practice, and should be an implementation detail (it's something I'm moving towards).
Can you remove this function and directly use `setVar` instead? I"ll write a comment on the use of that function to explain how we could do it.

Veir/Interpreter/Basic.lean
```diff
@@ -87,541 +94,560 @@ def InterpreterState.setResultValues (state : InterpreterState) (ctx : IRContext
...
+def InterpreterState.setArgumentValues (state : InterpreterState) (ctx : IRContext OpInfo)
+    (block : BlockPtr) (values : Array RuntimeValue) : InterpreterState :=
+  let args := (block.get! ctx).arguments
+  let rec loop (state : InterpreterState) (i : Nat) :=
+    match i with
+    | 0 => state
+    | i + 1 =>
+      let arg := args[i]!
+      let value := values[i]!
+      let newState := state.setBlockArgument arg value
```
math-fehr: Here, it would be better to have `state.setVar (block.getArgument! ctx i) value`.
It is true that a `block.getArguments` is missing that would return an `Array BlockPtr`.

Veir/Interpreter/Basic.lean
```diff
@@ -87,541 +94,560 @@ def InterpreterState.setResultValues (state : InterpreterState) (ctx : IRContext
...
-  Return the result runtime values and a control flow action indicating how
+  Return the result runtime values and an optional control flow action indicating how
   to continue the interpretation.
   If any error occurs during interpretation (e.g., unknown operation, missing variable),
   returns `none`.
 -/
 def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
-    (resultTypes : Array TypeAttr) (operands : Array RuntimeValue)
-    : Option ((Array RuntimeValue) × ControlFlowAction) :=
+    (resultTypes : Array TypeAttr) (operands : Array RuntimeValue) (blockOperands : Array BlockOperand)
```
math-fehr: Can you pass an `Array BlockPtr` instead of an `Array BlockOperand`?
`BlockOperand` exposes too many fields that we should not rely on, and in general we should avoid using these structs directly (I want to make them private).

Veir/Interpreter/Basic.lean
```diff
@@ -633,43 +659,58 @@ def interpretOp' (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
...
+-/
+def interpretRegion (ctx : IRContext OpCode) (blockPtr : BlockPtr) (state : InterpreterState) (blockInBounds : blockPtr.InBounds ctx := by grind) (wf : ctx.WellFormed := by grind) : Option (Array RuntimeValue) := do
+  match interpretBlock ctx blockPtr state blockInBounds wf with
+  | some (.return res) => some res
+  | some (.branch res succ) =>
+    if h : succ.InBounds ctx then
+      let state := state.setArgumentValues ctx succ res
+      interpretRegion ctx succ state h wf else none
+  | none => none
+partial_fixpoint
```
math-fehr: Should this be renamed something like `interpretBlockCFG`, and then having `interpretRegion` call `interpretBlockCFG` on the first block?
Just a naming issue, as `interpretRegion` is called on blocks, rather than on an region.

math-fehr (commented):
Just added a few comments on naming or uses of `BlockArgument`, but otherwise that is good to me!

nchappe:
Pushed a new version that addresses Mathieu's comments as suggested.

PR #421, opened 2026-04-15
Make sure maxHeartBeats fires in docs

tobiasgrosser:


tobiasgrosser:
This does not resolve the issue. #422 is the final solution.

PR #422, opened 2026-04-15
Move axioms to separate file

tobiasgrosser:
The doc builds currently fail with an error that axioms are not permitted in a module. Moving axioms to a separate file that is not a module fixes the doc build.

This PR also removes the previously introduced maxHeartBeats flag, which is not needed and has (for unknown reasons), no effect on the lean build. As the maxHeartBeats issues are only warnings, they do not affect the docs build.


Veir/Axioms.lean
```diff
@@ -0,0 +1,35 @@
...
+info: 'Veir.Rewriter.pushBlockOperand_DefUse' depends on axioms: [propext, Classical.choice, Quot.sound]
+-/
+#guard_msgs in
+#print axioms Rewriter.pushBlockOperand_DefUse
+
+/--
+info: 'Veir.Rewriter.pushBlockOperand_WellFormed' depends on axioms: [propext, Classical.choice, Quot.sound]
+-/
+#guard_msgs in
+#print axioms Rewriter.pushBlockOperand_WellFormed
```
tobiasgrosser:
```suggestion
#print axioms Rewriter.pushBlockOperand_WellFormed

```

luisacicolini (approved):

PR #424, opened 2026-04-16
Use the WfIRContext API for PatternRewriter

math-fehr:
Change `PatternRewriter` and `Pass` to use an `WfIRContext` instead of an `IRContext` with a proof of well-formedness.

I benchmarked it, and there seems to be no changes.


tobiasgrosser (approved):
Nice!

ineol:
I'm not a big fan of the `.val` because it exposes the fact that `WfContext` is a sigma type. How about `.ctx` or `.inner`?

PR #429, opened 2026-04-18
Ensure docs lean-toolchain version matches the version used for non-doc builds

tobiasgrosser:
Docs builds should use the same `lean-toolchain` as the non-doc builds. We implement this by making the doc lean-toolchain a symlink to the main lean-toolchain. The current version mismatch has caused some inconsistencies, including a failure when axioms in modules were not yet supported in the doc builds. We also remove in this PR the workaround we committed in #422, which was due to not identifying this version mismatch.

We also change the nightly update script to update the `docbuild` directory.


.github/workflows/update-lean-nightly-version.yml
```diff
@@ -52,6 +52,7 @@ jobs:
            sed -e "s/nightly.*/${{steps.leandate.outputs.date}}/" -i lean-toolchain
            cat lean-toolchain
            lake update --no-cache
+           (cd docbuild; lake update --no-cache)
```
math-fehr, line 55: Is this needed since `docbuild` refers to the main lean-toolchain anyway?

math-fehr (approved):
Nice! So the axiom problem was fixed in a lean update?

.github/workflows/update-lean-nightly-version.yml
tobiasgrosser: There is a `lake-manifest.json` that must be updated, AFAIU.

tobiasgrosser:
> Nice! So the axiom problem was fixed in a lean update?

It seems like. This PR is good to go. Just let me know if you are OK with the last comment.

math-fehr:
All good to me, feel free to merge it!

PR #430, opened 2026-04-19
Use ExtHashMap for the interpreter state

math-fehr:
This will allow us to more easily reason about the interpreter state, as we can then use equality between states instead of defining a property for it.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -63,14 +63,11 @@ def InterpreterState.getVar? (state : InterpreterState) (var : ValuePtr)
   state.variables[var]?
 
 @[ext]
-theorem InterpreterState.ext {state1 state2 : InterpreterState} :
-    (∀ var, state1.getVar? var = state2.getVar? var) →
-    state1 = state2 := by
-  intro h
-  have ⟨state1⟩ := state1; have ⟨state2⟩ := state2
-  simp only [mk.injEq]
-  ext key value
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
LGTM. A nice modular patch.

Veir/Interpreter/Basic.lean
```diff
@@ -62,6 +62,17 @@ def InterpreterState.getVar? (state : InterpreterState) (var : ValuePtr)
...
+@[ext]
+theorem InterpreterState.ext {state1 state2 : InterpreterState} :
+    (∀ var, state1.getVar? var = state2.getVar? var) →
+    state1 = state2 := by
+  intro h
+  have ⟨state1⟩ := state1; have ⟨state2⟩ := state2
+  simp only [mk.injEq]
+  ext key value
+  simp only [InterpreterState.getVar?] at h
+  grind
```
```suggestion
theorem InterpreterState.ext {s₁ s₂ : InterpreterState} :
    (∀ var, s₁.getVar? var = s₂.getVar? var) →
    s₁ = s₂ := by
  rcases s₁; rcases s₂
  simp only [getVar?, mk.injEq]
  grind
```
a tiny bit tidier

ineol (approved):

ineol:
Can we still dump the contents of the hashmap even with the extensional version?

math-fehr:
Talked to Leo offline about it, we can find a way to dump it later (it is possible).

PR #431, opened 2026-04-19
Add InsertPoint.after

math-fehr:
This is a constructor to build an `InsertPoint` that points after an operation. It requires to pass the operation parent block, as we can only get the InsertPoint after an operation if it has a parent.

Also, add two `simp` annotations that were missing.

### DIFF
--- a/Veir/Rewriter/InsertPoint.lean
+++ b/Veir/Rewriter/InsertPoint.lean
@@ -27,10 +27,22 @@ def InsertPoint.InBounds (ip : InsertPoint) (ctx : IRContext OpInfo) : Prop :=
   | before op => op.InBounds ctx
   | atEnd bl => bl.InBounds ctx
 
-@[grind =]
+@[simp, grind =]
 theorem InsertPoint.inBounds_before : (before op).InBounds ctx ↔ op.InBounds ctx := by rfl
-@[grind =]
+@[simp, grind =]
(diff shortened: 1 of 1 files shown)


Veir/Rewriter/InsertPoint.lean
```diff
@@ -27,11 +27,24 @@ def InsertPoint.InBounds (ip : InsertPoint) (ctx : IRContext OpInfo) : Prop :=
...
   | atEnd bl => bl.InBounds ctx
 
-@[grind =]
+@[simp, grind =]
 theorem InsertPoint.inBounds_before : (before op).InBounds ctx ↔ op.InBounds ctx := by rfl
-@[grind =]
+@[simp, grind =]
 theorem InsertPoint.inBounds_atEnd : (atEnd bl).InBounds ctx ↔ bl.InBounds ctx := by rfl
 
+set_option linter.unusedVariables false in
```
tobiasgrosser: Alternatively, you can use argument names that start with an `_`.

tobiasgrosser (approved):

ineol (approved):

Veir/Rewriter/InsertPoint.lean
math-fehr: What would you prefer here?
I used the linter thing so that the name is still nice when passing the argument explicitent at call site.

ineol: Personally it's easier with the `_` because it's clear which argument is unused. Otherwise a comment to explain is also fine.

PR #433, opened 2026-04-19
Definition of the equation lemma

math-fehr:
Defines the equation lemma, as well as an invariant on `InterpreterState` to state that every dominating operation satisfies the equation lemma.

### DIFF
--- a/Veir/Interpreter/EquationLemma.lean
+++ b/Veir/Interpreter/EquationLemma.lean
@@ -1,0 +1,96 @@
+import Veir.Interpreter.Basic
+import Veir.Interpreter.Lemmas
+import Veir.Dominance
+
+/-!
+# Equation Lemma and SSA Invariant
+
+This file contains the definition of the equation lemma (`EquationLemmaAt`) and the proof of
+preservation when interpreting an operation.
+
+The equation lemma is based on the paper "A formally verified SSA-based middle-end" from
+Gilles Barthe, Delphine Demange, and David Pichardie. Our definition is slightly different as
+CompCertSSA semantics are based on small-step operational semantics.
+-/
+
+namespace Veir
+
+variable {OpInfo : Type} [HasOpInfo OpInfo]
+variable {state : InterpreterState}
+
+/-!
+## Equation Lemma
+-/
+
+/--
+`state.EquationHolds ctx op` holds when `state` records the result of interpreting `op`.
+This is encoded by stating that interpreting `op` in `state` produces `state` itself, which is
+equivalent to saying that the results of interpreting `op` on the given `state` are present in
+`state` itself.
+-/
+def InterpreterState.EquationHolds (state : InterpreterState) (ctx : WfIRContext OpCode)
+    (op : OperationPtr) : Prop :=
+  ∃ controlFlow, interpretOp ctx op state = some (state, controlFlow)
+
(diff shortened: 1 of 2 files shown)


Veir/Interpreter/WellFormed.lean
```diff
@@ -0,0 +1,95 @@
...
+    (op : OperationPtr) : Prop :=
+  ∃ controlFlow, interpretOp ctx op state = some (state, controlFlow)
+
+set_option linter.unusedVariables false in
+/--
+  A well-formedness condition for the interpreter state at a given location.
+  It states that every operation that dominates the given operation has its equation
+  lemma satisfied.
+-/
+def InterpreterState.WellFormed (state : InterpreterState) (ctx : WfIRContext OpCode)
```
ineol: Shouldn't we relate `location` and `state` somehow?

math-fehr:
This is now ready for review!

Veir/Interpreter/EquationLemma.lean
```diff
@@ -0,0 +1,91 @@
+import Veir.Interpreter.Basic
+import Veir.Interpreter.Lemmas
+import Veir.Dominance
+
+/-!
+# Equation Lemma and SSA Invariant
+
+This file contains the definition of the equation lemma (`SatisfiesEquationLemma`), and a
+proposition that states that the interpreter state satisfies the equation lemma for all operations
+that dominate a given program point (`SSAInvAt`).
```
tobiasgrosser: If this relates to a lemma introduced in the CompCert world, can you link to the paper that introduced it an explain the relationship between your lemma and the one in CompCert?

tobiasgrosser (approved):
Very cool!

math-fehr:
I'm really not sure why this has conflicts, let's see what happens after we merge the first PR

Veir/Interpreter/EquationLemma.lean
```diff
@@ -0,0 +1,96 @@
...
+/--
+`state.HasInterpretedOperation ctx op` holds when `state` records the result of interpreting `op`.
+This is encoded by stating that interpreting `op` in `state` produces `state` itself, which is
+equivalent to saying that the results of interpreting `op` on the given `state` are present in
+`state` itself.
+-/
+def InterpreterState.HasInterpretedOperation (state : InterpreterState) (ctx : WfIRContext OpCode)
+    (op : OperationPtr) : Prop :=
+  ∃ controlFlow, interpretOp ctx op state = some (state, controlFlow)
+
```
ineol: I'm not sure about this name... if you had a noop operation it would be considered as having been interpreted everywhere which is strange?

ineol: Maybe something like `EquationHolds` or something?

ineol (approved):

Veir/Interpreter/EquationLemma.lean
math-fehr: Yes EquationHolds is perfect!

Veir/Interpreter/WellFormed.lean
math-fehr: How should we do that? My understanding is that `EquationLemmaAt` is now the thing that now relates the location and the state.

luisacicolini (approved):

PR #434, opened 2026-04-19
Decidable instances for (Block)InsertPoint.InBounds

GZGavinZhao:
~~Required by #222.~~
Required by #435.

### DIFF
--- a/Veir/Rewriter/InsertPoint.lean
+++ b/Veir/Rewriter/InsertPoint.lean
@@ -27,6 +27,13 @@ def InsertPoint.InBounds (ip : InsertPoint) (ctx : IRContext OpInfo) : Prop :=
   | before op => op.InBounds ctx
   | atEnd bl => bl.InBounds ctx
 
+theorem InsertPoint.inBounds_def :
+    ip.InBounds ctx = (match ip with | before op => op.InBounds ctx | atEnd bl => bl.InBounds ctx) :=
+  by rfl
+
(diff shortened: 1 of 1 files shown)


Veir/Rewriter/InsertPoint.lean
```diff
@@ -21,12 +21,15 @@ inductive InsertPoint where
   | atEnd (block: BlockPtr)
 deriving DecidableEq
 
-@[grind]
+@[grind, expose]
```
GZGavinZhao: Is it ok to add `expose` here? This was the easiest fix I know to make `simp only [InsertPoint.InBounds]` work below.

Veir/Rewriter/InsertPoint.lean
```diff
@@ -332,10 +335,14 @@ inductive BlockInsertPoint where
 
 namespace BlockInsertPoint
 
-@[grind]
+@[grind, expose]
```
GZGavinZhao: Same question as the `expose` annotation on `InsertPoint.InBounds`.

Veir/Rewriter/InsertPoint.lean
ineol: If you look at `IR/Basic.lean` instead we define theorems `InBounds_def` that we rewrite by hand.

tobiasgrosser:
@GZGavinZhao, can you rebase the IR and resolve @ineol's comment.

GZGavinZhao:
Will rebase and address the feedback tonight!

math-fehr (approved):

PR #438, opened 2026-04-20
Implement `OperationPtr.getOpResults` and `OperationPtr.getResults`

math-fehr:
As well as their associated lemmas. This is mostly a copy-paste of `OperationPtr.getOperands`.

There is two versions of this since we might want to return an array of `ValuePtr` or `OpResultPtr`.


ineol (approved):
I wonder if we should try marking some functions as `noncomputable` to document which ones are not meant to be used in programs

math-fehr:
I do agree with the `noncomputable`, though there we should use these functions in programs. At least until we have iterators

PR #445, opened 2026-04-20
Fix printing of operations that have both successors and properties

nchappe:
As per https://mlir.llvm.org/docs/LangRef/#operations, successors should go first.
The parser is correct, but the printer inverts them.

As there is a bug with the current parsing of unregistered operations, the test will be added when we add `cf.cond_br`

### DIFF
--- a/Veir/Printer.lean
+++ b/Veir/Printer.lean
@@ -186,8 +186,8 @@ partial def printOperation (ctx: IRContext OpCode) (op: OperationPtr) (indent: N
   printOpResults ctx op
   IO.print s!"\"{String.fromUTF8! opStruct.opType.name}\""
   printOpOperands ctx op
-  printOpProperties ctx op
   printBlockOperands ctx op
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
Nice. Can you add a test case?

nchappe:
My upcoming `cf.cond_br` PR will contain a test with both successors and properties.
Do you want me to add another such test in `Test/Parsing/` to be safe?

math-fehr:
There is actually a bug in our code right now where we remove properties of programs, is it fine if we only add a test in Nicolas next PR, so we don't have to fix all of this right now?

tobiasgrosser (approved):
> There is actually a bug in our code right now where we remove properties of programs, is it fine if we only add a test in Nicolas next PR, so we don't have to fix all of this right now?

Sure. Then just mention this in the PR message. If no test case is added in a PR, the PR message should explain the strategy.

PR #446, opened 2026-04-20
Parse and verify `cf.cond_br`

nchappe:
This depends on #445.
Because `cf.cond_br` takes variadic operands there is some magic with the `operandSegmentSizes` property, I handle that in an ad-hoc way for now.


math-fehr (approved):

tobiasgrosser (approved):
Nice!

PR #447, opened 2026-04-20
Fix compilation of `Interpreter/Lemmas.lean`

math-fehr:
The file was not imported from anywhere.

### DIFF
--- a/Veir.lean
+++ b/Veir.lean
@@ -11,3 +11,4 @@ import Veir.Printer
 import Veir.PatternRewriter.Basic
 import Veir.Benchmarks
 import Veir.Parser.Lexer
+import Veir.Interpreter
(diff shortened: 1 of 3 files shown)


Veir.lean
```diff
@@ -11,3 +11,4 @@ import Veir.Printer
 import Veir.PatternRewriter.Basic
 import Veir.Benchmarks
 import Veir.Parser.Lexer
+import Veir.Interpreter.Lemmas
```
tobiasgrosser: Should `Veir.lean` not interpret `Veir.Interpreter` and `Veir.Interpreter` would then import `Veir.Interpreter.Lemmas`?

math-fehr: So do you suggest that we should have a `Veir.Interpreter` file that imports all the `Interpreter` folder?

tobiasgrosser (approved):
LGTM.

PR #449, opened 2026-04-20
Implement `ValuePtr.getDefiningOp`

math-fehr:
This function is quite useful to check if a value is an operation result, and get the operation defining it at the same time.


luisacicolini (approved):
awesome, thanks!

PR #450, opened 2026-04-20
riscv: combine and proof automation

luisacicolini:
We add a `right_identity_op_add` combine that folds `riscv.add lhs 0 -> lhs`, including an automation simp-set to prove the semantics equivalence. The name is the same used in [LLVM](https://github.com/llvm/llvm-project/blob/ab4283959fd10fa347e0a60507b7101ad273fc57/llvm/include/llvm/Target/GlobalISel/Combine.td#L554). 
In LLVM, this combine lives at the generic machine IR level (GMIR), and is therefore not riscv-specific. However, at the moment, we introduce a new namespace `Veir.RISCV` where these combines live, together with the proofs and the matching methods. In the future we can replace this with, e.g., a GMIR namespace that can match different backends. 

Concerning the proof automation, we introduce a new `reg_toBitVec` simp set, copying the design from the way Lean treats `UInt*` types. This set unfolds the semantics of `Reg` operations to `BitVec` operations.


Test/Passes/Combines/right_identity_op_add.mlir
```diff
@@ -0,0 +1,15 @@
...
+    "func.func"() ({
+    ^bb(%a : !reg):
+        %c0 = "riscv.li"() <{"value" = 0 : i64}>: () -> !reg
+        %add = "riscv.add"(%a, %c0) : (!reg, !reg) -> !reg
+        // The remaining value is unused and can be eliminated by DCE.
+        // CHECK:      %[[zero:.*]] = "riscv.li"() <{"value" = 0 : i64}> : () -> !reg
+        %c1 = "riscv.li"() <{"value" = 1 : i64}>: () -> !reg
+        %add1 = "riscv.add"(%a, %c1) : (!reg, !reg) -> !reg
+        // CHECK-NEXT:  %{{.*}} = "riscv.li"() <{"value" = 1 : i64}> : () -> !reg
+        // CHECK-NEXT:  %{{.*}} =  "riscv.add"(%{{.*}}, %{{.*}}) : (!reg, !reg) -> !reg
```
math-fehr: Can you add regex values here so that it's obvious that the value was replaced?

Veir/Passes/Combines/Combine.lean
```diff
@@ -0,0 +1,37 @@
...
+/-! # Pass implementation -/
+
+def RISCVCombines.impl (ctx : WfIRContext OpCode) (op : OperationPtr) (_ : op.InBounds ctx.raw) :
+    ExceptT String IO (WfIRContext OpCode) := do
+  let pattern := RewritePattern.GreedyRewritePattern #[right_identity_zero_add]
+  match RewritePattern.applyInContext pattern ctx with
+  | none => throw "Error while applying pattern rewrites"
+  | some ctx => pure ctx
+
+public def RISCVCombines : Pass OpCode :=
```
math-fehr:
```suggestion
public def RISCVCombine : Pass OpCode :=
```
?

Veir/Data/RISCV/Reg/Simp.lean
```diff
@@ -0,0 +1,6 @@
+
```
math-fehr:
```suggestion
```

Veir/Data/RISCV/Reg/Simp.lean
```diff
@@ -0,0 +1,6 @@
...
+module
+
```
math-fehr, line 2: I named a similar file `Simp.lean`.
Is `Tactic.lean` the expected name (I don't know the conventions)

math-fehr (commented):
Looks good to me, I just had some nit comments!

Veir/Data/RISCV/Reg/Simp.lean
luisacicolini: that makes more sense, let me change that

Test/Passes/Combines/right_identity_op_add.mlir
luisacicolini: done

Veir/Passes/Combines/Combine.lean
luisacicolini: updated

Veir/Data/RISCV/Reg/Simp.lean
luisacicolini: fixed

luisacicolini:
thank you @math-fehr for the review - everything is addressed!

math-fehr (approved):

Veir/Passes/RISCVMatching.lean
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Veir
+
+/--
+  Match an operation that has a single result, a specific opcode,
+  and a specific number of operands.
+  Returns the operands and the properties of the operation if it matches, or `none` otherwise.
+-/
+
+def matchRISCVAdd (op : OperationPtr) (ctx : IRContext OpCode) : Option (ValuePtr × ValuePtr) := do
```
tobiasgrosser:
```suggestion
def matchRISCVAdd (op : OperationPtr) (ctx : IRContext OpCode) : Option (ValuePtr × ValuePtr) := do
```

Do we need the empty line here?

Veir/Passes/RISCVMatching.lean
```diff
@@ -0,0 +1,22 @@
...
+  let (op, _) ← matchOp op ctx (.riscv .add) 2
+  return (op[0]!, op[1]!)
+
+
+def matchRISCVLi (op : OperationPtr) (ctx : IRContext OpCode) : Option ( propertiesOf (.riscv .li)) := do
```
tobiasgrosser: Rather than having RISCV here in every operation, should we use a namespace, e.g., `VEIR.Dialect.RISCV.matchAdd` and `VEIR.Dialect.RISCV.matchLI`?

Veir/Passes/Combines/Proofs.lean
```diff
@@ -0,0 +1,20 @@
+import Veir.Data.RISCV.Reg.Basic
+import Veir.Data.RISCV.Reg.Simp
+import Veir.Data.Casting
+import Veir.Data.Refinement
```
tobiasgrosser: Do we need all these imports at the moment? I suggest to only add imports when they are needed here.

Veir/Passes/Combines/Combine.lean
```diff
@@ -0,0 +1,37 @@
...
+
+/-!
+  This file replicates LLVM's pre- and post-legalization GlobalISel combines.
+-/
+
+/-! # Lowering Patterns -/
+
+set_option warn.sorry false in
+/-- riscv.add x 0 -> x -/
+def right_identity_zero_add (rewriter: PatternRewriter OpCode) (op: OperationPtr) :
```
tobiasgrosser, line 15: Should this be `add_zero` instead of `zero_add`? Also, why do we have `right_identity` in the name? Is this inspired from somewhere?

Veir/Passes/Combines/Combine.lean
```diff
@@ -0,0 +1,37 @@
+import Veir.Pass
+import Veir.PatternRewriter.Basic
+import Veir.Passes.RISCVMatching
+
+namespace Veir
+
```
tobiasgrosser, line 6: Should this be in a specific namespace, e.g., `Veir.Passes.Combines.RISCV`? It may make sense to briefly explain in the PR message the folder structure you have in mind.

Veir/Data/RISCV/Reg/Basic.lean
```diff
@@ -24,6 +26,7 @@ instance : ToString Reg where
 /--
   Load a 64-wide immediate into the destination register rd
 -/
+@[riscv_semantics]
```
tobiasgrosser: Does `BitVec.UInt8` use simp sets on the definitions themselves, or where did you find inspiration for this style?

tobiasgrosser (commented):
Amazing. We start to prove things here. 🙏

Veir/Passes/RISCVMatching.lean
luisacicolini: added namespace

Veir/Passes/Combines/Proofs.lean
luisacicolini: removed

Veir/Passes/Combines/Combine.lean
luisacicolini: added message, updated PR message

Veir/Data/RISCV/Reg/Basic.lean
luisacicolini: I used the same strategy we adopted in lean-mlir. AFAIU in lean it's in the `simp` set directly

Veir/Passes/Combines/Combine.lean
luisacicolini: I used the same names as LLVM. Updated PR message to explain this.

Veir/Data/RISCV/Reg/Basic.lean
tobiasgrosser: Can you double-check how Lean does this for `UInt8`. In particular, I believe UInt8 uses a function `toBitVec`. You do not use it here, no?

luisacicolini: afaiu `toBitVec` is a field of the data type itself:
```
structure UInt8 where
  /--
  Creates a `UInt8` from a `BitVec 8`. This function is overridden with a native implementation.
  -/
  ofBitVec ::
  /--
  Unpacks a `UInt8` into a `BitVec 8`. This function is overridden with a native implementation.
  -/
  toBitVec : BitVec 8
 ```
 in the `Reg` we just have a `value : BitVec 64`. 
 
 There is also a macro [`declare_uint_theorems` ](https://github.com/leanprover/lean4/blob/master/src/Init/Data/UInt/Lemmas.lean), but I don't think that's what we want. Imho what we want is something like [`int_toBitVec`](https://github.com/leanprover/lean4/blob/master/src/Lean/Elab/Tactic/BVDecide/Frontend/Attr.lean), which contains the `"simp theorems used to convert UIntX/IntX statements into BitVec ones"` and is what we are doing with `riscv_semantics` as well. AFAIU the only difference is that the uint tactic is part of `bv_decide`, while our simpproc is not.

Comment (math-fehr) [reply Veir/Data/RISCV/Reg/Basic.lean]: To me that makes sense, and I think we should do that more in VeIR in general.
Having a simp set for a specific goal is quite good in general, and gives you some automation for free (that's why we have the `eq_bang` simp set).

File: Veir/Passes/Combines/Combine.lean [hunk 11]
```diff
@@ -0,0 +1,37 @@
...
+def Combine.impl (ctx : WfIRContext OpCode) (op : OperationPtr) (_ : op.InBounds ctx.raw) :
+    ExceptT String IO (WfIRContext OpCode) := do
+  let pattern := RewritePattern.GreedyRewritePattern #[right_identity_zero_add]
+  match RewritePattern.applyInContext pattern ctx with
+  | none => throw "Error while applying pattern rewrites"
+  | some ctx => pure ctx
+
+public def Combine : Pass OpCode :=
+  { name := "riscv-combine"
+    description :=
```
ineol: Spurious newline?

ineol (approved):

Veir/Passes/Combines/Combine.lean
luisacicolini: fixed

luisacicolini:
Addressed all the comments and introduced equational and injectivity lemmas to turn `Reg` operations into `BitVec` ones.

tobiasgrosser (approved):
Thank you, Luisa!

PR #451, opened 2026-04-20
Add lemmas about the interpreter and dominance

math-fehr:
Since we do not have a dominance theory yet, we add a list of axioms for now.

### DIFF
--- a/Veir/Dominance.lean
+++ b/Veir/Dominance.lean
@@ -1,0 +1,121 @@
+module
+
+public import Veir.IR.WellFormed
+public import Veir.IR.OpInfo
+public import Veir.Rewriter.InsertPoint
+
+/-!
+  # Dominance
+
+  This file is a placeholder for the dominance relation between IR constructs.
+  It currently only contains axioms, and will be filled in with actual definitions and proofs
+  in the future.
+-/
+
+public section
+
+namespace Veir
+
+variable {OpInfo : Type} [HasOpInfo OpInfo]
+variable {ctx : WfIRContext OpInfo}
+variable {op op₁ op₂ : OperationPtr}
+
+/-!
+## Definition of Dominance
+-/
+
+/--
+  The dominance relation between operations.
+  An operation `op` dominates itself.
+-/
+axiom OperationPtr.dominates (op₁ op₂ : OperationPtr) (ctx : WfIRContext OpInfo) : Prop
+
+/--
+  The strict dominance relation between operations.
+  An operation `op` does not strictly dominate itself.
+-/
+def OperationPtr.strictlyDominates (op₁ op₂ : OperationPtr) (ctx : WfIRContext OpInfo) : Prop :=
+  op₁.dominates op₂ ctx ∧ op₁ ≠ op₂
+
+/--
+  An operation `op₁` strictly dominates an operation `op₂` if it dominates it
+  and the operations are not equal.
+-/
+theorem OperationPtr.strictlyDominates_def :
+    op₁.strictlyDominates op₂ ctx ↔ op₁.dominates op₂ ctx ∧ op₁ ≠ op₂ := by
+  grind [strictlyDominates]
+
+/--
+  The dominance relation between an operation and an insertion point.
+-/
+axiom OperationPtr.dominatesIp (op : OperationPtr) (ip : InsertPoint) (ctx : WfIRContext OpInfo) : Prop
+
+/-!
+## Lemmas about Dominance
+-/
+
+/--
+An operation `op₁` dominates an operation `op₂` if it strictly dominates it.
+-/
+theorem OperationPtr.dominates_of_strictlyDominates :
+    op₁.strictlyDominates op₂ ctx → op₁.dominates op₂ ctx := by
+  grind [strictlyDominates]
+
+/--
+An operation dominates itself.
+-/
+@[grind .]
+axiom OperationPtr.dominates_refl : op.dominates op ctx
+
+/--
+An operation `op₁` dominates an operation `op₂` if and only if
+`op₁` stricly dominates `op₂` or if `op₁` is `op₂`.
+-/
+theorem OperationPtr.dominates_iff_strictlyDominates_or_eq :
+    op₁.dominates op₂ ctx ↔ op₁.strictlyDominates op₂ ctx ∨ op₁ = op₂ := by
+  grind [OperationPtr.strictlyDominates]
+
+/--
+An operation `op₁` dominates the program point after a given operation `op₂` if it
+either dominates the `op₂`, or is `op₂`.
+-/
+axiom OperationPtr.dominatesIp_iff :
+    op₁.dominatesIp (InsertPoint.after op₂ ctx.raw block op₂HasParent op₂InBounds) ctx ↔
+    op₁.dominates op₂ ctx
+
+/--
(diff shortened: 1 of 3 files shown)


Veir/Dominance.lean
```diff
@@ -0,0 +1,72 @@
...
+    op₁.dominates op₂ ctx ∨ op₁ = op₂
+
+axiom OperationPtr.dominatesIp_before : op₁.dominatesIp (.before op₂) ctx ↔ op₁.dominates op₂ ctx
+
+axiom dominates_next :
+  op₁.dominates op₂ ctx →
+  (op₂.get! ctx.raw).next = some nextOp →
+  op₁.dominates nextOp ctx
+
+axiom IRContext.Dom.value_not_in_results_of_in_dominated_op_operands :
```
tobiasgrosser: Is there a reason you make the arguments here (also in others) not explicitly named arguments?

@luisacicolini, could you look at the naming convention here?

tobiasgrosser (approved):
Nice. Thank you!

Veir/Dominance.lean
```diff
@@ -0,0 +1,72 @@
...
+axiom OperationPtr.dominatesIp_iff :
+    op₁.dominatesIp (InsertPoint.after op₂ ctx.raw block op₂HasParent op₂InBounds) ctx ↔
+    op₁.dominates op₂ ctx ∨ op₁ = op₂
+
+axiom OperationPtr.dominatesIp_before : op₁.dominatesIp (.before op₂) ctx ↔ op₁.dominates op₂ ctx
+
+axiom dominates_next :
+  op₁.dominates op₂ ctx →
+  (op₂.get! ctx.raw).next = some nextOp →
+  op₁.dominates nextOp ctx
```
tobiasgrosser: Why does this not have any namespace/prefix associated?

Veir/Dominance.lean
luisacicolini: `value_not_in_results_of_values_in_operands_of_dominates_of_dom` maybe? The idea is to have an `of` for each hypothesis you require. It's a bit hard to guess though, I dont think in Lean there are such long names, but we clearly need it.

math-fehr: I changed it to `IRContext.Dom.value_not_in_results_of_values_in_operands_of_dominates`. `dom` is the namespace here, so we shouldn't add `of_dom`. Is that good to you?

Also, why should the arguments be explicitely named here? In lemmas we usually keep most arguments implicit, so then we don't have to pass all arguments (since a lot of them can be deduced).

Veir/Dominance.lean
math-fehr: Fixed!

Veir/Dominance.lean
luisacicolini: my reasoning is as follows: 
* 1 `of` for `ctx.Dom`
* 1 `of` for `op₁.dominates op₂ ctx`
* 1 `of` for `∀ (value : ValuePtr), value ∈ op₁.getOperands! ctx.raw`

that's the reason for the last `Dom`... alternatively, we could have `Dom.value_not_in_results_of_forall_in_operands_of_dominates`? maybe that is both more coincise and closer to lean. To be fair, I don't have a strong opinion. This really looks like a complex name regardless, so we should have a meaningful comment and just live with it (btw. could you add a comment?)

Veir/Dominance.lean
ineol: Maybe choose more consistent/self-documenting names for ops?

Veir/Dominance.lean
```diff
@@ -0,0 +1,72 @@
...
+
+axiom OperationPtr.dominatesIp_before : op₁.dominatesIp (.before op₂) ctx ↔ op₁.dominates op₂ ctx
+
+axiom OperationPtr.dominates_next :
+  op₁.dominates op₂ ctx →
+  (op₂.get! ctx.raw).next = some nextOp →
+  op₁.dominates nextOp ctx
+
+axiom IRContext.Dom.value_not_in_results_of_values_in_operands_of_dominates :
+    ctx.Dom →
```
tobiasgrosser: Is there a reason you pass 'ctx:Dom` via an implication instead of nambed argument?

math-fehr: Just changed it!

Veir/Dominance.lean
math-fehr: Went with your version!

math-fehr:
I added a lot of documentation on `Dominance.lean`, let me know if this looks good!

Veir/Interpreter/Lemmas.lean
```diff
@@ -80,3 +82,42 @@ theorem InterpreterState.setResultValues_comm {ctx : WfIRContext OpInfo}
...
+  case blockArgument blockArg => grind
+  case opResult opRes =>
+    simp only [ite_eq_right_iff, and_imp]
+    simp only [OperationPtr.getResults!.mem_iff_exists_index, ValuePtr.opResult.injEq, not_exists,
+      not_and] at this
+    grind [OperationPtr.getResult_def, cases OpResultPtr]
+
+@[grind =>]
+theorem InterpreterState.getOperandValues_setResultValues_of_dominates {ctx : WfIRContext OpInfo}
+    (ctxDom : ctx.Dom) (hdom : op'.dominates op ctx)  :
```
tobiasgrosser:
```suggestion
    (ctxDom : ctx.Dom) (hdom : op'.dominates op ctx) :
```

tobiasgrosser (approved):

ineol (approved):

luisacicolini (approved):

Veir/Dominance.lean
```diff
@@ -0,0 +1,109 @@
...
+  The dominance relation between operations.
+  An operation `op` dominates itself.
+-/
+axiom OperationPtr.dominates (op₁ op₂ : OperationPtr) (ctx : WfIRContext OpInfo) : Prop
+
+/--
+  The strict dominance relation between operations.
+  An operation `op` does not strictly dominate itself.
+-/
+axiom OperationPtr.strictlyDominates (op₁ op₂ : OperationPtr) (ctx : WfIRContext OpInfo) : Prop
```
ineol: Can't you define it as `dominates` and `op1 \neq op2`?

ineol: This would make the following few axioms trivial lemmas

Veir/Interpreter/Lemmas.lean
```diff
@@ -80,3 +82,42 @@ theorem InterpreterState.setResultValues_comm {ctx : WfIRContext OpInfo}
     (state.setResultValues ctx.raw op₂ resValues₂).setResultValues ctx.raw op₁ resValues₁ := by
   ext val runtimeVal
   cases val <;> grind
+
+theorem InterpreterState.getVar?_setResultValues_of_dominates {ctx : WfIRContext OpInfo}
```
ineol: I'd add the word "operand" in the lemma name since that's what makes the lemma true

ineol (approved):

Veir/Interpreter/Lemmas.lean
math-fehr: Is this `InterpreterState.getVar?_setResultValues_operand_of_dominates` better?

PR #453, opened 2026-04-22
Remove unpacking of array

math-fehr:
Unpacking arrays doesn't produce the right lemmas to reason about it, so we should convert it to a list first.


tobiasgrosser (approved):

tobiasgrosser:
That's a trivial PR. I think this can be merged "as obvious".

PR #457, opened 2026-04-23
bitblastable `Int` type and infra

luisacicolini:
Inspired by #178, we introduce a bitblastable integer type `IntBv` and define some basic infrastructure to convert an `LLVM.Int w` into an `IntBv w`, namely `toIntBv : Int w -> IntBv w`. 
We also introduce two helper definitions `isPoison` and `getValue`, to extract the poison flag and bitvector value from an `Int w`, and a simp-set `llvm_toBitVec` that converts LLVM types and operations to bitblastable ones.
Finally, we generalize the previous refinement definition for LLVM integers `Int w` and add a refinement definition for the bitblastable type `IntBv`, useful to reason about the LLVM semantics, and we prove it equivalent to the standard one, such that `bv_decide` can unfold it and reason about it. We use `⊑` as a notation for refinement.

The `Bitblast.lean` file is organized as follows:  

**generic lemmas**: we add some lemmas to `bv_normalize` and/or to the simp-set `llvm_toBitVec`, these are rather generic and useful to unfold and simplify constructs. I tried my best to make the `llvm_toBitVec` confluent, but I can not guarantee it is such. 

**LLVM-operations-specific lemmas**: for each LLVM `op` we have three lemmas
1. a `toIntBv_op` lemma that unfolds `(op ...).toIntBv` and pushes the `toIntBv` conversion to the leaves. These lemmas mirror what happens already for registers (with the [`reg_toBitVec`](https://github.com/opencompl/veir/blob/ca1dd404e93e93ac2eb94ab7a36da578bd20b6f1/Veir/Data/RISCV/Reg/Lemmas.lean#L20) simp-set) and, more broadly, what happens in Lean for fixed-width integers (with the [`int_toBitVec`](https://github.com/leanprover/lean4/blob/master/src/Lean/Elab/Tactic/BVDecide/Frontend/Attr.lean) simp-set). This first lemma uses the `getValue` and `isPoison` definitions, for succintness.
2. a `isPoison_op` lemma, that defines when the operation returns poison, and is used by the simp-set to unfold what happens in lemma no. (1)
3. a `getValue_op` lemma, that defines when the operation returns poison, and is used by the simp-set to unfold what happens in lemma no. (1)

Additionally, to test this infrastructure, we include the tests retrieved from lean-mlir's [Alive lemmas](https://github.com/opencompl/lean-mlir/blob/main/SSA/Projects/InstCombine/AliveStatements.lean), excluding 3 lemmas with `LLVM.not` and `LLVM.neg` operations, which we do not support yet. 
We solve **all** the lemmas by calling `simp [llvm_toBitVec]; bv_decide`, except for 1 lemma where `bv_decide` times out (imho because it invokes the bitblasting multiplication circuit with width `64`). 
At the moment the file is not properly tested in CI, we should ideally find a proper location and then add it.

The PR ended up quite big, but I think it explores a broad enough surface (with enough design choices and their consequences) for us to make an informed decision about whether this is what we are looking for or not. I am happy to split it up in reasonable chunks once consensus is reached.


Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
-    mul x (mul y z) = mul (mul x y) z := by
-  simp only [HMul.hMul, Mul.mul, mul, Id.run]
-  cases x <;> cases y <;> cases z <;> simp [BitVec.mul_assoc]
-
-@[grind =]
-theorem mul_comm {w : Nat} {nsw nuw : Bool} (x y : Int w) :
-    mul x y nsw nuw = mul y x nsw nuw := by
-  simp only [Id.run, Veir.Data.LLVM.Int.mul]
-  cases x <;> cases y <;>
-  simp [BitVec.mul_comm, BitVec.smulOverflow_comm, BitVec.umulOverflow_comm]
```
tobiasgrosser, line 54: Is there a reason you removed the existing theory on `LLVM.Int`?

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+  the `poison` field indicates whether the value is poison,
+  and `inv` preserves a proof that if the value is poison its corresponding bitvector
+  value is `0#w`.
+-/
+@[ext]
+structure IntBv (w : Nat) where
+  toBitVec : BitVec w
+  poison : Bool
+  inv : poison → (toBitVec = 0#w) := by simp
+deriving DecidableEq
```
tobiasgrosser: I wonder if this would be better placed in a file `Int/BitBlast.lean`?

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+
+/-- An `LLVM.Int w` is converted into a structure `IntBv`, where
+  the `poison` field indicates whether the `Int` is poison. -/
+def toIntBv (x : Int w) : IntBv w :=
```
tobiasgrosser:
```suggestion
def toIntBv {w : Nat} (x : Int w) : IntBv w :=
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+  match h : x with
+  | .val v => ⟨v, false, by simp⟩
+  | .poison => ⟨0#w, true, by simp⟩
```
tobiasgrosser:
```suggestion
  | .val v => {toBitVec := v, poison := false}
  | .poison => {toBitVec := 0#w, poison := true}
```
Record syntax allows us to elide `inv`. Not sure if this is more readable?

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    simp only [toIntBv, IntBv.mk.injEq, and_true] at h
+    simp [h]
+  | .poison, .poison, _ => rfl
+  | .val v,  .poison, h => by
+    simp [toIntBv] at h
+  | .poison, .val v,  h => by
+    simp [toIntBv] at h
+
+@[llvm_toBitVec]
+theorem int_inj (i1 i2 : Int w) :
```
tobiasgrosser:
```suggestion
theorem int_inj {w : Nat} (i1 i2 : Int w) :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    i1 = i2 ↔ i1.toIntBv = i2.toIntBv := ⟨(· ▸ rfl), Int.toIntBv.inj⟩
+
+@[llvm_toBitVec]
+theorem toIntBv_poison :
```
tobiasgrosser:
```suggestion
theorem toIntBv_poison {w : Nat} :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    poison.toIntBv = ⟨0#w, true, by simp⟩ := by
+  simp [toIntBv]
+
+@[llvm_toBitVec]
+theorem toIntBv_val :
```
tobiasgrosser:
```suggestion
theorem toIntBv_val {w : Nat} {v : BitVec w} :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+  obtain ⟨bv, poison, h⟩ := x
+  exact h
+
+@[bv_normalize]
+theorem eq_iff {w : Nat} {l r : IntBv w} :
+    l = r ↔ l.toBitVec = r.toBitVec ∧ l.poison = r.poison :=
+  IntBv.ext_iff
+
+@[bv_normalize]
+theorem toBitVec_ite {w} (b : Prop) [Decidable b] (x y : IntBv w) :
```
tobiasgrosser:
```suggestion
theorem toBitVec_ite {w : Nat} (b : Prop) [Decidable b] (x y : IntBv w) :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    (if b then x else y).toBitVec = if b then x.toBitVec else y.toBitVec := by
+  split <;> rfl
+
+@[bv_normalize]
+theorem poison_ite {w} (b : Prop) [Decidable b] (x y : IntBv w) :
```
tobiasgrosser:
```suggestion
theorem poison_ite {w : Nat} (b : Prop) [Decidable b] (x y : IntBv w) :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    (if b then x else y).poison = if b then x.poison else y.poison := by
+  split <;> rfl
+
+@[llvm_toBitVec]
+theorem ite_poison_eq (x : Int w) :
```
tobiasgrosser:
```suggestion
theorem ite_poison_eq {w : Nat} (x : Int w) :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+@[bv_normalize]
+theorem poison_toBitVec_constraint {w : Nat} (x : IntBv w) :
+    x.poison = true → x.toBitVec = 0#w := x.inv
+
+theorem ext (l r : IntBv w) (h : l.toBitVec = r.toBitVec ∧ l.poison = r.poison) :
+    l = r := by
+  exact IntBv.ext_iff.mpr h
+
+@[bv_normalize]
+theorem IntBv.toBitVec_bif {w} (b : Bool) (x y : IntBv w) :
```
tobiasgrosser:
```suggestion
theorem IntBv.toBitVec_bif {w : Nat} (b : Bool) (x y : IntBv w) :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    (bif b then x else y).toBitVec = bif b then x.toBitVec else y.toBitVec := by
+  cases b <;> rfl
+
+@[bv_normalize]
+theorem IntBv.poison_bif {w} (b : Bool) (x y : IntBv w) :
```
tobiasgrosser:
```suggestion
theorem IntBv.poison_bif {w : Nat} (b : Bool) (x y : IntBv w) :
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    (bif b then x else y).poison = bif b then x.poison else y.poison := by
+  cases b <;> rfl
+
+/- We enable `bv_decide` to normalize `toIntBv` for values and poison. -/
+attribute [bv_normalize] toIntBv_poison
+attribute [bv_normalize] toIntBv_val
+attribute [bv_normalize] toBitVec_zero_of_poison
```
tobiasgrosser: Do we need this if we explicitly call `llvm_toBitVec`?

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+
+@[bv_normalize]
+theorem poison_ite {w} (b : Prop) [Decidable b] (x y : IntBv w) :
+    (if b then x else y).poison = if b then x.poison else y.poison := by
+  split <;> rfl
+
+@[llvm_toBitVec]
+theorem ite_poison_eq (x : Int w) :
+  ((if x.toIntBv.poison = true then
+    {toBitVec := 0#w, poison := true, inv := by simp}
```
tobiasgrosser:
```suggestion
    {toBitVec := 0#w, poison := true}
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+  else { toBitVec := x.toIntBv.toBitVec, poison := false, inv := by simp }) : IntBv w )=
```
tobiasgrosser:
```suggestion
  else { toBitVec := x.toIntBv.toBitVec, poison := false}) : IntBv w )=
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+
+@[llvm_toBitVec]
+theorem toIntBv_constant {w : Nat} (v : _root_.Int) :
+    (constant w v).toIntBv = ⟨BitVec.ofInt w v, false, by simp⟩ := by
+  simp [constant, toIntBv]
+
+@[llvm_toBitVec]
+theorem toIntBv_add {w : Nat} (x y : Int w) :
+    (add x y).toIntBv =
+      if (x.toIntBv.poison ∨ y.toIntBv.poison) = true then ⟨0#w, true, by simp⟩
```
tobiasgrosser:
```suggestion
      if (x.toIntBv.poison ∨ y.toIntBv.poison) = true then {toBitVec := 0#w, poison := true}
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+        else ⟨x.toIntBv.toBitVec + y.toIntBv.toBitVec, false, by simp⟩ := by
```
tobiasgrosser:
```suggestion
        else {toBitVec := x.toIntBv.toBitVec + y.toIntBv.toBitVec, poison := false} := by
```

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+    (val v).toIntBv = ⟨v, false, by simp⟩ := by
+  simp [toIntBv]
+
+theorem toBitVec_zero_of_poison (x : IntBv w) :
+    x.poison = true → x.toBitVec = 0#w := by
+  obtain ⟨bv, poison, h⟩ := x
+  exact h
+
+@[bv_normalize]
+theorem eq_iff {w : Nat} {l r : IntBv w} :
```
tobiasgrosser: You use a mix of `l r` and `x y`. Should we stay with `x y`?

tobiasgrosser (commented):
Thank you for pushing this ahead. I like the idea of going through a struct.

I feel these are exactly the kind of experiments we would want to have. See also https://github.com/opencompl/veir/pull/178/changes, one of my earlier attempts.

I left a couple of style comments. I may still want to play a bit more with the proofs.

Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -1,56 +1,133 @@
...
+  the `poison` field indicates whether the `Int` is poison. -/
+def toIntBv (x : Int w) : IntBv w :=
+  match h : x with
+  | .val v => ⟨v, false, by simp⟩
+  | .poison => ⟨0#w, true, by simp⟩
+
+/--
+  We prove the injectivity of `toIntBv`.
+-/
+theorem Int.toIntBv.inj {w : Nat} : ∀ {x y : Int w}, x.toIntBv = y.toIntBv → x = y
```
tobiasgrosser: Is there a reason you do not use 
```suggestion
theorem Int.toIntBv.inj {w : Nat} {x y : Int w} (h : x.toIntBv = y.toIntBv) : x = y
```
I saw @math-fehr also use implications, but its not clear to me why this is a good idea over just using normal arguments. Did I miss sth?

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: restored in `Lemmas`, moved the bitblasting infra to `Bitblast`

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: addressed

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: elided

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: addressed

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: addressed

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: addressed

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: I think so, because these lemmas are useful for bv_decide to figure out to handle `toIntBv`

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: done

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: replaced with `x` and `y` only

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: no, replaced with explicit hypothesis.

Veir/Data/LLVM/Int/Lemmas.lean
luisacicolini: double checked and this was not necessary. cleaned

luisacicolini:
thank you @tobiasgrosser for the review. I have addressed all the comments, and will next experiment with alternatives that only require one `bv_decide` call.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,715 @@
...
+  match h : x with
+  | .val v => {toBitVec := v, poison := false}
+  | .poison => {toBitVec := 0#w, poison := true}
+
+/-- Return a boolean if the LLVM.int `x` is poison. -/
+def isPoison {w : Nat} (x : Int w) : Bool := x.toIntBv.poison
+
+/-- Return a concrete bitvector value given an LLVM.Int. -/
+def getValue {w : Nat} (x : Int w) : BitVec w := x.toIntBv.toBitVec
+
```
alexkeizer, line 39: Do you need `IntBv`? I would expect you can just define `isPoison` and `getValue` directly in terms of the inductive (i.e., `isPoison` is defined the same as `Option.isNone` and `getValue` the same as `Option.getD _ 0`)

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,715 @@
...
+  exact or_assoc
+
+@[llvm_toBitVec]
+theorem getValue_sub {w : Nat} (x y : Int w) {nsw nuw : Bool} :
+    (sub x y nsw nuw).getValue =
+      if x.isPoison ∨ y.isPoison then 0#w
+        else if nsw ∧ BitVec.ssubOverflow x.getValue y.getValue then 0#w
+          else if nuw ∧ BitVec.usubOverflow x.getValue y.getValue then 0#w
+            else x.getValue - y.getValue := by
+  simp [getValue, llvm_toBitVec]
```
alexkeizer, line 238: Continueing my earlier comment: the statements here are much larger than the proofs, so I feel like the potential sharing that `toIntBv` might give might not be worth it.

alexkeizer, line 238: I would expect this rewrite to be conditional on the fact that the result in non-poison (see the Zulip thread), to result in a much cleaner rhs

alexkeizer (commented):
There seems to be a _lot_ of boilerplate here, so I just skimmed it. Perhaps pulling out an initial PR that does just, say, addition and subtraction might help to have a more focussed discussion on the basic setup

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: In general, the idea I am pursuing for the API is to have a bitblastable definition for every inductive one (e.g., the same thing happens for the refinement def), proving their equivalence and attributing the equivalence to `bv_normalize`, and building the bitblastable-friendly API around this bitblastable definition. 

For this specific definition, in my mind `IntBv` should act as a middle point to separate as much as possible the inductive-related lemmas and the structure-related ones (based on #178 and on `fp-lean`). To me, having an `IntBv` and methodically adding to the correct simp-set the lemmas that all refer to the same, bitblastable structure seemed more natural. In particular, I felt this would make the definitions and API easier to manipulate - also due to my limited knowledge of Lean: for example, I am not familiar with `Option` and such, and find it hard to reason about them, which is why I did not use them.

With that said, I can work on a parallel PR that uses the `Option` types, and maybe we can then compare the approaches?

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: In my mind, these were also very mechanical unfoldings and lemmas that only serve as building blocks (and are never to be touched). I was trying to do with these lemmas the same thing that happens in other parts of Veir, where we have lemmas that have complex statements and live in the right grind-sets, making proofs more automated. I used the same if conditions (and not `isPoison`, as we could have done as well) to simplify the proofs and make their structure very mechanical - also, I could not make it work the other way around (e.g., using `isPoison` in the def of `toIntBv_sub` the proof was significantly more complex. But also, it might well be that I'm missing some intermediate lemma to connect the dots - if you have any advice or intuition in this direction it would be super helpful).

luisacicolini: Yes, I see this point - I unfortunately could not make it work without ugly proof, but it might well be that I'm missing something here.

luisacicolini:
Thank you @alexkeizer for the review! I answered the comments. I let this PR grow big because I wanted to see how many lemmas can interact and what do we need to make this interaction sustainable (which is also why there is a lot of boilerplate, in my experience building a lot of theory around the objects we manipulate has been useful), but I can pull out a smaller one if we want to have a more specific discussion, e.g. on the structure or on the shape of the lemmas.

Veir/Data/Refinement.lean
```diff
@@ -15,10 +15,12 @@ public section
   In particular, any concrete `i'` refines a poison `i`, but a poison `i'` does *not* refine
   any `i`.
 -/
-def isRefinedBy (i i' : Veir.Data.LLVM.Int 64) : Prop :=
+def isRefinedBy {w : Nat} (i i' : Veir.Data.LLVM.Int w) :=
```
tobiasgrosser: This can probably be pulled out into a separate PR.

luisacicolini: done in #485

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,715 @@
...
+  We define a bitblastable structure `IntBv`, where the `toBitVec` fields contains a bitvector,
+  the `poison` field indicates whether the value is poison,
+  and `inv` preserves a proof that if the value is poison its corresponding bitvector
+  value is `0#w`.
+-/
+@[ext]
+structure IntBv (w : Nat) where
+  toBitVec : BitVec w
+  poison : Bool
+  inv : poison → (toBitVec = 0#w) := by simp
```
georgerennie, line 24: Maybe it is discussed elsewhere, but what is the reasoning for requiring poisoned bitvecs to have a 0 value? Is it so that equality works out of the box?

I'm not 100% certain but it strikes me that this somewhat forces you to tie the data and poison production circuits together directly, which potentially hinders the optimizations `bv_normalize` can do directly on the data component. If you allowed any value when poisoned, the data circuit would generally not need to reference the poison circuit so could potentially be better optimized (I can try to create a concrete example if it helps).

georgerennie: quick example [here](https://live.lean-lang.org/#codez=JYWwDg9gTgLgBAZRgEwHQBUCGBjGxuoBCAagCICm2wy5AULQM4xQCuuLU5cAkgHYyEAbnAAUAdzgAuOADlMMAJRwxAC3KdacOACNh0wsBjFKyzXEjAGEXlLiEIEADZmVti1ZuAkwh3CAvHAAGekFMKGBMbUcuAG8JaTkYAF96AAEAbQZQMABdWhoAMzhMZGRRAA84AE9bPgFhMSVpWqFlKV8zaLMtXTa4YEKy1HdrOEAKIiqhiEsRmDUbAIBiCXJHBi5BnoBqCd0u8ymPXsHhm3HKyenePddJf21qgHMw3mQzZNp0zPBcgrgGFm05SqNX4LQaILqrVuHT2PVufQGF0OZyRMzmgSWcBWazgG2EAFodoI9icjqjThMTtdevc4E9gC83qkMlkfuRCthrExRDZ9IZjNhlI0eKD6m0YVpunp%2FFdJftLr18phVnQ5Tc7o9nq8tO9Pqy8uy%2BgwAErshnkZCEaoiTA6CFg4UABSgEDA4q0mHJYzMIgANdovYByIiKqB6d1DggU9EYzDYMA4XGaggA5KI4rJ5EpVOpVT5bAYjCYxGZSfoHM4PizvgbCsVkKmRBVqk1RanwS26m33XBOnK4f48XBtuddnLSwOvSiqTrmV8cjW%2FgCG037Smhauu9CtL3Jf3cRG4ISR8SxwcRvDjmeKedp3BdVX579ObwmA3eXZ%2BUXhUnN%2B1t7DpTgWVJXHOAlRVJlKznNlCksU18nNS1KgbW1AQ7IQ2ydF03S3ENSVGH1%2FSDEMwx0CMo1ofFCXQNRoHIEA4E4f5HDwXgHjAl0GM4ABHFhgGediTgtOAQkcFhyAYOAYAgHQuCCWZyDohjdAAfQAQRKBAARUgBGABWABmABOKQ9kAACJRC4chdNXOAADYABYFAAGkXQFrJ00R%2FkBERn25AIlHIJQAB8AD44LNXgLStAAeYKihKLEbK4eF7jMOc4GsRxqjSOtXO81y%2FJgVyIoQqKkNyKUVN4aAQGVYAAC9cysKAoEqegqLgGjFM4LiJJYFiGXY%2FJOLgTlwBYVj2IUuBkHkW0xEMFQIEmsauWodQhvlDxaAUpSfHUzTtP04zU0kczLKSzz0LXRyXLc1MPK8pdRCK1MAqxELwpNSLouQuKEvrK6sRpdqtAyrKcrrZN8qXQquRgGGjXgxCrWTSqDpqqA6scRryCAA) based on `bv_AddSub_1539` (`a - (0 - b)) = a + b`). With the definition that requires zeroed poison values the logic for that prevents `bv_normalize` optimizing, whereas when you don't, `bv_normalize` can solve the goal.

luisacicolini: Thank you for taking a look and for the example, it was very useful and I played with it a bit to see if we could use it in #490 if you would like to take a look! I found it a bit more complex to define the relations between the `LLVM.Int` and the `IntBv`, but it is definitely interesting - would be great if you have time to take a look there!

alexkeizer: So this goes back to my other comment, you probably should only talk about `getValue` when there is already a proof that it is not `poison`. If you have that extra hypothesis, then whatever default value you pick should no longer matter.

luisacicolini: Ah I see, I misunderstood, I’m sorry! If you want to, feel free to look at #490 and let me know if that is the direction you meant

Veir/Data/LLVM/Int/Bitblast.lean
alexkeizer: AH, I should clarify: I'm not suggesting you actually use `Option`. I merely mentioned `Option` since the current inductive definition ought to be Option-shaped, and thus looking at the analogous Option definitions might show you how to define `isPoison` and `getValue` directly on the Int structure.

To be precise, that is like so:
```lean
def isPoison : Int w -> Bool
  | .val _ => false
  | .poison => true

def getValue : Int w -> BitVec w
  | .val x => x
  | .poison => 0#0
``` 

Then there is no need to have a strictly bitblastable "definition" per se, rather you can prove that `x = y` iff `x.isPoison = y.isPoison /\ (!x.isPoison -> !y.isPoison -> x.getValue = y.getValue)`, whence making a particular operation bitblastable then means to have the lemmas you already have about `getValue` and `isPoison` for that operation.

There is no need to explicitly materialize the structure.

Veir/Data/LLVM/Int/Bitblast.lean
alexkeizer: Sure, what I'm suggesting here is that you can drop `toIntBv_sub` and just prove `isPoison_sub` and `getValue_sub` directly

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: Ah, I see, thank you for clarifying! I got confused because of the `Option`, but it is much clearer now!

PR #458, opened 2026-04-24
[do not merge] fastNTT instructions

luisacicolini:
We investigate what portion of LLVM IR and RISCV ISA we need to parse and interpret to lower the FastNTT algorithm. 
Based on the [`heir`](https://github.com/google/heir/blob/1210ad37dc9531d6e60d8ddbce81dbd79f7626a6/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp#L1060) pseudocode, we include: 
* `fastntt.c`: the C implementation of the algorithm 
* `fastntt.ll`: the LLVM IR lowering, obtained via `clang -S -emit-llvm fastntt.c -o fastntt.ll`
* `fastntt_inlined.ll`: the inlined LLVM IR lowering, obtained via `opt -passes='always-inline,mem2reg' fastntt.ll -S -o fastntt_inlined.ll`
* `fastntt.asm`: the ASM lowered with [clang](https://godbolt.org/z/zdKrcheMj)
* `fastntt.mlir`: the MLIR encoding in generic syntax, obtained via `mlir-translate --import-llvm fastntt_inlined.ll --mlir-print-op-generic --mlir-print-local-scope -o fastntt.mlir`

it is not a particularly smart implementation, but should give us an idea of what's missing (see #459)


tobiasgrosser:
> fastntt_inlined.ll: the inlined LLVM IR lowering, obtained via opt -passes='always-inline' fastntt.ll -S -o fastntt_inlined.ll

Can you also run `-mem2reg`? Otherwise, will we have way too many `alloca`s. I wonder if we need any allocas in the end.

tobiasgrosser:
Also, can this file already be parsed or is sth still missing. If we still miss syntax, can you extract a small self-contained test case and create an issue?

luisacicolini:
added `mem2reg` and generic syntax printing. Will extract an MWE by the end of the week

tobiasgrosser:
@luisacicolini, can you use `--mlir-print-local-scope` to inline `#loop_annotation` and rewrite the example to use 64-bit integers (which is the only bitwidth supported by the instruction selector so far). AFAIU, this file should then already compile with VEIR and maybe even instruction-select. 🥳

tobiasgrosser:
Rather than having this as `do-not-merge` can we clean this up and add this as a slightly-larger-test to the `instruction selection folder`?

tobiasgrosser:
This should be meanwhile part of our eval repository. Let's close this PR for now.

PR #460, opened 2026-04-24
(cuda-tile) Added cuda_tile pointer type

hhkit:
Scaffolding work to support CUDA Tile IR, a block-level DSL for interfacing with Nvidia GPUs.

This is just the first of four types, implemented with reference to:
https://github.com/NVIDIA/cuda-tile/blob/0859212ad19f71133a9b940c05323286cbf28a05/test/Dialect/CudaTile/types.mlir

Interpreter support not included.

Tech debt is that pointeeType should be any tensor element (inclusive of f16/f32), rather than just Integer. I assume this is fine for now. If this implementation is fine, I'm happy to oneshot the remaining three types.


Veir/IR/Attribute.lean
```diff
@@ -140,6 +153,7 @@ structure DictionaryAttr where
   /- TODO: figure out how to maintain a proof of sorted-ness and uniqueness. -/
 deriving Inhabited, Repr, Hashable
 
+
```
tobiasgrosser: This is unrelated.

tobiasgrosser (approved):
This looks good. However, please polish the commit message describing what cuda-tile is, which parts this PR is adding, and which parts you leave for later. Instead of writing "The main issue is", just explain that some parts are not generalized as of yet. That is fine. Also state that interpreter support is left for later.

Let's wait for a 2nd reviewer.

Veir/IR/Attribute.lean
```diff
@@ -107,6 +107,19 @@ structure ModArithType where
   modulusType : Option IntegerType
 deriving Inhabited, Repr, DecidableEq, Hashable
 
+/-!
+  # cuda-tile types go here
```
math-fehr:
```suggestion
  # Cuda Tile types
```

Veir/IR/Attribute.lean
```diff
@@ -170,6 +184,8 @@ inductive Attribute
 | unregisteredAttr (attr : UnregisteredAttr)
 /-- HEIR modarith type -/
 | modArithType (type : ModArithType)
+/-- cuda-tile pointer type -/
```
math-fehr:
```suggestion
/-- Cuda Tile pointer type -/
```

Veir/IR/Attribute.lean
```diff
@@ -107,6 +107,19 @@ structure ModArithType where
...
+/-!
+  # cuda-tile types go here
+-/
+
+/--
+  An elemental pointer type represents a single location in global device memory.
+  Pointers are typed, i.e., they carry the type they point to.
+-/
+
+structure PointerType where
```
math-fehr: Can you rename this to `CudaTilePointerType`, or maybe even `CudaTile.PointerType`?
Because once we will have `!ptr.ptr`, there will be a name clash here.

math-fehr (commented):
Just added 1-2 comments, but otherwise looks good, thanks!

luisacicolini (approved):
lgtm, thank you!

math-fehr (approved):
Thanks!

PR #463, opened 2026-04-27
feat: new packed float datatype

bollu:
This PR adds a new `PackedFloat e s` datatype that represents an IEEE
formatted floating point number with `e` bits of exponent and `s` bits
of significand.

### DIFF
--- a/Veir/Data/FP/PackedFloat/Basic.lean
+++ b/Veir/Data/FP/PackedFloat/Basic.lean
@@ -7,19 +7,21 @@ namespace PackedFloat
 public section
 
 /--
-A packed floating point number,
+A packed floating point number of radix 2,
 with exponent and significand widths encoded at the type level.
-This implements the IEEE-754 floating point specification [1] of the bit layout.
-[1] Muller, Jean-Michel, et al. Handbook of floating-point arithmetic.
+This definition implements Section 3.4, Figure 3.1 of the IEEE-754 standard [1].
+We recommend [2] for a comprehensive introduction.
+[1] IEEE 754-2019, https://standards.ieee.org/ieee/754/6210/
+[2] Muller, Jean-Michel, et al. Handbook of floating-point arithmetic.
     Vol. 1. Basel, Switzerland:: Birkhäuser, 2018.
-
 -/
 @[ext]
 structure PackedFloat (exWidth sigWidth : Nat) where
     /-- Sign bit. -/
     sign : Bool
-    /-- Exponent of the packed float. -/
+    /-- Biased exponent of the packed float. -/
     ex : BitVec exWidth
-    /-- Significand (mantissa) of the packed float. -/
+    /-- Trailing significand (mantissa) of the packed float.
+      Leading digit implicitly encoded by biased exponent. -/
     sig : BitVec sigWidth
 deriving DecidableEq, Repr, Inhabited
--- a/Veir/Data/FP/PackedFloat/Basic.lean
+++ b/Veir/Data/FP/PackedFloat/Basic.lean
@@ -1,0 +1,21 @@
+module
+
+namespace Veir.Data.FP
+
+namespace PackedFloat
+
+public section
+
+/--
+A packed floating point number,
+with exponent and significand widths encoded at the type level.
+-/
+@[ext]
+structure PackedFloat (exWidth sigWidth : Nat) where
+    /-- Sign bit. -/
+    sign : Bool
+    /-- Exponent of the packed float. -/
+    ex : BitVec exWidth
+    /-- Significand (mantissa) of the packed float. -/
+    sig : BitVec sigWidth
+deriving DecidableEq, Repr, Inhabited
--- a/Veir/Data/FP/PackedFloat/Basic.lean
+++ b/Veir/Data/FP/PackedFloat/Basic.lean
@@ -9,6 +9,10 @@ public section
 /--
 A packed floating point number,
 with exponent and significand widths encoded at the type level.
+This implements the IEEE-754 floating point specification [1] of the bit layout.
+[1] Muller, Jean-Michel, et al. Handbook of floating-point arithmetic.
+    Vol. 1. Basel, Switzerland:: Birkhäuser, 2018.
+
 -/
 @[ext]
 structure PackedFloat (exWidth sigWidth : Nat) where
--- a/Veir/Data/FP/PackedFloat/Basic.lean
+++ b/Veir/Data/FP/PackedFloat/Basic.lean
@@ -7,7 +7,7 @@ namespace PackedFloat
 public section
 
 /--
-A packed floating point number of radix 2,
+A packed floating point number of radix 2 following IEEE 754 [1],
 with exponent and significand widths encoded at the type level.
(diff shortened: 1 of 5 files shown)


Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,21 @@
...
+
+namespace PackedFloat
+
+public section
+
+/--
+A packed floating point number,
+with exponent and significand widths encoded at the type level.
+-/
+@[ext]
```
tobiasgrosser, line 18: Out of interest, why is this `ext`? Just that I learn why one typically uses such annotation.

Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Veir.Data.FP
+
+namespace PackedFloat
+
+public section
+
+/--
+A packed floating point number,
+with exponent and significand widths encoded at the type level.
```
tobiasgrosser, line 11: You probably want to flesh out the comment here a bit. In particular, it probably makes sense to reference the corresponding IEEE standard and state clearly that this `PackedFloat` matches the bitencoding of IEEE floating point numbers. You probably also want to add some information, how this encoding behaves precisely.

Very cool.

tobiasgrosser (commented):
Very nice. What a fast turnaround!

I would like to have another look at this PR after you added the relevant documentation. It would also be great if @luisacicolini could take a look.

Veir/Data/FP/PackedFloat/Basic.lean
bollu: Integrates with the `ext` tactic, to allow one to write `ext` to show that two packed floats are equal when their component-wise values are equal. This is exactly how one wants to prove packed float equality.

Veir/Data/FP/PackedFloat/Basic.lean
bollu: Added a link, but the precise meaning is too verbose, due to the various nan, infinity, subnormal, normal cases. Happy to explain the various intricasies as we pull in more theory to implement the conversions and rounding.

Veir/Data/FP/PackedFloat/Basic.lean
tobiasgrosser: Perfect. I got this now.

Veir/Data/FP/PackedFloat/Basic.lean
tobiasgrosser: Thank you for the reference. We certainly want to cite `IEEE 754-2019` as available at https://standards.ieee.org/ieee/754/6210/.

We then want to follow IEEE-754 very carefully and continue to add both references to relevant subsections as well as relevant theory into this documentation. When finished, this file should offer the key specs from IEEE-754 inline in the comments plus references to all relevant subsections in the standard for complete details.

I agree we do not want to add all details now. However, this PR should establish a first skeleton of documentation, which we can add to over time. As we upstream further PRs, we can then continue to add documentation. The PR that establishes `isNorm` can then add information for subnormals.

bollu: Done

Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,27 @@
...
+namespace Veir.Data.FP
+
+namespace PackedFloat
+
+public section
+
+/--
+A packed floating point number of radix 2,
+with exponent and significand widths encoded at the type level.
+This definition implements Section 3.4, Figure 3.1 of the IEEE-754 standard [1].
```
tobiasgrosser:
```suggestion
This definition implements Section 3.4, Figure 3.1 of the IEEE-754-2009 standard [1].
```

Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,27 @@
...
+    Vol. 1. Basel, Switzerland:: Birkhäuser, 2018.
+-/
+@[ext]
+structure PackedFloat (exWidth sigWidth : Nat) where
+    /-- Sign bit. -/
+    sign : Bool
+    /-- Biased exponent of the packed float. -/
+    ex : BitVec exWidth
+    /-- Trailing significand (mantissa) of the packed float.
+      Leading digit implicitly encoded by biased exponent. -/
```
luisacicolini:
```suggestion
     The leading bit of the significand is implicitly encoded in the biased exponent. -/
```
nit

luisacicolini (approved):
thank you for the PR, LGTM!

Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,27 @@
+module
+
+namespace Veir.Data.FP
+
+namespace PackedFloat
+
+public section
+
+/--
+A packed floating point number of radix 2,
```
tobiasgrosser:
```suggestion
A packed floating point number of radix 2 following IEEE 754 [1],
```

Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,27 @@
...
+with exponent and significand widths encoded at the type level.
+This definition implements Section 3.4, Figure 3.1 of the IEEE-754 standard [1].
+We recommend [2] for a comprehensive introduction.
+[1] IEEE 754-2019, https://standards.ieee.org/ieee/754/6210/
+[2] Muller, Jean-Michel, et al. Handbook of floating-point arithmetic.
+    Vol. 1. Basel, Switzerland:: Birkhäuser, 2018.
+-/
+@[ext]
+structure PackedFloat (exWidth sigWidth : Nat) where
+    /-- Sign bit. -/
```
tobiasgrosser:
```suggestion
    /-- Sign bit -/
```

tobiasgrosser: Remove `.` as this is not a sentence, so no terminator punctuation is required.

Veir/Data/FP/PackedFloat/Basic.lean
```diff
@@ -0,0 +1,27 @@
...
+    sign : Bool
+    /-- Biased exponent of the packed float. -/
```
tobiasgrosser:
```suggestion
    /-- Biased exponent -/
```

`of the packed float` seems redundant. Also, drop the `.` as this is not a complete sentence (it has no noun).

Veir/Data/FP/PackedFloat/Basic.lean
tobiasgrosser: `by` -> `in the`

tobiasgrosser: `Leading digit` -> `The leading bit of the significand is implicitly encoded in the biased exponent.

tobiasgrosser (approved):
Nice. I added a couple of nits, but am otherwise happy to see this merged. Thank you for the quick turnaround.

bollu:
suggestions applied

PR #465, opened 2026-04-27
Factor out handling of unit attrs

nchappe:
This will make it easier to add LLVM memory operations, as they have quite a few unit attributes.

### DIFF
--- a/Veir/Properties.lean
+++ b/Veir/Properties.lean
@@ -10,6 +10,13 @@ namespace Veir
 
 public section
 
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
Nice. This is a beautiful refactor -- clean and simple. This is ready from my side. Let's see if we can get one more review.

ineol (approved):

math-fehr (approved):

PR #466, opened 2026-04-27
comb: Add CIRCT's comb dialect

georgerennie:
This adds the basic definitions of operations from CIRCT's comb dialect, providing only the basic syntactic definitions/verifier, not semantic definitions. The `comb.truth_table` operation is currently missing as it requires `DenseBoolArrayAttr` which is not implemented in VEIR yet.

This also adds a basic filecheck test that just parses/prints each of the added ops. The variadic ops are repeated for varying numbers of arguments and `comb.icmp` instantiated for each valid `predicate` value.

### DIFF
--- a/Test/Comb/identity.mlir
+++ b/Test/Comb/identity.mlir
@@ -1,0 +1,109 @@
+// RUN: veir-opt %s | filecheck %s
+
+"builtin.module"() ({
+^4():
+  %a = "arith.constant"() <{ "value" = 13 : i32 }> : () -> i32
+  %b = "arith.constant"() <{ "value" = 3 : i32 }> : () -> i32
+  %c = "arith.constant"() <{ "value" = 42 : i32 }> : () -> i32
+  %true = "arith.constant"() <{ "value" = 1 : i1 }> : () -> i1
+  %add1 = "comb.add"(%a) : (i32) -> i32
+  %add2 = "comb.add"(%a, %b) : (i32, i32) -> i32
+  %add3 = "comb.add"(%a, %b, %c) : (i32, i32, i32) -> i32
+  %and1 = "comb.and"(%a) : (i32) -> i32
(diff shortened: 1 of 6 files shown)


tobiasgrosser:
@georgerennie, thank you for this. This PR looks great. It clearly needs some tests. Otherwise, please note in the PR message, that this is syntax-only, but no semantics. I would skip the truth_table operation in this PR and just note that it still needs to be added. Adding it in a separate PR will be easy.

osmanyasar05:
Hey, I was just about to create a very similar PR for the Comb dialect as well and saw yours haha. Happy to work together on the semantics though:) I was working on it to get a sense of how things would be implemented for the Datapath dialect.

georgerennie:
> Hey, I was just about to create a very similar PR for the Comb dialect as well and saw yours haha. Happy to work together on the semantics though:) I was working on it to get a sense of how things would be implemented for the Datapath dialect.

Perfect, am happy to talk this through, I have not worked out how semantics should be done yet!

> @georgerennie, thank you for this. This PR looks great. It clearly needs some tests. Otherwise, please note in the PR message, that this is syntax-only, but no semantics. I would skip the truth_table operation in this PR and just note that it still needs to be added. Adding it in a separate PR will be easy.

Have done so
- `truth_table` removed for now
- `Test/Comb/identity.mlir` that checks the added ops all roundtrip through the parser/printer correctly
- Have updated the PR message to reflect the actual changes

math-fehr (approved):
Nice, thanks!

tobiasgrosser (approved):
LGTM.

luisacicolini:
btw, @georgerennie @osmanyasar05 lean-MLIR had already some [comb semantics](https://github.com/opencompl/lean-mlir/blob/main/SSA/Projects/CIRCT/Comb/Semantics.lean) - not sure they're the best semantics we could have but it might be useful as a starting point

PR #467, opened 2026-04-28
Add GetSet lemmas for `Rewriter.createOp`

math-fehr:
Also, add a few lemmas that were missing from the file that were needed to write these get-set lemmas.


tobiasgrosser (approved):
Nice.

luisacicolini (approved):

PR #468, opened 2026-04-29
Add support for `getSuccessor` in the rewriter get-set lemmas

math-fehr:
While `getSuccessor op ctx index` is just a wrapper around `((BlockOperandPtr.mk op index).get! ctx).value`, I think it is still worth to provide get-set lemmas for it, especially since we will need `getSuccessors` lemmas (that are quite needed from what I've seen).


tobiasgrosser (approved):
LGTM.

luisacicolini (approved):

PR #472, opened 2026-04-29
Add get/set lemmas for `WfRewriter.createOp`

math-fehr:
The getters we consider are slightly different than the ones we previously had, because there are a few getters that do not make any sense to manipulate directly at that level of abstraction (like first uses).


luisacicolini (approved):
reviewed mostly for style, lgtm! and these lemma look super useful!!!

PR #473, opened 2026-04-29
feat: add unpacked floats

bollu:
This PR adds support for unpacked floating point numbers,
which represent a nonzero rational number as two bitvectors,
the exponent and the significand.

### DIFF
--- a/Veir/Data/FP/Sign.lean
+++ b/Veir/Data/FP/Sign.lean
@@ -4,7 +4,7 @@ namespace Veir.Data.FP
 
 /--
 Treat a boolean as a sign,
-where `true` corresponds to -1 and `false` corresponds to `1`.
+where `true` corresponds to `-1` and `false` corresponds to `1`.
 -/
 public def signToInt (s : Bool) : Rat :=
   if s then -1 else 1
--- a/Veir/Data/FP/Sign.lean
+++ b/Veir/Data/FP/Sign.lean
@@ -1,0 +1,10 @@
+module
+
+namespace Veir.Data.FP
+
+/--
+Treat a boolean as a sign,
+where `true` corresponds to -1 and `false` corresponds to `1`.
+-/
+public def signToInt (s : Bool) : Rat :=
+  if s then -1 else 1
--- a/Veir/Data/FP/UnpackedFloat/Basic.lean
+++ b/Veir/Data/FP/UnpackedFloat/Basic.lean
@@ -42,7 +42,7 @@ def sigToRat (s : BitVec s) : Rat :=
 
 /--
 Treat an exponent bitvector as a rational number,
-where the exponent is interpreted as an integer.
+power of two where the exponent is the bitvector interpreted as an integer.
 -/
 def exToRat (e : BitVec e) : Rat :=
   2 ^ e.toInt
--- a/Veir/Data/FP/UnpackedFloat/Basic.lean
+++ b/Veir/Data/FP/UnpackedFloat/Basic.lean
@@ -1,0 +1,58 @@
+module
+public import Veir.Data.FP.Sign
+
+namespace Veir.Data.FP
+
+namespace UnpackedFloat
+
+public section
+
+/--
+`UnpackedFloat e s` is the *working* (unpacked) representation of a floating-point
+number with `e` bits for the exponent and `s` bits for the significand.
+
+The unpacked float `uf` is interpreted as the rational number
+given by the formula:
+
+```
+(-1)^uf.sign * 2^(uf.ex.toInt) * (1 + uf.sig.toNat / 2^s)
+```
+
+Crucially, this means that we can only represent *non-zero*
+numbers in the unpacked format,
+since the  implicit leading bit of the significand is always 1,
+so the significand is interpreted as a number in [1, 2).
+-/
+@[ext]
+structure UnpackedFloat (e s : Nat) where
+  /-- Sign of the floating-point number. -/
+  sign : Bool
+  /-- Unbiased exponent, interpreted as an integer. -/
+  ex : BitVec e
+  /-- Significand, interpreted as a number in [1, 2).-/
+  sig : BitVec s
+  deriving Inhabited, Repr
+
+/--
+Treat a significand bitvector as a rational number in [1, 2),
+where the leading bit is implicitly 1 and the trailing bits are interpreted as a fraction.
+-/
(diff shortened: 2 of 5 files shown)


bollu:
@luisacicolini lmk if the explanation of what this structure is trying to do makes sense!

bollu:
There's a design decision, where in theory, this could really just be called `NonzeroRatBV` or some such, since this data structure really just represents a nonzero rational number as a bitblastable pair of bitvectors.

Veir/Data/FP/UnpackedFloat/Basic.lean
```diff
@@ -0,0 +1,58 @@
...
+/--
+Treat a significand bitvector as a rational number in [1, 2),
+where the leading bit is implicitly 1 and the trailing bits are interpreted as a fraction.
+-/
+def sigToRat (s : BitVec s) : Rat :=
+  1 + s.toNat / (2 ^ s.toNat)
+
+/--
+Treat an exponent bitvector as a rational number,
+where the exponent is interpreted as an integer.
```
luisacicolini:
```suggestion
power of two where the exponent is the bitvector interpreted as an integer.
```

luisacicolini:
It makes sense to me. One thing I am wondering is whether we could have some testing and/or examples? Either in the comments or with some `#check` objects, to show how the definitions actually behave? I'm not sure where they should live, maybe even in `UnitTest`, but it would be useful to have them imho

luisacicolini:
I also find `NonzeroRatBV` more intuitive, as it explains more clearly what object we're dealing with, but let's hear others' opinions on it.

tobiasgrosser:
> I also find `NonzeroRatBV` more intuitive, as it explains more clearly what object we're dealing with, but let's hear others' opinions on it.

Why is it `non-zero`. I like the `RatBv` naming, though! On the other hand, I would expect `RatBv` to be a pair of numerand and divisor. Maybe this exponent-mantissa representation could have a different name?

bollu:
@tobiasgrosser I felt that having the hidden bit implicit actually makes it slightly cleaner. This means that we can only represent nonzero numbers this way, because zero numbers create degeneracy: $$0 \times 2^1 = 0 \times 2^2 = 0 \times 2^3 = \dots$$.

 This deviates a little bit from `fp-lean`, but this feels like something we learnt from our experience, and we should learn from the `fp-lean` experience.

bollu:
> It makes sense to me. One thing I am wondering is whether we could have some testing and/or examples? Either in the comments or with some `#check` objects, to show how the definitions actually behave? I'm not sure where they should live, maybe even in `UnitTest`, but it would be useful to have them imho

I agree! I was planning to add examples in the next PR, where I will provide a function that unpacks packed floats. This way, we can test the pipeline :) Without the pipeline, it's a little hard to interpret what the floats are saying, has been my experience 😅

tobiasgrosser:
> @tobiasgrosser I felt that having the hidden bit implicit actually makes it slightly cleaner. This means that we can only represent nonzero numbers this way, because zero numbers create degeneracy: 0 × 2 1 = 0 × 2 2 = 0 × 2 3 = … .
> 
> This deviates a little bit from `fp-lean`, but this feels like something we learnt from our experience, and we should learn from the `fp-lean` experience.

I find it very surprising to have a type that does not have a zero element. However, I am happy to go along with this type and building up theory without a zero. While the definition seems partial, it matches precisely what floating point numbers seem to be about.

> I also find NonzeroRatBV more intuitive, as it explains more clearly what object we're dealing with, but let's hear others' opinions on it.

>  I like the RatBv naming, though! On the other hand, I would expect RatBv to be a pair of numerand and divisor. Maybe this exponent-mantissa representation could have a different name?

I am wondering if there is name that describes the type rather than the process how we get to the type. I am OK with UnpackedFloat, but wanted to make sure you deliberately reflect on the name. Maybe brainstorm with others over lunch?

bollu:
> I am wondering if there is name that describes the type rather than the process how we get to the type. I am OK with UnpackedFloat, but wanted to make sure you deliberately reflect on the name. Maybe brainstorm with others over lunch?



Yes, I agree :) The type represents a number of the form $$2^e \times 1.b_0 b_1 \dots b_n$$, which is to say, a number in "binary scientific notation". The key point(s) are that: (1) enforcing the invariant of the form $$1.b_0 b_1 \dots$$ forces all the numbers that can be represented to be nonzero. Secondly, since the strings are finite, we can only represent a subset of the rational numbers (for example, $$1/3 = 0.01010101\dots$$ which in scientific notation would be $$2^{-2} \times 1.010101...$$.

tobiasgrosser:
> > I am wondering if there is name that describes the type rather than the process how we get to the type. I am OK with UnpackedFloat, but wanted to make sure you deliberately reflect on the name. Maybe brainstorm with others over lunch?
> 
> Yes, I agree :) The type represents a number of the form 2 e × 1. b 0 b 1 … b n , which is to say, a number in "binary scientific notation". The key point(s) are that: (1) enforcing the invariant of the form 1. b 0 b 1 … forces all the numbers that can be represented to be nonzero. Secondly, since the strings are finite, we can only represent a subset of the rational numbers (for example, 1 / 3 = 0.01010101 … which in scientific notation would be 2 − 2 × 1.010101 . . . .

What about `SciRat`?

bollu:
I guess the only deficiency of that name is that it doesn't point out that the definition is a BV definition that _represents_ a `Rat`. So, `ScientificNotationBV`? `BinaryScientific`? `ScientificBV`?

tobiasgrosser:
> I guess the only deficiency of that name is that it doesn't point out that the definition is a BV definition that _represents_ a `Rat`. So, `ScientificNotationBV`? `BinaryScientific`? `ScientificBV`?

I would also be happy with `ScientificBV`.

Veir/Data/FP/Sign.lean
```diff
@@ -0,0 +1,10 @@
+module
+
+namespace Veir.Data.FP
+
+/--
+Treat a boolean as a sign,
+where `true` corresponds to -1 and `false` corresponds to `1`.
```
tobiasgrosser:
```suggestion
where `true` corresponds to `-1` and `false` corresponds to `1`.
```

tobiasgrosser (approved):
OK, assuming Luisa is OK, lets merge this with either `UnpackedFloat` or `ScientificBV` as the name of this object?

PR #480, opened 2026-04-30
Parse pretty dialect types

nchappe:
Currently only dialect types of the form `!a.b<c>` are supported.

This adds supports for dialect types of the form `!a.b`.


math-fehr (approved):
Thanks!

PR #435, opened 2026-04-19
Use runtime checks for WellFormed preconditions in parser

GZGavinZhao:
This is the first step towards a `sorry`-less parser. The difficult part of making the parser `sorry`-less is to prove that the `IRContext` is `WellFormed` at all times, which normally requires carrying around the proof of the `InBounds` of blocks, operands, regions, insertion points, etc., at all times in the parser. While these proofs are possible to prove directly, doing this likely requires very complicated changes to the parser and/or the addition of many helper lemmas that definitely don't fit within a single PR.

Instead, the approach is to start with only runtime checks and migrate to direct proofs one step at a time. Right now, instead of carrying `InBounds` proofs with us all the time, we use runtime checks to construct proofs that things are `InBounds` only when we need them to prove `WellFormed` of `IRContext` (runtime checks are possible because all the `XXX.InBounds` are `Decidable`). Then, we can gradually swap out the runtime checks with direct proofs as we add more helper lemmas and as I understand how to utilize all the nice properties that `WfIRContext` provides. Note that the parser remains `sorry`-less throughout this entire process; the only difference is whether the `InBounds` proofs are proved directly or checked at runtime using `Decidable`.

The only thing remaining that's making the parser not `sorry`-less is the lack of some helper lemmas for proving `WellFormed` of `IRContext`, which @math-fehr is working on. This will also be addressed in future PRs. This PR is just to setup the runtime check infrastructure and make the parser use it.

I have benchmarked that runtime checks only makes the parser ~5% slower on super large MLIR files, which is within the acceptable range to me. The benchmark below compares between this PR and current HEAD (1b8c8ccb).
```
> hyperfine -w 2 "./old-veir-opt sqlite3-stripped.mlir" "./runtime-checks-veir-opt sqlite3-stripped.mlir"
Benchmark 1: ./old-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.209 s ±  0.020 s    [User: 2.186 s, System: 0.020 s]
  Range (min … max):    2.180 s …  2.247 s    10 runs

Benchmark 2: ./runtime-checks-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.351 s ±  0.015 s    [User: 2.331 s, System: 0.017 s]
  Range (min … max):    2.326 s …  2.371 s    10 runs

Summary
  ./old-veir-opt sqlite3-stripped.mlir ran
    1.06 ± 0.01 times faster than ./runtime-checks-veir-opt sqlite3-stripped.mlir
```

PR #481, opened 2026-04-30
Parser linearity fix and optimization

GZGavinZhao:
Increases parser performance by 9.9x in benchmark and likely by a magnitude from time complexity perspective. Original code by @ineol. 

Required by #435.

Benchmarked against [`sqlite-3-stripped.mlir`](https://github.com/opencompl/veir/blob/math-fehr/sqlite-test/sqlite3-stripped.mlir) from `math-fehr` branch. Old commit is https://github.com/opencompl/veir/commit/da74b917befb9688e6f959ee222f947c3eca92f0 and new commit is current PR.

```
> hyperfine "./old-veir-opt sqlite3-stripped.mlir" "./new-veir-opt sqlite3-stripped.mlir"
Benchmark 1: ./old-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):     22.361 s ±  0.138 s    [User: 22.213 s, System: 0.062 s]
  Range (min … max):   22.059 s … 22.501 s    10 runs

Benchmark 2: ./new-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.259 s ±  0.027 s    [User: 2.225 s, System: 0.028 s]
  Range (min … max):    2.222 s …  2.298 s    10 runs

Summary
  ./new-veir-opt sqlite3-stripped.mlir ran
    9.90 ± 0.13 times faster than ./old-veir-opt sqlite3-stripped.mlir
```

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -25,6 +25,7 @@ structure MlirParserState where
     (true), or has been forward declared (false).
   -/
   blocks : Std.HashMap ByteArray (BlockPtr × Bool)
+  deriving Inhabited
 
 def MlirParserState.fromContext (ctx : IRContext OpCode) : MlirParserState :=
   {ctx := ctx, values := Std.HashMap.emptyWithCapacity 128, blocks := Std.HashMap.emptyWithCapacity 1}
@@ -84,6 +85,35 @@ def registerValueDef (name : ByteArray) (value : ValuePtr) : MlirParserM Unit :=
 def setContext (ctx : IRContext OpCode) : MlirParserM Unit := do
   modify fun s => {s with ctx := ctx}
 
+/--
+  Modifies the current IR context.
+-/
+def modifyContext (f : IRContext OpCode → IRContext OpCode) : MlirParserM Unit := do
+  modify fun s => {s with ctx :=  f s.ctx}
+
+/--
+  Modifies the current IR context.
+
+  This function should be used instead of modifying the context with
+  `get`/`getContext` and `set`/`setContext` in order to preserve linearity.
+-/
+def modifyContextM' (f : IRContext OpCode → MlirParserM (α × IRContext OpCode)) : MlirParserM α := do
+  let ctx ← getContext
+  -- This `setContext` is required to preserve the linearity of the state
+  setContext default
+  let (r, ctx) ← f ctx
+  setContext ctx
+  pure r
+
+/--
+  Modifies the current IR context.
+
+  This function should be used instead of modifying the context with
+  `get`/`getContext` and `set`/`setContext` in order to preserve linearity.
+-/
+def modifyContextM (f : IRContext OpCode → MlirParserM (IRContext OpCode)) : MlirParserM Unit :=
+  modifyContextM' (fun ctx => do pure ((), ← f ctx))
+
 set_option warn.sorry false in
 /--
   Create a block at the given insert point and register its name in the parsing context.
@@ -97,10 +127,10 @@ def defineBlock (name : ByteArray) (ip : BlockInsertPoint) : MlirParserM BlockPt
     throw s!"block %{String.fromUTF8! name} has already been defined"
   | some (block, false) => -- Block of this name was forward declared.
     /- Insert the block at the given location. -/
-    let ctx ← getContext
-    let some ctx := Rewriter.insertBlock? ctx block ip (by sorry) (by sorry) (by sorry)
-      | throw "internal error: failed to insert block"
-    setContext ctx
+    modifyContextM fun ctx => do
+      let some ctx := Rewriter.insertBlock? ctx block ip (by sorry) (by sorry) (by sorry)
+        | throw "internal error: failed to insert block"
+      pure ctx
     /- Notify the parsing context that the block is defined. -/
     modifyThe MlirParserState (fun state =>
     {state with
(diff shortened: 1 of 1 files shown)


Veir/Meta/OpCode.lean
```diff
@@ -76,7 +76,7 @@ meta def emitFromName (ds : Array Dialect) : TermElabM Command := do
   for d in ds do
     for op in d.operations do
       if d.getName = "builtin" ∧ op = "unregistered" then continue
-      res ← `(if name = $(Syntax.mkStrLit (d.mkOpName op)).toByteArray then ($(mkIdent d.mkDialectCode) $(mkIdent (.mkStr2 d.name op))) else $res)
+      res ← `(if name.eq $(Syntax.mkStrLit (d.mkOpName op)).toByteArray then ($(mkIdent d.mkDialectCode) $(mkIdent (.mkStr2 d.name op))) else $res)
```
ineol: this is not necessary anymore since Henrik fixed the lean implementation upstream. You can also drop the `eq` implementation

GZGavinZhao:
Benchmarking whether `[inline]` has a measurable impact or not, should take ~10 mins.

GZGavinZhao:
`[inline]` doesn't have measurable performance improvement, dropping it:
```
> hyperfine "./no-inline-veir-opt sqlite3-stripped.mlir" "./inline-veir-opt sqlite3-stripped.mlir"
Benchmark 1: ./no-inline-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.205 s ±  0.028 s    [User: 2.183 s, System: 0.018 s]
  Range (min … max):    2.173 s …  2.247 s    10 runs

Benchmark 2: ./inline-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.179 s ±  0.024 s    [User: 2.158 s, System: 0.018 s]
  Range (min … max):    2.157 s …  2.227 s    10 runs

Summary
  ./inline-veir-opt sqlite3-stripped.mlir ran
    1.01 ± 0.02 times faster than ./no-inline-veir-opt sqlite3-stripped.mlir
```

Veir/Parser/MlirParser.lean
```diff
@@ -84,6 +85,28 @@ def registerValueDef (name : ByteArray) (value : ValuePtr) : MlirParserM Unit :=
...
+-/
+def modifyContext (f : IRContext OpCode → IRContext OpCode) : MlirParserM Unit := do
+  modify fun s => {s with ctx :=  f s.ctx}
+
+/--
+  Modifies the current IR context.
+-/
+def modifyGetContextM (f : IRContext OpCode → MlirParserM (α × IRContext OpCode)) : MlirParserM α := do
+  let s ← get
+  set (default : MlirParserState)
```
math-fehr: Can you add a comment here to explicitely mention we do this to preserve linearity?

Veir/Parser/MlirParser.lean
```diff
@@ -84,6 +85,28 @@ def registerValueDef (name : ByteArray) (value : ValuePtr) : MlirParserM Unit :=
...
+/--
+  Modifies the current IR context.
+-/
+def modifyContext (f : IRContext OpCode → IRContext OpCode) : MlirParserM Unit := do
+  modify fun s => {s with ctx :=  f s.ctx}
+
+/--
+  Modifies the current IR context.
+-/
+def modifyGetContextM (f : IRContext OpCode → MlirParserM (α × IRContext OpCode)) : MlirParserM α := do
```
math-fehr: Maybe renaming it `modifyContextM\2` or `modifyContextM'`?

math-fehr (approved):
Nice! Just had 2 style comments but otherwise perfect!

Veir/Parser/MlirParser.lean
GZGavinZhao: Do you mean above this `set` or for the function? Currently I'm doing the latter.

math-fehr: I would do it for both actually! Most importantly the `set` because it might seems weird from an outside perspective!

GZGavinZhao: Do the current comments sound clear to you? Added the same comment to `modifyContextM` as well since it's used for the same purpose.

math-fehr: I think it's fine, I would probably write it in the style `This function should be used instead of modifying the context with `get` and `set` in order to preserve linearity.`.

math-fehr:
I have a question actually, is there a reason we use `get` and `set` in `modifyContextM` instead of `getContext` and `setContext`? That should have the same effect without having to create a new state right?

GZGavinZhao:
Like this?
```lean
def modifyContextM' (f : IRContext OpCode → MlirParserM (α × IRContext OpCode)) : MlirParserM α := do
  let ctx ← getContext
  -- Writing it this way preserves linearity of the state
  let (r, ctx) ← f ctx
  setContext ctx
  pure r
```
This still has linearity issues:
```
> hyperfine "./no-inline-veir-opt sqlite3-stripped.mlir" "./setContext-veir-opt sqlite3-stripped.mlir"
Benchmark 1: ./no-inline-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.226 s ±  0.048 s    [User: 2.195 s, System: 0.017 s]
  Range (min … max):    2.177 s …  2.327 s    10 runs

Benchmark 2: ./setContext-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):     15.694 s ±  0.228 s    [User: 15.649 s, System: 0.017 s]
  Range (min … max):   15.435 s … 16.051 s    10 runs

Summary
  ./no-inline-veir-opt sqlite3-stripped.mlir ran
    7.05 ± 0.18 times faster than ./setContext-veir-opt sqlite3-stripped.mlir
```

GZGavinZhao:
This version works, is this what you're looking for?
```lean
def modifyContextM' (f : IRContext OpCode → MlirParserM (α × IRContext OpCode)) : MlirParserM α := do
  let ctx ← getContext
  -- THIS LINE IS REQUIRED
  setContext default
  let (r, ctx) ← f ctx
  setContext ctx
  pure r
```

math-fehr:
> This version works, is this what you're looking for?
> 
> ```lean
> def modifyContextM' (f : IRContext OpCode → MlirParserM (α × IRContext OpCode)) : MlirParserM α := do
>   let ctx ← getContext
>   -- THIS LINE IS REQUIRED
>   setContext default
>   let (r, ctx) ← f ctx
>   setContext ctx
>   pure r
> ```

Yes, that's what I was thinking of, here we just solve the linearity issue on the context, we don't need to solve it on the entire parser state.

Veir/Parser/MlirParser.lean
GZGavinZhao: Updated, what do you think of the current state?

GZGavinZhao:
Oops misclick, sorry for the ping Leo!

math-fehr (approved):
Nice, thanks a lot!

PR #483, opened 2026-05-01
Add get-set lemmas for `getOpType`

math-fehr:
We should favor the use of `op.getOpType! ctx` instead of `(op.get! ctx).opType`. This PR changes the get-set lemmas to use `getOpType!`.


tobiasgrosser (approved):
LGTM

luisacicolini (approved):

PR #484, opened 2026-05-01
Add a new axiom for dominance, and improve grind pattern

math-fehr:
The new axiom states that any user of an operation is being strictly dominated by it.

### DIFF
--- a/Veir/Dominance.lean
+++ b/Veir/Dominance.lean
@@ -10,6 +10,9 @@ public import Veir.Rewriter.InsertPoint
   This file is a placeholder for the dominance relation between IR constructs.
   It currently only contains axioms, and will be filled in with actual definitions and proofs
   in the future.
+
+  This formalization assumes that all regions are SSACFG regions, so it particular it doesn't
+  support graph regions.
 -/
 
 public section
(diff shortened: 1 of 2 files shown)


tobiasgrosser (approved):
Interesting.

This may not be true for graph regions and only for valid IR, not for any IR.

math-fehr:
Yes, this is only valid for IR that satisfies dominance (`ctx.Dom`), and for now we do not encode graph regions.

luisacicolini:
> Yes, this is only valid for IR that satisfies dominance (`ctx.Dom`), and for now we do not encode graph regions.

maybe worth adding these as caveats in the comment of the axiom? Other than that LGTM!

math-fehr:
> > Yes, this is only valid for IR that satisfies dominance (`ctx.Dom`), and for now we do not encode graph regions.
> 
> maybe worth adding these as caveats in the comment of the axiom? Other than that LGTM!



> > Yes, this is only valid for IR that satisfies dominance (`ctx.Dom`), and for now we do not encode graph regions.
> 
> maybe worth adding these as caveats in the comment of the axiom? Other than that LGTM!

Good point, I added it in the file description as other axioms are also having this assumption.

PR #487, opened 2026-05-01
Add `OperationPtr.getResultTypes`

math-fehr:
This helper is useful in the interpreter, and makes it easier to write get-set lemmas for high-level rewriting operations.

### DIFF
--- a/Veir/IR/Basic.lean
+++ b/Veir/IR/Basic.lean
@@ -668,6 +668,23 @@ theorem getResults!.getElem_eq_getResult
   simp only [getResults!, getResult]
   grind
 
+def getResultTypes (op : OperationPtr) (ctx : IRContext OpInfo)
+    (inBounds : op.InBounds ctx := by grind) : Array TypeAttr :=
+  (op.get ctx).results.map (·.type)
+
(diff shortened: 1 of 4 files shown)


Veir/IR/Basic.lean
```diff
@@ -668,6 +668,23 @@ theorem getResults!.getElem_eq_getResult
   simp only [getResults!, getResult]
   grind
 
+def getResultTypes (op : OperationPtr) (ctx : IRContext OpInfo)
+    (inBounds : op.InBounds ctx := by grind) : Array TypeAttr :=
+  Array.map (fun i => (op.get ctx).results[i]!.type) (Array.range (op.getNumResults ctx inBounds))
```
nchappe: Shouldn't `[]!` be avoided in non-`!` functions? `(op.get ctx).results.map (·.type)` could be more concise.

tobiasgrosser (approved):
This LGTM after @nchappe's comment is addressed.

Veir/IR/Basic.lean
math-fehr: Oh yes, sorry about this, I fixed it

PR #489, opened 2026-05-01
Parse Location attributes

georgerennie:
This parses MLIR's location attributes of the form `loc(body)` where body is a string with balanced delimiters inside into a LocationAttr.

LocationAttr just stores `body` as a string for now, without trying to parse it further.


tobiasgrosser (approved):
This looks sensible to me. We should however still wait for @math-fehr to comment on this patch.

math-fehr (approved):
Perfect, thanks!

PR #491, opened 2026-05-01
hw: Parse `!hw.modty`, `hw.module` and `hw.output`

georgerennie:
This PR adds parsing logic to parse the `!hw.modty` type, a description of the ports of a hardware module, as well as `hw.module` and `hw.output`.

To parse `!hw.modty` we introduce `HW.ModuleType`, `HW.ModulePort` and `HW.ModulePort.Direction` with parsers for each.

This currently ignores the `result_locs` and `comment` fields of `hw.module`.

I'm happy to split out the `!hw.modty` parser from the usage in `hw.module` and `hw.output` into a separate PR if it makes it easier to review, but it was gonna end up with stacked PRs so I thought I would leave them for now.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -172,7 +172,7 @@ deriving Inhabited, Repr, DecidableEq, Hashable
 
 /--
   The `!hw.modty` type from CIRCT's hw dialect.
-  This represents a list of ports to a module
+  This represents a list of ports to a module.
 -/
 structure ModuleType where
   ports : Array ModulePort
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -1,6 +1,7 @@
 module
 
 import Veir.ForLean
+public import Std.Data.Iterators.Producers.Array
 
 /-!
   # Attributes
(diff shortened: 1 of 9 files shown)


tobiasgrosser:
@georgerennie, #489 has been merged. What is the status here?

georgerennie:
> @georgerennie, #489 has been merged. What is the status here?


I've updated the PR to add a parser for the `!hw.modty` type. This means we can now parse `hw.module` and `hw.output` properly. More `hw` operations will be needed in due course to do more with it (particularly `hw.instance`) but this is a start.

Veir/IR/Attribute.lean
```diff
@@ -1,6 +1,7 @@
 module
 
 import Veir.ForLean
+public import Std.Data.Iterators.Producers.Array
```
math-fehr, line 4: Does this need to be a public import?

math-fehr (approved):
All good to me, only one question about a public import!

Veir/IR/Attribute.lean
georgerennie: I use it in
```lean4
instance : ToString HW.ModuleType where
  toString attr :=
    let values := attr.ports.iter.map ToString.toString |>.intercalateString ", "
    s!"!hw.modty<{values}>"
```
to let me map and intercalate without bothering to create another array in the middle.

Because this is an `instance`, the definition is exposed so it needs imports it relies on to be `public`.

PR #492, opened 2026-05-02
[WIP] dependently typed implementation of peephole rewrite verification

math-fehr:
Exploration of using dependent types in the interpreter to make proofs much simpler in the peephole rewrite verification


math-fehr:
Superceeded by PDL implementation

PR #494, opened 2026-05-04
Fix definition of `.WellFormed` to only take pointers instead of structures.

### LINKED ISSUE #44: Do not have both pointers and structure in WellFormed structure
Currently, you need to pass both `opPtr` and `op` to `Operation.WellFormed`.

math-fehr:
`Operation.WellFormed`, `Block.WellFormed`, `Region.WellFormed` were taking both the pointer and the structure as argument, despite the structure not being used in any of the fields.

This changes their names to `OperationPtr.WellFormed` (etc..), and remove the structure argument. This doesn't have any other changes.

Fixes #44


tobiasgrosser (approved):
This LGTM. Thank you, @math-fehr!

luisacicolini (approved):

ineol (approved):

PR #501, opened 2026-05-05
feat: unpack a PackedFloat into EScientificBV

bollu:
We add a concept of 'State' to streamline the unpacking of PackedFloat.
The central theorem to be proven (that will be left with a `proof_wanted`)
is that unpacking does not change the `toExtRat` value.

### DIFF
--- a/State.lean
+++ b/State.lean
@@ -1,0 +1,2 @@
+public import Veir.Data.FP.State.Basic
+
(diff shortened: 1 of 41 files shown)


tobiasgrosser:
Nice to see these PRs coming in swiftly.

> We add a concept of 'State' to streamline the unpacking of PackedFloat. The central theorem to be proven (that will be left with a `proof_wanted`) is that unpacking does not change the `toExtRat` value.

The commit message needs an update and the PR rebase.

PR #502, opened 2026-05-05
feat: add FP state inductive

bollu:
This PR adds a new C-style enum-inductive called 'state' which is
bitblastable, and enumerates the possible states a floating point number
can be in. This will be used to define unpacking, and to redefine the
rational interpretation of a packed number in a way that allows for
reasoning across abstraction layers (packed / escientific) of the status
of a given object.


tobiasgrosser (approved):
LGTM.

Issue #504, opened 2026-05-06
Computational Floating Point Design


bollu:
1. We have the lean `Float` which are doubles.
2. These are converted into a `PackedFloat`, that separates the IEEE components of a float into `(sign, mantissa, exponent)`. 
3. Following this, for computation, these are converted into an `EDyadic`, which is an inductive that can either be `nan`, `infinity (sign : Bool)`, `zero (sign : Bool)`, or `number (n : Dyadic)`. This provides the ability to perfectly represent numbers of the form $$k \times 2^n$$.
4. Now, on `Dyadic`, we will implement `round : Dyadic -> (enew : Int) (snew : Nat) -> ExtDyadic`, which provides an `ExtDyadic` that is rounded, i.e., there exits a IEEE floating point in `(enew, snew)` that is represented by this `ExtDyadic`. 
5. Finally, we have `pack : ExtDyadic -> Option PackedFloat` that packs a number back into an IEEE float. 

Overall, this representation uses `Dyadic` for scientific notation representation, and builds the theory of packing and unpacking.

PR #505, opened 2026-05-06
feat: add EDyadic for computable FP

bollu:
As discussed with @tobiasgrosser, for computable rounding,
it suffices to use the dyadics as a representation,
since it is a compressed, arbitrary-precision representation of numbers
of the form `k/2^n` where `k` is not divisible by `2`.

Therefore, we add an `EDyadic` type, and show how to convert it to a
rational. As the next step, we will show how to convert a `PackedFloat` into
an `EDyadic`.

Part of https://github.com/opencompl/veir/issues/504

### DIFF
--- a/Veir/Data/FP/EDyadic/Basic.lean
+++ b/Veir/Data/FP/EDyadic/Basic.lean
@@ -1,6 +1,4 @@
 module
-public import Veir.Data.FP.Sign
-public import Veir.Data.FP.ExtRat
 public import Veir.Data.FP.ScientificBV
 
 namespace Veir.Data.FP
@@ -8,14 +6,20 @@ namespace Veir.Data.FP
 public section
 
 /--
-An extended dyadic number (EScientificBV)
+An extended dyadic number (EDyadic)
 is a representation of a floating-point number
 that includes the special values zero, infinity, and NaN.
 -/
 inductive EDyadic where
   /-- A zero with a sign bit. -/
   | zero (sign : Bool)
-  /-- A nonzero finite number in scientific notation. -/
+  /-- A nonzero finite number in dyadic notation.
+  While dyadics can be zero, we use the `zero` constructor to
+  represent zero values (so the sign is represent),
+  and we use `nonzeroFinite` to represent nonzero values.
+  Thus, there is an implicit invariant that the `nonzeroFinite`
+  constructor should only be used with nonzero dyadics
+  -/
   | nonzeroFinite (d : Dyadic)
   /-- An infinite number with a sign bit. -/
   | infinity (sign : Bool)
--- a/Veir/Data/FP/EDyadic/Basic.lean
+++ b/Veir/Data/FP/EDyadic/Basic.lean
@@ -18,7 +18,7 @@ inductive EDyadic where
   represent zero values (so the sign is represent),
   and we use `nonzeroFinite` to represent nonzero values.
   Thus, there is an implicit invariant that the `nonzeroFinite`
-  constructor should only be used with nonzero dyadics
+  constructor should only be used with nonzero dyadics.
   -/
   | nonzeroFinite (d : Dyadic)
   /-- An infinite number with a sign bit. -/
--- a/Veir/Data/FP/EDyadic/Basic.lean
+++ b/Veir/Data/FP/EDyadic/Basic.lean
@@ -1,0 +1,36 @@
+module
+public import Veir.Data.FP.Sign
+public import Veir.Data.FP.ExtRat
+public import Veir.Data.FP.ScientificBV
+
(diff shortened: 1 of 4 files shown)


Veir/Data/FP/EDyadic/Basic.lean
```diff
@@ -0,0 +1,36 @@
...
+public import Veir.Data.FP.Sign
+public import Veir.Data.FP.ExtRat
+public import Veir.Data.FP.ScientificBV
+
+namespace Veir.Data.FP
+
+public section
+
+/--
+An extended dyadic number (EScientificBV)
```
tobiasgrosser: Why is there `EScientificBV` in this line?

Veir/Data/FP/EDyadic/Basic.lean
```diff
@@ -0,0 +1,36 @@
...
+is a representation of a floating-point number
+that includes the special values zero, infinity, and NaN.
+-/
+inductive EDyadic where
+  /-- A zero with a sign bit. -/
+  | zero (sign : Bool)
+  /-- A nonzero finite number in scientific notation. -/
+  | nonzeroFinite (d : Dyadic)
```
tobiasgrosser, line 23: Dyadics can be zero, no?

Veir/Data/FP/EDyadic/Basic.lean
```diff
@@ -0,0 +1,36 @@
+module
+public import Veir.Data.FP.Sign
+public import Veir.Data.FP.ExtRat
+public import Veir.Data.FP.ScientificBV
```
tobiasgrosser, line 2: Do we need all these includes?

Veir/Data/FP/EDyadic/Basic.lean
bollu: You're right, dyadics can represent zero, but they lose the uniqueness of representation (just like exponential notation), since $$0 / 1 = 0 / 2 = 0 / 4 = 0 \dots$$

For the specific case of floating point, it helps to separate out the zero case from the nonzero case, as floating point values behave very different at zero (rounding, signs, many things are different, so conceptually it's nicer to handle the case separately). Eventually, in our theorems, we will have an invariant that we will only store a nonzero dyadic in the `nonzeroFinite` slot.

Veir/Data/FP/EDyadic/Basic.lean
bollu: Thanks, fixed

Veir/Data/FP/EDyadic/Basic.lean
bollu: Thanks, fixed

bollu:
Also waiting on a review from @luisacicolini , as one of the purposes of doing this was a knowledge share.

Veir/Data/FP/EDyadic/Basic.lean
```diff
@@ -0,0 +1,40 @@
...
+-/
+inductive EDyadic where
+  /-- A zero with a sign bit. -/
+  | zero (sign : Bool)
+  /-- A nonzero finite number in dyadic notation.
+  While dyadics can be zero, we use the `zero` constructor to
+  represent zero values (so the sign is represent),
+  and we use `nonzeroFinite` to represent nonzero values.
+  Thus, there is an implicit invariant that the `nonzeroFinite`
+  constructor should only be used with nonzero dyadics
```
luisacicolini:
```suggestion
  constructor should only be used with nonzero dyadics.
```

Veir/Data/FP/EDyadic/ToExtRat.lean
```diff
@@ -0,0 +1,26 @@
...
+When the input is ±0, the result is `0`, losing the sign bit.
+Everywhere else, the function is injective.
+-/
+def toExtRat : EDyadic → ExtRat
+  | .zero _ => .number 0
+  | .nonzeroFinite d => .number d.toRat
+  | .infinity sign => .infinity sign
+  | .nan => .nan
+
+  end -- public section
```
luisacicolini:
```suggestion
  end
```
i don't think we usually have those?

Veir/Data/FP/EDyadic/Basic.lean
tobiasgrosser: Lean's Dyadic has an explicit zero case.

bollu: I agree it has a zero case, but it loses (a) the uniqueness of the representation, and more seriously, (b) does not distinguish between $$\pm 0$$. So we choose to fold this into a separate case (called `zero`)

tobiasgrosser (approved):

PR #508, opened 2026-05-06
Naive LLVM memory interpretation

nchappe:
Does not support poison, nor non-int64 accesses

### DIFF
--- a/Veir/ForLean.lean
+++ b/Veir/ForLean.lean
@@ -25,9 +25,28 @@ def UInt8.isDigit (c : UInt8) : Bool :=
 def UInt8.isHexDigit (c : UInt8) : Bool :=
   c.isDigit || (c >= 'a'.toUInt8 && c <= 'f'.toUInt8) || (c >= 'A'.toUInt8 && c <= 'F'.toUInt8)
 
+def UInt64.toByteArrayLE (u : UInt64) : ByteArray :=
+  ByteArray.mk (Array.mk [
+    u.toUInt8,
+    (u >>> 0x08).toUInt8,
+    (u >>> 0x10).toUInt8,
+    (u >>> 0x18).toUInt8,
+    (u >>> 0x20).toUInt8,
+    (u >>> 0x28).toUInt8,
+    (u >>> 0x30).toUInt8,
+    (u >>> 0x38).toUInt8,
+  ])
+
+namespace ByteArray
+
+def extend (ba : ByteArray) (n : Nat) (val : UInt8) : ByteArray :=
+  n.fold (init := ba) fun _ _ ba => ba.push val
+
 @[inline]
-def ByteArray.getD (ba : ByteArray) (i : Nat) (default : UInt8) : UInt8 :=
+def getD (ba : ByteArray) (i : Nat) (default : UInt8) : UInt8 :=
   if h : i < ba.size then ba[i] else default
+
+end ByteArray
 
 /--
   Convert a hexadecimal digit character to its Nat value.
@@ -298,11 +317,4 @@ namespace Rat
 def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
-
-namespace ByteArray
-
-def rightpad (n : Nat) (x : UInt8) (xs : ByteArray) : ByteArray where
-  data := Array.rightpad n x xs.data
-
-end ByteArray
 
--- a/Veir/ForLean.lean
+++ b/Veir/ForLean.lean
@@ -299,3 +299,10 @@ def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
 
+namespace ByteArray
+
+def rightpad (n : Nat) (x : UInt8) (xs : ByteArray) : ByteArray where
+  data := Array.rightpad n x xs.data
+
+end ByteArray
+
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -1,3 +1,4 @@
+import Lean.Compiler
 import Veir.IR.Basic
 import Veir.Rewriter.Basic
 import Veir.ForLean
@@ -35,13 +36,50 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
 -/
 inductive RuntimeValue where
 | int (bitwidth : Nat) (value : LLVM.Int bitwidth)
+| addr (value : UInt64)
 | reg (value : RISCV.Reg)
 deriving Inhabited
 
 instance : ToString (RuntimeValue) where
   toString
     | .int _ val => ToString.toString val
+    | .addr val => ToString.toString val
     | .reg val => ToString.toString val
+
+structure MemoryState where
+  contents : ByteArray
+  size : UInt64
+
+def MemoryState.empty : MemoryState :=
+  { contents := ByteArray.emptyWithCapacity 1024, size := 8 }
+
+def MemoryState.alloc (state : MemoryState) (size : UInt64)
+    : MemoryState × UInt64 :=
+  ({contents := (ByteArray.empty.push 0).copySlice 0 state.contents (state.size + size).toNat 1, size := state.size + size}, state.size)
+
+def MemoryState.store (state : MemoryState) (addr : UInt64) (val : ByteArray)
+    : MemoryState :=
+  {state with contents := val.copySlice 0 state.contents addr.toNat val.size false}
+
+def MemoryState.storeValue (state : MemoryState) (addr : UInt64) (val : RuntimeValue)
+    : MemoryState :=
+  match val with
+  | .int 64 (.val v) | .reg {val := v} => state.store addr (Lean.Compiler.LCNF.uint64ToByteArrayLE v.toNat.toUInt64).toList.toByteArray
+  | .int _ .poison => state
+  | .addr v => state.store addr (Lean.Compiler.LCNF.uint64ToByteArrayLE v).toList.toByteArray
+  | _ => state
+
+def MemoryState.load (state : MemoryState) (addr size : UInt64)
+    : ByteArray :=
+  state.contents.extract addr.toNat (addr + size).toNat
+
+def MemoryState.loadValue (state : MemoryState) (addr : UInt64) (type : TypeAttr)
+    : Option RuntimeValue := do
+  match type.val with
+  | Attribute.integerType { bitwidth := 8 } => some (.int 8 (.val (BitVec.ofNat 8 (state.load addr 1)[0]!.toNat)))
+  | Attribute.integerType { bitwidth := 64 } => some (.int 64 (.val (BitVec.ofNat 64 (state.load 1 8).toUInt64LE!.toNat)))
+  --| Attribute.llvmPointerType _ => some (.addr (state.load addr 8).toUInt64LE!)
+  | _ => none
 
 /--
   The state of the interpreter at a given point in time.
@@ -49,6 +87,7 @@ instance : ToString (RuntimeValue) where
 -/
 structure InterpreterState where
   variables : Std.ExtHashMap ValuePtr RuntimeValue
+  memory : MemoryState
 
 /--
   Set the value of a variable.
@@ -67,6 +106,7 @@ def InterpreterState.getVar? (state : InterpreterState) (var : ValuePtr)
 @[ext]
 theorem InterpreterState.ext {s₁ s₂ : InterpreterState} :
     (∀ var, s₁.getVar? var = s₂.getVar? var) →
+    s₁.memory = s₂.memory →
     s₁ = s₂ := by
   rcases s₁; rcases s₂
   simp only [getVar?, mk.injEq]
@@ -116,7 +156,7 @@ def InterpreterState.setArgumentValues (state : InterpreterState) (ctx : IRConte
   Create an interpreter state with no variables defined.
 -/
 def InterpreterState.empty : InterpreterState :=
-  { variables := Std.ExtHashMap.emptyWithCapacity 8 }
+  { variables := Std.ExtHashMap.emptyWithCapacity 8, memory := .empty }
 
 /--
   How the control flow should proceed after interpreting a terminator.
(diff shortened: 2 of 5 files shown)


nchappe:
- The test doesn't pass for now because [ByteArray.copySlice](https://lean-lang.org/doc/reference/latest/Basic-Types/Byte-Arrays/#ByteArray___copySlice) doesn't extend the destination ByteArray as I thought it would (if the destination ByteArray is not long enough it just appends the source ByteArray at the end and ignores the given indices).
- There is still one sorry left.

Veir/Interpreter/Basic.lean
```diff
@@ -33,20 +34,58 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
+def MemoryState.empty : MemoryState :=
+  { contents := ByteArray.emptyWithCapacity 1024, size := 8 }
+
+def MemoryState.alloc (state : MemoryState) (size : UInt64)
+    : MemoryState × UInt64 :=
+  ({contents := (ByteArray.empty.push 0).copySlice 0 state.contents (state.size + size).toNat 1, size := state.size + size}, state.size)
+
+def MemoryState.store (state : MemoryState) (addr : UInt64) (val : ByteArray)
+    : MemoryState :=
+  {state with contents := val.copySlice 0 state.contents addr.toNat val.size false}
```
luisacicolini:
```suggestion
  { state with contents := val.copySlice 0 state.contents addr.toNat val.size false }
```
super nit for consistency with the above

luisacicolini (commented):
very cool and useful for fastntt, thanks a lot for working on it!!

nchappe:
I fixed the interpreter, addressed Luisa's comments, rebased and cleaned up. The PR is now ready for review.
The `sorry`s will be non-trivial to fix, this will be for a later PR.

Veir/ForLean.lean
```diff
@@ -299,3 +299,10 @@ def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
 
+namespace ByteArray
+
+def rightpad (n : Nat) (x : UInt8) (xs : ByteArray) : ByteArray where
+  data := Array.rightpad n x xs.data
```
ineol: I'm afraid going thought the underlying `Array` will have to copy and box all the bytes and then copy them back to a `ByteArray`, since the actual memory representation of a `ByteArray` is not the same as an `Array UInt8`. My hunch is that we should reimplement it using push.

Veir/Interpreter/Basic.lean
```diff
@@ -34,22 +35,83 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
+def MemoryState.empty : MemoryState :=
+  { contents := ByteArray.emptyWithCapacity 1024 }
+
+/--
+  Allocate the given number of bytes of memory.
+  Return the updated memory state and the freshly allocated address.
+-/
+def MemoryState.alloc (state : MemoryState) (size : UInt64)
+    : MemoryState × UInt64 :=
+  ({ contents := ByteArray.rightpad (state.contents.size + size.toNat) 0 state.contents }, state.contents.size.toUInt64)
```
ineol: Very long line

ineol: Maybe define an extend function since you're always extending?

Veir/Interpreter/Basic.lean
```diff
@@ -34,22 +35,83 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
+  { state with contents := val.copySlice 0 state.contents addr.toNat val.size false }
+
+/--
+  Store a value to memory.
+-/
+def MemoryState.storeValue (state : MemoryState) (addr : UInt64) (val : RuntimeValue)
+    : MemoryState :=
+  match val with
+  | .int 8 (.val v) => state.store addr (ByteArray.empty.push v.toNat.toUInt8)
+  | .int 64 (.val v) | .reg {val := v} => state.store addr (Lean.Compiler.LCNF.uint64ToByteArrayLE v.toNat.toUInt64).toList.toByteArray
```
ineol: I don't think we should use internal Lean compiler functions here.

Why do we do `toByteArrayLE` then `toList` then `toByteArray`?

Veir/Interpreter/Basic.lean
```diff
@@ -34,22 +35,83 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
+  | .int _ .poison => state
+  | .addr v => state.store addr (Lean.Compiler.LCNF.uint64ToByteArrayLE v).toList.toByteArray
```
ineol: idem

Veir/Interpreter/Basic.lean
```diff
@@ -34,22 +35,83 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
+  | _ => state
```
ineol: pass through or error?

Veir/Interpreter/Basic.lean
```diff
@@ -34,22 +35,83 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
+
+/--
+  Load raw bytes from the given memory address.
+  Returns a truncated result if the access is out of bounds.
+-/
+def MemoryState.load (state : MemoryState) (addr size : UInt64)
```
ineol: Should we really silently fail when an access is out of bounds?

Veir/Interpreter/Basic.lean
```diff
@@ -439,11 +503,29 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
...
     | .int 1 .poison => Interp.ub
     | _ => none
+  | .alloca => do
+    let [.int _ (.val count)] := operands.toList | none
+    let size ← match properties.elem_type.val with
+    | Attribute.integerType { bitwidth := bw } => some (.ok (bw / 8))
+    | .llvmPointerType _ => some (.ok 8)
+    | _ => none
+    let totalSize := (size * count.toNat).toUInt64
+    let (mem, addr) := mem.alloc totalSize
```
ineol, line 535: If we exceed 2^64 does this overwrite the beginning of the memory?

Veir/Interpreter/Lemmas.lean
```diff
@@ -80,12 +81,20 @@ theorem InterpreterState.getOperandValues_eq_of_getVar?_eq {ctx : IRContext OpIn
...
+  sorry
+
 theorem InterpreterState.setResultValues_comm {ctx : WfIRContext OpInfo}
     {state : InterpreterState} (hOp : op₁ ≠ op₂) :
     (state.setResultValues ctx.raw op₁ resValues₁).setResultValues ctx.raw op₂ resValues₂ =
     (state.setResultValues ctx.raw op₂ resValues₂).setResultValues ctx.raw op₁ resValues₁ := by
   ext val runtimeVal
   cases val <;> grind
+  grind [InterpreterState.setResultValues_memory]
+  grind [InterpreterState.setResultValues_memory]
```
ineol: Add bullet points for the subgoals?

Veir/Interpreter/Lemmas.lean
```diff
@@ -125,3 +134,5 @@ theorem InterpreterState.setResultValues_setResultValues_self {ctx : WfIRContext
   ext val runtimeVal
   simp only [InterpreterState.getVar?_setResultValues]
   grind
+  grind [InterpreterState.setResultValues_memory]
+  grind [InterpreterState.setResultValues_memory]
```
ineol: idem

ineol (commented):
Thank you!

I have a high level question: once we add RISC-V memory, will the interpreter state have 2 memories or one that can contain both types of values? (of course now they have bytes but eventually LLVM will need to track provenance etc)

nchappe:
> Thank you!
> 
> I have a high level question: once we add RISC-V memory, will the interpreter state have 2 memories or one that can contain both types of values? (of course now they have bytes but eventually LLVM will need to track provenance etc)

I expect the memory state to be simpler on the RISC-V side. My hope is that even with a more realistic memory model, it will be possible to see the RISC-V memory state as a "sub-state" of the LLVM memory state (by ignoring provenance and poison information) with no significant performance cost.

Veir/ForLean.lean
nchappe: I agree, this was meant to be a short-term workaround. The plan was to switch to ExArray in the short/medium-term, I could make the switch right now but the repo is still private?

ineol: You can just define it as the following for now I think?
```suggestion
def extend (ba : ByteArray) (n : Nat) (val : UInt8) : ByteArray :=
  n.fold (init := ba) fun _ _ ba => ba.push val
```

nchappe: changed, thanks

Veir/Interpreter/Basic.lean
nchappe: simplified

Veir/Interpreter/Basic.lean
nchappe: wrapped in the Interp monad, the unsupported cases now fail

Veir/Interpreter/Lemmas.lean
nchappe: done

Veir/Interpreter/Basic.lean
nchappe: Now it raises UB! (using the Interp monad)

Veir/Interpreter/Basic.lean
nchappe: The only way to exceed 2^64 is to allocate an absurdly large amount of memory that will make the program crash anyways. I don't think it's worth adding logic to handle this in this short-term memory interpreter.

nchappe:
Cleaned up, and added bounds checks so that out-of-bounds load/store raises UB.

Veir/Interpreter/Basic.lean
math-fehr: The only issue I would see is if we start proving this, but I guess once we get there we'll modify it

Veir/Interpreter/Basic.lean
```diff
@@ -34,22 +34,35 @@ variable {OpInfo : Type} [HasOpInfo OpInfo]
...
 
+/--
+  Memory state during interpretation
+-/
+@[ext]
+structure MemoryState where
+  contents : ByteArray
+
+def MemoryState.empty : MemoryState :=
+  { contents := (ByteArray.emptyWithCapacity 1024).extend 8 0xff }
```
math-fehr, line 56: Can you comment why you are extending the memory at the beginning?
Is it illegal to allocate/write at address 0? Is it only on user mode?

Veir/Interpreter/Lemmas.lean
```diff
@@ -124,4 +133,6 @@ theorem InterpreterState.setResultValues_setResultValues_self {ctx : WfIRContext
     state.setResultValues ctx.raw op resValues' := by
   ext val runtimeVal
   simp only [InterpreterState.getVar?_setResultValues]
```
math-fehr: Why is this line not indentated?

Veir/Interpreter/Lemmas.lean
```diff
@@ -80,12 +81,20 @@ theorem InterpreterState.getOperandValues_eq_of_getVar?_eq {ctx : IRContext OpIn
   intro l hl
   induction l <;> grind
 
+set_option warn.sorry false in
+theorem InterpreterState.setResultValues_memory {ctx : WfIRContext OpInfo}
+    {state : InterpreterState} :
+    (state.setResultValues ctx.raw op resValues).memory = state.memory := by
+  sorry
```
math-fehr: Is there a reason why this is not proven here?

Veir/Interpreter/Lemmas.lean
```diff
@@ -55,20 +55,21 @@ theorem InterpreterState.getVar?_setResultValues_of_value_inBounds
...
     state' = state.setResultValues ctx op resValues := by
   simp only [interpretOp, bind, pure]
   rcases hOps : state.getOperandValues ctx op with _ | operands
   · grind
   rcases hOp : interpretOp' (op.getOpType! ctx) (op.getProperties! ctx (op.getOpType! ctx))
-      (op.getResultTypes! ctx) operands (op.getSuccessors! ctx) with _ | u
+      (op.getResultTypes! ctx) operands (op.getSuccessors! ctx) state.memory with _ | u
   · grind
-  cases u <;> grind
+  · sorry
```
math-fehr, line 72: Is there something missing to finish this proof?

Veir/Interpreter/EquationLemma.lean
```diff
@@ -40,8 +41,9 @@ theorem interpretOp_equationHolds_self
   intro hInterp
   exists controlFlow
   have ⟨operandValues, resValues, hOperandValues, hInterp', hResValues⟩ := interpretOp_some_iff.mp hInterp
-  grind [interpretOp]
```
math-fehr, line 43: I can take care of these proofs in a separate PR, I'll need to add a `isPure` function to make this work!

math-fehr (approved):
Very nice! I just had 1-2 comments, but otherwise this is good to me once these are addressed!

Veir/Interpreter/Lemmas.lean
tobiasgrosser: The idea was to get a skeleton of this PR in to ensure we can get testing feedback on it. We can then close open proves incrementally.

Veir/Interpreter/Basic.lean
tobiasgrosser: Maybe add a comment here to document this? I don't think we need to address this in code for now.

Veir/Interpreter/EquationLemma.lean
tobiasgrosser: Yes. It would be great to close these proves in subsequent PRs to not hold this one back. The current PR touches a lof of the interpreter code, so merging it will protect @nchappe from costly rebase work.

tobiasgrosser (approved):
Amazing. This is good-to-go from my side. 🙏

Veir/Interpreter/Lemmas.lean
math-fehr: Yes, but my understanding is that this should one should be rather trivial to prove, which is why I'm asking if there is some subtelties here or not.

Veir/Interpreter/Lemmas.lean
ineol: I wonder if there is a linter for this kind of things somewere

ineol (approved):
Thank you for the fixes!

Veir/Interpreter/Basic.lean
nchappe: Dereferencing a null address can be valid at the RISC-V level, but this is LLVM-only for now. Can a null LLVM pointer ever have valid provenance? Either way, I'd prefer to leave this corner case for when we have support for provenance.

Veir/Interpreter/Lemmas.lean
nchappe: this is now proven

Veir/Interpreter/Lemmas.lean
nchappe: I expect a few lemmas about memory would be needed, and possibly the statement would have to be adjusted.

nchappe:
Addressed Mathieu's comments, will merge soon, thanks everyone.

PR #509, opened 2026-05-06
Add `IRContext.Verified` and `OperationPtr.Verified`

math-fehr:
`OperationPtr.Verified` asserts that the operation verifies its local invariants, and `IRContext.Verified` asserts that all contained operations verify their local invariants.


tobiasgrosser (approved):
Nice

luisacicolini (approved):

PR #510, opened 2026-05-06
Add type information in arith.constant and arith.addi verifiers

math-fehr:
This checks that the operands and result of an `arith.addi` have the same integer type, and that the result of an `arith.constant` has the same type as the property it holds.

Also, add a lemma that states the constraints of a verified `arith.constant` or `arith.addi`. While I would love to have these lemmas in `grind`, I do not know how to have a pattern that only triggers if `getOpType! ctx = .arith .addi`. Currently, I can only make this trigger when `getOpType! ctx` is in the context, which is way too weak and might trigger this too many times (especially when we are going to have more of these lemmas).

### DIFF
--- a/Veir/Verifier.lean
+++ b/Veir/Verifier.lean
@@ -30,15 +30,22 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
     pure ()
   /- ARITH -/
   | .arith .addi => do
-    if op.getNumOperands ctx.raw opIn ≠ 2 then
+    if _ : op.getNumOperands ctx.raw opIn ≠ 2 then
       throw "Expected 2 operands"
-    if op.getNumResults ctx.raw opIn ≠ 1 then
+    else if _ : op.getNumResults ctx.raw opIn ≠ 1 then
       throw "Expected 1 result"
-    if op.getNumRegions ctx.raw opIn ≠ 0 then
+    else if op.getNumRegions ctx.raw opIn ≠ 0 then
       throw "Expected 0 regions"
-    if op.getNumSuccessors ctx.raw opIn ≠ 0 then
+    else if op.getNumSuccessors ctx.raw opIn ≠ 0 then
       throw "Expected 0 successors"
-    pure ()
+    else
+      let .integerType integerType := ((op.getOperand ctx.raw 0).getType ctx.raw).val
+        | throw "Expected integer type result"
+      if ((op.getOperand ctx.raw 0).getType ctx.raw).val ≠ ((op.getOperand ctx.raw 1).getType ctx.raw).val then
+        throw "Expected operands to have the same type"
+      if ((op.getResult 0).get ctx.raw).type.val ≠ ((op.getOperand ctx.raw 0).getType ctx.raw).val then
+        throw "Expected result type to match operand type"
+      pure ()
   | .arith .addui_extended => do
     if op.getNumOperands ctx.raw opIn ≠ 2 then
       throw "Expected 2 operands"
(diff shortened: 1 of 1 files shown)


Veir/Verifier.lean
```diff
@@ -38,6 +38,18 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
...
+      if _ : op.getNumResults ctx.raw opIn ≠ 1 then
+        throw "Expected 1 result"
+      else
+        let .integerType integerType := ((op.getOperand ctx.raw 0).getType ctx.raw).val
+          | throw "Expected integer type result"
+        if ((op.getOperand ctx.raw 0).getType ctx.raw).val ≠ ((op.getOperand ctx.raw 1).getType ctx.raw).val then
+          throw "Expected operands to have the same type"
+        if ((op.getResult 0).get ctx.raw).type.val ≠ ((op.getOperand ctx.raw 0).getType ctx.raw).val then
+          throw "Expected result type to match operand type"
     pure ()
```
tobiasgrosser: This could read:

```lean4
| .arith .addi => do
    if _ : op.getNumOperands ctx.raw opIn ≠ 2 then
      throw "Expected 2 operands"
    else if _ : op.getNumResults ctx.raw opIn ≠ 1 then
      throw "Expected 1 result"
    else if op.getNumRegions ctx.raw opIn ≠ 0 then
      throw "Expected 0 regions"
    else if op.getNumSuccessors ctx.raw opIn ≠ 0 then
      throw "Expected 0 successors"
    else
      let .integerType integerType := ((op.getOperand ctx.raw 0).getType ctx.raw).val
        | throw "Expected integer type result"
      if ((op.getOperand ctx.raw 0).getType ctx.raw).val ≠ ((op.getOperand ctx.raw 1).getType ctx.raw).val then
        throw "Expected operands to have the same type"
      if ((op.getResult 0).get ctx.raw).type.val ≠ ((op.getOperand ctx.raw 0).getType ctx.raw).val then
        throw "Expected result type to match operand type"
      pure ()
```

Which avoids adding two obvious redundant cases.

Veir/Verifier.lean
```diff
@@ -98,6 +110,12 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
       throw "Expected 0 regions"
     if op.getNumSuccessors ctx.raw opIn ≠ 0 then
       throw "Expected 0 successors"
+    if _ : op.getNumResults ctx.raw opIn ≠ 1 then
+      throw "Expected 1 result"
+    else
+      if (op.getProperties! ctx.raw (.arith .constant)).value.type ≠
+          ((op.getResult 0).get ctx.raw).type.val then
+        throw "Expected result type to be equal to the constant's type"
```
tobiasgrosser, line 110: Same as above. You can avoid adding a redundant case here.

Veir/Verifier.lean
```diff
@@ -1780,5 +1798,42 @@ theorem OperationPtr.satisfyInvariants_of_IRContext_satisfyOpInvariants {ctx : W
     op.Verified ctx opInBounds := by
   sorry -- This requires to reason about `IRContext.forOpsDepM`.
 
+/-!
+## Lemmas for verified operations
+
+These are the lemmas that give the information about the structure of verified operations.
+There is one lemma per operation, and they are all of the same form: given that an operation
+satisfies its local invariants, we can conclude that it has the expected number of operands, results, regions, and successors,
```
tobiasgrosser: This line is likely too long.

tobiasgrosser (approved):

luisacicolini (approved):

PR #515, opened 2026-05-07
rename llvm.constant to llvm.mlir.constant

tobiasgrosser:
MLIR uses the name `mlir.constant` for the constant operation in the `llvm` dialect to indicate that this is an MLIR-specific extension not present in LLVM. LLVM does not have constants as explicit operations in the IR but instead allows constant expressions as separate data structures as an alternative to SSA values.

As lean does not like `.` in names of inductive cases, we instead use `__` as a token for `.`.


math-fehr (approved):
We don't have any test with `llvm.constant`?

PR #523, opened 2026-05-08
Migrate parser to use WfIRContext

GZGavinZhao:
Required by #435. Makes writing proofs in verified parser much easier.

### DIFF
--- a/UnitTest/MlirParser.lean
+++ b/UnitTest/MlirParser.lean
@@ -1,6 +1,7 @@
-import Veir.Parser.MlirParser
 import Veir.IR.Basic
+import Veir.Parser.MlirParser
 import Veir.Printer
+import Veir.Rewriter.WfRewriter
 
 open Veir
 open Veir.Parser
@@ -9,7 +10,7 @@ open Veir.Parser
   Parse an operation and print it.
 -/
 def testParseOp (s : String) : IO Unit :=
-  match IRContext.create OpCode with
+  match WfIRContext.create OpCode with
   | some (ctx, _) =>
     match ParserState.fromInput (s.toByteArray) with
     | .ok parser =>
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -3,6 +3,8 @@ import Veir.Parser.AttrParser
 import Veir.IR.Basic
 import Veir.Rewriter.InsertPoint
 import Veir.Rewriter.Basic
+import Veir.Rewriter.GetSet
+import Veir.Rewriter.WellFormed
 import Veir.Properties
 import Veir.GlobalOpInfo
 
@@ -16,7 +18,7 @@ variable {ctx : IRContext OpCode}
 
 structure MlirParserState where
   /-- The current IR context. -/
-  ctx : IRContext OpCode
+  ctx : WfIRContext OpCode
   /-- The values that have been defined at that point in the parser. -/
   values : Std.HashMap ByteArray ValuePtr
   /--
@@ -27,7 +29,7 @@ structure MlirParserState where
   blocks : Std.HashMap ByteArray (BlockPtr × Bool)
   deriving Inhabited
 
-def MlirParserState.fromContext (ctx : IRContext OpCode) : MlirParserState :=
+def MlirParserState.fromContext (ctx : WfIRContext OpCode) : MlirParserState :=
   {ctx := ctx, values := Std.HashMap.emptyWithCapacity 128, blocks := Std.HashMap.emptyWithCapacity 1}
 
 abbrev MlirParserM := StateT MlirParserState (EStateM String ParserState)
@@ -55,7 +57,7 @@ def MlirParserM.run' (self : MlirParserM α)
 /--
   Get the current IR context that is stored in the parser state.
 -/
-def getContext : MlirParserM (IRContext OpCode) := do
+def getContext : MlirParserM (WfIRContext OpCode) := do
   return (← get).ctx
 
 /--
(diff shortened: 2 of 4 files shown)


GZGavinZhao:
The main thing I'd like to check here is distinguishing naming between `IRContext` and `WfIRContext`. We sometimes needs to use both and calling both `ctx` feels confusing to me, so the convention I used here is to use `wfctx` for any variable or function that represents or returns `WfIRContext`, and similarly use `ctx` for anything that's just plain `IRContext`. Open to changing this if this makes the API sounds weird/verbose.

tobiasgrosser (approved):
Please keep `ctx` as the name for WfIRContext. Otherwise, this looks great.

Let's wait for @math-fehr's feedback.

Veir/Parser/MlirParser.lean
```diff
@@ -385,19 +389,16 @@ partial def parseOptionalOp (ip : Option InsertPoint) : MlirParserM (Option Oper
...
-  let some (ctx, op) := Rewriter.createOp ctx opId outputTypes operands blockOperands regions properties ip (by sorry) (by sorry) (by sorry) (by sorry) (by sorry)
-      | throw "internal error: failed to create operation"
-  let ctx := op.setAttributes! ctx attrs
+  let op ← modifyContextM' fun ctx => do
+    match hctx' : Rewriter.createOp ctx.raw opId outputTypes operands blockOperands regions properties ip (by sorry) (by sorry) (by sorry) (by sorry) (by sorry) with
+    | none => throw "internal error: failed to create operation"
+    | some (ctx', op) =>
+      let ctx'' := op.setAttributes ctx' attrs (by sorry)
+      /- Update the parser context. -/
+      pure ⟨op, ⟨ctx'', by sorry⟩⟩
```
tobiasgrosser, line 398: What is this about? This seems to be more than changing the IRContext to WfIRContext. Can you add a comment to the PR message explaining this change.

GZGavinZhao: Oops, this is part of the linearity changes that I accidentally left out of #481. The logic of the original code was not changed, just rewritten to use the linearity-preserving `modifyContextM'`. Noted in the PR description.

tobiasgrosser: This should be a separate PR then. Do not sneak them into this PR.

UnitTest/MlirParser.lean
```diff
@@ -1,6 +1,7 @@
 import Veir.Parser.MlirParser
 import Veir.IR.Basic
 import Veir.Printer
+import Veir.Rewriter.WfRewriter
```
math-fehr, line 4:
```suggestion
import Veir.IR.WellFormed
```
This import should be better (imports WfIRContext without WfRewriter)

math-fehr (approved):
Nice! Once Tobias comment is resolved, this is good for me to go!

Veir/Parser/MlirParser.lean
GZGavinZhao: Opened #526.

UnitTest/MlirParser.lean
GZGavinZhao: Test uses `WfIRContext.create`, which is defined in `Veir.Rewriter.WfRewriter.Basic`, so we still need to import `WfRewriter`. Should I restore the original `import Veir.Rewriter.WfRewriter` or is there a better import to use?

tobiasgrosser:
The commit message should not include the linearity fix discussion and the build is currently broken. After both issues are resolved, feel invited to merge.

UnitTest/MlirParser.lean
math-fehr: oh my bad, I forgot it was defined there! Please restore it to what it was!

Veir/Parser/MlirParser.lean
```diff
@@ -162,9 +164,9 @@ def defineBlockUse (name : ByteArray) : MlirParserM BlockPtr := do
   | none => -- Block not yet encountered
     /- Create the block. -/
     let block ← modifyContextM' fun ctx => do
-      let some (ctx, block) := Rewriter.createBlock ctx none (by sorry) (by sorry)
-        | throw "internal error: failed to create block"
-      pure (block, ctx)
+      match hctx' : Rewriter.createBlock ctx.raw none (by grind) (by sorry) with
+      | none => throw "internal error: failed to create block"
+      | some (ctx', block) => pure ⟨block, ⟨ctx', by grind [Rewriter.createBlock_WellFormed]⟩⟩
```
tobiasgrosser, line 169: Why do you not use the nice `let ... | none` syntax here anymore?

GZGavinZhao: `grind` needs `hctx'`, which is only available in the (Lean) context if I do it in the `match` way. I hope there's someway to do it using `let` but couldn't figure it out.

GZGavinZhao: IIRC @ineol also investigated a nicer syntax but couldn't find one. This seems to be the best we have right now.

tobiasgrosser: @math-fehr and @ineol introduced a `hlet` syntax?

tobiasgrosser: If it is useful, feel free to use it. I am also happy to see this merged as is.

GZGavinZhao: Can someone elaborate what/where is the `hlet` syntax? I couldn't find definition or usage of `hlet` in the current codebase.

tobiasgrosser: Ask on zulip. :-) I may have gotten the name wrong.

tobiasgrosser:
Let's merge this as is. We can later see if an `hlet` is nicer.

PR #525, opened 2026-05-08
Add FlatSymbolRefAttr and BoolAttr; fix GEP verifier operand count

regehr:
@math-fehr: this is Claude's attempt to follow your attribute playbook. I read the code and it seems plausible, but I'm very much out of my depth here. the result of this patch is that VeIR can parse in a sudodu solver that I wrote, compiled to LLVM. however, the resulting MLIR has a substantial number of "unregistered" in it, so there's definitely more work that needs to be done here.

If this PR looks decent on its own, then I can work on those unregistereds after we merge this, but I'm also happy to close this and do some more work if this doesn't make sense to merge right now.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -107,6 +107,21 @@ structure UnregisteredAttr where
 deriving Inhabited, Repr, DecidableEq, Hashable
 
 /--
+  A flat symbol reference attribute, e.g. `@foo` or `@"my.func"`.
+  The value stores the raw text including the `@` prefix.
+-/
+structure FlatSymbolRefAttr where
+  value : String
+deriving Inhabited, Repr, DecidableEq, Hashable
+
+/--
+  A boolean attribute, written as `true` or `false`.
+-/
+structure BoolAttr where
+  value : Bool
+deriving Inhabited, Repr, DecidableEq, Hashable
+
+/--
   The `!mod_arith.int` type from HEIR's modarith dialect.
   The modulus type annotation is optional in syntax.
 -/
@@ -202,6 +217,10 @@ inductive Attribute
 | functionType (type : FunctionType)
 /-- An attribute from an unknown dialect. -/
 | unregisteredAttr (attr : UnregisteredAttr)
+/-- A flat symbol reference, e.g. `@foo` or `@"my.func"`. -/
+| flatSymbolRefAttr (attr : FlatSymbolRefAttr)
+/-- A boolean attribute, written as `true` or `false`. -/
+| boolAttr (attr : BoolAttr)
 /-- HEIR modarith type -/
 | modArithType (type : ModArithType)
 /-- LLVM pointer type -/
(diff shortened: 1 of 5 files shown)


Veir/IR/Attribute.lean
```diff
@@ -106,6 +106,21 @@ structure UnregisteredAttr where
   isType : Bool
 deriving Inhabited, Repr, DecidableEq, Hashable
 
+/--
+  A flat symbol reference attribute, e.g. `@foo` or `@"my.func"`.
```
tobiasgrosser, line 110:
```suggestion
  A flat symbol reference attribute, e.g., `@foo` or `@"my.func"`.
```

Veir/IR/Attribute.lean
```diff
@@ -202,6 +217,10 @@ inductive Attribute
 | functionType (type : FunctionType)
 /-- An attribute from an unknown dialect. -/
 | unregisteredAttr (attr : UnregisteredAttr)
+/-- A flat symbol reference, e.g. `@foo` or `@"my.func"`. -/
```
tobiasgrosser, line 220:
```suggestion
/-- A flat symbol reference, e.g., `@foo` or `@"my.func"`. -/
```

tobiasgrosser (commented):
Nice.

I am happy with the changes. However, can you:

1) Remove your Claude comment from the PR message (and pull it down into the chat)
2) Create a separate PR + test cases for the getelementptr changes?

Veir/Parser/AttrParser.lean
```diff
@@ -238,6 +238,29 @@ partial def parseOptionalDialectAttr : AttrParserM (Option Attribute) := do
...
+    return none
+
+/--
+  Parse a flat symbol reference attribute, if present.
+  Its syntax is `@ident` or `@"string"`.
+-/
+def parseOptionalFlatSymbolRefAttr : AttrParserM (Option FlatSymbolRefAttr) := do
+  let token ← peekToken
+  let .atIdent := token.kind | return none
+  let _ ← consumeToken
```
math-fehr, line 260: `parsePrefixedKeyword .atIdent` should directly parse the `@ident` syntax and returns you an `Option String`

math-fehr (commented):
I just had one comment on the parser, but otherwise everything looks good.

One issue though, is that both `BoolAttr` and `FlatSymbolRefAttr` are not actual types defined in MLIR IR. Rather, they are C++ wrapper around `IntegerType` and `SymbolRefAttr` with additional constraints (that the integer type has width 1, and that the symbol isn't nested).

I think it's fine to add `FlatSymbolRefAttr` for now, unless you want to add `SymbolRefAttr` directly, but I would at least remove `BoolAttr`, as it might cause problems later on. One example is that `arith.constant true` should correctly pass `true` as an integer attribute, and not as a bool attribute.

Making booleans work on your examples though should be fairly easy, you should just tweak a bit `parseOptionalIntegerAttr`.

regehr:
Claude's commentary on this patch follows.

-----------------------------------

Add two new attribute types needed to parse real-world LLVM MLIR:
  - FlatSymbolRefAttr: handles @ident and @"string" symbol references (e.g. callee = @printf, global_name = @".str.3")
  - BoolAttr: handles true/false boolean literals (e.g. isVolatile = false)

Both follow the standard attribute addition process across Attribute.lean, AttrParser.lean, UnitTest/AttrParser.lean, and Test/Builtin/attributes.mlir.

Also fix the llvm.getelementptr verifier, which required exactly 2 operands. GEP takes 1 base pointer plus one operand per dynamic index (rawConstantIndices entry == kDynamic). The verifier now computes the expected operand count from the properties instead of hardcoding 2.

Together these changes allow sud-opt.mlir to parse and run through the instcombine+dce pipeline.

PR #526, opened 2026-05-08
parseOptionalOp linearity fix

GZGavinZhao:
Missed by #481, required by #523.

Current `parseOptionalOp` doesn't have linearity issues but it achieves this by manually resetting the context with `setContext Inhabited.default` in the middle of modifying the context, which is confusing. In #481, the rest of the parser has migrated over to `modifyContext` helpers to achieve linearity while being easier to understand. `parseOptionalOp` was forgotten in #481, hence this PR is created to migrate `parseOptionalOp` to use `modifyContextM'`. This way the code is both consistent with the rest of the parser and also easier to read.

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -385,18 +385,15 @@ partial def parseOptionalOp (ip : Option InsertPoint) : MlirParserM (Option Oper
     throw s!"operation '{opName}' declares {inputTypes.size} operand types, but {operands.size} operands were provided"
   let operands ← operands.zip inputTypes |>.mapM (fun (operand, type) => resolveOperand operand type)
 
-  /- Set context in monad to default to preserve linearity whilst modifying -/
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
> Missed by #481, required by #523.

This PR misses context. The PR message should explain (a) if the current code base still has a linearity problem and then (b) state that the parser has now no known linearity problems. Instead of referring to #523, the PR message should refer to the previous commit which has claimed to have this addressed and state that this was forgotten and that this is a follow-up to the previous commit.

math-fehr (approved):
Thanks, all good to me!

tobiasgrosser (approved):

tobiasgrosser:
Please update the PR message. After, this is good-to-go. Thank you for the fast turnaround.

GZGavinZhao:
Updated PR description, please let me know if you have further comments/suggestions!

PR #529, opened 2026-05-08
Bitblastable `Int` type using contextual simp

georgerennie:
This is an another attempt at what bitblasting rules for `LLVM.Int` could look like (see also
@luisacicolini 's work in #457 and #490).

In this we don't actually bother creating a separate `IntBV` structure, instead we just add
`LLVM.Int.isPoison` and `LLVM.Int.getValue`. The latter returns the current value or 0 if the
int is poisoned.

For each `LLVM.Int` operation, we then create a pair of theorems, one defining `isPoison`
and the other defining `getValue`, taking `(op ...).isPoison = false` as an argument. So that
these `getValue` theorems can be used when simplifying `isPoison`, if `getValue` is used in the
definition of an `isPoison` it is nested within `(op ...).isPoison = false`. For example:

```lean4
@[llvm_toBitVec]
theorem isPoison_add {w : Nat} (x y : Int w) {nsw nuw : Bool} :
    (add x y nsw nuw).isPoison =
      decide
        (x.isPoison = true ∨ y.isPoison = true ∨
        (x.isPoison = false → y.isPoison = false →
          (nsw ∧ BitVec.saddOverflow x.getValue y.getValue) ∨
          (nuw ∧ BitVec.uaddOverflow x.getValue y.getValue))) := by
  ...

@[llvm_toBitVec]
theorem getValue_add {w : Nat} (x y : Int w) {nsw nuw : Bool} (h : (add x y nsw nuw).isPoison = false) :
    (add x y nsw nuw).getValue = x.getValue + y.getValue := by
  ...
```

These rules can be used with `simp (contextual := true) [llvm_toBitVec]`. Using these in the
bitblasting unit tests, they seem to work (every theorem gets proven) and when I looked at the result
of `bv_normalize` it seems to be able to optimize the `getValue` part as expected.

The proofs are all hacky drafts, this is more to show the structure. Another idea I had that could
be explored is explicitly making `x.getValue` take a proof `x.isPoison = false` which would ensure
that whenever `getValue` is used `x.isPoison = false` is definitely in the context.


georgerennie:
Closed in favour of #530

PR #530, opened 2026-05-08
Bitblastable `Int` type using dependently typed `getValue`

georgerennie:
Apologies for yet another draft PR with an attempt at the bitblast semantics. This one is
based on #529, but instead of `getValue` returning `0` for poison values, it instead requires
a proof that the value isn't poison (like `Option.get` vs `Option.getD`).

This means it no longer requires `(contextual := true)` in the `simp` call, although adding it can
clear up the state slightly.

This seems to work really nicely, it makes the `isPoison` theorems cleaner and results in pretty
clear small goals, e.g:

```lean4
example (e e_1 : Veir.Data.LLVM.Int 64) : add (and e e_1) (or e e_1) ⊑ add e e_1 := by
  simp (contextual := true) [llvm_toBitVec]

  -- Resulting proof state:
  -- ∀ (x : e.isPoison = false ∧ e_1.isPoison = false) (x_1 : True),
  --   (e.getValue ⋯ &&& e_1.getValue ⋯) + (e.getValue ⋯ ||| e_1.getValue ⋯) = e.getValue ⋯ + e_1.getValue ⋯
```

~~Unfortunately, where this goes wrong is that `bv_decide` does not seem to like the dependently
typed `getValue`, so if you try to call `bv_decide` on this you get:~~
```
The prover found a potentially spurious counterexample:
- It abstracted the following unsupported expressions as opaque variables: [e.getValue ⋯
 e.getValue ⋯,
 e.getValue ⋯,
 e_1.getValue ⋯,
 e.isPoison,
 e_1.getValue ⋯,
 e_1.getValue ⋯,
 e_1.isPoison]
Consider the following assignment:
e.getValue ⋯ = 18446744073709551615#64
e_1.getValue ⋯ = 18446744073709551615#64
e.getValue ⋯ = 18446744073709551615#64
e_1.getValue ⋯ = 18446744073709551615#64
e.getValue ⋯ = 18446744073709551615#64
e_1.getValue ⋯ = 9223372036854775807#64
e.isPoison = false
e_1.isPoison = false
```

~~It looks like it is abstracting each `e_1.getValue` as a separate term, whereas as they are only
dependent on a proof term (which has the same type each time) it should be possible to abstract
them as the same term. This seems like something that could be fixed in upstream `bv_decide`?~~

edit: I've fixed this by adding a low priority rule that replaces `getValue` with `getValueD` after the other rules run.


georgerennie:
Thanks to advice from @hargoniX and @alexkeizer, it seems the easiest way for now to make this work with `bv_decide` is to replace `getValue` with a non-proof carrying version `getValueD` after the other rules have run, making it possible for `bv_decide` to see the terms as equal. This seems to work well and the tactic is able to solve all of the examples from the unit tests.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,399 @@
...
+/--
+  Low priority rule to convert getValue to getValueD once no more simplification can be done.
+  This is needed because `bv_decide` cannot see different instantiations of `x.getValue proof`
+  as the same and abstracts them to separate terms.
+-/
+@[llvm_toBitVec 1]
+theorem getValue_eq_getValueD {w : Nat} (x : Int w) (h : x.isPoison = false) :
+    x.getValue h = x.getValueD := by
+  unfold getValue getValueD
+  grind
```
alexkeizer, line 39: Ah, I hadn't even thought about using priorities for this: nice trick!

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,399 @@
...
+    (constant w v).isPoison = false := rfl
+
+@[llvm_toBitVec, grind =]
+theorem isPoison_add {w : Nat} (x y : Int w) {nsw nuw : Bool} :
+    (add x y nsw nuw).isPoison =
+      if h : x.isPoison = true ∨ y.isPoison = true then
+        true
+      else
+        (nsw ∧ BitVec.saddOverflow x.getValue y.getValue) ∨
+        (nuw ∧ BitVec.uaddOverflow x.getValue y.getValue) := by
```
alexkeizer, line 86: `if ... then true` feels a bit off to me. Would `x.isPoison || y.isPoison || (nsw && ...) || (nuw && ...)` also work?

I could imagine that `bv_decide` might work just a little bit better with a plain Boolean formula instead of potentially triggering some if-then-else case-split, although it's just as likely bv_decide is smart enough to recognize `if ... then true` as a special case and it's just a matter of aestethics

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,399 @@
...
+
+
+/-! # Examples
+  The following section includes examples we are using to compare across all the different implementations.
+-/
+
+/-- We can prove some basic properties about LLVM operations. -/
+example (x y : Int 64) : (add x y (nsw := true)) ⊑ (add y x) := by
+  simp [llvm_toBitVec]
+  bv_decide
```
alexkeizer, line 393: you could wrap this up into a simple macro:

```
macro "llvm_decide" : tactic => `(tactic| simp only [llvm_bitvec] <;> bv_decide)
```

Caveat: I haven't tested the above code, so not sure if it's all syntactically right

Veir/Data/LLVM/Int/Bitblast.lean
georgerennie: The issue with that is that you need `x.isPoison = false` in `x.getValue` (its hidden in a grind autoparam so you don't see it explicitly but its happening). you could write it like `x.isPoison || y.isPoison || (!x.isPoison -> !y.isPoison -> ...)` but i think that can be kinda unwieldy to write too.

georgerennie: Having said that i'm also not 100% on what best practice for mixing bool/prop operators is and where exactly calls to `decide` etc should be inserted.

alexkeizer: Ah, right, I'd completely missed that dependency; it might be worth to add a comment explaining this, especially since it is hidden in that autoParam.

You can sometimes eliminate the `if-then-else` still, using an existential as a kind of "dependent and", but here you actually do the extra prop in the else-branch, so there is not a super clean mapping to an existential either. With that in mind, the if-then-else is probably the cleanest, yeah.

PR #532, opened 2026-05-08
support GEPs with more than one index

regehr:
we were not supporting GEPs that do multi-dimensional indexing, this adds support for that

### DIFF
--- a/Test/LLVM/getelementptr.mlir
+++ b/Test/LLVM/getelementptr.mlir
@@ -1,0 +1,41 @@
+// RUN: veir-opt %s | filecheck %s
+//
+// Tests for llvm.getelementptr with 1, 2, 3, and 4 dynamic indices.
+// The operand count must equal 1 (base pointer) + number of kDynamic
+// (-2147483648) entries in rawConstantIndices.
+
+"builtin.module"() ({
+  "llvm.func"() ({
+    ^bb0(%ptr: !llvm.ptr, %i: i64, %j: i64, %k: i64, %l: i64):
+
+      // 1 dynamic index: ptr[i]
+      %g1 = "llvm.getelementptr"(%ptr, %i) <{elem_type = i8, rawConstantIndices = array<i32: -2147483648>}> : (!llvm.ptr, i64) -> !llvm.ptr
+
+      // 2 dynamic indices: ptr[i][j]
+      %g2 = "llvm.getelementptr"(%ptr, %i, %j) <{elem_type = !llvm.array<10 x i8>, rawConstantIndices = array<i32: -2147483648, -2147483648>}> : (!llvm.ptr, i64, i64) -> !llvm.ptr
+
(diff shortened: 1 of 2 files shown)


regehr:
claude's explanation follows. I think it looks plausible. and, in fact, this patch is needed to support running VeIR on a sudoku solver I wrote, which of course indexes into a 2D array

----------------------------------------

## What this patch does and why                                                                                       
                                         
  ### Background: GEP in LLVM IR                                                                                        
   
  `getelementptr` (GEP) is LLVM's instruction for pointer arithmetic into aggregates. It takes a base pointer plus a    
  sequence of indices that navigate a type hierarchy — one index per level of nesting. Crucially, any individual index
  can be either a compile-time constant or a runtime SSA value.                                                         
                                                            
  ### How MLIR's LLVM dialect encodes GEP

  MLIR segregates constant and dynamic indices. The `rawConstantIndices` property is a dense `i32` array covering the   
  full index sequence, where:
                                                                                                                        
  - A non-sentinel value encodes a compile-time constant index inline.
  - The sentinel `kDynamic` (`-2147483648`, i.e. `INT32_MIN`) means the index is a runtime value drawn from the operand
  list.                                                                                                                 
   
  The total operand count is therefore:                                                                                 
                                                            
  1 (base pointer) + number of kDynamic entries in rawConstantIndices

  ### Why real C code produces multi-index GEPs

  The `sud` struct has a 2D board:                                                                                      
   
  ```c                                                                                                                  
  struct sud { cell board[9][9]; };                         
```
  Accessing s->board[i][j] navigates two array levels. LLVM lowers this to a GEP with two variable indices, which in    
  MLIR becomes:

```mlir
  "llvm.getelementptr"(%board, %i, %j)                      
    <{elem_type = !llvm.array<9 x i8>,
      rawConstantIndices = array<i32: -2147483648, -2147483648>}>
    : (!llvm.ptr, i64, i64) -> !llvm.ptr                                                                                
```
   
  This has 3 operands: base pointer + two dynamic indices. Multi-index GEPs arise whenever C code accesses nested       
  arrays, multi-dimensional arrays, or structs indexed at more than one level — all routine in real programs.
                                                                                                                        
  What Veir's old verifier assumed                          

  The verifier required getNumOperands == 2 for every llvm.getelementptr. This covers only the simplest case — one      
  variable index into a flat array or struct. It rejects any GEP with two or more dynamic indices, which is common in
  real LLVM-IR-derived MLIR.                                                                                            
                                                            
  What this patch does

  Instead of hardcoding ≠ 2, the verifier reads rawConstantIndices from the operation's properties, counts the entries  
  equal to kDynamic, and checks:
                                                                                                                        
  getNumOperands == 1 + dynamicCount                        

  This correctly validates GEPs with any number of dynamic indices while still rejecting operand-count mismatches.

tobiasgrosser (approved):
LGTM. Thank you, @regehr!

PR #533, opened 2026-05-08
Add FlatSymbolRefAttr

regehr:
ok this is a revised version of #525 that only adds support for FlatSymbolRefAttr! I believe I've addressed all feedback!


tobiasgrosser (approved):
This LGTM from my side. We should be able to get this in after @math-fehr's feedback.

math-fehr (approved):
Thanks, all good to me!

PR #535, opened 2026-05-09
Adaptive sampling and stddev reporting in benchmark script

snarang181:
Replace fixed-iteration runs with adaptive sampling: warmup, minimum samples, then extra samples until per-phase CV drops below 5% or the sample/time caps (30 / 180s) are hit. Emit stddev as the `range` field so github-action-benchmark renders error bars on the chart.

### DIFF
--- a/.github/scripts/run_benchmarks.sh
+++ b/.github/scripts/run_benchmarks.sh
@@ -42,6 +42,8 @@ BENCHMARKS=(
 
 OUTPUT_FILE="${BENCHMARK_OUTPUT:-benchmark-results.json}"
 
+lake build run-benchmarks >&2
+
 median() {
   local sorted
   sorted=($(printf '%s\n' "$@" | sort -g))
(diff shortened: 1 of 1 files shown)


.github/scripts/run_benchmarks.sh
```diff
@@ -41,27 +54,83 @@ median() {
...
+  awk -v m="$m" -v sd="$sd" 'BEGIN { if (m+0==0) print "0"; else printf "%.4f", (sd/m)*100 }'
+}
+
+flt_lt() {
+  awk -v a="$1" -v b="$2" 'BEGIN { exit !(a+0 < b+0) }'
+}
+
+run_one() {
+  local bench="$1" out cs rs
+  out=$(lake exe run-benchmarks "$bench" "$COUNT" "$PC" 2>&1) || return 1
```
GZGavinZhao: I think it's better to use `lake env run-benchmarks` instead of `lake exe` because `lake exe` checks whether a rebuild is needed before running the command, which means that the time your measuring includes `lake`'s whether-I-need-to-rebuild-anything check. On the other hand, with `lake env`,` lake` just runs whatever executable you pass to it directly without checking for file changes for rebuilds. Note that this does mean you must manually `lake build run-benchmarks` before running `lake env`.

tobiasgrosser: Yes, i agree.

snarang181: Done

GZGavinZhao (approved):
Looks fine from my side!

snarang181:
@tobiasgrosser, can we land?

GZGavinZhao:
Submitted #582 solely for testing since this PR is submitted by a fork and GH bot can't comment on it. The new benchmark comment would look like [this](https://github.com/opencompl/veir/pull/582#pullrequestreview-4302485290), which looks fine to me.

snarang181:
> Submitted #582 solely for testing since this PR is submitted by a fork and GH bot can't comment on it. The new benchmark comment would look like [this](https://github.com/opencompl/veir/pull/582#pullrequestreview-4302485290), which looks fine to me.

Thank you for testing @GZGavinZhao!

tobiasgrosser:
Sorry for missing this. Yes, I think this can go in. Thank you!

PR #546, opened 2026-05-10
Add `WfRewriter.setBlockArguments`

math-fehr:


ineol (approved):
I'm curious, when would it be used? at creation time?

luisacicolini (approved):

GZGavinZhao:
> I'm curious, when would it be used? at creation time?

It's going to be used in the parser. `noUses` becomes trivial when you have no arguments to start with.

math-fehr:
Yeah, this is almost only for the parser, though it might become useful somewhere else one day (but probably not any time soon)

PR #547, opened 2026-05-11
Allow to set argument types in `createBlock`

math-fehr:
`createBlock` only allowed to create block with no arguments.


ineol (approved):
Despite the stacking, this PR does not use `setArguments` right?

luisacicolini (approved):

math-fehr:
Yeah sorry, this could have been stacked differently, we don't use `setArguments` here

PR #550, opened 2026-05-11
Bitblasting `LLVM.Int`

luisacicolini:
We introduce the minimal infrastructure to make `LLVM.Int` bitblastable. This includes:
* an `isPoison` method that returns `true` if the value is poison 
* a `getValueD` method that returns a bitvector with the `LLVM.Int`'s value if it is not poison, or a bitvector `0#w` otherwise 
* a `getValue` method that returns a bitvector with the `LLVM.Int`'s value, given a non-poison hypothesis. 

We introduce lemmas to rewrite the equality and refinement relations using these lemmas, and an additional `getValue_eq_getValueD` lemma that rewrites `getValue` to `getValueD`. We show how this infrastructure enables bitblasting for `add`, `sub`, and `constant` operations, already solving some of the lemmas in `Bitblasting`. In a subsequent PR we will introduce the same infra for all the `LLVM.Int` operations, and we will pack the sequence 
```
simp [llvm_toBitVec]
simp only [getValue_eq_getValueD]
bv_decide
```
in a tactic that shall solve all the goals in the `Bitblasting` test.

Co-authored-by: @georgerennie @alexkeizer


Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,77 @@
+module
```
georgerennie, line 1: Should this all be wrapped in a `public section`? the definitions are all private currently I think.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,77 @@
...
+
+/-- The bitvector value of a non-poison `LLVM.Int` via `getValue` is equal to the default one,
+  obtained by `getValueD`.
+
+  We set the priority of the lemma to `1` (low priority), such that it is only applied when no
+  more simplifications are available. This ultimately allows `bv_decide` to reason about different
+  instantiations of `getValue` without abstracting them separately.
+-/
+@[llvm_toBitVec 1]
+theorem getValue_eq_getValueD {w : Nat} (x : Int w) (h : x.isPoison = false) :
```
georgerennie: I picked a priority slightly at random with this, but I think its slightly more ergonomic lean/clearer to write `llvm_toBitVec low` instead of `1` (which is a very low priority, i think default is 1000 and low is 100).

There is also a question of whether we even want to run this rule always or not - if you are simping to use `bv_decide` definitely but if you are simping to do the proof manually maybe its useful to you to still be working with `getValue` and the extra information that gives.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,77 @@
...
+    x.getValue h = x.getValueD := by
+  simp [getValue, getValueD]
+  cases x <;> grind
+
+/-- Two elements `a b : LLVM.Int` are equal if and only if they return the same `isPoison` and,
+  the same `getValue` in case they are *not* poison. -/
+@[llvm_toBitVec]
+theorem eq_iff {w : Nat} (a b : Int w) :
```
georgerennie, line 45: At the moment this is only used by the simp rules, I wonder if we want a version of this that grind can use too (a theorem suitable for `grind ext` maybe?)?

georgerennie (commented):
Looks pretty good, just a few comments that stuck out to me to think about.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,77 @@
...
+  a = b ↔
+    (a.isPoison = b.isPoison) ∧
+    ((_ : a.isPoison = false) → (_ : b.isPoison = false) → a.getValue = b.getValue) := by
+  simp [isPoison, getValue, llvm_toBitVec]
```
tobiasgrosser, line 49: Does this need to be `simp only` before `grind`?

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,77 @@
...
+/-- A `poison` element is poison. -/
+@[llvm_toBitVec, grind =]
+theorem isPoison_of_poison {w : Nat} :
+    poison.isPoison (w := w) = true := rfl
+
+/-- An element `b : LLVM.Int` refines an element `a : LLVM.Int` if either `a` is a poison value
+  (in which case, any concrete or poison value refines it) or if `a` is not a poison value,
+  `b` is not a poison value, and their concrete bitvector values are the same. -/
+@[llvm_toBitVec, grind =]
+theorem isRefinedBy_iff {w : Nat} (a b : Int w) :
```
tobiasgrosser, line 80: Does this need to be `simp only` before `grind`?

tobiasgrosser (commented):
This looks good.

The PR message could benefit from a refinement:

1) While it is clearly related to bitblasting, the PR message does not make clear how all the features it describe relate to bitblasting. You probably want to explain that for bitblasting we will eliminate the `IntBv` by using `eq_iff` to split the `Int` into a `Poison` and `Val` and to then push these outwards such that `bv_decide` can treat them as atoms that it can abstract on.

2) The PR talks abstractly about `will be bitblastable` in the future, but does not say what is achieved until now. Can we even bitblast something at this point?

3) There is also no testing of this feature?

I wonder if we should add propagation support for getValue and getPoison for one operation, e.g., `add` plus associated test cases. Then you can say, "This PR adds bitblasting support for `add`, the remain operations will be added incrementally in future PRs."

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,77 @@
...
+
+open Veir.Data.LLVM
+
+namespace Veir.Data.LLVM.Int
+
+/-- Return true if the LLVM.Int `x` is poison. -/
+def isPoison {w : Nat} (x : Int w) : Bool :=
+  match x with
+  | .poison => true
+  | .val _ => false
```
alexkeizer, line 18: Minor nit, but you can drop the `match` here (and the next three definitions also)

```suggestion
/-- Return true if the LLVM.Int `x` is poison. -/
def isPoison {w : Nat} : (x : Int w) -> Bool
  | .poison => true
  | .val _ => false
```

Veir/Data/LLVM/Int/Bitblast.lean
alexkeizer: Perhaps we should just have a `llvm_decide` macro, like I suggested on the other PR, which does 
```
simp only [llvm_toBitVec] 
simp only [getValue_eq_getValueD]
bv_decide
```

That way, if we ever need to debug the simp-set, you can do just `simp only [llvm_toBitVec]`.

Veir/Data/LLVM/Int/Bitblast.lean
alexkeizer: `simp; grind` is generally accepted as a terminal-ish use of simp

Veir/Data/LLVM/Int/Bitblast.lean
tobiasgrosser: I like the approach suggested by Alex.

Veir/Data/LLVM/Int/Bitblast.lean
tobiasgrosser: Perfect. Thank you. Then let's follow this convention.

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: I would not expect other modules to use these defs, though? Do you think they could be imported somewhere else?

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: sounds good to me, removec the priority, will implement the macro in the next PR

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: I think it would be good to have one, I am not super familiar with `grind ext`, do you have a sense how this such a lemma would look like?

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: done

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: ah nvm I see what you mean. fixed, thanks!

Veir/Data/LLVM/Int/Bitblast.lean
georgerennie: I had one like this in [valaig](https://github.com/opencompl/valaig/blob/68f5da9d5fc253575402f00f1c88ad1d48d72189/Valaig/Aig/Refs.lean#L184-L187).

I think you would want something like
```lean4
@[ext, grind ext]
theorem ext {w : Nat} {a b : Int w} (poisonEq : a.isPoison = b.isPoison) (valEq : ((_ : a.isPoison = false) → (_ : b.isPoison = false) → a.getValue = b.getValue)) :
    a = b := by
  ...
```

luisacicolini: added!

georgerennie (approved):
Looks good, thanks!

UnitTest/Bitblasting/Bitblasting.lean
```diff
@@ -1,9 +1,11 @@
 import Veir.Data.LLVM.Int.Basic
+import Veir.Data.LLVM.Int.Bitblast
 import Veir.Data.LLVM.Int.Simp
 import Veir.Data.Refinement
 
 open Veir.Data.LLVM.Int
 
+
```
tobiasgrosser: This is an unrelated change.

UnitTest/Bitblasting/Bitblasting.lean
```diff
@@ -35,16 +37,22 @@ example
 
 example
     (e e_1 : Veir.Data.LLVM.Int 64) : add (sub (constant 64 0) e) e_1 ⊑ sub e_1 e := by
-  llvm_bv_decide
+  simp [llvm_toBitVec]
+  simp only [getValue_eq_getValueD]
+  bv_decide
```
tobiasgrosser: Why this change? Should we not add this very pattern into `llvm_bv_decide`?

The same for the subsequent three changes.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,134 @@
+module
+
+public import Veir.Data.LLVM.Int.Basic
+public import Veir.Data.Refinement
+public import Veir.Data.Refinement
+public import Veir.Data.LLVM.Int.Simp
+import all Veir.Data.LLVM.Int.Basic
+import all Veir.Data.Refinement
```
tobiasgrosser, line 7: Please minimize the imports. Currently `public import Veir.Data.Refinement` is added twice / three times. Also check that other imports are all needed and that the public is needed for all.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,134 @@
...
+  match x with
+  | .poison => 0#w
+  | .val v => v
+
+/-- The bitvector value of a non-poison `LLVM.Int` via `getValue` is equal to the default one,
+  obtained by `getValueD`.
+
+  This lemma is intended to be applied when no more simplifications are available, such that
+  ultimately allows `bv_decide` to reason about different
+  instantiations of `getValue` without abstracting them separately.
```
tobiasgrosser: Please double-check grammar and layout of this comment.

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -0,0 +1,134 @@
...
+@[llvm_toBitVec, grind =]
+theorem isPoison_sub {w : Nat} (x y : Int w) {nsw nuw : Bool} :
+    (sub x y nsw nuw).isPoison =
+      if h : x.isPoison = true ∨ y.isPoison = true then true
+      else
+        (nsw ∧ BitVec.ssubOverflow x.getValue y.getValue) ∨
+        (nuw ∧ BitVec.usubOverflow x.getValue y.getValue) := by
+  simp only [isPoison, sub, Id.run, pure_bind, getValue, Bool.decide_or, Bool.decide_and,
+    Bool.decide_eq_true]
+  simp only [pure]
```
tobiasgrosser: Why do you not merge these two simps? Also, does this need to be `simp only`?

tobiasgrosser (commented):
Just a couple of comments, but I would like to have a final look.

Also, the tactic`llvm_bv_decide` already exists. I am curious why you do not update it here already.

UnitTest/Bitblasting/Bitblasting.lean
luisacicolini: AFAIU from @alexkeizer's comment this is the alternative to specifying the lemma's priority. Either we specify it with low priority, or we use it separately once everything else has been simped. AFAIU, if we add `getValue_eq_getValueD` to the simp set everything will be unfolded to `getValueD`, which we want to avoid (?)

UnitTest/Bitblasting/Bitblasting.lean
luisacicolini: removed

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: done - AFAIU we need botht eh `public import` and the `import all` because of module dependencies (i.e.: without both of them various things break throughout the file), but I'm not super confident with modules so maybe I'm missing something there.

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: fixed

Veir/Data/LLVM/Int/Bitblast.lean
luisacicolini: removed the `simp only`. Unfortunately, I could not find a way to merge the two `simp`s without changing the proof structure (merging them breaks the following `grind` call).

UnitTest/Bitblasting/Bitblasting.lean
tobiasgrosser:
```suggestion
  `(tactic| (try simp only [llvm_toBitVec]; try simp only [getValue_eq_getValueD]; try bv_decide); all_goals sorry)
```

We should add the pattern to the tactic above, no?

UnitTest/Bitblasting/Bitblasting.lean
tobiasgrosser: See my suggested change to the tactic above. We indeed want this sequence of three tactic calls. However, we probably want it hidden within the `llvm_bv_decide` tactic.

Veir/Data/LLVM/Int/Bitblast.lean
tobiasgrosser: Thank you.

UnitTest/Bitblasting/Bitblasting.lean
luisacicolini: oh I see, I was planning to replace the tactic once all the lemmas are solved - in this PR I thought it was nice to show how many lemmas we can already solve with the minimal add/sub/const lemmas, to explain why they are useful. Happy to replace immediately.

UnitTest/Bitblasting/Bitblasting.lean
luisacicolini: answered above

tobiasgrosser (approved):

UnitTest/Bitblasting/Bitblasting.lean
```diff
@@ -14,7 +15,7 @@ open Veir.Data.LLVM.Int
 
 /-- We introduce a tactic to automatically prove all the lemmas. -/
 macro "llvm_bv_decide" : tactic =>
-  `(tactic| ((try simp [llvm_toBitVec]; try bv_decide); all_goals sorry))
+  `(tactic| ((try simp only [llvm_toBitVec]; try simp only [getValue_eq_getValueD]; try bv_decide); all_goals sorry))
```
alexkeizer, line 18:
```suggestion
  `(tactic| (
      simp -failIfUnchanged only [llvm_toBitVec]; 
      simp -failIfUnchanged only [getValue_eq_getValueD];
      bv_decide))
```

Two things: 
* I assume you use `try` to prevent the tactic failing when there's no applicable rewrite; if that is indeed the only reason for using it,  `-failIfUnchanged` has the same effect and has the benefit of not swallowing other errors (making it easier to debug in the future, if anything else goes wrong)
* similarly, I would not put `all_goals sorry` in this particular tactic. We did this in LeanMLIR with the Alive file, which then meant we had to make that cursed CI job that tried to count how many sorry-lints we got to detect when we accidentally stopped solving a goal that we solved before; I would suggest not repeating our mistake there. Instead, if you add the sorry to each individual test, then you can gradually remove the sorries when expanding the capabilities, and you actually get CI to catch any regressions out of the box.

UnitTest/Bitblasting/Bitblasting.lean
alexkeizer, line 30: E.g, here I would put `try llvm_bv_decide; sorry` if the goal is not yet solved by the tactic. 

This will throw an error at the sorry once the tactic actually becomes powerful enough to solve the goal, at which point you can replace the tactic with `llvm_bv_decide` (and then if you have a regression where the tactic stops working for this goal, you will get an error again).

UnitTest/Bitblasting/Bitblasting.lean
luisacicolini: thank you! this was ultimately solved in #574. Afaiu, we still use `try` also because we don't want to run the tactic if there is no goal left to solve (some lemmas are solved by simp only).

PR #552, opened 2026-05-11
Add support for multiple result syntax to parser

georgerennie:
This PR adds support to the parser for the `%name:numResults = ...` and `%name#index` syntax for declaring and accessing multiple results under the same SSA name.

This is implemented by extending the value tracking in the parser from `Std.HashMap ByteArray ValuePtr` to `Std.HashMap ByteArray (Array ValuePtr)`. `%name` as an argument is treated as sugar for `%name#0` but we use an `Option Nat` to store this index in `UnresolvedOperation` for more precise error messages.

### DIFF
--- a/UnitTest/MlirParser.lean
+++ b/UnitTest/MlirParser.lean
@@ -188,7 +188,7 @@ def testParseOp (s : String) : IO Unit :=
 }) : () -> ()"
 
 /--
-  error: invalid result number 2 for %a
+  error: invalid result index 2 for %a
 -/
 #guard_msgs in
 #eval! testParseOp "\"builtin.module\"() ({
--- a/UnitTest/MlirParser.lean
+++ b/UnitTest/MlirParser.lean
@@ -162,3 +162,45 @@ def testParseOp (s : String) : IO Unit :=
     \"test.test\"() [^bb0] : () -> ()
   }) : () -> ()
 }) : () -> ()"
+
+/--
+  info: "builtin.module"() ({
+  ^4():
+    %5_0, %5_1 = "test.test"() : () -> (i32, i64)
+}) : () -> ()
+-/
+#guard_msgs in
+#eval! testParseOp "\"builtin.module\"() ({
+  %x:2 = \"test.test\"() : () -> (i32, i64)
+}) : () -> ()"
+
+/--
+  info: "builtin.module"() ({
+  ^4():
+    %5_0, %5_1, %5_2 = "test.test"() : () -> (i10, i32, i64)
+    %6 = "test.test"(%5_2, %5_0) : (i64, i10) -> i1
+}) : () -> ()
+-/
+#guard_msgs in
+#eval! testParseOp "\"builtin.module\"() ({
+  %a, %x:2 = \"test.test\"() : () -> (i10, i32, i64)
+  %b = \"test.test\"(%x#1, %a#0) : (i64, i10) -> i1
+}) : () -> ()"
+
+/--
+  error: use of undefined value %a#2
+-/
+#guard_msgs in
+#eval! testParseOp "\"builtin.module\"() ({
+  %a:2 = \"test.test\"() : () -> (i32, i64)
+  %b = \"test.test\"(%a#2) : (i32) -> i1
+}) : () -> ()"
+
+/--
+  error: type mismatch for value %a#1: expected i32, got i64
+-/
+#guard_msgs in
+#eval! testParseOp "\"builtin.module\"() ({
+  %a:2 = \"test.test\"() : () -> (i32, i64)
+  %b = \"test.test\"(%a#1) : (i32) -> i1
+}) : () -> ()"
--- a/UnitTest/MlirParser.lean
+++ b/UnitTest/MlirParser.lean
@@ -188,7 +188,7 @@ def testParseOp (s : String) : IO Unit :=
 }) : () -> ()"
 
 /--
-  error: use of undefined value %a#2
+  error: invalid result number 2 for %a
 -/
 #guard_msgs in
 #eval! testParseOp "\"builtin.module\"() ({
@@ -203,4 +203,20 @@ def testParseOp (s : String) : IO Unit :=
 #eval! testParseOp "\"builtin.module\"() ({
   %a:2 = \"test.test\"() : () -> (i32, i64)
   %b = \"test.test\"(%a#1) : (i32) -> i1
+}) : () -> ()"
+
+/--
+  error: operation 'test.test' declares 4 result types, but 3 result values were provided
+-/
+#guard_msgs in
+#eval! testParseOp "\"builtin.module\"() ({
+  %a, %b:2 = \"test.test\"() : () -> (i1, i2, i3, i4)
+}) : () -> ()"
+
+/--
+  error: operation 'test.test' declares 2 result types, but 3 result values were provided
+-/
+#guard_msgs in
+#eval! testParseOp "\"builtin.module\"() ({
+  %a:2, %b = \"test.test\"() : () -> (i1, i2)
 }) : () -> ()"
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -217,12 +217,12 @@ def parseOpResults : MlirParserM (Array (ByteArray × Nat)) := do
   Once the operation type is known, `resolveOperand` can be used to create an SSA value and
   check that the type matches with previous uses.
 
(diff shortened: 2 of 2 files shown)


Veir/Printer.lean
```diff
@@ -41,22 +41,16 @@ def printValue (ctx : IRContext OpCode) (value : ValuePtr) : IO Unit := do
     if opStruct.results.size = 1 then
```
math-fehr, line 41: Can you actually revert the changes in the printer?
Most of the time, mlir will print the values one by ones, and that's usually easier to read.
I'm not even sure the cases for which it prints them as a pack, but in general that's not something you want to have

UnitTest/MlirParser.lean
```diff
@@ -162,3 +162,45 @@ def testParseOp (s : String) : IO Unit :=
     \"test.test\"() [^bb0] : () -> ()
   }) : () -> ()
 }) : () -> ()"
+
```
math-fehr, line 165: Can you add tests for the following cases:
* If we define `%a`, `%a#0` should still not exist, as `%a` is not a packed value
* If we have 4 results, but only returning `%a, %b:2`, there should be an error. Same for 2 results.

Veir/Parser/MlirParser.lean
```diff
@@ -17,11 +17,30 @@ namespace Veir.Parser
...
+    match name.index with
+    | 0 => s!"%{string}"
+    | n => s!"%{string}#{n}"
+
 structure MlirParserState where
   /-- The current IR context. -/
   ctx : WfIRContext OpCode
   /-- The values that have been defined at that point in the parser. -/
-  values : Std.HashMap ByteArray ValuePtr
+  values : Std.HashMap IndexedName ValuePtr
```
math-fehr: I think instead of having this data structure, we should have something like `values : Std.HashMap ByteArray (Array ValuePtr)`, so then we can get better errors when we access a value out of bounds?

Veir/Parser/MlirParser.lean
```diff
@@ -175,20 +194,42 @@ def defineBlockUse (name : ByteArray) : MlirParserM BlockPtr := do
...
   let slice := { nameToken.slice with start := nameToken.slice.start + 1 } -- skip % character
-  return slice.of (← getInput)
+  let name : ByteArray := slice.of (← getInput)
+
+  -- If there is only one result defined, push it and return
+  let some _ ← parseOptionalToken .colon
+    | return results.push name
+
+  -- Multiple results are defined, so push each with a separate index
+  let count := (← parseInteger false false).toNat
```
math-fehr, line 198: Can you check that the integer is not `0`?

Veir/Parser/MlirParser.lean
```diff
@@ -199,14 +240,25 @@ def parseOpResults : MlirParserM (Array ByteArray) := do
...
+  This has the syntax `%name` or `%name#index`.
 -/
 def parseOperand : MlirParserM UnresolvedOperand := do
   let nameToken ← parseToken .percentIdent "operand expected"
-  return UnresolvedOperand.mk ({ nameToken.slice with start := nameToken.slice.start + 1 }.of (← getInput))
+  let name : ByteArray := { nameToken.slice with start := nameToken.slice.start + 1 }.of (← getInput)
+
+  -- If no index is specified, return with the default index
+  let some index ← parseOptionalToken .hashIdent
+    | return UnresolvedOperand.mk name
```
math-fehr: I forgot how this works, but if I recall, if no index is specified, this should return the pack itself. So if you define `%x:3`, you should return 3 values here.

math-fehr (commented):
I had some comments because I think there are some differences in features compared to MLIR/xDSL here, but otherwise thanks for this!

Veir/Printer.lean
georgerennie: Sure i'll revert. So it seems `mlir` generally tries to name the multiple return values, e.g.:
```
%low, %high = arith.mulsi_extended %c13_i32, %c13_i32 : i32
%low_0, %high_1 = arith.mului_extended %c13_i32, %c13_i32 : i32
```

It seems when it is unable to do that it reverts to the pack syntax. either way, at the moment veir produces
```
%25_0, %25_1 = "arith.mulsi_extended"(%5, %5) : (i32, i32) -> (i32, i32)
%26_0, %26_1 = "arith.mului_extended"(%5, %5) : (i32, i32) -> (i32, i32)
```

which is invalid MLIR. If you try parsing it with either veir or mlir you get
```
Error: Error parsing operation: '=' expected after operation results
```

Perhaps a solution without adding a mechanism to get names for op results for now could be to add a letter before the numbers in the name?

Veir/Parser/MlirParser.lean
georgerennie: Sure, I can switch to this style

Veir/Parser/MlirParser.lean
georgerennie: MLIR seems to treat no index as equivalent to index 0:
```
"builtin.module"() ({
^4():
  %5 = "arith.constant"() <{ "value" = 13 : i32 }> : () -> i32
  %6:2 = "arith.mului_extended"(%5, %5) : (i32, i32) -> (i32, i32)
  %7 = "arith.addi"(%6, %6) : (i32, i32) -> i32
}) : () -> ()
```

gets parsed successfully and results in
```
module {
  %c13_i32 = arith.constant 13 : i32
  %low, %high = arith.mului_extended %c13_i32, %c13_i32 : i32
  %0 = arith.addi %low, %low : i32
}
```

UnitTest/MlirParser.lean
georgerennie: The former case is allowed by MLIR and there is a test covering it:
```
"builtin.module"() ({
^4():
  %5 = "arith.constant"() <{ "value" = 13 : i32 }> : () -> i32
  %7 = "arith.addi"(%5#0, %5#0) : (i32, i32) -> i32
}) : () -> ()
```
results in
```
module {
  %c13_i32 = arith.constant 13 : i32
  %0 = arith.addi %c13_i32, %c13_i32 : i32
}
```

I will add tests for the other cases.

georgerennie: Added tests with the exception of `%a#0` for `%a = op ...` (this is already tested and doesn't throw an error, as described in above comment)

tobiasgrosser:
@GZGavinZhao, can you also do a review?

Veir/Parser/MlirParser.lean
```diff
@@ -198,16 +216,50 @@ def parseOpResults : MlirParserM (Array ByteArray) := do
   This is used during parsing to allow parsing operands before their types.
   Once the operation type is known, `resolveOperand` can be used to create an SSA value and
   check that the type matches with previous uses.
+
+  `resultNumber` is used for the `%name#resultNumber` syntax to refer to an indexed result
+  when multiple are defined for the same value.
 -/
 structure UnresolvedOperand where
   name : ByteArray
+  resultNumber : Option Nat
```
math-fehr: Maybe renaming this to `index`? `resultNumber` feels quite weird to me

math-fehr (approved):
Thanks a lot! I just added one nit comment about the naming of a field, but otherwise that's good to me!

GZGavinZhao (approved):
Parser code looks reasonable. No performance impact after benchmarking (vs e08626db).
```
> hyperfine -w 2 "./old-veir-opt sqlite3-stripped.mlir" "./new-veir-opt sqlite3-stripped.mlir"
Benchmark 1: ./old-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.591 s ±  0.020 s    [User: 2.568 s, System: 0.018 s]
  Range (min … max):    2.563 s …  2.622 s    10 runs

Benchmark 2: ./new-veir-opt sqlite3-stripped.mlir
  Time (mean ± σ):      2.610 s ±  0.026 s    [User: 2.587 s, System: 0.018 s]
  Range (min … max):    2.568 s …  2.639 s    10 runs

Summary
  ./old-veir-opt sqlite3-stripped.mlir ran
    1.01 ± 0.01 times faster than ./new-veir-opt sqlite3-stripped.mlir
```

Veir/Parser/MlirParser.lean
georgerennie: Yep makes sense, I had copied it out of the MLIR implementation but wasn't really using it right

tobiasgrosser (approved):

tobiasgrosser:
Thank you, @georgerennie!

PR #555, opened 2026-05-11
refactor(EDyadic): enforce d ≠ 0 invariant in nonzeroFinite [1/x]

bollu:
This PR makes `EDyadic.nonzeroFinite` carry a proof `hne : d ≠ 0`, lifting the no-zero invariant to the type level. This enforces that signed zeros live in `zero`, `nonzeroFinite` can no longer wrap `Dyadic.zero`. This makes many of the definitions in rounding be way clearer, as they can now manipulate a `Dyadic` while ruling out the zero case statically.

### DIFF
--- a/Veir/Data/FP/EDyadic/Basic.lean
+++ b/Veir/Data/FP/EDyadic/Basic.lean
@@ -14,13 +14,10 @@ inductive EDyadic where
   /-- A zero with a sign bit. -/
   | zero (sign : Bool)
   /-- A nonzero finite number in dyadic notation.
-  While dyadics can be zero, we use the `zero` constructor to
-  represent zero values (so the sign is represent),
-  and we use `nonzeroFinite` to represent nonzero values.
-  Thus, there is an implicit invariant that the `nonzeroFinite`
-  constructor should only be used with nonzero dyadics.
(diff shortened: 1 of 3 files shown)


Veir/Data/FP/EDyadic/Basic.lean
```diff
@@ -14,26 +14,37 @@ inductive EDyadic where
   /-- A zero with a sign bit. -/
   | zero (sign : Bool)
   /-- A nonzero finite number in dyadic notation.
-  While dyadics can be zero, we use the `zero` constructor to
-  represent zero values (so the sign is represent),
-  and we use `nonzeroFinite` to represent nonzero values.
-  Thus, there is an implicit invariant that the `nonzeroFinite`
-  constructor should only be used with nonzero dyadics.
+  The proof `hne : d ≠ 0` enforces that `nonzeroFinite` is never used to
+  represent zero: signed zeros live in the `zero` constructor.
```
luisacicolini, line 18: I find this a bit confusing, why would a `nonzeroFinite` represent zero in the first place?

bollu: It shouldn't, but unfortunately for us, `Dyadic` tries to represent a zero.

luisacicolini:
lgtm, thank you!

PR #556, opened 2026-05-12
Add `simp_getset` tactic for `WfIRContext`

math-fehr:
This PR is adding the `simp_getset` tactic that is used to automatically find the get-set lemmas that can be applied in a goal.

The way it works is by first registering a set of lemmas defined by the `@[simp_getset]` annotation. These lemmas are expected to have a single explicit argument, of the form `f ... = ...`. The registration groups these lemmas by the `f` constant.

The tactic goes over all hypotheses, and tries to find all hypotheses of the form `f ... = ...` where `f` is a constant registered by one of the `@[simp_getset]` lemmas. Once all hypotheses are scanned, it adds all the lemmas that were matched, applied to the hypotheses they got matched with, to a `simp only` tactic.

A second version, `simp_getset?`, is also defined, with uses a `simp? only` tactic, so that `[apply]` can be used in the IDE to convert `simp_getset?` to a more efficient `simp only` tactic.

For now, `simp_getset` is only defined for the `WfIRContext` get-set lemmas. We can think of porting it to the `IRContext` get-set lemmas as well.

### DIFF
--- a/Veir/Rewriter/WfRewriter/GetSet.lean
+++ b/Veir/Rewriter/WfRewriter/GetSet.lean
@@ -2,6 +2,7 @@ module
 
 public import Veir.Rewriter.WfRewriter.Basic
 public import Veir.Rewriter.GetSet
+public import Veir.Rewriter.WfRewriter.GetSetTactic
 
 import all Veir.Rewriter.WfRewriter.Basic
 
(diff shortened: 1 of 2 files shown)


ineol (commented):
Looks cool!

Did it improve proofs?

ineol (approved):

luisacicolini (approved):

math-fehr:
So this will be slower than running `simp`, its use is mostly to discover what get-set lemma to fire, and then use the `[apply] simp only` shown in the IDE.

PR #559, opened 2026-05-13
fix a combine bug

regehr:


regehr:
`deMorganOrToAnd` is incorrect for the same reason, argh!!! I'll patch that tomorrow, unless it might be fun to leave it as a canary for subsequent verification work, in which case I would just submit a comment to that effect?

Test/Passes/instcombine.mlir
```diff
@@ -70,6 +70,12 @@
     "test.test"(%or_self) : (i32) -> ()
     // CHECK-NEXT: "test.test"(%[[X]]) : (i32) -> ()
 
+    // or disjoint x | x must not fold: it can produce poison for nonzero x.
```
tobiasgrosser, line 73: I do not get this example. Can you give specific input values and what the output is? Let's say, `x` = `4`? Then the result would be poision before the rewriter and otherwise some concrete value. This seems to permisable, no?

math-fehr:
I agree with Tobias here, this seems like it should be correct, as it is valid to compile `poison` to a value? Unless the point is that you want to fold this to `poison` because it is better?

regehr:
argh, you're right of course!

regehr:
every few months I get twisted around on refinement. it's one of the reasons I wanted to work on Alive in the first place.

PR #560, opened 2026-05-13
Implement a dependently-typed interface for `RuntimeValue`

math-fehr:
This should allow us to use dependent types in the interpreter while still allowing to use a default, non-typed version.


tobiasgrosser (approved):
Nice!

PR #564, opened 2026-05-13
hw: `hw.constant` semantics

osmanyasar05:
This PR introduces the semantics for `hw.constant` operation.

### DIFF
--- a/Veir/Data/HW/Basic.lean
+++ b/Veir/Data/HW/Basic.lean
@@ -5,11 +5,11 @@ namespace Veir.Data.HW
 public section
 
 /-- A width `w` bitvector, representing signless integer type. -/
-structure Bvint (w : Nat) where
+structure BvInt (w : Nat) where
   val : BitVec w
 deriving DecidableEq
 
-instance (w : Nat) : ToString (Bvint w) where
+instance (w : Nat) : ToString (BvInt w) where
   toString r := toString r.val
 
 /-! # CIRCT HW Dialect Semantics -/
@@ -17,5 +17,5 @@ instance (w : Nat) : ToString (Bvint w) where
 /--
   The constant operation produces a constant value of standard integer type without a sign.
 -/
-def constant {w : Nat} (imm : BitVec w) : Bvint w :=
+def constant {w : Nat} (imm : BitVec w) : BvInt w :=
   ⟨imm⟩
--- a/Veir/Data/HW/Basic.lean
+++ b/Veir/Data/HW/Basic.lean
@@ -1,0 +1,21 @@
+module
+
+namespace Veir.Data.HW
+
+public section
+
+/-- A width `w` bitvector, representing signless integer type. -/
+structure Bvint (w : Nat) where
+  val : BitVec w
+deriving DecidableEq
+
+instance (w : Nat) : ToString (Bvint w) where
+  toString r := toString r.val
+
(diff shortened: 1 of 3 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -671,6 +674,16 @@ def Cf.interpretOp' (opType : Veir.Cf) (properties : HasDialectOpInfo.properties
...
 
+def HW.interpretOp' (opType : Veir.HW) (properties : HasDialectOpInfo.propertiesOf opType)
+    (resultTypes : Array TypeAttr) (_blockOperands : Array BlockPtr)
+    : Option ((Array RuntimeValue) × Option ControlFlowAction) :=
+  match opType with
+  | .constant => do
+    let resType ← resultTypes[0]?
+    let .integerType bw := resType.val
+      | none
+    return (#[.bvint bw.bitwidth (Veir.Data.HW.constant (BitVec.ofInt bw.bitwidth properties.value.value)).val], none)
```
osmanyasar05: I can't do HW.constant here because there is a name collision with Veir.HW. I can name Veir.Data.Hw, but then I think the convention is having capital letters for namespaces under Veir.Data? We need to change one of them @georgerennie

math-fehr: Hmmm, this is a bit annoying, and I think the way we are currently handling it is a bit of a Hack (because we format the acronyms in two different ways).
Let me see if we can find a solution, but otherwise I think we will have to keep `Veir.Data.HW`, or we should move where these things are defined.

math-fehr (approved):
To me that looks good, I'll let at least one more person approve it though.
We should figure out a solution for the namespace thing, but we don't have to do this in this PR.
Thanks!

math-fehr:
Oh I forgot about this, should we have a separate type that englobes an `BitVec w`, or should we expose `BitVec w` directly?

Veir/Data/HW/Basic.lean
```diff
@@ -0,0 +1,21 @@
+module
+
+namespace Veir.Data.HW
+
+public section
+
+/-- A width `w` bitvector, representing signless integer type. -/
+structure Bvint (w : Nat) where
```
luisacicolini:
```suggestion
structure BvUInt (w : Nat) where
```
for consistency with the capitalization in lean?

luisacicolini:
lgtm, if we are fine with the `Data.HW` (let's maybe wait for @georgerennie's thoughts on this?). Regarding not exposing the bv type: I think this unsigned integer bitvector will be useful for many CIRCT dialects (e.g. `comb`), so it is good to have it already.

luisacicolini (approved):

Veir/Data/HW/Basic.lean
osmanyasar05: I changed it here and in Interpreter/Basic.lean as well

luisacicolini:
Why not `BvUInt`? afaict they are unsigned?

osmanyasar05:
> Why not `BvUInt`? afaict they are unsigned?

I think I was not precise enough when we looked at them together. They are actually signless, rather than being unsigned. Therefore the interpretation is left to the operations that consume the number.

georgerennie (approved):
I think the `Data.HW` is fine at least for now - I wouldn't let namespacing block this.

Otherwise looks fine to me.

PR #565, opened 2026-05-13
Regehr/interpreter multiple block fix

regehr:
this fix follows @math-fehr's suggestion for threading the interpreter state across basic blocks. it all looks sensible to me but the code is by LLM. the test cases are ones that have come up for me in practice when I'm trying to fuzz VeIR.


Veir/Interpreter/Basic.lean
```diff
@@ -766,11 +766,21 @@ partial_fixpoint
...
-  interpretBlockCFG ctx block state
+  rlet firstOp ← (block.get ctx).firstOp
+  match firstOp.getOpType! ctx with
+  | .func .func | .llvm .func =>
+    if h : firstOp.getNumRegions ctx ≠ 1 then
+      none
+    else
+      interpretRegion ctx (firstOp.getRegion ctx 0) state
+  | _ =>
+    interpretBlockCFG ctx block state
```
math-fehr: Can you revert this to how it was?
Here that would mean that for every region we interpret, if it has a function as first operation then the interpreter behaves differently.

I think we want to move this logic to `veir-interpret`, but we probably want to do this in another PR.

math-fehr (commented):
Looks good to me, but can you just remove the change on the interpretation of regions? You might have to change `func_region_state.mlir` to not have a function.

I think we can add this feature in another PR, but we should add it in `veir-interpret` rather than here!

Veir/Interpreter/Basic.lean
regehr: ok-- I think this patch is just the part you want now.

it might want to be you who does the other part, I'm not sure how to best do it

math-fehr (approved):
Perfect, thanks!

PR #566, opened 2026-05-14
make the veir interpreter report immediate UB explicitly, instead of simply failing

regehr:
sorry, this is a big patch. and surely the details here have implications for our proof strategy that I don't foresee or understand. but, I'm pretty sure the new behavior as indicated by the test cases is solid.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -125,12 +125,48 @@ inductive ControlFlowAction where
   | return (vals : Array RuntimeValue)
   | branch (vals : Array RuntimeValue) (dest : BlockPtr)
 
+/--
+  Wrapper for interpreter step results: either a successful value `ok` or the
+  program has triggered undefined behaviour (`ub`). UB is a property of the
+  execution, not of any value, so it lives here rather than inside `RuntimeValue`
+  or `LLVM.Int`.
+-/
+inductive UBOr (α : Type) where
+  | ok : α → UBOr α
+  | ub : UBOr α
+deriving Inhabited
+
+/--
+  The interpreter monad. `Option (UBOr α)` has three states:
+  - `some (.ok x)` — successful execution producing `x`.
+  - `some .ub`     — execution triggered undefined behaviour.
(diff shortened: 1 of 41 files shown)


regehr:
this patch also includes real semantic changes! but I believe they're the right ones to bring behavior into line with LLVM

luisacicolini (commented):
this looks good to me, thank you! After #550 is merged we should be able to replace some of the infrastructure with the `isPoison` methods (`if op_isPoison then Interp.ub`, we have them for all ops.). Maybe it's worth to have that in and use them already? We should be able to merge by today :)

Veir/Interpreter/Basic.lean
```diff
@@ -125,12 +125,48 @@ inductive ControlFlowAction where
   | return (vals : Array RuntimeValue)
   | branch (vals : Array RuntimeValue) (dest : BlockPtr)
 
+/--
+  Wrapper for interpreter step results: either a successful value `ok` or the
```
luisacicolini, line 129:
```suggestion
  Wrapper for interpreter step results: it either returns a successful value `ok` or the
```

Veir/Interpreter/Basic.lean
```diff
@@ -154,22 +190,56 @@ def Arith.interpretOp' (opType : Veir.Arith) (properties : HasDialectOpInfo.prop
...
+    | .poison => Interp.ub
+    | .val v' =>
+      if v' = 0 then Interp.ub
+      else if bw ≠ 1 ∧ v' = -1 then
+        -- Divisor is concretely -1; if the dividend could be intMin, the
+        -- overflow case applies and is UB.
+        match lhs with
+        | .poison => Interp.ub
+        | .val v =>
+          if v = BitVec.intMin bw then Interp.ub
```
luisacicolini, line 214: this is the type of changes that could be replaced by `sdiv_isPoison` that I am referring to in the comment

luisacicolini:
#550 is merged but I realized there are some things still missing on my side - let's merge this!

luisacicolini (approved):

tobiasgrosser:
Thank you @regehr.

PR #567, opened 2026-05-14
Prove sorry in parseOptionalBlockLabel

GZGavinZhao:
Prove one of the remaining two `sorry`s in the parser using the newly added `WfRewriter.setBlockArguments` from #546.


math-fehr (approved):
Very nice!

PR #569, opened 2026-05-14
Ensure unregistered operations do not lose information

math-fehr:
Previously, unregistered operations were discarding the parsed operation name and properties.

Now, unregistered operations store in their properties the original properties, as well as the parsed operation name.

The parser and printer are now using these properties to pass the original operation information.

### DIFF
--- a/Test/Parsing/unregistered_ops.mlir
+++ b/Test/Parsing/unregistered_ops.mlir
@@ -1,10 +1,12 @@
 // RUN: veir-opt %s | filecheck %s
 
 "builtin.module"() ({
-    "unregistered.op"() : () -> ()
-    %1 = "unregistered.op"() : () -> !foo.bar
-    %2 = "unregistered.op"() : () -> !foo.bar<baz>
-    // CHECK:     "builtin.unregistered"() : () -> ()
-    // CHECK-NEXT: %{{.*}} = "builtin.unregistered"() : () -> !foo.bar
-    // CHECK-NEXT: %{{.*}} = "builtin.unregistered"() : () -> !foo.bar<baz>
+    "unregistered.op_one"() : () -> ()
+    %1 = "unregistered.op_two"() : () -> !foo.bar
+    %2 = "unregistered.op_three"() : () -> !foo.bar<baz>
+    "unregistered.op_four"() <{foo = 1 : i32}> : () -> ()
+    // CHECK:      "unregistered.op_one"() : () -> ()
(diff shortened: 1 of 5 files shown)


Veir/Properties.lean
```diff
@@ -301,7 +318,7 @@ def GetelementptrProperties.fromAttrDict (attrDict : Std.HashMap ByteArray Attri
     | none => .ok { value := 0, type := { bitwidth := 32 } }
   let rawConstantIndices ← match attrDict["rawConstantIndices".toUTF8]? with
     | some (.denseArrayAttr arr) => .ok arr
-    | some attr => .error s!"getelementptr: expected 'rawConstantIndices' to be a dense array attribute, 
+    | some attr => .error s!"getelementptr: expected 'rawConstantIndices' to be a dense array attribute,
```
tobiasgrosser: This is an unrelated change. I would prefer for those to be committed without review as an obvious fix.

Test/Parsing/unregistered_ops.mlir
```diff
@@ -4,7 +4,9 @@
...
     %1 = "unregistered.op"() : () -> !foo.bar
     %2 = "unregistered.op"() : () -> !foo.bar<baz>
-    // CHECK:     "builtin.unregistered"() : () -> ()
-    // CHECK-NEXT: %{{.*}} = "builtin.unregistered"() : () -> !foo.bar
-    // CHECK-NEXT: %{{.*}} = "builtin.unregistered"() : () -> !foo.bar<baz>
+    "unregistered.op"() <{foo = 1 : i32}> : () -> ()
+    // CHECK:      "unregistered.op"() : () -> ()
+    // CHECK-NEXT: %{{.*}} = "unregistered.op"() : () -> !foo.bar
+    // CHECK-NEXT: %{{.*}} = "unregistered.op"() : () -> !foo.bar<baz>
+    // CHECK-NEXT: "unregistered.op"() <{"foo" = 1 : i32}> : () -> ()
```
tobiasgrosser: Should we add a couple of operations with a different name here into the test cases.

tobiasgrosser (approved):

Veir/Properties.lean
math-fehr: I'm going to add a linter in a different PR, I think this is my IDE that automatically removes trailing spaces.
I'll remove this change.

Test/Parsing/unregistered_ops.mlir
math-fehr: Changed!

tobiasgrosser:
Perfect

PR #570, opened 2026-05-14
refactor(FP): introduce Veir.Data.FP.FloatFormat namespace

bollu:
## Summary
- Add `Veir/Data/FP/FloatFormat.lean`: a consolidated home for IEEE-754 format arithmetic over `(e, s)`. Eventually this will become a structure bundling those parameters; for now it is a namespace whose helpers take them as explicit arguments.
- Move `bias` out of `Veir.Data.FP.PackedFloat` (the definition has no `PackedFloat`-specific content) and into `FloatFormat`. Update the two callers (`PackedFloat.toExtRat`, `PackedFloat.toEDyadic`) to use `FloatFormat.bias`.
- The new module also ships `#guard`s that pin down `bias 8 = 127` and `bias 11 = 1023` (single and double precision).

The namespace is intentionally thin today. Follow-on work introducing `unbiasedExp`, `biasedExp`, `minBiasedExponent`, `maxBiasedExponent`, and the various significand-shift helpers will land here as their callers (e.g. `EDyadic.pack`) arrive.

## Test plan
- [x] `lake build` is green.

### DIFF
--- a/Veir/Data/FP/FloatFormat.lean
+++ b/Veir/Data/FP/FloatFormat.lean
@@ -27,8 +27,6 @@ def maxBiasedExponent (e : Nat) : Nat := 2 ^ e - 2
 #guard maxBiasedExponent 8 = 254
 -- Double precision: `maxBiasedExponent 11 = 2^11 - 2 = 2046`.
 #guard maxBiasedExponent 11 = 2046
-
-
 /--
 The smallest biased exponent representing a normal number: `1`.
 
--- a/Veir/Data/FP/FloatFormat.lean
+++ b/Veir/Data/FP/FloatFormat.lean
@@ -4,29 +4,9 @@ namespace Veir.Data.FP.FloatFormat
 
 public section
 
-/-! # IEEE-754 floating-point format computations
-
-A consolidated namespace for the integer/`BitVec` arithmetic that
-describes an IEEE-754 binary format with `e` exponent bits and `s`
-trailing-significand bits. Eventually this will become a structure
-`FloatFormat` bundling `e` and `s`; for now, each helper takes them
-as explicit arguments.
-
-This module currently exposes only the format constants that exist
-elsewhere in the codebase. Encoding/decoding arithmetic
-(`unbiasedExp`, `biasedExp`, shift helpers, …) will be added here as
-the corresponding callers land.
--/
 
 /--
 The IEEE-754 exponent bias for `e` exponent bits: `bias e = 2^(e-1) - 1`.
-
-A stored biased exponent `ex` encodes the unbiased exponent `ex - bias e`.
-The bias is chosen so the unbiased exponent range is roughly symmetric
-around `0`.
-
-For single precision (`e = 8`),  `bias = 127`.
-For double precision (`e = 11`), `bias = 1023`.
 -/
 def bias (e : Nat) : Nat := 2 ^ (e - 1) - 1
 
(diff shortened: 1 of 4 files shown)


Veir/Data/FP/FloatFormat.lean
```diff
@@ -0,0 +1,103 @@
...
+/--
+The largest biased exponent representing a finite normal number: `2^e - 2`.
+
+Recall that the all-ones bitpattern of `2^e - 1` is reserved for `±∞` / NaN.
+Hence the largest finite biased exponent is `(2^e - 1) - 1 = 2^e - 2`.
+-/
+def maxBiasedExponent (e : Nat) : Nat := 2 ^ e - 2
+
+-- Single precision: `maxBiasedExponent 8 = 2^8 - 2 = 254`.
+#guard maxBiasedExponent 8 = 254
```
luisacicolini, line 27: shouldn't this be in the tests?

Veir/Data/FP/FloatFormat.lean
```diff
@@ -0,0 +1,103 @@
...
+
+/--
+The IEEE-754 biased exponent, corresponding to a given unbiased exponent.
+Obtained by adding the bias to the unbiased exponent.
+-/
+def biasedExp (e : Nat) (n : Int) (k : Int) : Int :=
+  (bias e : Int) + unbiasedExp n k
+
+/--
+Left shift required to `n` such that when interprted as a `s`bit bitvetor,
```
luisacicolini:
```suggestion
Left shift required to `n` such that when interprted as a `s`bit bitvector,
```

luisacicolini (approved):
thank you! left a couple comments, other than that lgtm

Veir/Data/FP/FloatFormat.lean
```diff
@@ -0,0 +1,103 @@
...
+/--
+The mathematical (unbiased) base-2 exponent of `n* 2^(-k)`
+
+This gives us the exponent of `n * 2^(-k) = s * 2^unbiasedExp`,
+where `s ∈ [1, 2)` is the significand after normalizing `n`.
+-/
+def unbiasedExp (n : Int) (k : Int) : Int :=
+  (n.natAbs.log2 : Int) - k
+
+
```
tobiasgrosser:
```suggestion

```

Veir/Data/FP/FloatFormat.lean
```diff
@@ -0,0 +1,103 @@
...
+Used to drop the hidden bit from a fully-aligned significand
+`alignedSig ∈ [2^s, 2^(s+1))`. Since bit `s` is set and all higher bits
+are clear in that range,
+  `alignedSig &&& (2^s - 1) = alignedSig - 2^s`,
+which is exactly the stored `s`-bit trailing significand.
+-/
+def trailingSigMask (s : Nat) : Nat := 2 ^ s - 1
+
+end -- public section
```
tobiasgrosser:
```suggestion
end
```

Do we want to have these or not? @luisacicolini has requested them to be dropped, no?

tobiasgrosser (approved):

Veir/Data/FP/FloatFormat.lean
luisacicolini: ah yes, that I did not notice - I think we should drop them in general, they don't seem to be used in veir

Veir/Data/FP/FloatFormat.lean
bollu: I considered it, but this is kind of a 'sanity check'/'documentation', so I left it here.

bollu: I suggest we keep it here, as it's really more documentation than test.

Veir/Data/FP/FloatFormat.lean
```diff
@@ -0,0 +1,101 @@
...
+Hence the largest finite biased exponent is `(2^e - 1) - 1 = 2^e - 2`.
+-/
+def maxBiasedExponent (e : Nat) : Nat := 2 ^ e - 2
+
+-- Single precision: `maxBiasedExponent 8 = 2^8 - 2 = 254`.
+#guard maxBiasedExponent 8 = 254
+-- Double precision: `maxBiasedExponent 11 = 2^11 - 2 = 2046`.
+#guard maxBiasedExponent 11 = 2046
+
+
```
luisacicolini:
```suggestion

```

luisacicolini (approved):

PR #571, opened 2026-05-14
comb: add `comb.add` semantics

osmanyasar05:
This PR introduces the semantics for the `comb.add` operation. The monadic do-block is implemented this way since comb operations accept variadic input sizes.

Co-authored-by: Luisa Cicolini luisacicolini@gmail.com


Test/Interpreter/Comb/add.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: veir-interpret %s | filecheck %s
+
+"builtin.module"() ({
+  %lhs = "hw.constant"() <{ "value" = 3 : i32 }> : () -> i32
+  %rhs = "hw.constant"() <{ "value" = 5 : i32 }> : () -> i32
+  %x = "comb.add"(%lhs, %rhs) : (i32, i32) -> i32
```
georgerennie: Could we have tests for add with 1 argument and more than 2 arguments

Veir/Interpreter/Basic.lean
```diff
@@ -3,6 +3,7 @@ import Veir.Rewriter.Basic
 import Veir.ForLean
 import Veir.IR.WellFormed
 import Veir.PatternRewriter.Basic
+import Veir.Data.COMB.Basic
```
georgerennie: I think we should do `Veir.Data.Comb` (the others are capitalised but they are acronyms).

Veir/Interpreter/Basic.lean
```diff
@@ -763,24 +660,35 @@ def Riscv.interpretOp' (opType : Veir.Riscv) (properties : HasDialectOpInfo.prop
...
+    else
+      return (#[], some (.branch (operands.extract (trueSize + 1) operands.size) destFalse))
+
+def Comb.interpretOp' (opType : Veir.Comb) (properties : HasDialectOpInfo.propertiesOf opType)
+    (operands : Array RuntimeValue) (_blockOperands : Array BlockPtr)
+    : Option ((Array RuntimeValue) × Option ControlFlowAction) :=
+  match opType with
+  | .add => do
+    let l : List _ := operands.toList
+    let .bvInt w fst := l[0]! | none
```
georgerennie: It seems arbitrary to just add all the terms with the same type as the first argument (why first rather than any other). I think it would be better to pass in the result type and then filter by that.

Veir/Interpreter/Basic.lean
```diff
@@ -763,24 +660,35 @@ def Riscv.interpretOp' (opType : Veir.Riscv) (properties : HasDialectOpInfo.prop
...
+    let some nl := l.mapM (
+        fun e => do
+          let .bvInt w' val := e | none
+          if h : w' ≠ w then none else
+          return val.cast (by simpa using h)) | none
+    return (#[.bvInt w (Veir.Data.COMB.add nl)], none)
```
georgerennie: This definition feels like it would be a bit of a pain to work with for proofs (i think mapM could be annoying to reason about) and is not super clear to read. I think it would be better to pass in the result type to the function and it might even be nice to separate the check of widths from the comptuation. e.g. you could have something like (pseudocode-y):

```lean4
if (operands.any (fun x => x not bvInt with width w)) then
  return none

let nl := operands.toList.map (fun x => get x as bvint with width w else 0)
```

Veir/Data/COMB/Basic.lean
```diff
@@ -0,0 +1,15 @@
...
+namespace Veir.Data.COMB
+
+public section
+
+/-! # CIRCT Comb Dialect Semantics -/
+
+/--
+Variadic `add` operation with a list of bitvectors with width `w` as input
+-/
+def add {w : Nat} (l : List (BitVec w)) : BitVec w :=
```
georgerennie: Maybe worth thinking about whether this should take Array instead (given `interpretOp'` gets an Array of values), or maybe even `Std.Iter`.

georgerennie (commented):
A couple thoughts on the design of this.

Veir/Interpreter/Basic.lean
georgerennie: Oh actually thinking about it more I guess it fails if they don't all have the same type. Nonethless I think we should do it based on the result type - if all the args have the same type but its different from the result type we should return none.

luisacicolini (approved):

math-fehr (approved):
Thanks, looks good to me now!

PR #575, opened 2026-05-15
Test: tidy some testcases to pass MLIR verification

georgerennie:
This fixes some of the testcases up so that they don't fail verification
in MLIR which makes it slightly easier to compare parser behaviour with
MLIR when modifying the tests.

### DIFF
--- a/Test/Arith/identity.mlir
+++ b/Test/Arith/identity.mlir
@@ -7,14 +7,14 @@
   %addi_nsw = "arith.addi"(%5, %5) <{nws}> : (i32, i32) -> i32
   %addi_nuw = "arith.addi"(%5, %5) <{nuw}> : (i32, i32) -> i32
   %addi_nsw_nuw = "arith.addi"(%5, %5) <{nsw, nuw}> : (i32, i32) -> i32
-  %70, %71 = "arith.addui_extended"(%5, %5) : (i32, i32) -> (i32, i32)
+  %70, %71 = "arith.addui_extended"(%5, %5) : (i32, i32) -> (i32, i1)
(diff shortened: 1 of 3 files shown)


Test/Arith/identity.mlir
```diff
@@ -7,14 +7,14 @@
   %addi_nsw = "arith.addi"(%5, %5) <{nws}> : (i32, i32) -> i32
   %addi_nuw = "arith.addi"(%5, %5) <{nuw}> : (i32, i32) -> i32
   %addi_nsw_nuw = "arith.addi"(%5, %5) <{nsw, nuw}> : (i32, i32) -> i32
-  %70, %71 = "arith.addui_extended"(%5, %5) : (i32, i32) -> (i32, i32)
+  %70, %71 = "arith.addui_extended"(%5, %5) : (i32, i32) -> (i32, i1)
   %8 = "arith.andi"(%5, %5) : (i32, i32) -> i32
   %9 = "arith.ceildivsi"(%5, %5) : (i32, i32) -> i32
   %10 = "arith.ceildivui"(%5, %5) : (i32, i32) -> i32
-  %11 = "arith.cmpi"(%5, %5) : (i32, i32) -> i1
+  %11 = "arith.cmpi"(%5, %5) <{ "predicate" = 0 : i64 }> : (i32, i32) -> i1
```
georgerennie, line 14: I just realised that we don't actually store the `predicate` field for `arith.cmpi` yet, will add this in another PR but it just gets ignored for now which seems fine.

tobiasgrosser:
Nice

PR #576, opened 2026-05-15
Parser: support nested scopes and error on redefinition

georgerennie:
This PR adds nested scope tracking of variable definitions to the parser. This changes `values`
from `Std.HashMap ByteArray (Array ValuePtr)` to `Array (Std.HashMap ByteArray (Array ValuePtr))`
to represent the stack of scopes. When a new region is parsed, a scope is pushed to the stack
whilst parsing it. An error is now thrown on redefinition of variables and the tests have been
updated to avoid this.

This does not support `IsolatedFromAbove` scopes that allow reuse of variable names that appear
above the scope.

This updates existing tests which reused variables names to use a different name each time, and adds
`Test/Parsing/scopes.mlir` to test accessing/redefining variables in different scopes as well as new tests
in `UnitTest/MlirParser.lean` for the error cases.


tobiasgrosser (commented):
Beautiful. Let's wait for @math-fehr's feedback.

Veir/Parser/MlirParser.lean
```diff
@@ -32,7 +37,7 @@ structure MlirParserState where
   deriving Inhabited
 
 def MlirParserState.fromContext (ctx : WfIRContext OpCode) : MlirParserState :=
-  {ctx := ctx, values := Std.HashMap.emptyWithCapacity 128, blocks := Std.HashMap.emptyWithCapacity 1}
+  {ctx := ctx, values := #[Std.HashMap.emptyWithCapacity 128], blocks := Std.HashMap.emptyWithCapacity 1}
```
math-fehr: Can we change this to have a capacity of 0, or 2?
Because this will only be used for the toplevel operation, which 99.9999% of the time is a `builtin.module`, we will almost never actually use it in practice?

Veir/Parser/MlirParser.lean
```diff
@@ -66,7 +71,7 @@ def getContext : MlirParserM (WfIRContext OpCode) := do
   Get the array of values associated with a previously parsed name.
 -/
 def getValues? (name : ByteArray) : MlirParserM (Option (Array ValuePtr)) := do
-  return (← get).values[name]?
+  return (← get).values.findSomeRev? (·[name]?)
```
math-fehr: Should we iterate the values in the opposite order here? It is way more likely that the value we are looking for is in the end of the array, rather than at the beginning I would guess?
We can also leave this for later if we want to.

Veir/Parser/MlirParser.lean
```diff
@@ -503,6 +522,9 @@ partial def parseRegion : MlirParserM RegionPtr := do
   if (← parseOptionalPunctuation "}") then
     return region
 
+  /- Ensure variables defined in this region do not leak out of it. -/
+  inChildScope do
```
math-fehr: Should we add an indentation here to make the scope more visible?
Alternatively, should we move the scope at the beginning of this function?

math-fehr (approved):
Very cool! I just have 2-3 nit comments, but otherwise all is good!
Quick question otherwise, do you know if this `Array` is the data structure used in MLIR?
I'm fine merging this even if it's not the MLIR version, but we should know about it.

Veir/Parser/MlirParser.lean
georgerennie: `findSomeRev?` will start searching from the back of the array, so it should do this already I think

Veir/Parser/MlirParser.lean
georgerennie: yeah I can put it up at the beginning of the function to make it clear

georgerennie:
> Quick question otherwise, do you know if this `Array` is the data structure used in MLIR? I'm fine merging this even if it's not the MLIR version, but we should know about it.

https://github.com/llvm/llvm-project/blob/bf7d6fe85eadf45b65e416ad1c8dba3333d00e8a/mlir/lib/AsmParser/Parser.cpp#L827-L848 its similar but they have both a hashmap from names to values (`values`) and an array of hashsets of names for each scope (`definitionsPerScope`). I guess this means that looking up a value only requires looking in one hashmap rather than an array of them - I'm happy to switch to this style if we think it is better.

Veir/Parser/MlirParser.lean
math-fehr: My bad, i misdunderstood the function

math-fehr:
I think we should follow MLIR style, because they have thought of this quite a lot and that's probably best (cost of everything is `O(1)` amortized, while for us it is not).
I'm fine in doing this in a separate PR if you want

georgerennie:
Yeah agreed, I've just updated it to use the MLIR style because it was easy enough.

tobiasgrosser:
@georgerennie, this seems to have broken some filecheck test cases.

GZGavinZhao (approved):
Parser changes look good!

georgerennie:
Rebased and fixed tests, assuming CI passes now I will merge.

PR #577, opened 2026-05-15
select now blocks poison from the not-selected arm

regehr:
also, remove two tests that are definitely incorrect under the current LLVM rules. select *cannot* be lowered to logic unless non-poison can be proved.


regehr:
I have addressed @luisacicolini's comments (from zulip)

PR #579, opened 2026-05-15
Printer: use pack syntax for operations with multiple results

georgerennie:
This changes the formatting of operations with multiple results to use the pack syntax, e.g.
```
%5:3 = "test.test"() : () -> (i32, i32, i32)
"test.test"(%5#1) : (i32) -> ()
```

instead of the existing unpacked syntax
```
%5_0, %5_1, %5_2 = "test.test"() : () -> (i32, i32, i32)
"test.test"(%5_1) : (i32) -> ()
```
which is not valid MLIR as names beginning with a digit must only contain digits.


tobiasgrosser (approved):
LGTM. Let's see what @math-fehr and @GZGavinZhao say.

math-fehr (approved):
All good, thanks!

PR #580, opened 2026-05-15
Add !llvm.func type attribute for LLVM function signatures

regehr:
I'm unsure about this... the patch breaks parsing raw LLVM `ptr` and `void`, so I edited a test case to use the `!llvm.ptr` and `!llvm.void` syntax. if that's acceptable, then maybe this patch works.


math-fehr:
I think we should try to make this work with `ptr` and `void` without the `!llvm` before, as this is how MLIR prints it.
Have you tried to see if the file is parseable by MLIR?

Otherwise, what we should do is copy what xDSL is currently doing here: https://github.com/xdslproject/xdsl/blob/e596f41c99e7c20478240b8f386e9e5de6a4ee14/xdsl/dialects/llvm.py#L202

It shouldn't be too hard to implement

regehr:
> I think we should try to make this work with `ptr` and `void` without the `!llvm` before, as this is how MLIR prints it. Have you tried to see if the file is parseable by MLIR?

Ah, I didn't realize this was verbatim output. Ok, I agree.
 
> Otherwise, what we should do is copy what xDSL is currently doing here: https://github.com/xdslproject/xdsl/blob/e596f41c99e7c20478240b8f386e9e5de6a4ee14/xdsl/dialects/llvm.py#L202
> 
> It shouldn't be too hard to implement

Cool I will look. Claude had suggested several solutions but they all sounded like hacks to me :(

regehr:
(but yes, my edited file does parse cleanly using upsteam `mlir-opt`)

regehr:
ok, interesting approach! I hope this is better

UnitTest/AttrParser.lean
```diff
@@ -179,6 +179,20 @@ macro "#assert " e:term : command =>
...
+    #[(IntegerType.mk 32 : Attribute)] #[(IntegerType.mk 32 : Attribute)]), by rfl⟩
+#assert expectSuccessType "!llvm.func<i64 ()>"
+  ⟨.llvmFunctionType (FunctionType.mk #[] #[(IntegerType.mk 64 : Attribute)]), by rfl⟩
+#assert expectSuccessType "!llvm.func<i32 (i32, i64)>"
+  ⟨.llvmFunctionType (FunctionType.mk
+    #[(IntegerType.mk 32 : Attribute), (IntegerType.mk 64 : Attribute)]
+    #[(IntegerType.mk 32 : Attribute)]), by rfl⟩
+#assert expectSuccessType "!llvm.func<!llvm.ptr (!llvm.ptr)>"
+  ⟨.llvmFunctionType (FunctionType.mk
+    #[(LLVM.PointerType.mk : Attribute)] #[(LLVM.PointerType.mk : Attribute)]), by rfl⟩
```
math-fehr, line 194: Can you add tests for parsing it with `ptr` instead of `!llvm.ptr`, and same for `void`?

Veir/IR/Attribute.lean
```diff
@@ -249,6 +249,8 @@ inductive Attribute
 | modArithType (type : ModArithType)
 /-- LLVM pointer type -/
 | llvmPointerType (type : LLVM.PointerType)
+/-- LLVM function type -/
+| llvmFunctionType (type : FunctionType)
```
math-fehr, line 253: Can you also add a `LLVM.FunctionType` here?
This is not exactly the same as a `FunctionType`, as it has only one result, and may have variadic arguments (which we will implement later)

math-fehr (commented):
Much nicer, thanks!
I just have 2 comments, but otherwise this looks good!

PR #581, opened 2026-05-15
Make `hw.constant` use `LLVM.Int` instead of `BitVec`

math-fehr:
As `IntegerType` should map to a unique Lean Type, we are using `LLVM.Int` even though `comb` uses `IntegerType` as bitvectors.

As `RuntimeValue.intBv` is not used anymore, I removed it as well for the time being.


osmanyasar05 (approved):
LGTM

PR #582, opened 2026-05-16
DON'T MERGE Testing #535

GZGavinZhao:
Just testing what the new benchmark would look like given that #535 is submitted through a fork and therefore the GH bot doesn't have the permissions to post the results.


tobiasgrosser:
#535 has been merged.

PR #584, opened 2026-05-17
Remove the `liftExcept` function from the parser

math-fehr:
This function was only used to lift other functions from the `Except` monad to a generic monad. Instead, we can just implement all these other functions in the more generic monad directly.


GZGavinZhao (approved):
Ahh nice! Good catch. Thanks for removing this annoyance.

PR #585, opened 2026-05-17
Replace an OperationPtr.InBounds runtime check with proof

GZGavinZhao:
Replaces an unnecessary `OperationPtr.InBounds` runtime check with proof. The helper function `checkOpInBounds` is no longer used but still kept because we might need it in the future when expanding the parser.


math-fehr (approved):
Nice!

PR #588, opened 2026-05-17
Use `ParserError` instead of `String` in the parser

math-fehr:
In particular, no error output is being changed, as the `ParserError` is converted to a string at the end in `veir-opt`

This PR only contains the change from `String` to `ParserError`, and the temporary addition of `ParserError.throwString` and the coercion from `ParserError` to `String`. These two last additions will be removed once everything moves to `ParserError`.

### DIFF
--- a/UnitTest/AttrParser.lean
+++ b/UnitTest/AttrParser.lean
@@ -3,34 +3,35 @@ import Veir.IR.Basic
 
 open Veir
 open Veir.Parser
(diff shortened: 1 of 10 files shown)


tobiasgrosser (approved):
Nice. Thank you for breaking the PR in easy-to-understand pieces.

luisacicolini (approved):

GZGavinZhao (approved):

PR #591, opened 2026-05-17
Trivial whitespace fix [NFC]

tobiasgrosser:


tobiasgrosser:
Not sure if the whitespace in the string actually carries meaning. I initially thought this was a trailing whitespace in code.

PR #592, opened 2026-05-17
Add parsing for llvm.fadd operation

tobiasgrosser:


Veir/IR/Attribute.lean
```diff
@@ -44,6 +44,14 @@ structure IntegerType where
   bitwidth : Nat
 deriving Inhabited, Repr, DecidableEq, Hashable
 
+/--
+ A `!builtin.floating_point` attribute is a floating point literal with an
```
tobiasgrosser: Not sure if `!builtin.floating_point` is actually a thing. This was adapted from the `IntegerType` comment.

Veir/Properties.lean
```diff
@@ -104,6 +104,21 @@ def NnegProperties.fromAttrDict (attrDict : Std.HashMap ByteArray Attribute) :
   let nneg ← getUnitAttr "nneg" attrDict
   return { nneg := nneg }
 
+structure FastMathFlagsProperties where
+  fast : Bool
+  nan : Bool
```
luisacicolini: afaiu this is called `nnan` in [llvm](https://llvm.org/docs/LangRef.html#id2081), should we call it the same?

tobiasgrosser: Good catch. I fixed this.

luisacicolini (approved):

Veir/Properties.lean
```diff
@@ -104,6 +104,21 @@ def NnegProperties.fromAttrDict (attrDict : Std.HashMap ByteArray Attribute) :
   let nneg ← getUnitAttr "nneg" attrDict
   return { nneg := nneg }
 
+structure FastMathFlagsProperties where
+  fast : Bool
+  nnan : Bool
+  ninf : Bool
+  nsz : Bool
+deriving Inhabited, Repr, Hashable, DecidableEq
```
math-fehr, line 112: This does not correspond to the fastmath flags from LLVM.
This should contain a single `::mlir::LLVM::FastmathFlagsAttr`, instead of a set of unit attributes.
Also, this won't be parsed by MLIR as the LLVM dialect does not use properties currently, and rather use the attributes. I do not have a nice solution for that part though, we might have to hack the parser/printer to make this work, or to try to upstream a fix in MLIR.

tobiasgrosser: Ok, I will look into this. Thank you for your comment.

PR #596, opened 2026-05-17
change interpreter to run an llvm.main if it exists

regehr:
this PR modifies the vier interpreter to look for a zero-argument function called main(). if found, this is executed. if not, and if there are instructions at the module level, the those are executed. else, an error is raised.

the plan that I think @math-fehr is onboard with is to land this patch (or some version thereof-- I'm not at all sure about the best way to do things here). then in a separate patch, trivially rewrite all interpreter tests to put their instructions into a main() function. and finally, make a patch so that the interpreter executes only main() and never the module-level instructions.

presumably, later on, we would forbid module-level instructions.

### DIFF
--- a/Test/Interpreter/LLVM/main_func.mlir
+++ b/Test/Interpreter/LLVM/main_func.mlir
@@ -1,7 +1,7 @@
 // RUN: veir-interpret %s | filecheck %s
 
 "builtin.module"() ({
-  "llvm.func"() <{function_type = !llvm.func<i1 ()>, sym_name = "main"}> ({
+  "llvm.func"() ({
   ^bb0():
     %c0 = "llvm.mlir.constant"() <{"value" = 0 : i1}> : () -> i1
     %c1 = "llvm.mlir.constant"() <{"value" = 1 : i1}> : () -> i1
@@ -10,7 +10,7 @@
 
   ^bb1():
     "llvm.return"(%v0) : (i1) -> ()
-  }) : () -> ()
+  }) {function_type = !llvm.func<i1 ()>, sym_name = "main"} : () -> ()
 }) : () -> ()
 
 // CHECK: Program output: #[0x1#1]
--- a/Test/Interpreter/LLVM/main_func.mlir
+++ b/Test/Interpreter/LLVM/main_func.mlir
@@ -1,0 +1,16 @@
+// RUN: veir-interpret %s | filecheck %s
+
+"builtin.module"() ({
+  "llvm.func"() <{function_type = !llvm.func<i1 ()>, sym_name = "main"}> ({
+  ^bb0():
+    %c0 = "llvm.mlir.constant"() <{"value" = 0 : i1}> : () -> i1
+    %c1 = "llvm.mlir.constant"() <{"value" = 1 : i1}> : () -> i1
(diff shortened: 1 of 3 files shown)


VeirInterpret.lean
```diff
@@ -29,6 +29,42 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode 
...
 
+partial def topLevelOps (ctx : IRContext OpCode) (op : OperationPtr) : Array OperationPtr :=
+  if op.getNumRegions! ctx != 1 then #[] else
+  match (op.getRegion! ctx 0 |>.get! ctx).firstBlock with
+  | none => #[]
+  | some b =>
+    let rec go (o : Option OperationPtr) : Array OperationPtr :=
+      match o with
+      | none => #[]
+      | some o => #[o] ++ go (o.get! ctx).next
```
math-fehr, line 40: This will actually run in `O(n^2)` as we keep concatenating arrays.

Test/Interpreter/LLVM/main_func.mlir
```diff
@@ -0,0 +1,16 @@
...
+  "llvm.func"() ({
+  ^bb0():
+    %c0 = "llvm.mlir.constant"() <{"value" = 0 : i1}> : () -> i1
+    %c1 = "llvm.mlir.constant"() <{"value" = 1 : i1}> : () -> i1
+    %v0 = "llvm.add"(%c0, %c1) : (i1, i1) -> i1
+    "llvm.br"() [^bb1] : () -> ()
+
+  ^bb1():
+    "llvm.return"(%v0) : (i1) -> ()
+  }) {function_type = !llvm.func<i1 ()>, sym_name = "main"} : () -> ()
```
math-fehr, line 13: This test uses attributes, while the other one uses properties, something is likely broken in one of the tests?

VeirInterpret.lean
```diff
@@ -29,6 +29,42 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode 
...
+
+def isNonExecutableTopLevel (ctx : IRContext OpCode) (op : OperationPtr) : Bool :=
+  match op.getOpType! ctx with
+  | .llvm .func | .func .func | .llvm .module_flags => true
+  | _ => false
+
+def symName? (ctx : IRContext OpCode) (op : OperationPtr) : Option String :=
+  match (op.get! ctx).attrs.entries.find? (·.1 == "sym_name".toUTF8) with
+  | some (_, .stringAttr s) => some (String.fromUTF8! s.value)
+  | _ => none
```
math-fehr, line 51: Instead of returning a list of toplevel operations, maybe we should instead have a function that finds a toplevel function with the "main" symName. Then in `interpretInput` we just either run this one (if it finds it), or run `interpretModule`?

math-fehr, line 51: It looks weird to look for a `symName` without knowing the operation type.
I think it would make more sense probably to merge these two functions into `isFuncWithName` that checks that this is a function, and checks if it has the correct `sym_name`?

math-fehr (commented):
Nice! I added some comments on the code design, that I think should be changed a bit.
Let me know if it makes sense!

tobiasgrosser:
I like the PR and the overall direction. We should update the PR title to be more meaningful.

PR #597, opened 2026-05-17
support two function properties required by mlir-translate

regehr:
support for parsing/print the `sym_name` and  `function_type` LLVM properties.

this requires updating an existing test case, since these are no longer dropped.


tobiasgrosser (approved):
This looks sensible to me. Let's wait for @math-fehr and @luisacicolini, as both have more experience with LLVM properties than I do.

math-fehr (approved):
That looks good to me as well.
Note that LLVM does not actually uses properties, but let's uses them and figure out how to change the parser/printer to connect with MLIR correctly.

luisacicolini (approved):

PR #599, opened 2026-05-18
Introduce `VariableState` in the interpreter and fix remainnig sorries

math-fehr:
The idea is that getting/setting variables in a `VariableState` that is contained in the `InterpreterState` makes things easier for proofs. This is because we do not have to write many small lemmas about separation between memory and variable states. In general, it just makes things easier to manipulate.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -55,68 +55,71 @@ structure MemoryState where
 def MemoryState.empty : MemoryState :=
   { contents := (ByteArray.emptyWithCapacity 1024).extend 8 0xff }
 
+structure VariableState where
+  variables : Std.ExtHashMap ValuePtr RuntimeValue
+
 /--
   The state of the interpreter at a given point in time.
   It includes a mapping from IR values to their runtime values.
 -/
+@[ext]
 structure InterpreterState where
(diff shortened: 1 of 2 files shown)


tobiasgrosser (approved):
Nice.

ineol (approved):

Veir/Interpreter/Basic.lean
```diff
@@ -55,68 +55,77 @@ structure MemoryState where
...
+theorem VariableState.ext {s₁ s₂ : VariableState} :
     (∀ var, s₁.getVar? var = s₂.getVar? var) →
-    s₁.memory = s₂.memory →
     s₁ = s₂ := by
   rcases s₁; rcases s₂
-  simp only [getVar?, mk.injEq]
+  simp only [VariableState.getVar?, mk.injEq]
   grind
 
+@[ext]
```
nchappe: Couldn't you get away with just `@[ext]` on the structure definition at this point?

nchappe (approved):
This is much nicer, thanks!

Veir/Interpreter/Basic.lean
math-fehr: Yes good point, I fixed it now!

PR #600, opened 2026-05-18
Support memory in the equation lemma

math-fehr:
The equation lemma was previously defined in a pure world where operations would not modify memory. Now that we have memory in the interpreter, the equation lemma is now valid only on operations that are pure.

This PR adds the notion of purity on an operation. In particular, this is more general than defining purity on operation types. Then, the equation lemma is updated so that we can only derive information on pure operations. It also removes all remaining sorries about the equation lemma.

### DIFF
--- a/Veir/Interpreter/EquationLemma.lean
+++ b/Veir/Interpreter/EquationLemma.lean
@@ -23,6 +23,46 @@ variable {state : InterpreterState}
 -/
 
 /--
+An operation is *pure* when its interpretation does not depend on, and does not modify, the
+memory state: running it under any memory yields the same result values and control flow, with
+the memory threaded through unchanged.
+
+Concretely, the result under `memory₁` is the result under `memory₂` with the output memory
+rewritten to the input memory.
+-/
+def OperationPtr.Pure (op : OperationPtr) (ctx : IRContext OpCode) : Prop :=
+  ∀ operands memory₁ memory₂,
+    interpretOp' (op.getOpType! ctx) (op.getProperties! ctx (op.getOpType! ctx))
+      (op.getResultTypes! ctx) operands (op.getSuccessors! ctx) memory₁ =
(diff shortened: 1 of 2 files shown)


tobiasgrosser (approved):
Nice.

Veir/Interpreter/EquationLemma.lean
```diff
@@ -22,6 +22,65 @@ variable {state : InterpreterState}
...
+-/
+def OperationPtr.Pure (op : OperationPtr) (ctx : IRContext OpCode) : Prop :=
+  ∀ operands memory₁ memory₂,
+    interpretOp' (op.getOpType! ctx) (op.getProperties! ctx (op.getOpType! ctx))
+        (op.getResultTypes! ctx) operands (op.getSuccessors! ctx) memory₁ =
+    (do
+      let (resValues, _, cf) ←
+        interpretOp' (op.getOpType! ctx) (op.getProperties! ctx (op.getOpType! ctx))
+          (op.getResultTypes! ctx) operands (op.getSuccessors! ctx) memory₂
+      pure (resValues, memory₁, cf))
```
ineol: I'm not a big fan of the do notation on the RHS of the equality. Can you replace it with something like this?

```suggestion
        interpretOp' (op.getOpType! ctx) (op.getProperties! ctx (op.getOpType! ctx))
          (op.getResultTypes! ctx) operands (op.getSuccessors! ctx) memory₂
          |>.map (fun (r, _, cf) => (r, memory₁, cf))
```

ineol (approved):

nchappe (approved):

Veir/Interpreter/EquationLemma.lean
math-fehr: Is this better? I added `Interp.map`, which unfolds into `Option.map` and `UBOr.map`. Happy to also remove this indirection and directly write `Interp.map` as a `match`

PR #601, opened 2026-05-18
[do not merge] isel example

luisacicolini:
one of the patterns we used as an instruction selection examples.


tobiasgrosser:
@luisacicolini, I close this. AFAIU this is now handled in one of our evaluation repos.

PR #602, opened 2026-05-18
interpreter executes llvm.func called main, if it exists

regehr:
sigh... I closed the previous PR about this so I could work on this branch without bothering people, but after I force-pushed to squash, it would not let me re-open it. 

so anyway: I appreciate your patience @math-fehr, I think I've done a better job this time, and also I've replaced the test cases with fresh ones straight out of `mlir-translate` since as you said I had gotten something confused with properties and attributes

I looked into also trying to find a `func.func` called `main`, but the patch got too big. so I'll add that in a different PR.

### DIFF
--- a/VeirInterpret.lean
+++ b/VeirInterpret.lean
@@ -75,17 +75,18 @@ def main (args : List String) : IO Unit := do
       match ctx.verify with
       | .ok _ =>
         let rawCtx : IRContext OpCode := ctx
+        let report : Interp (Array RuntimeValue) → String → IO Unit
+          | some (.ok results), _ => IO.println s!"Program output: {results}"
+          | some .ub,           _ => IO.println "Undefined behavior"
+          | none,             msg => IO.eprintln msg
         match findMainFunc rawCtx op with
         | some mainOp =>
-          match interpretRegion rawCtx (mainOp.getRegion! rawCtx 0) InterpreterState.empty (by sorry) (by sorry) with
-          | some (.ok (_, results)) => IO.println s!"Program output: {results}"
-          | some .ub => IO.println "Undefined behavior"
-          | none => IO.eprintln "Error while interpreting module"
+          let result := bind (interpretRegion rawCtx (mainOp.getRegion! rawCtx 0) InterpreterState.empty (by sorry) (by sorry))
+                             (fun (_, r) => pure r)
+          report result "Error while interpreting module"
         | none =>
-          match interpretModule rawCtx op (by sorry) (by sorry) with
-          | some (.ok results) => IO.println s!"Program output: {results}"
-          | some .ub => IO.println "Undefined behavior"
-          | none => IO.eprintln "Error: No entry point: define a function named 'main' or use top-level executable ops"
+          report (interpretModule rawCtx op (by sorry) (by sorry))
+                 "Error: No entry point: define a function named 'main' or use top-level executable ops"
       | .error errMsg => IO.eprintln s!"Error verifying input program: {errMsg}"
     | .error errMsg =>
       IO.eprintln s!"Error: {errMsg}"
--- a/VeirInterpret.lean
+++ b/VeirInterpret.lean
@@ -29,6 +29,43 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode �
   | .error errMsg =>
     throw s!"Error reading file: {errMsg}"
 
+/-- Returns true if `op` is an `llvm.func` with the given `name`. -/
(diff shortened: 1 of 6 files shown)


tobiasgrosser (approved):
Very neat patch. Let's see what @math-fehr says.

VeirInterpret.lean
```diff
@@ -37,10 +74,19 @@ def main (args : List String) : IO Unit := do
...
       | .ok _ =>
-        match interpretModule ctx op (by sorry) (by sorry) with
-        | some (.ok results) => IO.println s!"Program output: {results}"
-        | some .ub => IO.println "Undefined behavior"
-        | none => IO.eprintln "Error while interpreting module"
+        let rawCtx : IRContext OpCode := ctx
+        let report : Interp (Array RuntimeValue) → String → IO Unit
+          | some (.ok results), _ => IO.println s!"Program output: {results}"
+          | some .ub,           _ => IO.println "Undefined behavior"
+          | none,             msg => IO.eprintln msg
```
math-fehr: Can you move this function declaration outside of this function? I don't think it captures anything so that should be fine (and makes it easier to read the main function).

VeirInterpret.lean
```diff
@@ -37,10 +74,19 @@ def main (args : List String) : IO Unit := do
...
+        match findMainFunc rawCtx op with
+        | some mainOp =>
+          let result := bind (interpretRegion rawCtx (mainOp.getRegion! rawCtx 0) InterpreterState.empty (by sorry) (by sorry))
+                             (fun (_, r) => pure r)
+          report result "Error while interpreting module"
+        | none =>
+          report (interpretModule rawCtx op (by sorry) (by sorry))
+                 "Error: No entry point: define a function named 'main' or use top-level executable ops"
```
math-fehr: This is a bit weird, because if you have executable operations in the main module, but they somehow fail, then you would get this error even though this is not the issue?
I guess this is fine because it is temporary, as we will remove `interpretModule` in favor of `findMainFunc`?

math-fehr (approved):
Looks good to me, I just have a comment about a function that should be moved out of the other function, and otherwise just an error message issue!

VeirInterpret.lean
regehr: hmm, ugh, I didn't think of that. but yes-- as long as the rest of the team is onboard, I would love to make module-level instructions illegal and move all of our tests into either an `llvm.func` or a `func.func` depending on what they're testing. I think those are the only kinds of function that we currently have? I didn't see anything else so far

math-fehr: Yep, fine by me!

regehr: ok, I have resolved this comment by writing earlier some code that I had being going to save for the next PR. what the code does now is to insist on exactly one entry point. that is, either exactly one zero-argument `llvm.func` named `main` or else some top-level instructions. but not both, and not multiple mains. this makes the code longer, but I think it's quite semantically clean now.

I think this extra complexity is OK. in a subsequent PR the module-level operations part will go away, but we'll add code looking for a `func.func` called main, and in that case we still want to enforce the "unique main" condition, and this structure will support that.

VeirInterpret.lean
```diff
@@ -29,6 +29,95 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode 
   | .error errMsg =>
     throw s!"Error reading file: {errMsg}"
 
+/-- Returns true if `op` is an `llvm.func` with the given `name`. -/
+def isFuncWithName (ctx : IRContext OpCode) (op : OperationPtr) (name : String) : Bool :=
+  let opType := op.getOpType! ctx
+  let check : (opCode : OpCode) → propertiesOf opCode → Bool
```
math-fehr, line 35: Can you inline this function? It is used only once AFAIK?

VeirInterpret.lean
```diff
@@ -29,6 +29,95 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode 
...
+  | none => false
+  | some ft =>
+    match ft.val with
+    | .llvmFunctionType funcType => funcType.inputs.isEmpty
+    | _ => false
+
+/-- Returns true if `op` is an `llvm.func @main` with no arguments. -/
+private def isZeroArgMainFunc (ctx : IRContext OpCode) (op : OperationPtr) : Bool :=
+  let opType := op.getOpType! ctx
+  let check : (opCode : OpCode) → propertiesOf opCode → Bool
```
math-fehr: Same here, you should inline this

math-fehr (approved):
Still good to me after the changes! Though most of this code will disappear once we remove support for top level interpretation

PR #603, opened 2026-05-18
C to MLIR compiler driver

regehr:
just a little convenience utility, but it can grow over time. 
eventually we can also invoke the risc-v backend and assembler and linker from here.

### DIFF
--- a/Test/Vcc/demorgan.c
+++ b/Test/Vcc/demorgan.c
@@ -1,0 +1,26 @@
+// REQUIRES: clang, opt, mlir-translate, mlir-opt
+// RUN: ./Tools/vcc -O %s -o - | filecheck %s
+
+unsigned d1(unsigned p, unsigned q) {
+  return ~(~p & ~q);
+}
+
+unsigned short d2(unsigned short p, unsigned short q) {
+  return ~(~p | ~q);
+}
+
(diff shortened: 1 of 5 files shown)


tobiasgrosser (approved):
Nice. Let's see what @math-fehr and others say.

Test/Vcc/demorgan.c
```diff
@@ -0,0 +1,26 @@
+// REQUIRES: clang, opt, mlir-translate, mlir-opt
```
math-fehr: Do you know if there is a way in `lit.cfg` to make these tools necessary for *any* `.c` file, instead of having to manually having to REQUIRES them here?
I'm fine if there is no way to do this, just asking.

math-fehr (approved):
Nice, that will be quite useful!

Test/Vcc/demorgan.c
regehr: ah... there might be a way to do this, I will look into it before merging this

math-fehr: If there is not it's fine btw! It's just that it would be one less thing to think about

regehr: ok, this is solved by at the directory level using a lit.local.cfg! much nicer, thanks

PR #604, opened 2026-05-18
fuzzer and fuzzer driver for the LLVM dialect

regehr:
this generates a random collection of LLVM dialect instructions. it is loop-free to ensure termination, but there is control flow. VeIR doesn't do yet do much of the stuff that would contain bugs that this script can find, but I still think it's useful.

I should add that this code already caught 2 `veir-interpret` bugs: the one where SSA values were not being passed across blocks, and also the one where gigantic shift amounts were causing the interpreter to terminate abnormally!

what I will do soon is modify this code so that it exports generated code to LLVM IR and then cross checks to make sure `veir-interpret` produces the same result as `llubi`. this should be very helpful in validating our semantics against LLVM's!

### DIFF
--- a/Tools/vsmith
+++ b/Tools/vsmith
@@ -1,0 +1,423 @@
+#!/usr/bin/env python3
+"""Differential fuzzer for VeIR optimization passes.
+
+Repeatedly generates random MLIR programs, interprets them before and after
+running an optimizer pipeline, and reports a reproducer if the observable
+outputs differ.
+"""
+
+# TODO
+#
+# - convert both src and tgt to LLVM IR and run llubi, check for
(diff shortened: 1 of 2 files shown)


tobiasgrosser (approved):
Amazing. @math-fehr and others should give feedback.

math-fehr (approved):
Nice! I will not comment on the code, as this is quite big and I think the code quality doesn't matter as much as veir "core" code!
So to me it's just nice to have one of these tools, however it is written!

regehr:
right-- I'm committing to maintain this. if it ever becomes a problem for any reason we can delete it, and still win if it has provided value in the meantime.

PR #605, opened 2026-05-19
`VEIR_ROUNDTRIP`: ensure `veir-opt` can reparse its output

tobiasgrosser:
For this we parse and reprint each mlir file used in `VEIR_ROUNDTRIP` using `veir-opt`.

We also extend `veir-opt` to support reading from stdin if either no command line argument is given or if the filename is `-`. This is consistent with the behaviour of `mlir-opt`.

### DIFF
--- a/Veir/Parser/ParserError.lean
+++ b/Veir/Parser/ParserError.lean
@@ -106,13 +106,12 @@ def lineContaining (input : ByteArray) (loc : Nat) : Option String := Id.run do
   When `loc` is `none`, the header uses `<unknown location>` and no source line
   is printed.
 -/
-def formatLabel (filename : Option String) (input : ByteArray) (severity : String) (loc : Option Nat)
+def formatLabel (filename : String) (input : ByteArray) (severity : String) (loc : Option Nat)
     (msg : String) : String :=
   match loc with
   | none => s!"<unknown location>: {severity}: {msg}"
   | some loc =>
     let (line, col) := byteOffsetToLineCol input loc
-    let filename := if let some f := filename then f else "<stdin>"
     let header := s!"{filename}:{line}:{col}: {severity}: {msg}"
     match lineContaining input loc with
     | some srcLine =>
@@ -128,7 +127,7 @@ public section
   Render one primary `error` label followed by a `note` label per entry in
   `e.notes`, each via `formatLabel`. The primary location is taken from `e.pos`.
 -/
-def format (e : ParserError) (filename : Option String) (input : ByteArray) : String :=
+def format (e : ParserError) (filename : String) (input : ByteArray) : String :=
   let primary := formatLabel filename input "error" e.pos e.msg
   e.notes.foldl
     (fun acc note => acc ++ "\n" ++ formatLabel filename input "note" (some note.pos) note.msg)
--- a/Veir/Parser/ParserError.lean
+++ b/Veir/Parser/ParserError.lean
@@ -106,12 +106,13 @@ def lineContaining (input : ByteArray) (loc : Nat) : Option String := Id.run do
   When `loc` is `none`, the header uses `<unknown location>` and no source line
   is printed.
 -/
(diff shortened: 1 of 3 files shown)


regehr:
LGTM!

VeirOpt.lean
```diff
@@ -59,14 +59,29 @@ def parsePipelineOption (args : List String) : Except String (PassPipeline OpCod
...
+
+  if positional.length == 0 then -- read from stdin
+    return VeirOptArgs.mk "-" pipeline
+
+  let [filename] := positional
+    | .error "Expected exactly one positional argument for the input filename."
+
   return VeirOptArgs.mk filename pipeline
 
+def getFileContent (filename : String) : ExceptT String IO ByteArray := do
```
math-fehr: Can you make this an Option String instead?

math-fehr (approved):
Looks good to me, can you just modify the string to an option string, to better convey the difference when we expect a file or not?

VeirOpt.lean
tobiasgrosser: This requires a number of changes. Feedback on those would be appreciated.

tobiasgrosser:
I think this is good-to-go. However, I would appreciate a final look from @math-fehr.

Veir/Parser/ParserError.lean
```diff
@@ -106,12 +106,13 @@ def lineContaining (input : ByteArray) (loc : Nat) : Option String := Id.run do
   When `loc` is `none`, the header uses `<unknown location>` and no source line
   is printed.
 -/
-def formatLabel (filename : String) (input : ByteArray) (severity : String) (loc : Option Nat)
+def formatLabel (filename : Option String) (input : ByteArray) (severity : String) (loc : Option Nat)
```
math-fehr: I think it would be better if we keep a string here, and then move the none logic in veir-opt?

VeirOpt.lean
math-fehr, line 98: So here we would check for none I guess, which makes sense since the caller knows it's stdin

math-fehr (commented):
What do you think of this?

Veir/Parser/ParserError.lean
tobiasgrosser: OK.

VeirOpt.lean
tobiasgrosser: Done.

tobiasgrosser:
@math-fehr, I gave this a shot. What do you think?

VeirOpt.lean
```diff
@@ -59,16 +59,36 @@ def parsePipelineOption (args : List String) : Except String (PassPipeline OpCod
...
-def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode × OperationPtr) := do
-  let fileContent ← IO.FS.readBinFile filename
+  if positional.length == 0 then -- read from stdin
+    return VeirOptArgs.mk none pipeline
+
+  let [filename] := positional
+    | .error "Expected exactly one positional argument for the input filename."
+
+  if filename == "-" then
+    return VeirOptArgs.mk none pipeline
```
math-fehr, line 72: This should be removed I think

math-fehr (approved):
Thanks! Sorry for the nits, should be good now (besides 2 lines that I think should be removed

VeirOpt.lean
tobiasgrosser: Why do you think this should be removed? `mlir-opt` reads from stdin when given `-`. Is there a reason to behave differently?

math-fehr: Oh, I did not know `mlir-opt` is doing this, but to me this is bad design, undocumented, and probably not something we should follow in general.
I think there is simply no reason to keep this feature to be honest, but if you want to keep it let's do this, I don't think this is that important (but then we should document this).

tobiasgrosser: Let's keep it. It`s actually pretty common, also exists in clang and gcc at least for `stdout`.

PR #609, opened 2026-05-19
Parse a few basic RISCV branching operations

nchappe:
This is based on the xDSL riscv_cf dialect: https://docs.xdsl.dev/reference/dialects/riscv_cf/

### DIFF
--- a/Test/RISCV_Cf/identity.mlir
+++ b/Test/RISCV_Cf/identity.mlir
@@ -1,0 +1,27 @@
+// RUN: VEIR_ROUNDTRIP
+
+"builtin.module"() ({
+  ^4():
+    "func.func"() ({
+      ^6(%arg6_0 : i32):
+        "riscv_cf.branch"(%arg6_0) [^7] : (i32) -> ()
+      ^7(%arg7_0 : i32):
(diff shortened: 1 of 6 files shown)


regehr:
did you mean to use `cf.*` instead of `riscv_cf.*` in the test? if not, then this will want to be adjusted! I'm not sure that you tested your new code at all

Test/RISCV_Cf/identity.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: VEIR_ROUNDTRIP
+
+"builtin.module"() ({
+  ^4():
+    "func.func"() ({
+      ^6(%arg6_0 : i32):
+        "cf.branch"(%arg6_0) [^7] : (i32) -> ()
```
math-fehr: These should be `riscv_cf` instead of `cf`.

I'll add a flag in another PR so that by default we don't accept unregistered operations anymore, that should catch these.

math-fehr (approved):
All good to me if the new test can pass!

nchappe:
Good catch, I was too used to veir-opt rejecting unknown operations so I didn't properly check the spelling. This should be fixed now.

PR #610, opened 2026-05-19
simple common subexpression elimination

regehr:
this is a single-BB value numbering implementation. I have tested it with vsmith and also it looks correct to me.

### DIFF
--- a/Test/Passes/CSE/exact_and_misc_flags.mlir
+++ b/Test/Passes/CSE/exact_and_misc_flags.mlir
@@ -1,0 +1,83 @@
+// RUN: veir-opt %s -p=cse | filecheck %s
+
+// Tests that the CSE pass treats exact/disjoint/nneg flags as part of an
+// operation's identity.
+
+"builtin.module"() ({
+  "llvm.func"() ({
+
+// `lshr exact` is distinct from plain `lshr`.
+^lshr_exact(%m : i32, %n : i32):
+    %lshr_exact_1 = "llvm.lshr"(%m, %n) <{exact}> : (i32, i32) -> i32
+    %lshr_exact_2 = "llvm.lshr"(%m, %n) <{exact}> : (i32, i32) -> i32
+    %lshr_plain   = "llvm.lshr"(%m, %n) : (i32, i32) -> i32
+    "test.test"(%lshr_exact_1, %lshr_exact_2, %lshr_plain) : (i32, i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i32, %{{.*}} : i32):
+    // CHECK-NEXT: %[[LSHR_EXACT:.*]] = "llvm.lshr"(%{{.*}}, %{{.*}}) <{exact}> : (i32, i32) -> i32
+    // CHECK-NEXT: %[[LSHR_PLAIN:.*]] = "llvm.lshr"(%{{.*}}, %{{.*}}) : (i32, i32) -> i32
+    // CHECK-NEXT: "test.test"(%[[LSHR_EXACT]], %[[LSHR_EXACT]], %[[LSHR_PLAIN]])
+
+// `ashr exact` is distinct from plain `ashr`.
+^ashr_exact(%o : i32, %p : i32):
+    %ashr_exact = "llvm.ashr"(%o, %p) <{exact}> : (i32, i32) -> i32
+    %ashr_plain = "llvm.ashr"(%o, %p) : (i32, i32) -> i32
+    "test.test"(%ashr_exact, %ashr_plain) : (i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i32, %{{.*}} : i32):
+    // CHECK-NEXT: %[[ASHR_EXACT:.*]] = "llvm.ashr"(%{{.*}}, %{{.*}}) <{exact}> : (i32, i32) -> i32
+    // CHECK-NEXT: %[[ASHR_PLAIN:.*]] = "llvm.ashr"(%{{.*}}, %{{.*}}) : (i32, i32) -> i32
+    // CHECK-NEXT: "test.test"(%[[ASHR_EXACT]], %[[ASHR_PLAIN]])
+
+// `udiv exact` and plain `udiv` are distinct; `udiv exact` duplicates merge.
+// Going from "not exact" to "exact" would introduce UB where there was none
+// if the dividend isn't actually divisible - keying them distinctly sidesteps
+// that.
+^udiv_exact(%udiv_a : i32, %udiv_b : i32):
+    %udiv_exact_1 = "llvm.udiv"(%udiv_a, %udiv_b) <{exact}> : (i32, i32) -> i32
+    %udiv_exact_2 = "llvm.udiv"(%udiv_a, %udiv_b) <{exact}> : (i32, i32) -> i32
+    %udiv_plain   = "llvm.udiv"(%udiv_a, %udiv_b) : (i32, i32) -> i32
+    "test.test"(%udiv_exact_1, %udiv_exact_2, %udiv_plain) : (i32, i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i32, %{{.*}} : i32):
+    // CHECK-NEXT: %[[UDIV_EXACT:.*]] = "llvm.udiv"(%{{.*}}, %{{.*}}) <{exact}> : (i32, i32) -> i32
+    // CHECK-NEXT: %[[UDIV_PLAIN:.*]] = "llvm.udiv"(%{{.*}}, %{{.*}}) : (i32, i32) -> i32
+    // CHECK-NEXT: "test.test"(%[[UDIV_EXACT]], %[[UDIV_EXACT]], %[[UDIV_PLAIN]])
+
+// Same for `sdiv exact`.
+^sdiv_exact(%sdiv_a : i32, %sdiv_b : i32):
+    %sdiv_exact_1 = "llvm.sdiv"(%sdiv_a, %sdiv_b) <{exact}> : (i32, i32) -> i32
+    %sdiv_exact_2 = "llvm.sdiv"(%sdiv_a, %sdiv_b) <{exact}> : (i32, i32) -> i32
+    %sdiv_plain   = "llvm.sdiv"(%sdiv_a, %sdiv_b) : (i32, i32) -> i32
+    "test.test"(%sdiv_exact_1, %sdiv_exact_2, %sdiv_plain) : (i32, i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i32, %{{.*}} : i32):
+    // CHECK-NEXT: %[[SDIV_EXACT:.*]] = "llvm.sdiv"(%{{.*}}, %{{.*}}) <{exact}> : (i32, i32) -> i32
+    // CHECK-NEXT: %[[SDIV_PLAIN:.*]] = "llvm.sdiv"(%{{.*}}, %{{.*}}) : (i32, i32) -> i32
+    // CHECK-NEXT: "test.test"(%[[SDIV_EXACT]], %[[SDIV_EXACT]], %[[SDIV_PLAIN]])
+
+// `or disjoint` is commutative and partitions identity by flag.
+^or_disjoint(%q : i32, %r : i32):
+    %or_disjoint_1 = "llvm.or"(%q, %r) <{disjoint}> : (i32, i32) -> i32
+    %or_disjoint_2 = "llvm.or"(%r, %q) <{disjoint}> : (i32, i32) -> i32
+    %or_plain      = "llvm.or"(%q, %r) : (i32, i32) -> i32
+    "test.test"(%or_disjoint_1, %or_disjoint_2, %or_plain) : (i32, i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i32, %{{.*}} : i32):
+    // CHECK-NEXT: %[[OR_DISJOINT:.*]] = "llvm.or"(%{{.*}}, %{{.*}}) <{disjoint}> : (i32, i32) -> i32
+    // CHECK-NEXT: %[[OR_PLAIN:.*]] = "llvm.or"(%{{.*}}, %{{.*}}) : (i32, i32) -> i32
+    // CHECK-NEXT: "test.test"(%[[OR_DISJOINT]], %[[OR_DISJOINT]], %[[OR_PLAIN]])
+
+// `zext nneg` is distinct from plain `zext`.
+^zext_nneg(%s : i8):
+    %zext_nneg_1 = "llvm.zext"(%s) <{nneg}> : (i8) -> i32
+    %zext_nneg_2 = "llvm.zext"(%s) <{nneg}> : (i8) -> i32
+    %zext_plain  = "llvm.zext"(%s) : (i8) -> i32
+    "test.test"(%zext_nneg_1, %zext_nneg_2, %zext_plain) : (i32, i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i8):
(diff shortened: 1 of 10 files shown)


Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  * it only supports the LLVM dialect (arith would be trivial to add)
+  * it does not use a worklist or iterate to fixpoint, so it may leave
+    work undone when it finishes
+-/
+
+abbrev CSEKind := (op : OpCode) × propertiesOf op
+
+instance (opCode : OpCode) : DecidableEq (propertiesOf opCode) := by
+  simp only [properties_of]
+  cases opCode <;> (try simp) <;> (rename_i op; cases op <;> infer_instance)
```
math-fehr:
```suggestion
  cases opCode <;> (try simp only) <;> (rename_i op; cases op <;> infer_instance)
```
This makes sure you are not using weird `simp` rules and instead only use `simp` to simplify `match`

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  This pass implements a small, conservative CSE:
+  * it only reasons within one basic block;
+  * it only considers arithmetic operations;
+  * distinct UB flags are treated as distinct instructions
+  * it only supports the LLVM dialect (arith would be trivial to add)
+  * it does not use a worklist or iterate to fixpoint, so it may leave
+    work undone when it finishes
+-/
+
+abbrev CSEKind := (op : OpCode) × propertiesOf op
```
math-fehr: Can you add a comment about what this is supposed to be used for?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  cases opCode <;> (try simp) <;> (rename_i op; cases op <;> infer_instance)
+
+instance : Hashable CSEKind where
+  hash k := mixHash (hash k.fst) (hash k.snd)
+
+structure CSEKey where
+  kind : CSEKind
+  resultType : TypeAttr
+  operands : Array ValuePtr
+deriving DecidableEq, BEq, Hashable
```
math-fehr: Should these maybe go in the `CSE` namespace as well?
You would probably want to remove the `CSE` prefix then though

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+abbrev CSEKind := (op : OpCode) × propertiesOf op
+
+instance (opCode : OpCode) : DecidableEq (propertiesOf opCode) := by
+  simp only [properties_of]
+  cases opCode <;> (try simp) <;> (rename_i op; cases op <;> infer_instance)
+
+instance : Hashable CSEKind where
+  hash k := mixHash (hash k.fst) (hash k.snd)
+
+structure CSEKey where
```
math-fehr: Can you add documentation on what this is supposed to be used for?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+deriving DecidableEq, BEq, Hashable
+
+namespace CSE
+
+def isIntegerType (type : TypeAttr) : Bool :=
+  match type.val with
+  | .integerType _ => true
+  | _ => false
+
+def valueSortKey (value : ValuePtr) : Nat × Nat × Nat :=
```
math-fehr, line 54: Can you add documentation on most of the functions actually? For instance here it would be quite useful to explain what this function returns (or more importantly what is the point of this function).

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  guard (isIntegerType resultType)
+  let operands := op.getOperands! ctx
+  guard (allOperandsInteger ctx operands)
+  let opType := op.getOpType! ctx
+  let kind : CSEKind := ⟨opType, op.getProperties! ctx opType⟩
+  match opType with
+  | .llvm .mlir__constant =>
+      return ordinaryKey ctx op kind
+  | .llvm .add | .llvm .mul | .llvm .and | .llvm .or | .llvm .xor =>
+      guard (operands.size = 2)
```
math-fehr: Are these guards about the number of operands matter here? I think it should still work without the guard right?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+      return ordinaryKey ctx op kind
+  | .llvm .zext | .llvm .sext | .llvm .trunc =>
+      guard (operands.size = 1)
+      return ordinaryKey ctx op kind
+  | .llvm .select =>
+      guard (operands.size = 3)
+      return ordinaryKey ctx op kind
+  | _ => none
+
+abbrev AvailableExprs := Std.HashMap CSEKey ValuePtr
```
math-fehr: Is there a reason why the values are `ValuePtr` instead of `OperationPtr`?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+
+set_option warn.sorry false in
+def processBlock
```
math-fehr, line 114: As far as I know, you are not using the worklist in the `PatternRewriter` here.
So instead, can you pass an `WfIRContext OpCode`, and use the `WfRewriter` operations?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+    match key? rewriter.ctx.raw op with
+    | some key =>
+        match available[key]? with
+        | some earlier =>
+            rewriter ← rewriter.replaceValue (op.getResult 0) earlier sorry sorry
+            if !op.hasUses! rewriter.ctx.raw then
+              rewriter ← rewriter.eraseOp op sorry sorry sorry
+        | none =>
+            available := available.insert key (op.getResult 0)
+    | none => pure ()
```
math-fehr: A better "syntax" to use for this `match` is:
```
if let some key := key? rewriter.ctx.raw op
  ...
```

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  let mut available : AvailableExprs := Std.HashMap.emptyWithCapacity
+  let mut current := (block.get! rewriter.ctx.raw).firstOp
+  while current.isSome do
+    let op := current.get!
+    let next := (op.get! rewriter.ctx.raw).next
+    match key? rewriter.ctx.raw op with
+    | some key =>
+        match available[key]? with
+        | some earlier =>
+            rewriter ← rewriter.replaceValue (op.getResult 0) earlier sorry sorry
```
math-fehr: Here you are assuming that operations have all 1 result. Can you either document this at the top of the file, or handle the multiple result case? I forgot if I did implement this but `rewriter.replaceOp` should exist.

math-fehr (commented):
Very cool! I added comments mostly for style, but otherwise this look very nice!

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  guard (allOperandsInteger ctx operands)
+  let opType := op.getOpType! ctx
+  let kind : CSEKind := ⟨opType, op.getProperties! ctx opType⟩
+  match opType with
+  | .llvm .mlir__constant =>
+      return ordinaryKey ctx op kind
+  | .llvm .add | .llvm .mul | .llvm .and | .llvm .or | .llvm .xor =>
+      guard (operands.size = 2)
+      commutativeKey ctx op kind
+  | .llvm .sub | .llvm .icmp
```
math-fehr: Note, `icmp` is commutative for certain predicates

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+  match type.val with
+  | .integerType _ => true
+  | _ => false
+
+def valueSortKey (value : ValuePtr) : Nat × Nat × Nat :=
+  match value with
+  | .opResult result => (0, result.op.id, result.index)
+  | .blockArgument arg => (1, arg.block.id, arg.index)
+
+def lexNat3LE (lhs rhs : Nat × Nat × Nat) : Bool :=
```
ineol: Shouldn't it be `lexNat3LT`?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,162 @@
...
+def orderedBinaryOperands (lhs rhs : ValuePtr) : Array ValuePtr :=
+  if valueLE lhs rhs then #[lhs, rhs] else #[rhs, lhs]
+
+def allOperandsInteger (ctx : IRContext OpCode) (operands : Array ValuePtr) : Bool :=
+  operands.all fun operand => isIntegerType (operand.getType! ctx)
+
+def commutativeKey
+    (ctx : IRContext OpCode) (op : OperationPtr) (kind : CSEKind) :
+    Option CSEKey := do
+  let operands := op.getOperands! ctx
```
ineol: the first parameters should start on the same line as `commutativeKey` I think

Veir/Passes/CSE.lean
ineol: If you rebase on `main` this instance should not be needed anymore

tobiasgrosser:
@regehr, it is not clear if you are still working on this or if you are waiting for another round of reviews. Given that some comments are still "unresolved" and there is no final comment where you hand this PR back to the reviewers, I assume you are still working on this. Can you add an explicit comment to indicate when this is ready for another round of reviews?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,161 @@
...
+  * distinct UB flags are treated as distinct instructions
+  * it only supports the LLVM dialect (arith would be trivial to add)
+  * it does not use a worklist or iterate to fixpoint, so it may leave
+    work undone when it finishes
+-/
+
+abbrev CSEKind := (op : OpCode) × propertiesOf op
+
+instance (opCode : OpCode) : DecidableEq (propertiesOf opCode) := by
+  infer_instance
```
ineol: You should be able to delete this instance completely, as it is defined in an imported module.

Veir/Passes/CSE.lean
ineol: Why not move this type in the `CSE` namespace and call it `Key`?

regehr:
> @regehr, it is not clear if you are still working on this or if you are waiting for another round of reviews. Given that some comments are still "unresolved" and there is no final comment where you hand this PR back to the reviewers, I assume you are still working on this. Can you add an explicit comment to indicate when this is ready for another round of reviews?

Yes, sorry, I've converted it to a draft for now. 
I'm really at the edge of what I know what to do here and I want to try to take time to learn from Mathieu's and Leo's comments!

Veir/Passes/CSE.lean
regehr: I think LE is accurate! the first two comparisons are LT, which is fine-- it's the third one being LE that makes the entire function LE. or am I missing something?

Test/Passes/CSE/flags.mlir
```diff
@@ -0,0 +1,161 @@
...
+"builtin.module"() ({
+
+// `add nsw` merges with `add nsw`; flagless `add` stays separate.
+^add_nsw(%a : i32, %b : i32):
+    %add_nsw_1 = "llvm.add"(%a, %b) <{nsw}> : (i32, i32) -> i32
+    %add_nsw_2 = "llvm.add"(%a, %b) <{nsw}> : (i32, i32) -> i32
+    %add_plain = "llvm.add"(%a, %b) : (i32, i32) -> i32
+    "test.test"(%add_nsw_1, %add_nsw_2, %add_plain) : (i32, i32, i32) -> ()
+
+    // CHECK:      %[[ADD_NSW:.*]] = "llvm.add"(%{{.*}}, %{{.*}}) <{nsw}> : (i32, i32) -> i32
```
tobiasgrosser: Do we need a `CHECK-LABEL` here to ensuer only two `llvm.add` are produced?

regehr:
ok-- I have substantially revised this pass, adding comments, addressing feedback, etc. I think it is a little nicer now!

Test/Passes/CSE/exact_and_misc_flags.mlir
```diff
@@ -0,0 +1,83 @@
...
+  "llvm.func"() ({
+
+// `lshr exact` is distinct from plain `lshr`.
+^lshr_exact(%m : i32, %n : i32):
+    %lshr_exact_1 = "llvm.lshr"(%m, %n) <{exact}> : (i32, i32) -> i32
+    %lshr_exact_2 = "llvm.lshr"(%m, %n) <{exact}> : (i32, i32) -> i32
+    %lshr_plain   = "llvm.lshr"(%m, %n) : (i32, i32) -> i32
+    "test.test"(%lshr_exact_1, %lshr_exact_2, %lshr_plain) : (i32, i32, i32) -> ()
+
+    // CHECK-LABEL: ^{{.*}}(%{{.*}} : i32, %{{.*}} : i32):
```
tobiasgrosser, line 16: In the future (when we support native syntax), we should likely use functions as boundaries. However, for now this may be too verbose. Happy to keep it this way.

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,158 @@
...
+    let op := current.get!
+    let next := (op.get! ctx.raw).next
+    if let some key := key? ctx.raw op then
+        match available[key]? with
+        | some earlier =>
+            match WfRewriter.replaceOp? ctx op earlier sorry sorry sorry sorry sorry with
+            | some ctx' =>
+                ctx := ctx'
+            -- TODO remove this panic as soon as possible!
+            | none => panic! "CSE: replaceOp failed"
```
tobiasgrosser: This can be resoled with a rebase, no?

tobiasgrosser (approved):

tobiasgrosser:
Very cool! Let's see what @math-fehr says.

Veir/Passes/CSE.lean
regehr: yes!

Test/Passes/CSE/exact_and_misc_flags.mlir
regehr: ok! I'll leave this for now but happy to rewrite it later

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,155 @@
...
+  let mut available : Std.HashMap Key OperationPtr := Std.HashMap.emptyWithCapacity
+  let mut current := (block.get! ctx.raw).firstOp
+  while current.isSome do
+    let op := current.get!
+    let next := (op.get! ctx.raw).next
+    if let some key := key? ctx.raw op then
+        match available[key]? with
+        | some earlier =>
+            ctx := WfRewriter.replaceValue ctx (op.getResult 0) (earlier.getResult 0) sorry sorry sorry
+            ctx := WfRewriter.eraseOp ctx op sorry sorry sorry
```
regehr, line 127: @math-fehr I think line 128 is a good use of your changed API?

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,155 @@
...
+abbrev Kind := (op : OpCode) × propertiesOf op
+
+instance : Hashable Kind where
+  hash k := mixHash (hash k.fst) (hash k.snd)
+
+/- This is the basis for CSE: if two instructions have the same Key,
+   then they compute the same value: they are interchangeable. In that
+   case we can remove one of them and switch all of its uses to the
+   other. Proving this will be the crux of the eventual correctness
+   proof for this pass. -/
```
ineol: All these comments should probably be doc comments with `/-- -/`

Veir/Passes/CSE.lean
```diff
@@ -0,0 +1,155 @@
...
+            available := available.insert key op
+    current := next
+  return ctx
+
+set_option warn.sorry false in
+def processAllBlocks (ctx : WfIRContext OpCode) :
+    WfIRContext OpCode := Id.run do
+  let mut ctx := ctx
+  for block in ctx.raw.blocks.keys do
+    if block.InBounds ctx.raw then
```
ineol: isn't this check redundant?

ineol (approved):

Veir/Passes/CSE.lean
math-fehr: Yes!

math-fehr (approved):
All good for me as well, thanks!

PR #611, opened 2026-05-19
properties for func.func

regehr:
we need sym_name on func.func if we're going to have the interpreter find main().

and of course, I hope we can support function calls in the interpreter soon!

this PR mirrors my earlier one for llvm.func.


tobiasgrosser (approved):
LGTM. Let's see if others have further input.

nchappe (approved):

luisacicolini (approved):

math-fehr (approved):
Nice! Thanks

PR #612, opened 2026-05-20
vsmith: put instructions into a function instead of directly into the module

regehr:
I don't think anyone needs to review this since it only changes vsmith, but I'll wait before merging it in case anyone objects


tobiasgrosser (approved):
LGTM

tobiasgrosser:
I am a fan of following LLVM's post-commit review policy. Mathieu prefers explicit reviews. Maybe we can clarify this on our call tomorrow?

PR #613, opened 2026-05-20
Add an extensible array as a new Lake package

ineol:


nchappe:
Why not a separate repository?

Also, the `.envrc` and the `compiler` script should probably not be there?

ineol:
> Also, the `.envrc` and the `compiler` script should probably not be there?

So the the envrc needs to have been sourced when compiling dependencies of `ExArray` so I thought it makes sense to put in in the top level. I put `compiler` in the subfolder to minimize the amount of noise. Where do you think it should be?

nchappe:
These files are Mac/Brew-specific, so I had assumed that you pushed them by mistake, but I've not looked into it in more detail

nchappe:
Ok they are actually not Mac-specific, I didn't see the bash conditional at first, sorry.
Does `pkg-config --cflags-only-I gmp` work with Mac/Brew? This could be a bit more platform-agnostic.

ineol:
It did :)

nchappe:
In `blit`/`buffed_ex_array_blit` there is a `numBytes` argument that is divided by 8 on the C++ side, why is that? Also, does it behave well when `numBytes` is not a multiple of 8?

ExArray/c/ffi_static.cpp
```diff
@@ -0,0 +1,238 @@
...
+    lean_object * r = buffed_ex_array_ensure_exclusive(a);
+    uint64_t * p = (uint64_t *)(buffed_ex_array_cptr(r) + offset);
+    *p = x;
+    return r;
+}
+
+lean_obj_res buffed_ex_array_read(b_lean_obj_arg buf, uint64_t n, uint64_t len) {
+    if (len < 8) { // scalar
+      uint64_t res = *(uint64_t *)(buffed_ex_array_cptr(buf) + n);
+      res = res & ~(-1 << (len * 8));
```
math-fehr:
```suggestion
      res = res & ~((uint64_t)(-1) << (len * 8));
```

ExArray/lean/ExArray/Basic.lean
```diff
@@ -0,0 +1,606 @@
+module
+
+public import Std.Tactic.BVDecide
+meta import Std.Tactic.BVDecide
+public import ExArray.Ranges
+
```
math-fehr, line 6: Can you add documentation on top of the file to explain what this file defines? More importantly that `ExArray` is the main data structure defined here.

ExArray/lean/ExArray/Basic.lean
```diff
@@ -0,0 +1,606 @@
...
+theorem Nat_mod_mod_eq_mod_mod_of_two_pow {a b c: Nat} :
+    a % 2 ^ b % 2 ^ c = a % 2 ^ c % 2 ^ b := by
+  grind only [Nat.pow_dvd_pow, Nat.mod_mod_eq_mod_mod_of_dvd]
+
+-- Upstream?
+def _root_.BitVec.ofFn {w : Nat} (f : Fin w → Bool) : BitVec w :=
+  BitVec.iunfoldr (fun i _ => ((), f i)) () |>.2
+
+structure ExArray where
```
math-fehr, line 35: Can you add documentation about what `ExArray` is supposed to be, and how it differs to a Lean `Array`?

math-fehr (commented):
Nice! Most of the code come from our `buffed` experiments, so they are already reviewed from then.

Can you add in the standard `README.md` the compilation instructions? For Mac and Linux if possible?
On a "fresh" mac I had to install `pkgconf`, `llvm`, and `gmp`

ExArray/c/ffi_static.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+// [[gnu::noinline]]
+lean_obj_res buffed_ex_array_blit(b_lean_obj_arg *w, lean_obj_arg a, size_t offset, uint64_t len, lean_obj_arg o) {
+    len /= 8;
+    lean_object *r = buffed_ex_array_ensure_exclusive(a);
+    uint8_t *p = (uint8_t *)(buffed_ex_array_cptr(r) + offset);
+    if (lean_is_scalar(o)) {
+      uint64_t val = lean_unbox(o);
+      for (int i = 0; i < len; i++) {
+        p[i] = (uint8_t)(val >> i);
```
math-fehr:
```suggestion
        p[i] = (uint8_t)(val >> 8*i);
```

ineol:
@math-fehr please review again!

math-fehr (approved):

GZGavinZhao:
@ineol Did you intend to commit `.envrc`? This breaks build on Linux and if I remove it I can still `lake build ExArray` normally.

GZGavinZhao:
> This breaks build on Linux 

Nevermind, it was my system configuration that was messed up. I don't need this file on Linux to build VeIR though. I assume this is only for MacOS only?

GZGavinZhao:
Oops PR conversation got folded for me, I see the rationale for this file after reading the conversation.

ineol:
> Oops PR conversation got folded for me, I see the rationale for this file after reading the conversation.

no worries! you need to run the test executable to see if it works though, building does not exercise C compiler compatibility

PR #615, opened 2026-05-20
fix width 1 division UB

regehr:
ugh this is embarrassing!! should be better now

this was one of my first veir patches and I leaned too hard on LLM output, I had read the code and thought the width-1 special case was legit, but I didn't read it closely enough nor did I specifically test the i1 case. 100% my fault, I'm being more careful now.

### DIFF
--- a/Test/Interpreter/Arith/divsi_i1_intmin_neg1.mlir
+++ b/Test/Interpreter/Arith/divsi_i1_intmin_neg1.mlir
@@ -1,0 +1,10 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// In i1, -1 is also intMin, so `divsi -1, -1` is immediate UB.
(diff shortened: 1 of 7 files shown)


tobiasgrosser:
> ugh this is embarrassing!! should be better now

Can you replace this with 1-2 sentences explaining what went wrong? How did this get introduced, what was the misunderstanding, and how it is now fixed? Can you also state how this was verified to match the LLVM semantics.

Otherwise, this LGTM.

luisacicolini (approved):

regehr:
> Can you replace this with 1-2 sentences explaining what went wrong? How did this get introduced, what was the misunderstanding, and how it is now fixed? Can you also state how this was verified to match the LLVM semantics.

done!

Issue #519, opened 2026-05-07
Split `instcombine` test file into one file per pattern


math-fehr:

sishtiaq:
A question about the granularity of the split. 

There are three obvious *big* parts to split it up into:
1. multzero_multtwo check; lines 5-12. 
2. id_annihilation check: lines 23-83. 
3. de_morgan check: 84-134. 

Or were you thinking of smaller files? So that, eg, even `~(~a | ~b) => a & b` over lines 123-134 would be in it's own file?

math-fehr:
I would think that having one file per pattern would make it easier to find what tests are defined per pattern, by just grepping the name.
We could separate it by "similar-looking patterns" like demorgan for instance, but then it's very easy to miss tests when you want to find them because you don't know where they are supposed to be.

sishtiaq:
Done in https://github.com/opencompl/veir/pull/616. 
@math-fehr: as the original author, you can close the issue?

sishtiaq:
Done in #616.

PR #616, opened 2026-05-20
Split instcombine test file into one file per pattern (#519).

sishtiaq:
I've split `Test/Passes/instcombine.mlir` into 16 files in `Test/Passes/InstCombine`. 
What I don't know, so would appreciate feedback on:
1. Is this the right granularity to split at? It's basically file-per-pattern, but one person's pattern is another person's multi-pattern:)
2. The names of the files. The other tests in `Test/Passes` aren't super descriptive, so I've kept the new names high-level.


regehr:
this LGTM! although I don't really have any particular strong feelings about files containing fewer vs. more test cases

luisacicolini (approved):
Thank you, lgtm! 
I don't have particularly strong feelings, although looking at [llvm's test organization](https://github.com/llvm/llvm-project/tree/main/llvm/test/Transforms/InstCombine) I wonder if we might want one test per operation and some "special" files for particular rewrites (e.g., de morgan)? Not sure.

sishtiaq:
Added `test.test` in` mult_two.mlir` to use the `mult_two` variable.

math-fehr (approved):
Nice, thanks!

PR #617, opened 2026-05-21
cross-check veir-interpret results against llubi and/or alive-exec

regehr:
this PR only changes vsmith and associated scripts.

this adds the ability to cross-check veir-interpret against llubi and alive-exec, although alive-exec doesn't work very well since it is solver based and is sometimes really slow.

unfortunately I was not able to reuse mlir-translate to convert llvm-dialect MLIR into LLVM IR. I wanted the translation to be an equivalence, but mlir-translate folds constants, eliminating UB, making it only a refinement. so this PR adds a custom llvm-dialect -> LLVM IR translator. it's totally vibe coded but if it ever messes up, a bug will be signaled, so this should not be a problem.

2 bugs found so far: one that I had added to veir earlier and one in llubi, an incorrect poison condition in the code for exact lshr and ashr.

### DIFF
--- a/Tools/veir2llvm
+++ b/Tools/veir2llvm
@@ -1,0 +1,736 @@
+#!/usr/bin/env python3
+"""Translate a small VEIR LLVM-dialect MLIR module to LLVM IR.
+
+This is intentionally not a general MLIR parser.  It accepts one
+``builtin.module`` containing exactly one ``llvm.func`` and supports the small
+integer LLVM subset used by ``Tools/vsmith_generate.py``.
+"""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from dataclasses import dataclass, field
+from pathlib import Path
+
+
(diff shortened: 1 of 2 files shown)


math-fehr (approved):
Nice!
Again, I'll let `veir2llvm` and `vsmith` to be without proper review, as these are the kind of tools where code quality doesn't really matter.

It's a bit sad that `mlir-translate` does these folding :/ I can see why for constants, but do you have an example of a case where it removes UB?

regehr:
> It's a bit sad that `mlir-translate` does these folding :/ I can see why for constants, but do you have an example of a case where it removes UB?

sure, it folds away something like `sdiv 5, 0`. it's probably not hard to avoid generating things that it folds, but remember that we're also feeding the output of `veir-opt` to `mlir-translate` so that tool would need to also not generate anything foldable.

PR #621, opened 2026-05-21
Interpret some RISCV control-flow operations

nchappe:


regehr:
otherwise this LGTM but I think I agree with @math-fehr that it's a bit awkward to have an invisible side-requirement that one of the target blocks is the successor.
one way to avoid this extra requirement would be to permit branching to any 2 blocks in the function, and then doing block shuffling during the conversion to assembly. or does that put too much work into that lowering step?

math-fehr (approved):
Nice, thanks!

tobiasgrosser:
Thank you for looking into this and ensuring we are aligned with xDSL and @superlopuh's dialects.

> otherwise this LGTM but I think I agree with @math-fehr that it's a bit awkward to have an invisible side-requirement that one of the target blocks is the successor. one way to avoid this extra requirement would be to permit branching to any 2 blocks in the function, and then doing block shuffling during the conversion to assembly. or does that put too much work into that lowering step?

I am confused about this side requirement, especially as it is neither mentioned in the PR message or in any comment. At the very least, we should document with comments why this is.

I personally find this requirement worth discussing with @superlopuh, to understand where it comes from. I somehow feel we want at least one version of the RV-IR which does not have this requirement (and maybe one that has). This seems to be an interesting design question clearly worth a conversation.

tobiasgrosser (approved):
(I am happy to see this merged with the appropriate comments)

superlopuh:
Depending on what you care about this may or may not be something that is actually important to you. If you care about the semantics/outputs of code then it's not so important, but if you're trying to actually do instruction selection and care about instruction encoding alignment or whatever then the order of the blocks is material, and should be reflected in your IR directly. One way to resolve this would be to have two levels of ops, maybe in different dialects, one that is "instruction-selected", meaning the exact instruction is fixed, and one that's just encoding jumps in control flow, and doesn't yet care about which instruction is used and in which order the blocks are.

nchappe:
There is no requirement that the second target block is the direct successor in the interpreter / this PR, so we can probably merge it as is.
However this is a requirement in the verifier (#609), because as suggested by the commit message I basically followed the xDSL spec.
We can further discuss the right course of action and I will then open a PR to document or remove the requirement.

PR #622, opened 2026-05-21
support a func.func called main as an interpreter entry point

regehr:
yet another preliminary patch, on our path towards not executing instructions that live directly in the module


Test/Interpreter/Func/main_with_args_no_entry.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: veir-interpret %s 2>&1 | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{sym_name = "main"}> ({
+    ^0(%arg0 : i32):
+      "func.return"(%arg0) : (i32) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Error: No entry point: define a function named 'main' or use top-level executable ops
```
math-fehr: The error should likely be different here? Is the problem that there is an argument?

math-fehr (approved):
Nice! I just added a comment on something that should probably be fixed in another PR (because it is a bit unrelated), but otherwise all good!

math-fehr:
Let me know if you want to fix this here, otherwise you can just merge it!

Test/Interpreter/Func/main_with_args_no_entry.mlir
regehr: ack... yes I forgot to do that. I'll just add "zero-argument" to this error and use it in both cases

regehr:
> Let me know if you want to fix this here, otherwise you can just merge it!

it's so simple that I just fixed it here-- let me know if you agree that just adding "zero-argument" gives users enough information

math-fehr (approved):
Thanks, all good! I guess the error message is fine for now!

PR #623, opened 2026-05-21
validate VeIR optimizations using alive-tv

regehr:
this program offers a way to cross-validate VeIR's LLVM dialect semantics against those from Alive2


regehr:
I can put tests on this script easily enough, but I'm not sure that I need to, since they'll increasing testing time.
if I don't put any tests on this, it might break, but then presumably someone will eventually notice and fix it.
fine with me either way, but I've omitted the tests for now.

math-fehr (approved):
Oh nice! I think that it is fine if we don't test it, because otherwise we would have to install alive as well?
I'm wondering what @tobiasgrosser thinks about it, but otherwise to me it's good to go!

regehr:
> Oh nice! I think that it is fine if we don't test it, because otherwise we would have to install alive as well? 

thanks!

yes, this would have to be an optional test that is skipped if `alive-tv` (or either of the MLIR tools) is not available in the user's `PATH`

tobiasgrosser (approved):
I think the `right` thing is to properly test this with sufficient checks that the test is only executed when `alive` is around. However, I doubt this is worth the effort at the moment (writing it, reviewing it, increased test setup complexity). I am happy for this to go in as is.

PR #624, opened 2026-05-22
trivial rewrite of some tests to put instructions into a main function

regehr:
This is another step towards eliminating instructions at the top level of MLIR modules. I rewrote 16 test cases to do this.

Unfortunately it's a bit annoying to read the diffs since the instructions wanted to be re-indented. But the intent is that no changes have been made to the bodies of any of these tests, it's purely wrapping the existing instructions in a function.

There's a slight abuse here where one LLVM dialect test is rewritten to use a `func.func` instead of an `llvm.func` because the latter does not support multiple return values. This should be perfectly fine for purposes of testing using `veir-interpret` but the native MLIR tooling rejects this function as is. It will accept it if we add the `function_type` property, but we can just do that later on if we want to process these test cases using `mlir-translate`.

If people are happy with this patch, then I'll continue rewriting tests until we're able to disable interpreter support for module-level instructions.

### DIFF
--- a/Test/Interpreter/Arith/addi.mlir
+++ b/Test/Interpreter/Arith/addi.mlir
@@ -1,10 +1,12 @@
 // RUN: veir-interpret %s | filecheck %s
 
 "builtin.module"() ({
-  %lhs = "arith.constant"() <{ "value" = 3 : i32 }> : () -> i32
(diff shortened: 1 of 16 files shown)


tobiasgrosser (approved):
LGTM

PR #625, opened 2026-05-22
return non-zero to the OS on errors

regehr:
`vcc` is having trouble figuring out when `veir-opt` has failed, because `veir-opt` is terminating with exit code 0 even when it runs into an error.

I have no idea if this specific code is the right way to do it, but: all command line tools need to return non-zero to the OS when they terminate with an error. so if we don't accept this patch, let's hopefully have someone else do this the right way! but this seems to work.

### DIFF
--- a/RunBenchmarks.lean
+++ b/RunBenchmarks.lean
@@ -33,4 +33,4 @@ def main (args : List String) : IO Unit := do
   | some bench => Veir.Benchmarks.runBenchmark bench count (getPCFrom args[2]?)
   | _ =>
     IO.eprintln "Please provide a benchmark name"
-    IO.Process.exit 1
+    IO.Process.exit 2
--- a/RunBenchmarks.lean
+++ b/RunBenchmarks.lean
@@ -31,4 +31,6 @@ def main (args : List String) : IO Unit := do
   let count := getCountFrom args[1]?
   match args[0]? with
   | some bench => Veir.Benchmarks.runBenchmark bench count (getPCFrom args[2]?)
-  | _ => IO.println "Please provide a benchmark name"
+  | _ =>
+    IO.eprintln "Please provide a benchmark name"
+    IO.Process.exit 1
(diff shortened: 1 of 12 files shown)


RunBenchmarks.lean
```diff
@@ -31,4 +31,6 @@ def main (args : List String) : IO Unit := do
   let count := getCountFrom args[1]?
   match args[0]? with
   | some bench => Veir.Benchmarks.runBenchmark bench count (getPCFrom args[2]?)
-  | _ => IO.println "Please provide a benchmark name"
+  | _ =>
+    IO.eprintln "Please provide a benchmark name"
+    IO.Process.exit 1
```
tobiasgrosser:
```suggestion
  | _ =>
    IO.eprintln "Please provide a benchmark name"
    IO.Process.exit 2
```

It seems as if some tools use `2` for being called with wrong command line arguments:
https://stackoverflow.com/a/40484670.

VeirInterpret.lean
```diff
@@ -134,11 +136,17 @@ def main (args : List String) : IO Unit := do
...
+      | .error errMsg =>
+        IO.eprintln s!"Error verifying input program: {errMsg}"
+        IO.Process.exit 1
     | .error errMsg =>
       IO.eprintln s!"Error: {errMsg}"
+      IO.Process.exit 1
   | _ =>
     IO.eprintln "Wrong number of arguments."
     IO.eprintln "Usage: veir-interpret <filename>"
+    IO.Process.exit 1
```
tobiasgrosser:
```suggestion
    IO.Process.exit 2
```
It seems as if some tools use 2 for being called with wrong command line arguments:
https://stackoverflow.com/a/40484670.

tobiasgrosser (approved):
LGTM.

RunBenchmarks.lean
ineol: I think an `if let some bench := args[0]?` would be slightly more idiomatic

math-fehr (approved):

PR #628, opened 2026-05-22
a small code style improvement suggested by Leo

regehr:
no functional change! just more idiomatic


regehr:
this is literally what Leo suggested and it's a tiny NFC so I'll merge w/o review

PR #634, opened 2026-05-23
interpreter: remove support for executing top-level ops

regehr:
we just make a list of entry points. if this list has one member we interpret it, otherwise emit an error.

I still don't have a great handle on idiomatic Lean but I think this is reasonably clean and simple now.

### DIFF
--- a/VeirInterpret.lean
+++ b/VeirInterpret.lean
@@ -30,6 +30,31 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode �
   | .error errMsg =>
     throw s!"Error reading file: {errMsg}"
 
+/-- Returns true if `op` is a viable zero-argument `@main` function. -/
+private def isZeroArgMainFunc (ctx : IRContext OpCode) (op : OperationPtr) : Bool :=
+  let opType := op.getOpType! ctx
+  let check : (opCode : OpCode) → propertiesOf opCode → Bool
+    | .llvm .func, props =>
+      if let some symName := props.sym_name then
+        String.fromUTF8! symName.value == "main" &&
+        match props.function_type with
+        | some ft =>
+          match ft.val with
+          | .llvmFunctionType funcType => funcType.inputs.isEmpty
+          | _ => false
+        | none => false
+      else false
+    | .func .func, props =>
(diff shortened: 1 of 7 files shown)


VeirInterpret.lean
```diff
@@ -29,93 +30,60 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode 
   | .error errMsg =>
     throw s!"Error reading file: {errMsg}"
 
-/-- Returns true if `op` is a viable zero-argument `@main` function. -/
```
math-fehr, line 32: Is there a reason to remove `isZeroArgMainFunc`? I felt the design was better with this logic put in a single function rather than inlined in `scanEntryPoints`

VeirInterpret.lean
```diff
@@ -29,93 +30,60 @@ def parseOperation (filename : String) : ExceptT String IO (WfIRContext OpCode 
...
+    match (region.get! ctx).firstBlock with
+    | none => pure []
+    | some blockPtr => scanEntryPoints ctx (blockPtr.get! ctx).firstOp
+  match entryPoints with
+  | [] =>
+    IO.eprintln "Error: No entry point: define a zero-argument func.func or llvm.func named 'main'"
+    IO.Process.exit 1
+  | [mainOp] => return mainOp
+  | _ =>
+    IO.eprintln "Error: Multiple entry points: define exactly one zero-argument func.func or llvm.func named 'main'"
```
math-fehr: Can we change back the error to ask for a zero-argument `function`, rather than asking for a `func.func` or a `llvm.func`? I am scared that once we add more functions we will have to change all tests all over again.

math-fehr (commented):
I think we should go back to using `isZeroArgMainFunc` here, and change one comment, but otherwise that's nice!

VeirInterpret.lean
regehr: yeah I thought about that, but was just trying to be very specific for people. but you're right that we don't need a ton of test churn, I'll fix this and the other thing later today!

regehr:
ok, feedback addressed, thanks!

tobiasgrosser (approved):

PR #637, opened 2026-05-24
Fix `WfRewriter.replaceValue` to not return an `Option`

math-fehr:
There is one remaining sorry for the function termination that will be removed in the future.


regehr:
let me first try to rephrase what you did here, before we decide whether a code review by me means anything :)

most of the changes are just replacing the ←, which as I understand is similar to `try` in other languages, with direct binding of the result, since it can no longer fail

then the actual change in `WfRewriter.replaceValue` is to replace the old implementation with a simple recursive call and then you've declined to actually prove that this terminates

assuming I understand, LGTM!

regehr (approved):

math-fehr:
Yes that exactly it! I also added a new argument in `replaceValue` which is a proof that the two values are different. Otherwise, this wouldn't be correct.

tobiasgrosser:
LGTM, indeed!

PR #638, opened 2026-05-25
Prove termination of `WfRewriter.replaceValue`

math-fehr:
This adds a few helpers lemmas here and there.

### DIFF
--- a/Veir/Rewriter/WfRewriter/Basic.lean
+++ b/Veir/Rewriter/WfRewriter/Basic.lean
@@ -71,7 +71,6 @@ def WfRewriter.replaceOperand (wfCtx : WfIRContext OpInfo) (use : OpOperandPtr) 
   ⟨Rewriter.replaceUse wfCtx use newValue useIn newIn (by grind),
     by grind [Rewriter.replaceUse_WellFormed]⟩
 
-set_option warn.sorry false in
 /-- Replace a value with another value. -/
 def WfRewriter.replaceValue (ctx: WfIRContext OpInfo) (oldValue: ValuePtr) (newValue: ValuePtr)
     (neValues : oldValue ≠ newValue := by grind)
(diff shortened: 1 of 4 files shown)


tobiasgrosser (approved):

regehr:
I don't understand these theorems at all! I hope you can explain them to me a bit when we talk on Tuesday

Veir/Rewriter/WfRewriter/Basic.lean
```diff
@@ -71,7 +71,6 @@ def WfRewriter.replaceOperand (wfCtx : WfIRContext OpInfo) (use : OpOperandPtr)
   ⟨Rewriter.replaceUse wfCtx use newValue useIn newIn (by grind),
     by grind [Rewriter.replaceUse_WellFormed]⟩
 
-set_option warn.sorry false in
 /-- Replace a value with another value. -/
 def WfRewriter.replaceValue (ctx: WfIRContext OpInfo) (oldValue: ValuePtr) (newValue: ValuePtr)
     (neValues : oldValue ≠ newValue := by grind)
```
ineol, line 76: Should it be a precondition or a dynamic check in the function?

math-fehr: If you want, I can make a `replaceValue?` function that will check this? Or maybe with a different name?

math-fehr: At least, if it is a dynamic check, the question is wether or not we should return an `Option` or do nothing and return the context. If we do nothing, then many lemmas will need to explicitely ask for `oldValue \ne newValue`, as for instance we cannot prove that the old value has no uses anymore.

ineol: What does the og MLIR equivalent do when they are equal?

math-fehr: it's an assert currently:
```
assert((!newValue || this != OperandType::getUseList(newValue)) &&
"cannot RAUW a value with itself");
```

ineol: LGTM then :)

PR #639, opened 2026-05-25
Test `mlir-opt` compatibility in CI

tobiasgrosser:
Add a lit abbreviation `MLIR_ROUNDTRIP` which passes the input file to`mlir-opt` to check that the input is accepted by MLIR. It then passes the output to `veir-opt` and subsequently to `mlir-opt` again, to check that the output of `mlir-opt` is accepted by `veir-opt`, and that the output of `veir-opt` is accepted by `mlir-opt`. Finally, we use `veir-opt` once more to ensure the textual output is identical between `VEIR_ROUNDTRIP` and `MLIR_ROUNDTRIP`.

We enable a single file and will add other files as they are `MLIR_ROUNDTRIP`-clean.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -25,7 +25,7 @@ jobs:
 
       - uses: gerlero/apt-install@v1
         with:
-          packages: mlir-opt-22
+          packages: mlir-tools-22
 
       - name: Rename MLIR-opt
         run: |
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -16,14 +16,16 @@ jobs:
       - name: Source environment variables
         run: source .envrc
 
-      - uses: gerlero/add-apt-repository@v1
+      - name: Add LLVM apt repository
+        uses: gerlero/add-apt-repository@v1
         with:
           uri: http://apt.llvm.org/noble/ 
           key: https://apt.llvm.org/llvm-snapshot.gpg.key
           component: main
           suite: llvm-toolchain-noble-22
 
-      - uses: gerlero/apt-install@v1
+      - name: Install mlir-opt
+        uses: gerlero/apt-install@v1
         with:
           packages: mlir-22-tools
 
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -21,6 +21,7 @@ jobs:
           uri: http://apt.llvm.org/noble/ 
           key: https://apt.llvm.org/llvm-snapshot.gpg.key
           component: main
+          suite: llvm-toolchain-noble
 
       - uses: gerlero/apt-install@v1
         with:
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -60,7 +60,7 @@ jobs:
         run: uv sync
 
       - name: Run filecheck tests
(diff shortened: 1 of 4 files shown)


regehr:
I don't have specific feedback about the code but in general I love roundtrip testing, it gives us a lot of testing power for not a lot of effort!

.github/workflows/lean_action_ci.yml
```diff
@@ -16,6 +16,14 @@ jobs:
       - name: Source environment variables
         run: source .envrc
 
+      - name: Install mlir-opt
+        run: |
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 22
+          sudo apt-get install mlir-22-tools
+          sudo ln -s /usr/bin/mlir-opt-22 /usr/bin/mlir-opt
```
math-fehr, line 34: How long does this take usually? Can we try to do something like in xDSL where we cache this?

tobiasgrosser: Thank you for pointing this out.

Initially, it took about 1m30s. I now added caching and it takes about 10s.

This seems to be as good as it can get.

tobiasgrosser:
@math-fehr, I addressed your comment and would appreciate another round of reviews.

.github/workflows/lean_action_ci.yml
math-fehr: Nice, 10s is great!

math-fehr (approved):
All good to me, thanks!
Should we ask for the user to provide `TEST_MLIR`, or should we do like in xDSL and only enable tests if `mlir-opt` is detected? I feel the second solution would make more sense?

tobiasgrosser:
> All good to me, thanks! Should we ask for the user to provide `TEST_MLIR`, or should we do like in xDSL and only enable tests if `mlir-opt` is detected? I feel the second solution would make more sense?

These tests only work with `mlir-opt` version 22. If someone has an older mlir-opt version installed, the tests may break . Alternatively, we could add a version-check on `mlir-opt` auto-detect it and emit a warning if it is skipped. Would that be preferable?

math-fehr:
If you can auto-detect it, that would be very cool. Maybe you can do a check directly in `lit.cfg.py`?

tobiasgrosser:
> If you can auto-detect it, that would be very cool. Maybe you can do a check directly in `lit.cfg.py`?

Done. Let me know if you have further comments.

Test/lit.cfg
```diff
@@ -16,3 +18,22 @@ config.test_format = lit.formats.ShTest(preamble_commands=[f"cd {veir_root}"])
...
+  match = re.search(r'version\s+([a-zA-Z0-9\.]+)', str(result))
+  if match:
+    version = match.group(1)
+
+  if version == '22.0.0git':
+    print(f"'mlir-opt' version {version} detected. Running MLIR compatibility tests")
+    config.substitutions.append(("MLIR_ROUNDTRIP",
+     "mlir-opt --mlir-print-op-generic --mlir-print-local-scope %s | lake exec veir-opt | mlir-opt --mlir-print-op-generic --mlir-print-local-scope %s | lake exec veir-opt | filecheck %s"))
+  else:
+    print(f"'mlir-opt' version {version} detected. Skipping MLIR compatibility tests")
```
math-fehr: Can you add in the message here that we only support `22.0.0git`?

math-fehr (approved):
Thanks! I just added one comment, feel free to merge this after addressing it

PR #641, opened 2026-05-25
Add properties to llvm.module_flags

tobiasgrosser:
These properties we previously silently discarded.


regehr:
LGTM!

Veir/GlobalOpInfo.lean
```diff
@@ -269,6 +270,10 @@ def Properties.toAttrDict (opCode : OpCode) (props : propertiesOf opCode) :
     if let some function_type := props.function_type then
       dict := dict.insert "function_type".toUTF8 function_type
     dict
+  | .llvm .module_flags => Id.run do
+    let mut dict := Std.HashMap.emptyWithCapacity 3
```
math-fehr, line 274:
```suggestion
    let mut dict := Std.HashMap.emptyWithCapacity 1
```
Since we only insert one thing?

math-fehr (approved):
Nice!

PR #643, opened 2026-05-25
llvm and arith: make the high-level representation for icmp predicates first-class in our IR

regehr:
I think that dealing with predicates represented as integers is awkward and error-prone, so let's not do that

for example, see the pleasing little cleanups this enables in CSE.lean and in RISCV64.lean, below

### DIFF
--- a/Veir/Data/LLVM/Int/Basic.lean
+++ b/Veir/Data/LLVM/Int/Basic.lean
@@ -27,7 +27,7 @@ inductive IntPred where
   | sge
   | slt
   | sle
-deriving DecidableEq, Inhabited
+deriving DecidableEq, Inhabited, Repr, Hashable
 
 /-- Mapped as in MLIR:
   https://github.com/llvm/llvm-project/blob/d3417c8bf35852af88f41aa721a719ea756fdd8c/mlir/include/mlir/Dialect/LLVMIR/LLVMEnums.td#L571 -/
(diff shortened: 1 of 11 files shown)


tobiasgrosser (approved):
Nice. Let's see what @math-fehr says.

Veir/Data/LLVM/Int/Basic.lean
```diff
@@ -45,6 +45,25 @@ def IntPred.fromNat (s : Nat) : Option IntPred :=
...
+  | .ult => 6
+  | .ule => 7
+  | .ugt => 8
+  | .uge => 9
+
+/-- Sanity check: A numeric code parses to a predicate exactly when it
+    is that predicate's MLIR code. -/
+theorem IntPred.fromNat_eq_some_iff {n : Nat} {p : IntPred} :
+    IntPred.fromNat n = some p ↔ p.toNat = n := by
+  cases p <;> simp [IntPred.fromNat, IntPred.toNat] <;> split <;> simp_all <;> omega
```
math-fehr: By any chance, would
```
cases p <;> grind [IntPred.fromNat, IntPred.toNat]
```
work here?
Otherwise, there are a lot of `simp` that may change in the future (I can explain you that tomorrow)

math-fehr (approved):
Nice!

Veir/Data/LLVM/Int/Basic.lean
regehr: alas it says "`grind` failed" and then prints many pages of what look like case splits

math-fehr: let me try to fix it quickly

math-fehr: Okay, this works: `cases p <;> simp only [IntPred.fromNat, IntPred.toNat] <;> grind`

regehr: yayyy- thanks!!

PR #644, opened 2026-05-26
teach CSE that x>y and y<x are the same thing and can be commoned

regehr:


tobiasgrosser (approved):
Nice! Let's see what the others say.

math-fehr (approved):
Nice! We'll have to think at some point if we want to continue deviating from MLIR here, or if we should have one LLVMCSE pass and one CSE pass

PR #645, opened 2026-05-26
Remove parsing support for top-level operations

### LINKED ISSUE #636: Remove parsing support for top-level operations
Previously, `veir-interpret` would run instructions sitting at the top level of a module. This is now disabled. 

Next, we should refuse to parse MLIR modules that contain executable instructions outside of functions. This will require a number of tests to be updated. I don't think this is particularly pressing.

GZGavinZhao:
Fixes #636.

~~I haven't manually reviewed the test case changes, they're coded by DeepSeek.~~ All tests have been updated by hand, AI was unwieldy for this type of mostly but not always copy-and-paste tasks.

### DIFF
--- a/Test/CudaTile/types.mlir
+++ b/Test/CudaTile/types.mlir
@@ -1,10 +1,20 @@
 // RUN: VEIR_ROUNDTRIP
 "builtin.module"() ({
-^bb0(%arg0: !cuda_tile.ptr<i1>):
-    %0 = "test.test"() : () -> !cuda_tile.ptr<i1>
+  // Note: the proper type should be without quotes, i.e.
+  // `!cuda_tile.ptr<i1> -> ()`. Our parser cannot handle this right now, it's a
+  // bug that should be fixed.
+  "func.func"() <{function_type = "!cuda_tile.ptr<i1> -> ()", sym_name = "main"}> ({
+    ^bb0(%arg0: !cuda_tile.ptr<i1>):
+      %0 = "test.test"() : () -> !cuda_tile.ptr<i1>
+      "func.return"() : () -> ()
+  }) : () -> ()
 }) : () -> ()
 
 // CHECK-NEXT: "builtin.module"() ({
-// CHECK-NEXT:   ^{{.*}}(%{{.*}} : !cuda_tile.ptr<i1>):
-// CHECK-NEXT:     %{{.*}} = "test.test"() : () -> !cuda_tile.ptr<i1>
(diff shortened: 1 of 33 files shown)


Test/CudaTile/types.mlir
```diff
@@ -1,10 +1,20 @@
 // RUN: VEIR_ROUNDTRIP
 "builtin.module"() ({
-^bb0(%arg0: !cuda_tile.ptr<i1>):
-    %0 = "test.test"() : () -> !cuda_tile.ptr<i1>
+  // Note: the proper type should be without quotes, i.e.
+  // `!cuda_tile.ptr<i1> -> ()`. Our parser cannot handle this right now, it's a
+  // bug that should be fixed.
+  "func.func"() <{function_type = "!cuda_tile.ptr<i1> -> ()", sym_name = "main"}> ({
```
GZGavinZhao, line 6: This is confirmed with the MLIR source code [here](https://github.com/llvm/llvm-project/blob/5acb952cb5d9c899692f44f4280af75a780a599c/mlir/lib/AsmParser/AttributeParser.cpp#L233): in the attribute value position, if the token isn't a known attribute keyword/literal, the last attempt (i.e. the `default` branch) is to try to parse it as a type.

As a side note, I'm surprised that `func.func` doesn't enforce that the `function_type` attribute must be a `TypeAttr`.

Test/Interpreter/main_and_top_level.mlir
```diff
@@ -9,4 +9,4 @@
   "func.return"(%0) : (i32) -> ()
 }) : () -> ()
 
-// CHECK: Error: Top-level operations are disallowed; define a zero-argument function named 'main'
+// CHECK: Error: Error parsing module: top-level operation with operands or results is not allowed; only function and module definitions may appear at the module level
```
GZGavinZhao, line 12: Just want to check on whether we like this error message or if something else is preferred.

tobiasgrosser (approved):
This LGTM. Let's wait for feedback from @math-fehr.

Test/CudaTile/types.mlir
math-fehr: > As a side note, I'm surprised that func.func doesn't enforce that the function_type attribute must be a TypeAttr.

This is a limitation of our current encoding, and that's because of Lean type theory. I don't understand much of it, but the idea is that if you have a recursive type, you cannot have a function `type -> A` contained in this type.

PR #646, opened 2026-05-26
Add support for LLVM array type

nchappe:
It consists of a size and a type.

Because of how attributes are mutually defined,
this inner type is an `Attribute`.

### DIFF
--- a/Test/LLVM/getelementptr.mlir
+++ b/Test/LLVM/getelementptr.mlir
@@ -15,10 +15,10 @@
       %g2 = "llvm.getelementptr"(%ptr, %i, %j) <{elem_type = !llvm.array<10 x i8>, rawConstantIndices = array<i32: -2147483648, -2147483648>}> : (!llvm.ptr, i64, i64) -> !llvm.ptr
 
       // 3 dynamic indices: ptr[i][j][k]
-      %g3 = "llvm.getelementptr"(%ptr, %i, %j, %k) <{elem_type = !llvm.array<10 x array<10 x i8>>, rawConstantIndices = array<i32: -2147483648, -2147483648, -2147483648>}> : (!llvm.ptr, i64, i64, i64) -> !llvm.ptr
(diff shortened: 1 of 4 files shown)


Test/LLVM/getelementptr.mlir
```diff
@@ -15,10 +15,10 @@
       %g2 = "llvm.getelementptr"(%ptr, %i, %j) <{elem_type = !llvm.array<10 x i8>, rawConstantIndices = array<i32: -2147483648, -2147483648>}> : (!llvm.ptr, i64, i64) -> !llvm.ptr
 
       // 3 dynamic indices: ptr[i][j][k]
-      %g3 = "llvm.getelementptr"(%ptr, %i, %j, %k) <{elem_type = !llvm.array<10 x array<10 x i8>>, rawConstantIndices = array<i32: -2147483648, -2147483648, -2147483648>}> : (!llvm.ptr, i64, i64, i64) -> !llvm.ptr
+      %g3 = "llvm.getelementptr"(%ptr, %i, %j, %k) <{elem_type = !llvm.array<10 x !llvm.array<10 x i8>>, rawConstantIndices = array<i32: -2147483648, -2147483648, -2147483648>}> : (!llvm.ptr, i64, i64, i64) -> !llvm.ptr
```
math-fehr, line 18: Note for another PR. This is technically not needed, we should be able to parse it without the `!llvm` later on.

Veir/IR/Attribute.lean
```diff
@@ -220,6 +220,14 @@ structure DictionaryAttr where
   /- TODO: figure out how to maintain a proof of sorted-ness and uniqueness. -/
 deriving Inhabited, Repr, Hashable
 
+/--
+  An attribute representing a fixed-sized array type
+-/
+structure LLVM.ArrayType where
+  size : Int
```
math-fehr: Should this be a `Nat` rather than an `Int`?

math-fehr (approved):
Nice! I would just change the size to be a `Nat` rather than an `Int`, but otherwise all good! (and I would test that a negative integer fails)

nchappe:
thanks, changed Int to Nat

I would prefer not to test that a negative integer fails, as MLIR does accept it

math-fehr:
Yuck, I looked at MLIR, and they automatically convert `-1` to `4294967295` :(
I'm fine skipping the tests then

PR #647, opened 2026-05-27
Add Parser.Location

math-fehr:
Instead of using `Nat` for representing the location in a file, we wrap it in a newtype. This makes the public-facing API nicer as it gives a better meaning than `Nat`.

### DIFF
--- a/UnitTest/ParserError.lean
+++ b/UnitTest/ParserError.lean
@@ -1,22 +1,23 @@
 module
 
 import all Veir.Parser.ParserError
+public meta import Veir.Parser.Location
 
 open Veir.Parser
 open Veir.Parser.ParserError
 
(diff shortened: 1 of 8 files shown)


UnitTest/ParserError.lean
```diff
@@ -1,22 +1,23 @@
 module
 
 import all Veir.Parser.ParserError
+public meta import Veir.Parser.Location
```
math-fehr, line 4: @ineol, do you know by any chance why this is needed? Without it it tells me that `Location.mk` is not exposed

GZGavinZhao (approved):
Yay! I was quite confused when reading through the this part of the codebase for the first time, trying to juggle all the different `Nat` in my head. This looks much easier to understand.

tobiasgrosser (approved):
Very beautiful. 👍

PR #648, opened 2026-05-27
Add better error messages for values

math-fehr:
For this, we store value definition locations in the parser state.
We now get these errors:
```
already-defined.mlir:5:3: error: value %a has already been defined
  %a = "test.test"() <{second}> : () -> i32
  ^
already-defined.mlir:4:3: note: previously defined here
  %a = "test.test"() <{first}> : () -> i32
  ^
```


tobiasgrosser (approved):
Amazing.

PR #649, opened 2026-05-27
fix inverted logic in DCE

regehr:
hasSideEffects was returning the opposite of what it should have returned. however, since all call sites also had the condition inverted, there was not an actual bug here, just a latent one. easy to fix.


math-fehr (approved):
Oops, good catch!

PR #651, opened 2026-05-27
Minimal interpreter support for `llvm.getelementptr`

nchappe:
For now, it is restricted to exactly one dynamic index


math-fehr (approved):
Nice!

PR #655, opened 2026-05-27
Add `LLVM.Int.cast_self`

math-fehr:
This lemma is useful when reasonning about the interpreter.


math-fehr:
This is stacked because I have a large PR that depends on many of these small PRs

tobiasgrosser (approved):

PR #657, opened 2026-05-27
Improve lemma about verified add and constant operations

math-fehr:
It is usually easier to reason about `typeattr = <_,_>` than `typeattr.val = _`.


math-fehr:
This is stacked because I have a large PR that depends on many of these small PRs

tobiasgrosser (approved):
LGTM

PR #658, opened 2026-05-28
Dominance Public API

axelcool1234:
Provides an API (which utilizes the dominance analysis) for querying dominance information in `IR/Dominance.lean`.

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -1,6 +1,7 @@
 import UnitTest.DataFlowFramework.Helpers
 
 import Veir.Analysis.DataFlow.DominanceAnalysis
+import Veir.IR.Dominance
 
 open Std (HashMap HashSet)
 open Veir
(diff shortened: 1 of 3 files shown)


axelcool1234:
Note, the reason I have both `BlockPtr.getIDom?` in `DominanceAnalysis.lean` and `BlockPtr.immediateDominator?` in `IR/Dominance.lean`, is because `DominanceAnalysis.lean` needs to grab the immediate dominator of blocks in several places (and we can't use `BlockPtr.immediateDominator?` as that'd cause a trivial import cycle). Ideally `getIDom?` could be a private method but only exposed to `IR/Dominance.lean` so `BlockPtr.immediateDominator?` remains as a wrapper over it or something. Would love your thoughts.

regehr:
@axelcool1234 the compare* family of functions is just for testing? if so, can you include comments to this effect?

axelcool1234:
> @axelcool1234 the compare* family of functions is just for testing? if so, can you include comments to this effect?

Yes exactly. I'll make sure to mention this in their documentation

regehr:
btw I love your ASCII art basic block drawings for the test cases!

axelcool1234:
> btw I love your ASCII art basic block drawings for the test cases!

Haha I knew you would 😂 What's great is they explain the test cases better than any description would

regehr:
I think others should also have a look, but this LGTM!

regehr (approved):

axelcool1234:
cc @math-fehr

axelcool1234:
Will roll out the dominance API in multiple PRs.

PR #659, opened 2026-05-28
feat(dataflow): dead code analysis

axelcool1234:
Implements a dataflow analysis for unreachable code.

Note - MLIR has a much more advanced understanding of region semantics I didn't know how to implement with VeIR's current capabilities. I have left a large number of TODOs in places where they should be handled.


Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,310 @@
...
+      pure ()
+
+  dfCtx
+
+def visit
+    (point : InsertPoint)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  if point.prev! irCtx = none then
+    return dfCtx
```
axelcool1234: Go back to MLIR and change this to a panic if needed.

axelcool1234 (commented):
Review for myself

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+    -- callables in nested regions.
+    let region := regionPtr.get! irCtx
+    let mut maybeBlock := region.firstBlock
+    while let some block := maybeBlock do
+      let mut maybeOp := (block.get! irCtx).firstOp
+      while let some nestedOp := maybeOp do
+        dfCtx := initializeRecursively nestedOp dfCtx irCtx
+        maybeOp := (nestedOp.get! irCtx).next
+      maybeBlock := (block.get! irCtx).next
+  dfCtx
```
axelcool1234, line 299: This is a common pattern that I have done many many times, and I know @regehr has done for CSE. Perhaps this should be a helper in `Basic.lean`? @math-fehr

regehr: this does seem like a useful helper. we don't want to return a list-- that's slow. rather is should be some sort of visitor / iterator / callback.

UnitTest/DataFlowFramework/Helpers.lean
```diff
@@ -165,3 +166,45 @@ def runWithAnalyses
       let some dfCtx := fixpointSolve top analyses parserState.ctx
         | return "analysis did not converge"
       return renderReport (check top dfCtx parserState)
+
+/-- Dead code helpers. -/
```
regehr: @axelcool1234 if this code is potentially useful for multiple analyses, then this file seems like a good place to put it. but if it is only helping dead code analysis, then just put it there and don't create this extra file

UnitTest/DataFlowFramework/Helpers.lean
```diff
@@ -165,3 +166,45 @@ def runWithAnalyses
...
+def isBlockLive (dfCtx : DataFlowContext) (block : BlockPtr) (irCtx : IRContext OpCode) : Bool :=
+  match dfCtx.getFact? .executable (.InsertPoint (InsertPoint.atStart! block irCtx)) with
+  | some fact => fact.live
+  | none => false
```
regehr: what are the circumstances in which there would not be a fact available here? can these circumstances be avoided? if not, should you signal a real error here instead of merely returning false?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
math-fehr: Yeah that would be a useful helper, in a file like `Walker.lean` or something.
Just pass a function as input of a function like this!

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -63,6 +65,7 @@ Tags to match on for different fact types.
 inductive FactKind where
   | dominator
   | regionMetadata
+  | executable
```
math-fehr: `liveness` maybe instead of executable?

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -82,22 +85,32 @@ Stored in the entry block of each region.
 structure RegionMetadataPayload where
   postOrderIndex : HashMap BlockPtr Nat := {}
 
+/--
+Tracks whether a control flow point or edge is executable.
```
math-fehr: Same here about `executable`.
In general, executable sounds like its capacity to be executed, not whether or not it is live or not.

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -137,10 +159,26 @@ def setPostOrderIndex (fact : Fact .regionMetadata)
...
+
+def latticeElement (fact : Fact .executable) : Liveness :=
+  fact.payload.latticeElement
+
+def setLatticeElement (fact : Fact .executable) (latticeElement : Liveness) : Fact .executable :=
+  { fact with payload := { fact.payload with latticeElement := latticeElement } }
+
+def setToLive (fact : Fact .executable) : Fact .executable :=
+  fact.setLatticeElement .live
+
```
math-fehr: Now that I'm thinking of it, you should probably move these to a different namespace.
Maybe something like `Fact.Executable.*`?

Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -137,10 +159,26 @@ def setPostOrderIndex (fact : Fact .regionMetadata)
...
+  | .live => true
+
+def latticeElement (fact : Fact .executable) : Liveness :=
+  fact.payload.latticeElement
+
+def setLatticeElement (fact : Fact .executable) (latticeElement : Liveness) : Fact .executable :=
+  { fact with payload := { fact.payload with latticeElement := latticeElement } }
+
+def setToLive (fact : Fact .executable) : Fact .executable :=
+  fact.setLatticeElement .live
```
math-fehr: Maybe a `setToDead` counterpart?

UnitTest/DataFlowFramework/Helpers.lean
math-fehr: This documentation should use `/-!` rather than `/--`, and be moved one line earlier

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+
+    -- Reinvoke analyses on all operations in the block
+    for analysisKind in state.subscribers do
+      let mut maybeOp := (block.get! irCtx).firstOp
+      while h : maybeOp.isSome do
+        let op := maybeOp.get h
+        let some point := InsertPoint.after? op irCtx
+          | panic! "Dead Code propagate: block operation without insertion point"
+        dfCtx := dfCtx.enqueue (point, analysisKind)
+        maybeOp := (op.get! irCtx).next
```
math-fehr, line 40: Can you explain me why we have to enqueue all operations of the block here?
Is it because their analyses might change compared because the block is dead? Are operations actually checking that the block is dead or alive?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+TODO: This function likely needs to be replaced with
+an interface much like what MLIR has.
+-/
+private def isBranchOp
+    (op : OperationPtr)
+    (irCtx : IRContext OpCode) : Bool :=
+  -- TODO: Replace this `.test .test` check once VeIR has proper branch ops.
+  match (op.get! irCtx).opType with
+  | .test .test => true
+  | _ => false
```
math-fehr, line 103: What is a `branch` op?
We may have new ones here, you should check now that the code has updated a bit

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+
+/--
+Read a literal constant directly from the defining operation when possible.
+-/
```
math-fehr, line 107: Is it just checking if an operation is an `arith.constant`, or will this handle things like `x - x`? Is it supposed to be more than that at some point?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+
+/--
+Get the constant domain lattice elements of the operands of an operation.
+Non-literal operands are conservatively treated as `top`, so standalone dead
+code analysis remains useful without any external constant information.
+-/
+private def getOperandValues
+    (op : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext × Option (Array AbstractConstant) := Id.run do
```
math-fehr, line 133: This shouldn't return the `dfCtx` back, as it is completely unchanged?
Actually it shouldn't even take is as an operand?

math-fehr: in general, I'm a bit confused here. Shouldn't we read `dfCtx` here to get the current information? Or is this what your SCCP analysis will do later on?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+        else
+          some (op.getSuccessor! irCtx 0)
+      | Data.LLVM.Int.poison =>
+        none
+    | _ =>
+      none
+  else
+    none
+
+def visitBranchOperation
```
math-fehr, line 178: Can you document what this function does?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+  match getSuccessorForOperands? branch operands irCtx with
+  | some successor =>
+    markEdgeLive parentBlock successor dfCtx irCtx
+  | none =>
+    let mut dfCtx := dfCtx
+    for successor in branch.getSuccessors! irCtx do
+      dfCtx := markEdgeLive parentBlock successor dfCtx irCtx
+    dfCtx
+
+private def visitOp
```
math-fehr, line 201: Can you document this function as well?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+
+    -- TODO: Check if this is a callable operation and use callsite information
+    -- to decide whether to mark the callable executable.
+
+    -- else:
+    dfCtx := markEntryBlocksLive op dfCtx irCtx
+
+  -- TODO: If `op` is a region or callable return, visit the corresponding
+  -- terminator semantics once VeIR has the necessary interfaces.
+
```
math-fehr, line 234: There are a lot of line skip here, maybe removing some?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+    let mut dfCtx := dfCtx
+    for successor in branch.getSuccessors! irCtx do
+      dfCtx := markEdgeLive parentBlock successor dfCtx irCtx
+    dfCtx
+
+private def visitOp
+    (op : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  -- If the parent block is not executable, there is nothing to do.
```
math-fehr: That's not really true, because you are not returning from this function?
I would change the comment a bit

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
math-fehr: Yeah I think this should be a panic

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+-/
+partial def initializeRecursively
+    (op : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  let mut dfCtx := dfCtx
+
+  -- Initialize the analysis by visiting every op with control-flow semantics.
+  if op.getNumRegions! irCtx ≠ 0 || op.getNumSuccessors! irCtx ≠ 0 then
+    -- TODO: || isRegionOrCallableReturn op || isACallOpInterface op
```
math-fehr, line 273: Very soon the operation interface will be merged, so we will be able to use that!

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+
+def init
+    (top : OperationPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  -- Mark the top level blocks as executable.
+  let dfCtx := markEntryBlocksLive top dfCtx irCtx
+
+  -- TODO: Mark as overdefined the predecessors of symbol callables with
+  -- potentially unknown predecessors.
```
math-fehr, line 309: I think that instead of having TODO here, what would be best is to have at the beginning of the file what features the DCE algorithm supports, and explicitely state that we don't support callables.
That would make it much better, since TODO comments tend to be stale after a bit.

math-fehr (commented):
Nice, and sorry again for the delay!
I added a few comments here and there, let me know when you want me to make a pass again!

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+  dfCtx
+
+def visit
+    (point : InsertPoint)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  if point.prev! irCtx = none then
+    return dfCtx
+
+  visitOp ((point.prev! irCtx).get!) dfCtx irCtx
```
math-fehr: Wait, why do you expect here an `InsertPoint` with a predecessor, where you use in the rest of the code only `InsertPoint` that have no predecessor (`atStart!`)?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
+module
+
```
math-fehr, line 2: You should add this file to `Veir/Analysis.lean`, otherwise it is not compiled outside of `UnitTest`?

UnitTest/DataFlowFramework/Helpers.lean
```diff
@@ -165,3 +166,45 @@ def runWithAnalyses
...
+    match blockMap[name]? with
+    | some block =>
+      let observed := isBlockLive dfCtx block irCtx
+      if observed != live then
+        report := report.push s!"block {name}: expected live={live}, observed live={observed}"
+    | none =>
+      report := report.push s!"block {name}: missing block label"
+  report
+
+def checkNamedEdgeLiveness
```
math-fehr: This and `isEdgeLive` seems dead code? Can you remove them?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,313 @@
...
+public section
+
+namespace Veir
+
+namespace ExecutableFact
+
+def mkDefault : ExecutableFact :=
+  { payload := { latticeElement := .dead } }
+
+def propagate (state : ExecutableFact) (anchor : LatticeAnchor)
```
math-fehr:
```suggestion
def propagate (state : ExecutableFact) (anchor : LatticeAnchor)
```
Trailing whitespace here.
Can you add some plugin in helix to remove these automatically upon saving? They are hard to catch in github.
I really need to add CI for that

UnitTest/DataFlowFramework/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,66 @@
...
+private def run
+    (mlir : String)
+    (expectedBlockLives : Array (String × Bool)) : String :=
+  runWithAnalyses mlir #[Veir.DeadCodeAnalysis] (fun top dfCtx parserState => Id.run do
+    match recoverNames top parserState.ctx mlir with
+    | Except.error err =>
+        return #[err]
+    | Except.ok recovered =>
+        checkNamedBlockLiveness dfCtx parserState.ctx recovered.blocks expectedBlockLives)
+
```
math-fehr, line 62: All of these tests technically only introspect the first block terminator. Can you add a test that needs to actually propagate anything? So for instance having bb0 -> bb1 -> bb2, where bb2 is not a successor of bb0

math-fehr (commented):
Some more comments I forgot about

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: Liveness is a backwards analysis in MLIR and differs from DeadCodeAnalysis which is forward:
https://github.com/llvm/llvm-project/blob/e5c19b0ebc54206348273fc2f7cf67cb195908ed/mlir/include/mlir/Analysis/DataFlow/LivenessAnalysis.h#L56

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: Yeah I have to sit down and figure out a better organization of Facts in general since I'm just lumping them all in `Facts.lean`. Thanks for reminding me.

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: Due to the monotonicity of dead code (like all the other analyses), I would never use this. We start at bottom (dead) and move up the lattice (to live).

UnitTest/DataFlowFramework/Helpers.lean
axelcool1234: This is an implicit assumption made in my dataflow framework that I should encode explicitly. Essentially, every fact in existence starts at the pessimistic bottom value. So if a value is not initialized in the dataflow context, it is not live since that's the bottom value. It is not an error.

This does the beg the question: If every fact uses my domain class, then I _could_ have `getFact?` be changed to `getFact` that returns bottom by default. This is a TODO I'll place above this function.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: @math-fehr I actually switched to Neovim and coded up Helix motions ontop (that way I have a Lean viewer, since Helix has no plugins)! I'll have to create an autocommand for lean files to remove trailing whitespaces upon buffer write.

UnitTest/DataFlowFramework/Helpers.lean
axelcool1234: I put it in `Helpers.lean` in preparation for SCCP, which will use both dead code and constant propagation helpers in the unittests.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: Note, dead code analysis is a two element lattice. Bottom is dead, Live is top. Propagate is only ever called when the lattice element changes, meaning when the block becomes live. So these invoked analyses, invoked on every op of the block, may change now. And yes, they will check that the block is dead or alive (which is why it may change now). My constant propagation, which is not merged yet, does make this check. All SSA transfer functions using my sparse analysis API gets this check for free!

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: https://github.com/llvm/llvm-project/blob/18d2b47a0d3b6bb5d55260e69bf51c47de8e0dc6/mlir/lib/Analysis/DataFlow/DeadCodeAnalysis.cpp#L393

Will check. I don't understand how `BranchOpInterface` works very well. But we need something like this to prevent having code in several places all independently checking for branch ops.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: Thanks for pointing this out. I need to use John's folder instead of this. It should handle things that are trivially constant. `x - x` is a good example.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: This was a mistake probably. I was messing around a lot with how to handle constants without constant propagation. Now that John has a folder, I'll have to revamp this code like I said in the previous comment.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: This is mirroring this code pretty closely:
https://github.com/llvm/llvm-project/blob/18d2b47a0d3b6bb5d55260e69bf51c47de8e0dc6/mlir/lib/Analysis/DataFlow/DeadCodeAnalysis.cpp#L331-L338

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: Thank goodness. I've been needing this stuff! Glad it's close!

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: I threw all of these TODO comments in places where they belong in the closely followed MLIR version of dead code analysis. It's mostly serving as markers for myself to remember where to look. I can leave a top level TODO comment as well summarizing it all. It'll just be useful since I intend on removing all of these once I have the ability to do so.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: I'd be able to remove several, and I really mean _several_ instances of this code across my entire dataflow framework if this happens.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: The newlines help dilineate that these TODOs are all separate steps that should be implemented when possible. Without the newlines I may later interpret them as one solid step. It's why I also have the newline between the TODO and the if statement.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: The hyperlinked lines is the entirety of `visit` (which is quite small here) in MLIR:
https://github.com/llvm/llvm-project/blob/b6059b6ce21886c3f8644355ae9017cf3ba04aad/mlir/lib/Analysis/DataFlow/DeadCodeAnalysis.cpp#L325-L327

It's just grabbing the previous op of the insert point. It's visiting that op.

UnitTest/DataFlowFramework/Helpers.lean
axelcool1234: I can just move it to the dead code analysis file for now and move it back come time for SCCP

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: You know what, I'm just gonna rename it anyways. I already created a Liveness Domain.

We will figure out if the code I sent in the previous comment is every relevant at a later date and deliberate on naming then.

Veir/Analysis/DataFlow/Facts.lean
axelcool1234: @math-fehr I am going to make another PR where I will split up Facts.lean into a folder, much like I've done for Domains. It'll be a little bit more messy due to dependency junk but it should still be useful. I'll leave it like this for now.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
axelcool1234: I think we should leave it in this very naive state and I'll replace it once https://github.com/opencompl/veir/pull/1084 is merged. Improving this stub is not going to be very useful for long.

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
math-fehr: Yeah, but also a folder should not be called in an analysis, otherwise you start mutating the IR during an analysis pass, which should not happen!
So what should this code do then?

Veir/Analysis/DataFlow/DeadCodeAnalysis.lean
```diff
@@ -0,0 +1,322 @@
...
+
+def visit
+    (point : InsertPoint)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  if point.prev! irCtx = none then
+    panic "Dead code visit called on non block start insertion point.
+           Should've already been marked live when initialized!"
+
+  visitOp ((point.prev! irCtx).get!) dfCtx irCtx
```
math-fehr: I would fix it with this code:
```
def visit
    (point : InsertPoint)
    (dfCtx : DataFlowContext)
    (irCtx : IRContext OpCode) : DataFlowContext :=
  match point.prev! irCtx with
  | none => dfCtx
  | some op => visitOp op dfCtx irCtx
```

Otherwise you have some issue when the block order is different than you wolud expect, like for this reproducer (can you add it as a test?):
```
import UnitTest.DataFlowFramework.DeadCodeAnalysis

open Veir

/--
Exercise reachability that is discovered against source order.

`bb1` is scanned while dead, so its terminator subscribes the dead-code analysis
to the block's liveness fact. Visiting the later `bb2` then makes `bb1` live.
-/
private def testReachabilityDiscoveredAfterSourceOrderScan : String :=
  let mlir := "\"builtin.module\"() ({\n\
^bb0:\n\
  \"test.test\"() [^bb2] : () -> ()\n\
^bb1:\n\
  \"test.test\"() [^bb0] : () -> ()\n\
^bb2:\n\
  \"test.test\"() [^bb1] : () -> ()\n\
}) : () -> ()"
  runWithAnalyses mlir #[Veir.DeadCodeAnalysis] (fun top dfCtx parserState => Id.run do
    match recoverNames top parserState.ctx mlir with
    | .error err =>
      return #[err]
    | .ok recovered =>
      return (
        checkNamedBlockLiveness dfCtx parserState.ctx recovered.blocks
          #[("bb0", true), ("bb1", true), ("bb2", true)] ++
        checkNamedEdgeLiveness dfCtx recovered.blocks
          #[ (("bb0", "bb2"), true)
           , (("bb2", "bb1"), true)
           , (("bb1", "bb0"), true)
           ]))

/--
info: "ok"
-/
#guard_msgs in
#eval! testReachabilityDiscoveredAfterSourceOrderScan

```

PR #660, opened 2026-05-28
feat(dataflow): sparse constant propagation analysis

axelcool1234:


Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,202 @@
...
+        | .poison => none) with
+    | some constant =>
+      if h : results.size > 0 then
+        propagateConstant (results[0]'h) constant dfCtx irCtx
+      else
+        dfCtx
+    | none =>
+      setAllToUnknownConstants results dfCtx irCtx
+  | _ =>
+    setAllToUnknownConstants results dfCtx irCtx
```
axelcool1234: I need to replace this with https://github.com/opencompl/veir/pull/1084

axelcool1234 (commented):
>

UnitTest/DataFlowFramework/Helpers.lean
```diff
@@ -166,3 +168,28 @@ def runWithAnalyses
       let some dfCtx := fixpointSolve top analyses parserState.ctx
         | return "analysis did not converge"
       return renderReport (check top dfCtx parserState)
+
+/-- Sparse constant propagation helpers. -/
+def showConstantDomain : AbstractConstant -> String
```
regehr, line 173: don't do it in this PR, but this code should be part of the abstract domain implementation

UnitTest/DataFlowFramework/SparseConstantPropagation.lean
```diff
@@ -0,0 +1,211 @@
...
+  "cf.br"(%x, %x) [^bb1] : (i32, i32) -> ()
+^bb1(%dead : i32, %y : i32):
+  %z = "arith.addi"(%y, %y) : (i32, i32) -> i32
+}) : () -> ()"#
+    #[ ("x", constInt 32 1)
+     , ("y", constInt 32 1)
+     , ("z", constInt 32 2)
+     ]
+
+/--
```
regehr, line 114: put each of these blocks next to the associated test, please

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,80 @@
...
+
+instance : SparseFactSpec .sparseConstant AbstractConstant where
+  payloadEq := rfl
+
+def kind : AnalysisKind :=
+  .sparseConstantPropagation
+
+private def abstractConstantOfRuntimeValue : RuntimeValue → AbstractConstant
+  | .int bitwidth value => .constant ⟨bitwidth, value⟩
+  | _ => ⊤
```
regehr: why do we sometimes use the top symbol and sometimes `.top`? can you just use the symbols everywhere?

UnitTest/DataFlowFramework/SparseConstantPropagation.lean
```diff
@@ -0,0 +1,211 @@
...
+    (expected : Array (String × AbstractConstant)) : String :=
+  runWithAnalyses mlir #[Veir.SparseConstantPropagationAnalysis]
+  (fun top dfCtx parserState => Id.run do
+      match recoverNames top parserState.ctx mlir with
+      | Except.error err =>
+          return #[err]
+      | Except.ok recovered =>
+          checkNamedConstants dfCtx recovered.values expected)
+
+private def testAddiAllConstant : String :=
```
regehr: I think this test (and others) is testing folding rather than testing propagation. please arrange for constants to be passed between blocks or something else that actually tests your code more meaningfully

regehr:
ok this looks good, plz make a few changes and hopefully Mathieu will also look at it.
also, please follow up with two more PRs:
- add printing abstract values to the dataflow analysis API
- invoke constant propagation from Veir's `canonicalize` optimization pass

axelcool1234:
> ok this looks good, plz make a few changes and hopefully Mathieu will also look at it. also, please follow up with two more PRs:
> 
>     * add printing abstract values to the dataflow analysis API
> 
>     * invoke constant propagation from Veir's `canonicalize` optimization pass

@math-fehr Would love your review on this

Veir.lean
```diff
@@ -27,8 +27,6 @@ import Veir.Passes.CastsReconciliation.Reconciliation
 import Veir.Passes.Legalization.Proofs
 
 -- FIXME: These modules are otherwise orphans and would not be compiled.
-import Veir.Analysis.DataFlow.SparseFact
-import Veir.Analysis.DataFlow.SparseForwardDataFlowAnalysis
```
math-fehr, line 31: If you remove these imports, these files are not compiled anymore

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,80 @@
+module
```
math-fehr: This file is not compiled by `Veir.lean` as far as I know

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,80 @@
...
+
+namespace SparseConstantPropagation
+
+instance : SparseFactSpec .sparseConstant AbstractConstant where
+  payloadEq := rfl
+
+def kind : AnalysisKind :=
+  .sparseConstantPropagation
+
+private def abstractConstantOfRuntimeValue : RuntimeValue → AbstractConstant
```
math-fehr: Can you rename this to `AbstractConstant.ofRuntimeValue`?

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,80 @@
...
+  | .int bitwidth value => .constant ⟨bitwidth, value⟩
+  | _ => ⊤
+
+private def abstractConstantOfFoldResult
```
math-fehr: Can you rename this to `AbstractConstant.ofFoldResult`?

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,80 @@
...
+    (result : FoldResult) (operands : Array AbstractConstant) : AbstractConstant :=
+  match result with
+  | .useOperand index => operands[index]?.getD ⊤
+  | .useConstant value => abstractConstantOfRuntimeValue value
+
+/--
+Sparse constant propagation transfer function.
+- region operations conservatively force results to the unknown state,
```
math-fehr: What is a "region operations"?

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
```diff
@@ -0,0 +1,80 @@
...
+  | _ => ⊤
+
+private def abstractConstantOfFoldResult
+    (result : FoldResult) (operands : Array AbstractConstant) : AbstractConstant :=
+  match result with
+  | .useOperand index => operands[index]?.getD ⊤
+  | .useConstant value => abstractConstantOfRuntimeValue value
+
+/--
+Sparse constant propagation transfer function.
```
math-fehr, line 18: Can you expand/rewrite this documentation? I find it quite unclear and cannot understand what it does from the description.

math-fehr (commented):
I just had a few comments, but otherwise all good I would say!
Also, you still have commits from another branch I think!

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
axelcool1234: I got it from here: https://github.com/llvm/llvm-project/blob/1db5b9c26f5c153244e7617ad95ef140a1bdd2a0/mlir/lib/Analysis/DataFlow/ConstantPropagationAnalysis.cpp#L52-L60

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
axelcool1234: Let me know if my modifications help

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
axelcool1234: An operation with more than 0 regions

regehr:
@axelcool1234 over on zulip I asked you to add a few MLIR-level tests, which in general you can get just by converting a couple of your unit tests to `.mlir` tests

but also, there's a specific thing I need you to test, which is that you're able to propagate a poison constant. the folder tries to very carefully preserve poison but it looks here like you might be losing that bit of information. please dig in and see if that works.

axelcool1234:
> @axelcool1234 over on zulip I asked you to add a few MLIR-level tests, which in general you can get just by converting a couple of your unit tests to `.mlir` tests
> 
> but also, there's a specific thing I need you to test, which is that you're able to propagate a poison constant. the folder tries to very carefully preserve poison but it looks here like you might be losing that bit of information. please dig in and see if that works.

Didn't think about poison at all, thanks. Will do. Also, should I just move all of these over to `.mlir` tests?

Veir/Analysis/DataFlow/SparseConstantPropagationAnalysis.lean
math-fehr: I would rename this to "operation with regions", it is not that much understandable otherwise!

math-fehr (approved):
All good to me, I think John wants you to add some tests, and we can merge it!

math-fehr:
And yes, if you can move some tests to `.mlir` files that would be cool!

regehr:
I think it was my mistake to ask for MLIR tests since this is just the analysis, it's not wired up to a transformation pass.

but Axel please do make sure that you can propagate poisons properly. I think it would be a mistake for you to explicitly deal with poison in your code, but rather you should propagate an LLVM.Int, which can be poison. let me know if you want me to look at code.

math-fehr:
We can do the MLIR thing later if it's too annoying!
But long term, it would be nice if most tests are in MLIR. We should have a system to put analysis data in MLIR files, so then we can easily check in an MLIR file what is the analysis of every value/operation

regehr:
> We can do the MLIR thing later if it's too annoying! But long term, it would be nice if most tests are in MLIR. We should have a system to put analysis data in MLIR files, so then we can easily check in an MLIR file what is the analysis of every value/operation

yes let's do MLIR later. I agree that's where most tests belong, but this PR is simply a static analysis without a transformation backing it up.

after this lands, I plan to make our `canonicalize` pass run constant propagation instead of folding-- this should be strictly more powerful!

and then of course once we have SCCP, we run that instead of SCP.

regehr (approved):
nice work, LGTM!

PR #662, opened 2026-05-28
only run benchmarks on opencompl repos

tobiasgrosser:
This should avoid CI failures on forks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
           .github/scripts/run_benchmarks.sh 1000 100 5
 
       - name: Store benchmark result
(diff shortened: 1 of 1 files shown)


regehr:
LGTM!

axelcool1234:
Thank you!

math-fehr (approved):
I'm fine with it, but considering that we are never actually using these benchmarks, I would prefer them to be removed at some point.

tobiasgrosser:
> I'm fine with it, but considering that we are never actually using these benchmarks, I would prefer them to be removed at some point.

We are gathering historic data: https://opencompl.github.io/veir/dev/bench/. We should probably still discuss at some point if we can make better use of these benchmarks (or drop them).

PR #666, opened 2026-05-29
feat(dataflow): definition of abstract domain and constant domain

axelcool1234:
This implements:
- `Set`
- `Top`, `Bot`, `OrderTop`, `OrderBot`, and `BoundedOrder`
- `Join`
- `JoinSemilattice`
- `AbstractDomain` (which is `BoundedOrder` + `JoinSemilattice` + `\gamma`)
- `ConstantDomain`, an instance of `AbstractDomain`
- notation for `Set`, `Top`, `Bot`, and `Join`


regehr (approved):
ok... so poison is hidden inside of LLVM.int, so this should be equivalent to the lattice I was working with yesterday. LGTM!

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
```diff
@@ -0,0 +1,55 @@
...
+  join : Domain → Domain → Domain
+
+  /-- The join is an upper bound on the first argument. -/
+  le_join_left (a b : Domain) : a ≤ join a b
+
+  /-- The join is an upper bound on the second argument. -/
+  le_join_right (a b : Domain) : b ≤ join a b
+
+  /-- The join is the least upper bound. -/
+  join_le (a b c : Domain) : a ≤ c → b ≤ c → join a b ≤ c
```
ineol: do you need top and bottom too?

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
```diff
@@ -0,0 +1,55 @@
...
+Each abstract value denotes a set of concrete values, represented as a predicate 
+`ConcreteValue → Prop`.
+-/
+class AbstractDomain (AbstractValue : Type) (ConcreteValue : Type)
+    extends JoinSemilattice AbstractValue where
+  /--
+  Concretization. Given an abstract value, returns the predicate describing the
+  set of concrete values it denotes.
+  -/
+  γ : AbstractValue → ConcreteValue → Prop
```
ineol: I think it would be more clear if you defined `def Set alpha := alpha -> Prop` and use it to denote a set of concrete values. Then you can define notation and a lattice structure on it. cf https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Defs.html#Set

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
```diff
@@ -0,0 +1,55 @@
...
+
+  /--
+  The order coincides with concretization inclusion.
+  Monotonicity (soundness of the order): a ≤ b → γ a ⊆ γ b
+  Reflection (completeness of the order): γ a ⊆ γ b → a ≤ b
+  -/
+  le_iff_γ (a b : AbstractValue) : a ≤ b ↔ ∀ c, γ a c → γ b c
```
ineol: This looks a bit fishy to me, because it's very strong.

If you look at page 3 there https://dl.acm.org/doi/pdf/10.1145/2676726.2676966 the figure gives an axiomatisation of gamma which seems more minimal.

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,165 @@
...
+inductive AbstractConstant where
+  | top
+  | bottom
+  | poison
+  | constant (value : KnownConstant)
+deriving BEq, DecidableEq, TypeName
+
+namespace AbstractConstant
+
+def γ (absVal : AbstractConstant) : ConcreteConstant → Prop :=
```
ineol: If we had nice notation for `Set` we could write emptyset and top instead of constant functions.

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,165 @@
...
+  value : Data.LLVM.Int bitwidth
+deriving BEq, DecidableEq
+
+/-- Abstract values used by sparse constant propagation. -/
+inductive AbstractConstant where
+  | top
+  | bottom
+  | poison
+  | constant (value : KnownConstant)
+deriving BEq, DecidableEq, TypeName
```
ineol, line 28: Why not this one?
```suggestion
inductive AbstractConstant where
  | top
  | bottom
  | constant (value : ConcreteConstant)
deriving BEq, DecidableEq, TypeName
```

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,165 @@
...
+  le_iff_γ := le_iff_γ
+  le_refl := le_refl
+  le_trans := le_trans
+  le_antisymm := le_antisymm
+  join := .join
+  le_join_left := le_join_left
+  le_join_right := le_join_right
+  join_le := join_le
+
+instance : AbstractDomain AbstractConstant ConcreteConstant := ConstantDomain
```
ineol: why not this?
```suggestion
instance : AbstractDomain AbstractConstant ConcreteConstant where
  le := le
  γ := γ
  le_iff_γ := le_iff_γ
  le_refl := le_refl
  le_trans := le_trans
  le_antisymm := le_antisymm
  join := .join
  le_join_left := le_join_left
  le_join_right := le_join_right
  join_le := join_le


```

ineol (commented):
Thank you! I have a few questions but it looks good already :)

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
```diff
@@ -0,0 +1,55 @@
+module
+
```
math-fehr, line 2: Can you add a toplevel comment with `/-! -/` to explain what this file contains?

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,165 @@
+module
+
+public import Veir.Data.LLVM.Int.Basic
+public import Veir.Analysis.DataFlow.Domains.AbstractDomain
+
```
math-fehr, line 5: Can you add a toplevel comment with `/-! -/` to explain what this file contains?

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,165 @@
...
+structure KnownConstant where
+  bitwidth : Nat
+  value : BitVec bitwidth
+deriving BEq, DecidableEq
+
+/-- Concrete integer values tracked by the constant domain. -/
+structure ConcreteConstant where
+  bitwidth : Nat
+  value : Data.LLVM.Int bitwidth
+deriving BEq, DecidableEq
```
math-fehr: What is the difference between these two structures? Is there a reason why we do not only use `ConcreteConstant`?

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,165 @@
...
+
+/-- Abstract values used by sparse constant propagation. -/
+inductive AbstractConstant where
+  | top
+  | bottom
+  | poison
+  | constant (value : KnownConstant)
```
math-fehr: This two values should be merged if we are using `ConcreteConstant` only

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
math-fehr: In general, having only one of these structures would make the code easier to reason about.

math-fehr (commented):
Nice! I added a few comments, but otherwise that looks good to me!
Leo is more knowledgeable on which properties to actually track, so I didn't review this part.

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
regehr: IIRC, a join semilattice does not actually require these. whether Axel needs to define a join semilattice without those elements is a separate question, that I don't know the answer to

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
axelcool1234: > In general, having only one of these structures would make the code easier to reason about.

So to explain why I had this - I nearly pulled my hair out trying to reason about LLVM Ints directly as a `constant` in the inductive because now we have to face proofs that involve, say, an LLVM Int with a bitwidth of 8 and an LLVM Int with a bitwidth of 32 both being poison. These poisons are NOT the same! So now we have infinite poisons instead of one. I am not knowledgeable on how to deal with this. Any suggestions? @math-fehr @ineol @regehr

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
axelcool1234: > why not this?
> ```suggestion
> instance : AbstractDomain AbstractConstant ConcreteConstant where
>   le := le
>   γ := γ
>   le_iff_γ := le_iff_γ
>   le_refl := le_refl
>   le_trans := le_trans
>   le_antisymm := le_antisymm
>   join := .join
>   le_join_left := le_join_left
>   le_join_right := le_join_right
>   join_le := join_le
> 
> 
> ```
How do I refer to `AbstractDomain AbstractConstant ConcreteConstant` in other files? It'd be nice if I could have it as a type as `ConstantDomain`. I think I have a misunderstanding of Lean's type system here. So for the SCCP analysis for example I need to be able to pass around `ConstantDomain`. How can I achieve this?

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
axelcool1234: > IIRC, a join semilattice does not actually require these. whether Axel needs to define a join semilattice without those elements is a separate question, that I don't know the answer to

We could leave them out I believe

axelcool1234: Although so far I have always used top and bottom

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
axelcool1234: Thanks for the reference. I'll look at it :)

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
ineol: Why do you need to pass around `ConstantDomain`? you can use `AbstractConstant` and the typeclass will allow you to use `join` on elements of `ConstantDomain`. Note that if you want to name the instance you can write `instance ConstantDomain : ...`.

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
ineol: Can you give a specific example where having `AbstractConstant = ConcreteConstant + top + bottom` is causing trouble?

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
ineol: So in this verasco project they consider join semi lattices with top in their axiomatisation, but not bottom. I think it would make sense to read some AI to know what's what's best. For example I think the join doesn't have to be the _least_ upper bound to get soundness, etc.

axelcool1234:
Note to self: 
```lean
/-- Alias for the top element: value is unknown or conflicting. -/
def unknown : AbstractConstant := ⊤

/-- Alias for the bottom element: no information has been learned yet. -/
def uninitialized : AbstractConstant := ⊥
```

I need to change my stacked PRs to not use this (will remove these defs to avoid confusion).

axelcool1234:
@regehr You'd probably like to see these changes too.

Veir/Analysis/DataFlow/Domains/AbstractDomain.lean
```diff
@@ -0,0 +1,120 @@
...
+  /--
+  The top element denotes every concrete value.
+  -/
+  γ_top : γ (⊤ : AbstractValue) = (⊤ : Set ConcreteValue)
+
+  /--
+  The bottom element denotes no concrete values.
+  -/
+  γ_bot : γ (⊥ : AbstractValue) = (⊥ : Set ConcreteValue)
```
ineol:
```suggestion
  γ_top : γ ⊤ = ⊤

  /--
  The bottom element denotes no concrete values.
  -/
  γ_bot : γ ⊥ = ⊥
```

ineol (commented):
I like the new definitions :)

I would consider, here or in a later PR, working out how to annotate them with simp/grind so that the proofs become more automatic. Right now you're doing a lot of manual proof automation with `try` and a lot of unfolding, like `le_join_right` in the constant domain file.

axelcool1234:
@ineol I actually tried providing lots of definitions, like defining commutativity, associativity, and more for JoinSemilattice (check the commit called "commented out simp defs") and others. But none of them helped... I'm unfortunately still too new to Lean proofs so I'll need some pointers on this

ineol:
Then let's postpone it to a later PR I think? I think it will be a non trivial amount of work because of the several layers of definitions.

axelcool1234:
> Then let's postpone it to a later PR I think? I think it will be a non trivial amount of work because of the several layers of definitions.

That sounds good with me :)

ineol:
so once you've written the top level comments LGTM

axelcool1234:
> so once you've written the top level comments LGTM

Done!

ineol:
Awesome! Let's at least wait for @math-fehr

regehr (approved):
Axel I've read this all and it's super nice and readable. It formalizes the stuff that I expect it to formalize. I don't have any issues.

However-- let's wait for our Lean4 / abstract interpretation heavy hitters @ineol and @math-fehr to approve it since they have a ton of experience that I'm missing here.

Separately, I asked Claude what it thinks about the design here and one piece of what it said is something that I totally do not understand. I'm pasting it here just in case this might be helpful to you.

> Latent LE diamond. instance : LE AbstractConstant is defined standalone, then JoinSemilattice extends LE and the instance re-supplies le :=
>   le, while BoundedOrder was proven against the standalone LE. AbstractDomain then merges both. It compiles only because the two LE paths are
>   defeq. Note that Std.IsPartialOrder — which you're extending — deliberately takes [LE α] as a parameter rather than extending it, exactly to
>   avoid this. I'd follow that idiom:
>   class JoinSemilattice (α : Type) [LE α] extends Std.IsPartialOrder α, Join α where ...
>   Then there's one LE source (the standalone instance), the le := field disappears from the instance, and the diamond is gone. BoundedOrder
>   already does this correctly.

axelcool1234:
> Axel I've read this all and it's super nice and readable. It formalizes the stuff that I expect it to formalize. I don't have any issues.
> 
> However-- let's wait for our Lean4 / abstract interpretation heavy hitters @ineol and @math-fehr to approve it since they have a ton of experience that I'm missing here.
> 
> Separately, I asked Claude what it thinks about the design here and one piece of what it said is something that I totally do not understand. I'm pasting it here just in case this might be helpful to you.
> 
> > Latent LE diamond. instance : LE AbstractConstant is defined standalone, then JoinSemilattice extends LE and the instance re-supplies le :=
> >   le, while BoundedOrder was proven against the standalone LE. AbstractDomain then merges both. It compiles only because the two LE paths are
> >   defeq. Note that Std.IsPartialOrder — which you're extending — deliberately takes [LE α] as a parameter rather than extending it, exactly to
> >   avoid this. I'd follow that idiom:
> >   class JoinSemilattice (α : Type) [LE α] extends Std.IsPartialOrder α, Join α where ...
> >   Then there's one LE source (the standalone instance), the le := field disappears from the instance, and the diamond is gone. BoundedOrder
> >   already does this correctly.

Ah, I think this is a very real Lean type system problem. Type inference is like Prolog if you're aware of it - unification and such. But there are issues with diamonds. I think their approach to avoid... conflicts I think... Is to always prefer the left hand side of a "type diamond" or whatever. As you can tell I am also very much weak on this. But I think Claude is telling a very real problem here

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,123 @@
...
+  match x, y with
+  | .bottom, _ => True
+  | _, .top => True
+  | .constant c, .constant d => c = d
+  | _, _ => False
+
+instance : LE AbstractConstant where
+  le := le
+
+@[simp] theorem le_def (a b : AbstractConstant) : (a ≤ b) ↔ le a b := Iff.rfl
```
math-fehr, line 43: @ineol @luisacicolini, is this supposed to be `simp`?
I'm asking because I just don't know if this is standard or not.

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,123 @@
...
+
+theorem le_top (a : AbstractConstant) : a ≤ .top := by
```
math-fehr, line 46: I guess this should be marked as `simp` and `grind .`

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,123 @@
...
+  cases a <;> trivial
+
+theorem bot_le (a : AbstractConstant) : .bottom ≤ a := by
```
math-fehr, line 50: Same here for `simp` and `grind .`

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
math-fehr: Especially since the `le` theorems below are shown in terms of `<=` and not in terms of `le`.

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,123 @@
...
+
+instance : Join AbstractConstant where
+  join := join
+
+theorem γ_monotone (a b : AbstractConstant) : a ≤ b → γ a ⊆ γ b := by
+  intro hab x hx
+  cases a <;> cases b <;> simp [γ, le] at hab hx ⊢
+  all_goals first | trivial | exact hx.trans hab
+
+theorem le_refl (a : AbstractConstant) : a ≤ a := by
```
math-fehr, line 82: You want to add `simp` and `grind .` for this theorem as well (not the `le_trans` one though).

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
```diff
@@ -0,0 +1,123 @@
...
+
+theorem le_join_left (a b : AbstractConstant) : a ≤ a ⊔ b := by
+  cases a <;> cases b <;> try simp [le, join]
+  case constant.constant c d =>
+    by_cases h : c = d <;> simp [h]
+
+theorem le_join_right (a b : AbstractConstant) : b ≤ a ⊔ b := by
+  cases a <;> cases b <;> try simp [le, join]
+  case constant.constant c d =>
+    by_cases h : c = d <;> simp [h]
```
math-fehr, line 103: Can you make these `simp` and `grind .` as well?

math-fehr (commented):
Very cool! The design is indeed much better now!
I just added comments about `simp` and `grind` on some lemmas.
In particular, I'm not sure if `le_def` should be `simp` (@luisacicolini and @ineol) would know though.

After that we should be able to merge it!

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
math-fehr: Also, I think this should be `grind =`?

axelcool1234: Without the simped `le_def` a lot of my theorems fail actually. That's why I have it

axelcool1234:
@math-fehr if you're up for it, I have a commit called "commented out simp defs" I think, that has a _ton_ of additional theorems/lemmas and such that I introduced and then removed. Could you tell me if any of them are remotely useful? None of them ended up helping with proving ConstantDomain.

math-fehr:
I think most of them should be useful? But, please add them AFTER this PR, not in this one.
Many people reviewed this PR already, you don't want to add more noise to it and make them review it again ;)
But as a second PR I think we should try to add most of these!

axelcool1234:
> I think most of them should be useful? But, please add them AFTER this PR, not in this one.
> Many people reviewed this PR already, you don't want to add more noise to it and make them review it again ;)
> But as a second PR I think we should try to add most of these!

Sounds good. I'll need help determining this simp and grind stuff for those. I'll ping you for that PR once this one gets merged :)

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
axelcool1234: Why no `le_trans`? Does it cause state explosion for `grind` or something?

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
axelcool1234: `le_antisymm` can't be simp:
```|
▼ 88:3-88:7: warning:
Left-hand side of simp theorem has a variable as head symbol. This means the theorem will be tried on every simp step, which can be expensive. This may be acceptable for `local` or `scoped` simp lemmas.
Use `set_option warning.simp.varHead false` to disable this warning.
```

Not sure what this means.

axelcool1234:
@math-fehr I have a few comments I left on the last of the unresolved review comments. After that I'm done, and I think this is ready for merging :)

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
ineol: @math-fehr the grind doc says

```
opaque R : [Int](https://lean-lang.org/doc/api/Init/Data/Int/Basic.html#Int) → [Int](https://lean-lang.org/doc/api/Init/Data/Int/Basic.html#Int) → Prop
@[grind →] axiom Rtrans {x y z : Int} : R x y → R y z → R x z

example {a b c d} : R a b → R b c → R c d → R a d := by
  grind
```
which only triggers when both R a b and R b c are in the context

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
ineol: For sure it shouldn't be here, but I think designing a good simp/grind set and a complete API is a fair amount of work and it might be good to do it in a later PR?

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
math-fehr: Ohh, yeah then `grind ->` ^^

Veir/Analysis/DataFlow/Domains/ConstantDomain.lean
math-fehr: Oh sorry, yeah let's not do that, that might be too expensive
So just having `grind .` for the left and right version, and `grind ->` for antisymm

axelcool1234:
Alright @math-fehr, `grind ->` for `le_antisymm` and `le_trans` added

math-fehr:
Can you add the unicode arrow, sorry ;)

axelcool1234:
> Can you add the unicode arrow, sorry ;)

Done :rofl:

math-fehr (approved):
Thanks!

PR #667, opened 2026-05-29
Ensure all values in the interpreter state conforms to their types

math-fehr:
In `VariableState`, we now add a field that asserts that all runtime values in the hashmap actually correspond to their associated MLIR types.
This means that `VariableState` and `InterpreterState` now have an WfIRContext parameter.
Additionally, `VariableState.setVar` and `VariableState.setResultVars` now return an optional, as we need to check wether or not the values correctly conforms to their respective types. This required to update a lot of our proofs.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -65,8 +65,8 @@ instance : Decidable (Conforms val ty) := by
 
 @[grind <=]
 theorem Conforms.integerType :
-    Conforms runtimeValue (⟨.integerType intType, h⟩ : TypeAttr)
-    → ∃ val, runtimeValue = .int intType.bitwidth val := by
+    Conforms runtimeValue ⟨.integerType intType, h⟩ →
+    ∃ val, runtimeValue = .int intType.bitwidth val := by
   simp only [Conforms]
   cases runtimeValue
   case int bw val =>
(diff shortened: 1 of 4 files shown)


regehr (approved):
I did look through this, and it seems fine -- but I'm well out of my depth here!

Veir/Interpreter/Basic.lean
```diff
@@ -1096,22 +1111,29 @@ decreasing_by grind
   to continue the interpretation.
   Return `none` if any errors occur during interpretation.
 -/
-def interpretBlock (ctx : IRContext OpCode) (blockPtr : BlockPtr) (state : InterpreterState) (blockInBounds : blockPtr.InBounds ctx := by grind) (wf : ctx.WellFormed := by grind) : Interp (InterpreterState × ControlFlowAction) := do
-  rlet firstOp ← (blockPtr.get ctx).firstOp
-  interpretOpList ctx firstOp state
+def interpretBlock (blockPtr : BlockPtr) {ctx : WfIRContext OpCode} (state : InterpreterState ctx)
+    (blockInBounds : blockPtr.InBounds ctx.raw := by grind) :
+  Interp (InterpreterState ctx × ControlFlowAction) := do
```
```suggestion
    Interp (InterpreterState ctx × ControlFlowAction) := do
```

ineol (approved):

PR #670, opened 2026-05-29
ci: make LLVM/identity.mlir compatible with mlir-opt

tobiasgrosser:
This works, but breaks test coverage. I will pull out changes bit-by-bit.


tobiasgrosser:
This has been incrementally upsteamed, with the last PR being #708.

PR #671, opened 2026-05-29
llvm: llvm.alloc alignment expects an i64 attribute

tobiasgrosser:
This removes an error that arises when trying to parse `LLVM/identity.mlir` with `mlir-opt`.


math-fehr (approved):
Nice, should we also add this in the verifier?

regehr:
LGTM!

tobiasgrosser:
> Nice, should we also add this in the verifier?

I added a verifier.

PR #673, opened 2026-05-29
test: Make sure all tests are in top-level `func.func`

GZGavinZhao:
Stemmed out of #636. Per discussion on Zulip, we probably don't want to encode into the parser the restriction that only functions are allowed at the top-level of a module, since spec-wise MLIR doesn't care what you put at the top-level of a `builtin.module`. However, that is still non-standard behavior and our tests should be as close to real-world MLIR as possible. This PR moves all tests that originally look like this:
```mlir
"builtin.module"() ({
^4():
  "foo.bar"() : () -> ()
}) : () -> ()
```
and wrap that top-level basic block inside a `main` function:
```mlir
"builtin.module"() ({
  "func.func"() <{function_type = () -> (), sym_name = "main"}> ({
    ^4():
      "foo.bar"() : () -> ()
      "func.return"() : () -> ()
  }) : () -> ()
}) : () -> ()
```
All changes are hand-crafted so there can be no copy-and-pasting errors caused by those ~~pesky stochastic machines~~ LLMs.

### DIFF
--- a/Test/CudaTile/types.mlir
+++ b/Test/CudaTile/types.mlir
@@ -1,10 +1,20 @@
 // RUN: VEIR_ROUNDTRIP
 "builtin.module"() ({
-^bb0(%arg0: !cuda_tile.ptr<i1>):
-    %0 = "test.test"() : () -> !cuda_tile.ptr<i1>
+  // TODO(#675): the proper type should be without quotes, i.e.
+  // `!cuda_tile.ptr<i1> -> ()`. Our parser cannot handle this right now, it's a
+  // bug that should be fixed.
+  "func.func"() <{function_type = "!cuda_tile.ptr<i1> -> ()", sym_name = "main"}> ({
+    ^bb0(%arg0: !cuda_tile.ptr<i1>):
+      %0 = "test.test"() : () -> !cuda_tile.ptr<i1>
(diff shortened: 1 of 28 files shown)


Test/CudaTile/types.mlir
```diff
@@ -1,10 +1,20 @@
 // RUN: VEIR_ROUNDTRIP
 "builtin.module"() ({
-^bb0(%arg0: !cuda_tile.ptr<i1>):
-    %0 = "test.test"() : () -> !cuda_tile.ptr<i1>
+  // Note: the proper type should be without quotes, i.e.
```
regehr: do we have a convention for this kind of thing? I usually mark stuff like this as TODO or FIXME so I can grep for those later on

regehr: (or file an issue, if we think this is something that someone should work on in the near future)

GZGavinZhao: I agree, TODO would be the standard. `Note:` is just what I use personally. Will update this to TODO.

regehr:
This LGTM!

regehr:
(I don't think I can click "approve" until you resolve the comment I left, but consider this approved)

PR #677, opened 2026-05-29
llvm: llvm.[load|store] alignment expects an i64 attribute

tobiasgrosser:
This removes an error that arises when trying to parse `LLVM/identity.mlir` with `mlir-opt`. To make `llvm.load` and `llvm.store` parse, we also drop `syncscope` which is only allowed for atomic load/stores.

Also match LLVM's error message for `llvm.alloc` with `mlir-opt`.


nchappe (approved):

nchappe:
There is just a small typo: all the error messages mention `llvm.store` instead of the relevant operation

PR #680, opened 2026-05-29
Minimal RISCV load/store interpretation

nchappe:
This should be sufficient for FastNTT

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -102,6 +102,12 @@ structure MemoryState where
 def MemoryState.empty : MemoryState :=
   { contents := (ByteArray.emptyWithCapacity 1024).extend 8 0xff }
 
+def MemoryState.ensureSize (mem : MemoryState) (size : Nat) : MemoryState :=
+  if mem.contents.size < size then
+    { contents := mem.contents.extend (size - mem.contents.size) 0 }
+  else
+    mem
+
 /--
   Property that a hash map from `ValuePtr` to `RuntimeValue` conforms to the value types in the
   IR context. This is an invariant that must be maintained by the variable state of the interpreter.
@@ -270,20 +276,21 @@ def MemoryState.alloc (state : MemoryState) (size : UInt64)
 -/
 def MemoryState.store (state : MemoryState) (addr : UInt64) (val : ByteArray)
     : Interp MemoryState :=
-  if addr.toNat != 0 && addr.toNat + val.size <= state.contents.size then
+  if addr.toNat + val.size ≤ state.contents.size then
     return { state with contents := val.copySlice 0 state.contents addr.toNat val.size false }
   else
     Interp.ub
 
 /--
-  Store a value to memory.
-  Yields UB if the access is out of bounds.
+  Store an LLVM value to memory.
+  Yields UB if the access is out of bounds or the address is 0.
 -/
 def MemoryState.storeValue (state : MemoryState) (addr : UInt64) (val : RuntimeValue)
     : Interp MemoryState :=
+  if addr.toNat == 0 then Interp.ub else
   match val with
   | .int 8 (.val v) => state.store addr (ByteArray.empty.push (UInt8.ofBitVec v))
-  | .int 64 (.val v) | .reg {val := v} => state.store addr (UInt64.ofBitVec v).toByteArrayLE
+  | .int 64 (.val v) => state.store addr (UInt64.ofBitVec v).toByteArrayLE
   | .int _ .poison => return state
   | .addr v => state.store addr v.toByteArrayLE
   | _ => none
@@ -294,17 +301,18 @@ def MemoryState.storeValue (state : MemoryState) (addr : UInt64) (val : RuntimeV
 -/
 def MemoryState.load (state : MemoryState) (addr size : UInt64)
     : Interp ByteArray :=
-  if addr.toNat != 0 && addr.toNat + size.toNat <= state.contents.size then
+  if addr.toNat + size.toNat <= state.contents.size then
     return state.contents.extract addr.toNat (addr + size).toNat
   else
     Interp.ub
 
 /--
-  Load a value from the given memory address.
-  Yields UB if access is out of bounds.
+  Load an LLVM value from the given memory address.
+  Yields UB if access is out of bounds or the address is 0.
 -/
 def MemoryState.loadValue (state : MemoryState) (addr : UInt64) (type : TypeAttr)
     : Interp RuntimeValue := do
+  if addr == 0 then Interp.ub else
   match type.val with
   | Attribute.integerType { bitwidth := 8 } =>
       let ba ← state.load addr 1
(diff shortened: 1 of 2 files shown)


tobiasgrosser (approved):
Nice. Let's see what @ineol thinks.

Veir/Interpreter/Basic.lean
```diff
@@ -201,7 +201,7 @@ def MemoryState.alloc (state : MemoryState) (size : UInt64)
 -/
 def MemoryState.store (state : MemoryState) (addr : UInt64) (val : ByteArray)
     : Interp MemoryState :=
-  if addr.toNat != 0 && addr.toNat + val.size <= state.contents.size then
+  if addr.toNat + val.size <= state.contents.size then
```
```suggestion
  if addr.toNat + val.size ≤ state.contents.size then
```

Veir/Interpreter/Basic.lean
```diff
@@ -575,318 +575,333 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
...
+    let val ← match val with
+    | some .ub => pure 0
+    | some (.ok v) => pure (BitVec.ofNat 64 v.toUInt64LE!.toNat)
+    | none => none
+    return (#[.reg { val }], mem, none)
+  | .sd => do
+    let [.reg addr, val] := operands.toList | none
+    let mem := if mem.contents.size < addr.val.toNat + 8 then
+      { contents := mem.contents.extend (addr.val.toNat - mem.contents.size + 8) 0 }
+    else mem
```
ineol: should be indented?

Veir/Interpreter/Basic.lean
```diff
@@ -575,318 +575,333 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
...
   | .sgtz => do
     let [.reg op] := operands.toList | none
-    return (#[.reg (RISCV.sgtz op)], none)
-  | _ => none
+    return (#[.reg (RISCV.sgtz op)], mem, none)
+  | .ld => do
+    let [.reg addr] := operands.toList | none
+    let val := mem.load addr.val.toNat.toUInt64 8
+    let val ← match val with
+    | some .ub => pure 0
```
ineol: Why are we fine with UB? it deserves a comment I think

Veir/Interpreter/Basic.lean
ineol: Maybe define a function `ensureSize` or something? this way we can write lemmas to reason about it without case analysis.

ineol (approved):
Thank you!

Veir/Interpreter/Basic.lean
math-fehr: This changes the semantics of LLVM and Arith as well.
What should be the semantics of storing at address 0 in LLVM and Arith?

regehr: loading or storing through the null pointer is immediate UB in LLVM, and I assume also in Arith.

but this should 100% be legal at the RISC-V level, there's nothing special about address zero there

math-fehr: Yes, I'm just mentionning this here because `MemoryState.store` was used by the LLVM/Arith interpreter, so this will change the semantics of Arith and LLVM.
I agree this function should be changed to this though, but we should then fix the arith/llvm intepreter.

Veir/Interpreter/Basic.lean
nchappe: for `mem.load` to raise UB the access has to be out of bounds
now I extend the memory before this point so this cannot happen

Veir/Interpreter/Basic.lean
nchappe: Now the 0 check is in `storeValue` only, and the RISCV store operation bypasses it by directly calling `store`.
I think dereferencing null pointers can be valid in LLVM with `-fno-delete-null-pointer-checks` but it probably makes more sense to disallow it for now.

nchappe:
thanks for the comments, they all should be resolved now

Veir/Interpreter/Basic.lean
```diff
@@ -575,318 +582,332 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
...
+    return (#[.reg (RISCV.sgtz op)], mem, none)
+  | .ld => do
+    let [.reg addr] := operands.toList | none
+    -- extend the memory so that the access is in bounds
+    let mem := mem.ensureSize (addr.val.toNat + 8)
+    let val := mem.load addr.val.toNat.toUInt64 8
+    let val ← match val with
+    | some (.ok v) => pure (BitVec.ofNat 64 v.toUInt64LE!.toNat)
+    | _ => none
+    return (#[.reg { val }], mem, none)
```
```suggestion
    let val ← mem.load addr.val.toNat.toUInt64 8
    pure (BitVec.ofNat 64 val.toUInt64LE!.toNat)
```

does some variation of that work by any chance?

nchappe: the match was intentional, the point was that UBs should not happen with RISCV memory accesses
the val UB case is unreachable but I still mapped it to none just in case

I just pushed a new version with your suggested edit and a comment stating that UB should not happen

Veir/Interpreter/Basic.lean
math-fehr:
```suggestion
  Yields UB if the access is out of bounds or the address is 0.
```

math-fehr, line 300:
```suggestion
  Yields UB if the access is out of bounds or the address is 0.
```

regehr:
this LGTM!

math-fehr (approved):
I just added 2 notes on comments that should be added, but otherwise all good.
Btw, I think we should change `storeValue`/`store` (in a different PR) to have probably a different name. Because currently they seem like they should be able to be called with `.reg`, but they shouldn't.

regehr (approved):
I would still say that we permit accesses to address 0 when interpreting RISC-V assembly, but not when interpreting higher-level dialects, since NULL pointers are a language-level concept, not a machine concept. but this will not actually matter until someone attempts verified compilation of an OS or something of that nature.

nchappe:
Null accesses are already allowed as is for RISCV memory accesses, but not for LLVM memory accesses.

regehr:
> Null accesses are already allowed as is for RISCV memory accesses, but not for LLVM memory accesses.

I didn't read carefully enough. thanks!

Veir/Interpreter/Basic.lean
```diff
@@ -270,20 +276,21 @@ def MemoryState.alloc (state : MemoryState) (size : UInt64)
...
   else
     Interp.ub
 
 /--
-  Store a value to memory.
-  Yields UB if the access is out of bounds.
+  Store an LLVM value to memory.
+  Yields UB if the access is out of bounds or the address is 0.
 -/
 def MemoryState.storeValue (state : MemoryState) (addr : UInt64) (val : RuntimeValue)
```
math-fehr, line 288: Should this be renamed `storeLLVMValue` then, to make this clearer to the caller?

math-fehr (approved):

nchappe:
Rebased and addressed everything except Mathieu's final comment.

This code will probably not survive more than a couple of weeks anyways.

PR #681, opened 2026-05-29
Fix missing grind lemma

nchappe:


tobiasgrosser (approved):
The PR title and body should probably explain in one sentence what the expected set of grind lemmas is and why this lemma completes this set. Are there potentially further lemmas missing, or did you verify that set of grind lemmas is now complete.

math-fehr (approved):

PR #683, opened 2026-05-29
Throw located errors in attribute parser

math-fehr:
attribute errors are now showing the location where the error was triggered, instead of just outputting the error message.


tobiasgrosser (approved):
Nice. Can we have an `*.mlir` file-level test case that demonstrates how these errors are emitted via `veir-opt`?

ineol (approved):

GZGavinZhao (approved):

math-fehr:
I added one test for attributes, and one test for types!

PR #688, opened 2026-05-29
chore(DataFlow): remove `anchor` from `Fact`

ineol:


ineol:
cc @axelcool1234 (I can't seem to add you as a reviewer for some reason)

axelcool1234:
I am unsure if this'll cause issues for dead code analysis and SCP/SCCP but it probably shouldn't. Good catch, LGTM!

ineol:
My intuition is that usually when we query a fact we know the anchor so it's redundant

axelcool1234:
> My intuition is that usually when we query a fact we know the anchor so it's redundant

Yeah that makes sense

PR #690, opened 2026-05-29
chore(DataFlow): simplify function

ineol:
`DominanceAnalyis.dominates` was using Lean dot-notation for functions in a tricky way that was confusing to me.

-----

cc @axelcool1234


axelcool1234:
LGTM. I think I changed this function to be simpler but I accidentally lost it when when I was doing some JJ stuff

math-fehr (approved):

PR #691, opened 2026-05-29
chore(parser): Improve block entries in the parser (NFC)

math-fehr:
Having an inductive makes the meaning more clear than having a pair with a boolean

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -23,7 +23,7 @@ variable {ctx : IRContext OpCode}
 /--
   The state of a block encountered during parsing.
   A block is either `Defined` (its label has been parsed) or `ForwardDeclared`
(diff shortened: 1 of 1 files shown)


Veir/Parser/MlirParser.lean
```diff
@@ -20,6 +20,21 @@ open Veir.Parser.ParserError
 
 variable {ctx : IRContext OpCode}
 
+/--
+  The state of a block encountered during parsing.
+  A block is either `Defined` (its label has been parsed) or `ForwardDeclared`
+  (it has only been referenced, e.g. as a block operand, but not yet defined).
```
tobiasgrosser:
```suggestion
  (it has only been referenced, e.g., as a block operand, but not yet defined).
```

tobiasgrosser (approved):
Beautiful!

PR #692, opened 2026-05-29
fix: check that blocks are terminated properly

regehr:
this PR makes 3 changes:
- lots of test cases trivially updated so that blocks have terminator instructions
- recognize a subset of instructions as terminators
- for regions where this is required, ensure that each block contains a terminator as its final operation, and contains terminators nowhere else


Test/Passes/InstCombine/and_x_x.mlir
```diff
@@ -11,6 +11,6 @@
       %and_self = "llvm.and"(%x, %x) : (i32, i32) -> i32
       "test.test"(%and_self) : (i32) -> ()
       // CHECK-NEXT: "test.test"(%[[X]]) : (i32) -> ()
-      "function.return"() : () -> ()
```
math-fehr, line 14: Ahah!

math-fehr: I'm... surprised that this wasn't catched by my `--allow-unregistered-dialect`? I need to look into it

Test/Verifier/empty_block.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() ({
+  ^bb0():
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: the block is empty
```
math-fehr:
```suggestion
// CHECK: Expected the block to end in a terminator, but the block is empty
```
To check the entire error?

Veir/Verifier.lean
```diff
@@ -1875,13 +1875,64 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
...
-  ctx.raw.forOpsDepM (fun op opIn => op.verifyLocalInvariants ctx opIn)
+def WfIRContext.verify (ctx : WfIRContext OpCode) : Except String Unit := do
+  ctx.raw.forOpsDepM (fun op opIn => do
+    op.verifyLocalInvariants ctx opIn
+    match (op.get ctx.raw opIn).parent with
+    | some block =>
+      match (block.get! ctx.raw).parent with
+      | some region =>
+        if region.requiresTerminator ctx then
+          op.verifyTerminatorPosition ctx opIn
```
math-fehr:
```suggestion
        op.verifyTerminatorPosition ctx opIn
```
This should be checked for graph regions and SSACFG regions!
Can you also add a test for let's say `builtin.module`?

math-fehr (commented):
Nice! I just have 2-3 comments, but otherwise all good!

Test/Passes/InstCombine/and_x_x.mlir
regehr: I had the same thought!!!

regehr:
@math-fehr what do you think about adding an interface that can be used to query whether a block is SSACFG? (not in this PR, but if we add it then the code here becomes a little nicer)

Veir/Verifier.lean
regehr: @math-fehr not totally sure what you mean-- but I didn't have a test on this particular case

```mlir
// RUN: veir-opt %s | filecheck %s

// A `builtin.module` does not require a terminator

"builtin.module"() ({
^bb0():
}) : () -> ()

// CHECK: "builtin.module"() ({
```
does that work for you?

math-fehr:
> @math-fehr what do you think about adding an interface that can be used to query whether a block is SSACFG? (not in this PR, but if we add it then the code here becomes a little nicer)

Yeah that makes sense to me, this is how MLIR does it as well!

regehr:
ok I looked at the MLIR code and it seems simple and obviously good, so I've updated this PR to do something similar.

Veir/Verifier.lean
```diff
@@ -1875,13 +1875,60 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
...
-def WfIRContext.verify (ctx : WfIRContext OpCode) : Except String Unit := Id.run do
-  ctx.raw.forOpsDepM (fun op opIn => op.verifyLocalInvariants ctx opIn)
+def WfIRContext.verify (ctx : WfIRContext OpCode) : Except String Unit := do
+  ctx.raw.forOpsDepM (fun op opIn => do
+    op.verifyLocalInvariants ctx opIn
+    match (op.get ctx.raw opIn).parent with
+    | some block =>
+      match (block.get! ctx.raw).parent with
+      | some _ =>
+        op.verifyTerminatorPosition ctx opIn
```
math-fehr: I think you are missing a condition here that the region is an SSACFG?

math-fehr (commented):
Nice! That looks better, but I think you are missing something in your `verify` function?

Veir/Verifier.lean
regehr: wait I'm confused... isn't this the same code where earlier you said "This should be checked for graph regions and SSACFG regions" and then I accepted your change removing the check for SSACFG?

math-fehr: Oh wait sorry I'm wrong! Though you can remove the check that the block is inside a region

regehr:
ok! maybe this is better now :)

math-fehr (approved):

PR #695, opened 2026-05-30
chore(interpreter): add `interpretFunction`

math-fehr:
`interpretFunction` interprets an operation representing a function by interpreting its region given a memory state and an array of values (that will be passed to its region arguments).

This function will be useful to define the semantics of a program, and more importantly the refinement relation between two functions.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -122,6 +122,12 @@ structure VariableState (ctx : WfIRContext OpInfo) where
   variablesIn : ∀ val, val ∈ variables → val.InBounds ctx.raw
 
 /--
+  Create a variable state with no variables defined.
+-/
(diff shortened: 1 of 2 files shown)


regehr (approved):
that's exciting to see progress towards function calls!
this LGTM.

regehr:
@math-fehr after my patch about terminators lands, I could work on making veir insist that function returns agree with function return types. that seems obviously on the critical path for supporting function calls. let me know if I should try to do that or if you wanted to.

tobiasgrosser:
> @math-fehr after my patch about terminators lands, I could work on making veir insist that function returns agree with function return types. that seems obviously on the critical path for supporting function calls. let me know if I should try to do that or if you wanted to.

I think that's a good idea. We have plenty of test cases that get this wrong. I encounter this as I add `mlir-opt` testing.

tobiasgrosser (approved):

nchappe (approved):

PR #698, opened 2026-05-30
chore: allow multiple MLIR versions

regehr:
I habitually keep a top-of-tree LLVM/MLIR in my PATH. 
ok for us to support this as well as 22.0.0?


math-fehr:
I would be fine with it, but I'll let @tobiasgrosser decide on it!
We might want to add it in the CI as well, I'm not sure.

tobiasgrosser (approved):
We should eventually have CI coverage for any MLIR version we support. Happy to add this for now, though.

PR #699, opened 2026-05-30
chore(parser): add better error for block definitions

math-fehr:
The parser now provides better error locations when a block is defined twice, or a block is used but never defined.


regehr (approved):
this looks great!

PR #701, opened 2026-05-30
fix(LLVM): parse and print support for C-style variadic function arguments

regehr:
this is needed for passing almost any real C code through VeIR

### DIFF
--- a/UnitTest/AttrParser.lean
+++ b/UnitTest/AttrParser.lean
@@ -219,36 +219,42 @@ macro "#assert " e:term : command =>
 /-! ## LLVM Function type -/
 #assert expectSuccessType "!llvm.func<i32 (i32)>"
   ⟨.llvmFunctionType (FunctionType.mk
-    #[(IntegerType.mk 32 : Attribute)] #[(IntegerType.mk 32 : Attribute)]), by rfl⟩
+    #[(IntegerType.mk 32 : Attribute)] #[(IntegerType.mk 32 : Attribute)] (isVarArg := false)), by rfl⟩
 #assert expectSuccessType "!llvm.func<i64 ()>"
-  ⟨.llvmFunctionType (FunctionType.mk #[] #[(IntegerType.mk 64 : Attribute)]), by rfl⟩
+  ⟨.llvmFunctionType (FunctionType.mk #[] #[(IntegerType.mk 64 : Attribute)] (isVarArg := false)), by rfl⟩
 #assert expectSuccessType "!llvm.func<i32 (i32, i64)>"
   ⟨.llvmFunctionType (FunctionType.mk
     #[(IntegerType.mk 32 : Attribute), (IntegerType.mk 64 : Attribute)]
-    #[(IntegerType.mk 32 : Attribute)]), by rfl⟩
+    #[(IntegerType.mk 32 : Attribute)] (isVarArg := false)), by rfl⟩
 #assert expectSuccessType "!llvm.func<!llvm.ptr (!llvm.ptr)>"
   ⟨.llvmFunctionType (FunctionType.mk
-    #[(LLVM.PointerType.mk : Attribute)] #[(LLVM.PointerType.mk : Attribute)]), by rfl⟩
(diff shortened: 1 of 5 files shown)


UnitTest/AttrParser.lean
```diff
@@ -219,36 +219,42 @@ macro "#assert " e:term : command =>
 /-! ## LLVM Function type -/
 #assert expectSuccessType "!llvm.func<i32 (i32)>"
   ⟨.llvmFunctionType (FunctionType.mk
-    #[(IntegerType.mk 32 : Attribute)] #[(IntegerType.mk 32 : Attribute)]), by rfl⟩
+    #[(IntegerType.mk 32 : Attribute)] #[(IntegerType.mk 32 : Attribute)] (isVarArg := false)), by rfl⟩
```
math-fehr, line 222: I guess you don't actually need to change these first tests, as `isVarArg := false` is the default?

Veir/Parser/AttrParser.lean
```diff
@@ -424,11 +429,11 @@ partial def parseOptionalFunctionType : AttrParserM (Option FunctionType) := do
     -- Convert the parsed types to attributes, as FunctionType stores attributes instead
     -- of types.
     let outputs := outputs.map (fun x => x.val)
-    return some (FunctionType.mk inputs outputs)
+    return some (FunctionType.mk inputs outputs (isVarArg := false))
   | none =>
     -- If there is no list of output, check for a single output type.
     let outputType ← parseType "function type output expected"
-    return some (FunctionType.mk inputs #[outputType])
+    return some (FunctionType.mk inputs #[outputType] (isVarArg := false))
```
math-fehr, line 436: Same here, you can probably remove the changes and it should just work?

Veir/Parser/AttrParser.lean
```diff
@@ -453,9 +460,16 @@ partial def parseOptionalLLVMFunctionType : AttrParserM (Option TypeAttr) := do
...
+    if (← parseOptionalToken .ellipsis).isSome then
+      return LLVMFuncParam.ellipsis
+    return LLVMFuncParam.type (← parseLLVMType)
   parsePunctuation ">"
-  let ft := FunctionType.mk (params.map (·.val)) #[result.val]
+  let isVarArg := params.any (· matches .ellipsis)
+  let paramTypes := params.filterMap fun
+    | .type ty => some ty.val
+    | .ellipsis => none
+  let ft := FunctionType.mk paramTypes #[result.val] isVarArg
```
math-fehr, line 474: I think this is a bit dangerous, because we wouldn't actually catch `(i32, ..., i32)` for instance, because all information is lost here?
Could we for instance check here that ellipsis can only appear as the last element?

math-fehr (commented):
Nice! I would just like that we actually check that the ellipsis appear at the very end though :/
Maybe instead of using `parseDelimitedList`, you could write your own while loop that will stop if it sees a `...`?

regehr:
> Nice! I would just like that we actually check that the ellipsis appear at the very end though :/ Maybe instead of using `parseDelimitedList`, you could write your own while loop that will stop if it sees a `...`?

ah, I had hoped to leave this for the verifier as it seemed to make the parse code uglier. but I'm sure I just did not figure out a way to make it look nice. I'll see what I can do.

UnitTest/AttrParser.lean
regehr: sigh... if there is a way to leave these out, I don't know how to make it work.
this particular kind of constructor does not get the default value, it requires all fields to be initialized. 
we can instead use the anonymous constructor but that likely results in a bigger patch than this.

math-fehr: Oh okay, let's keep them then, I'm not sure why this doesn't behave that way, but that's fine then!

Veir/Parser/AttrParser.lean
regehr: ah, right, this can work, thanks!

regehr:
ok I think this is ready for another super quick look Mathieu

math-fehr (approved):
All good, thanks!

PR #702, opened 2026-05-30
chore(IR): add `OperationPtr.getParentOp!`

math-fehr:
This is a useful method to have on `OperationPtr`. I do not define the non-! version of it yet, as it requires `FieldsInBounds` to be correct and we should figure out where to put these definitions first.


regehr (approved):
LGTM!

PR #703, opened 2026-05-30
feat(interpreter): add refinement definition for functions and modules

math-fehr:
This adds propositions that defines function and module-wide refinements.
First, a runtime value is refining another if either the values are equal, or the values are `LLVM.Int` and refines using the `LLVM.Int` refinement relation.
Then, a function refines another if, for all inputs, its results refine pointwise the other, and memory are equal.
Finally, a module refines another if, for all functions in the source module, it contains a function with the same name that refines the source function.

While this is probably not the long-term refinement relation we want to have, this is a good first step I believe.

### DIFF
--- a/Veir/Interpreter/Refinement.lean
+++ b/Veir/Interpreter/Refinement.lean
@@ -1,0 +1,1 @@
+import Veir.Interpreter.Refinement.Basic
--- a/Veir/Interpreter/Refinement/Basic.lean
+++ b/Veir/Interpreter/Refinement/Basic.lean
@@ -1,0 +1,118 @@
+import Veir.Interpreter.Basic
+import Veir.Data.Refinement
+
+/-!
+# Refinement of programs
+
+Defines when one program is *refined by* another across two `WfIRContext`s (which lets us
+relate a program to a rewritten or lowered version of it). Refinement is defined at three levels:
+
+* `RuntimeValue.isRefinedBy` relates two runtime values: integers refine via the `· ⊒ ·` ordering on
+  `LLVM.Int`, while other types of values must match exactly.
+* `OperationPtr.isRefinedByAsFunction` relates two function-like operations: interpreting the source
+  with any arguments and memory is refined by interpreting the target.
+* `OperationPtr.isRefinedByAsModule` relates two modules: every top-level `func.func` of the source
+  module must be refined, as a function, by a same-named top-level `func.func` of the target module.
+-/
+
+open Veir.Data
+
+namespace Veir
+
+/-- Refinement relation between two runtime values. -/
+def RuntimeValue.isRefinedBy (source target : RuntimeValue) : Prop :=
+  match source, target with
+  | .int bw s, .int bw' t => ∃ h : bw = bw', s.cast h ⊒ t
+  | .addr s, .addr t => s = t
+  | .reg s, .reg t => s = t
+  | .float bw s, .float bw' t => bw = bw' ∧ s = t
+  | _, _ => False
+
+@[inherit_doc] infix:50 " ⊒ " => RuntimeValue.isRefinedBy
+
+/--
+An array `source` of runtime values is refined by `target`. This asserts that the arrays have
+the same size, and that they refine pointwise.
+-/
+def RuntimeValue.arrayIsRefinedBy (source target : Array RuntimeValue) : Prop :=
+  source.size = target.size ∧
+    ∀ (i : Nat) (_ : i < source.size), source[i]! ⊒ target[i]!
+
+@[inherit_doc] infix:50 " ⊒ " => RuntimeValue.arrayIsRefinedBy
+
+/--
+A function interpretation `source` is refined by `target`. This asserts that the final memories
+are equal, and the returned values refine pointwise.
+-/
+def FunctionResult.isRefinedBy (source target : MemoryState × Array RuntimeValue) : Prop :=
+  source.1 = target.1 ∧ source.2 ⊒ target.2
+
+@[inherit_doc] infix:50 " ⊒ " => FunctionResult.isRefinedBy
+
+/--
+An interpretation result `source` is refined by `target` given a refinement relation `R`
+on the underlying values. This asserts:
+* every well-defined outcome `some (.ok a)` of `source` must be matched by an outcome
+  `some (.ok b)` of `target` with `R a b`;
+* when `source` is undefined behaviour (`some .ub`), `target` must succeed (i.e. not be `none`),
+  but may be either `some .ub` or `some (.ok _)`;
+* when `source` or `target` failed interpretation (i.e. are `none`), no refinement exists.
+-/
+def Interp.isRefinedBy (R : α → α → Prop) (source target : Interp α) : Prop :=
+  match source, target with
+  | some (.ok a), some (.ok b) => R a b
+  | some .ub, some _ => True
+  | _, _ => False
+
+/--
+The function described by source `op₁` (in `ctx₁`) is *refined by* target `op₂` (in `ctx₂`) when,
+for every argument `values` and initial memory `mem`, interpreting `op₁` is refined by interpreting
+`op₂`.
+-/
+def OperationPtr.isRefinedByAsFunction (op₁ : OperationPtr) (ctx₁ : WfIRContext OpCode)
+    (op₂ : OperationPtr) (ctx₂ : WfIRContext OpCode)
+    (op₁In : op₁.InBounds ctx₁.raw := by grind)
+    (op₂In : op₂.InBounds ctx₂.raw := by grind) : Prop :=
+  ∀ (valuesSource valuesTarget : Array RuntimeValue) (mem : MemoryState),
+    valuesSource ⊒ valuesTarget →
+    Interp.isRefinedBy FunctionResult.isRefinedBy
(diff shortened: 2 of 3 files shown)


Veir/Interpreter/Refinement.lean
```diff
@@ -0,0 +1,111 @@
...
+-/
+
+open Veir.Data
+
+namespace Veir
+
+/-- Refinement relation between two runtime values. -/
+def RuntimeValue.isRefinedBy (source target : RuntimeValue) : Prop :=
+  match source, target with
+  | .int bw s, .int bw' t => ∃ h : bw = bw', s.cast h ⊑ t
```
regehr: I don't understand the cast

Veir/Interpreter/Refinement.lean
```diff
@@ -0,0 +1,111 @@
...
+values. This asserts:
+* every well-defined outcome `some (.ok a)` of `source` must be matched by some `some (.ok b)` of `target`
+  with `R a b`;
+* when `source` is undefined behaviour (`some .ub`), `target` must succeed (i.e. not be `none`), but may
+  be either `some .ub` or `some (.ok _)`;
+* when `source` failed interpretation, `target` is unconstrained.
+-/
+def Interp.isRefinedBy (R : α → α → Prop) (source target : Interp α) : Prop :=
+  (∀ a, source = some (.ok a) → ∃ b, target = some (.ok b) ∧ R a b) ∧
+  (source = some .ub → target ≠ none)
```
regehr: just so I understand, `target != none` simply means that the interpreter fails to encounter an error on any execution of target?

Veir/Interpreter/Refinement.lean
```diff
@@ -0,0 +1,111 @@
...
+`op₂`.
+-/
+def OperationPtr.isRefinedByAsFunction (op₁ : OperationPtr) (ctx₁ : WfIRContext OpCode)
+    (op₂ : OperationPtr) (ctx₂ : WfIRContext OpCode)
+    (op₁In : op₁.InBounds ctx₁.raw := by grind)
+    (op₂In : op₂.InBounds ctx₂.raw := by grind) : Prop :=
+  ∀ (values : Array RuntimeValue) (mem : MemoryState),
+    Interp.isRefinedBy FunctionResult.isRefinedBy
+      (interpretFunction op₁ values mem (ctx := ctx₁) op₁In)
+      (interpretFunction op₂ values mem (ctx := ctx₂) op₂In)
```
regehr: isn't it necessary to somehow relate ctx1 and ctx2?

regehr (approved):
LGTM but there's stuff going on I don't understand :)

Veir/Interpreter/Refinement.lean
math-fehr: So here, `s` and `t` do not have the same type in Lean, so you need to cast `s` so it has the same type as `t`, so you can compare them.
Here, `s.cast h` means "cast s of type `BitVec bw` to type `BitVec bw'` using `h`"

Veir/Interpreter/Refinement.lean
math-fehr: Yes, we do not want to have `ub` as input, but an interpretation error as output.
Since `Interp` is a bit broken right now, that means that we do not really handle non-termination, because it is also encoded as `none` sadly. That's something we need to fix later.

Veir/Interpreter/Refinement.lean
math-fehr: I don't think so at this point! We only care about the function "semantics" here, so they can be from very different modules, as long as one refines the other, it should be good!

Veir/Interpreter/Refinement.lean
regehr: ah... since we're not using dependent types, the typechecker doesn't understand that bw=bw', got it

regehr: what an irritating set of tradeoffs!

Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -0,0 +1,111 @@
...
+`op₂`.
+-/
+def OperationPtr.isRefinedByAsFunction (op₁ : OperationPtr) (ctx₁ : WfIRContext OpCode)
+    (op₂ : OperationPtr) (ctx₂ : WfIRContext OpCode)
+    (op₁In : op₁.InBounds ctx₁.raw := by grind)
+    (op₂In : op₂.InBounds ctx₂.raw := by grind) : Prop :=
+  ∀ (values : Array RuntimeValue) (mem : MemoryState),
+    Interp.isRefinedBy FunctionResult.isRefinedBy
+      (interpretFunction op₁ values mem (ctx := ctx₁) op₁In)
+      (interpretFunction op₂ values mem (ctx := ctx₂) op₂In)
```
ineol: why not if the inputs refine each others, the output refine each others, logical relation style?

ineol (commented):
LGTM except the direction of the sign like Nicolas remarked and the question about definition for op refinement

tobiasgrosser (approved):
Thank you all for the nice reviews. I have nothing further to add, beyond the comments already raised.

Veir/Interpreter/Refinement/Basic.lean
math-fehr: Done!

Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -0,0 +1,111 @@
...
+@[inherit_doc] infix:50 " ⊒ " => FunctionResult.isRefinedBy
+
+/--
+An interpretation result `source` is refined by `target` given a refinement relation `R` on the underlying
+values. This asserts:
+* every well-defined outcome `some (.ok a)` of `source` must be matched by some `some (.ok b)` of `target`
+  with `R a b`;
+* when `source` is undefined behaviour (`some .ub`), `target` must succeed (i.e. not be `none`), but may
+  be either `some .ub` or `some (.ok _)`;
+* when `source` failed interpretation, `target` is unconstrained.
```
nchappe: I think when interpretation returns `none` (on either side), the refinement should not hold. Doing anything else means reasoning about unimplemented/unsupported semantics.

Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -0,0 +1,111 @@
...
+  source.1 = target.1 ∧ source.2 ⊒ target.2
+
+@[inherit_doc] infix:50 " ⊒ " => FunctionResult.isRefinedBy
+
+/--
+An interpretation result `source` is refined by `target` given a refinement relation `R` on the underlying
+values. This asserts:
+* every well-defined outcome `some (.ok a)` of `source` must be matched by some `some (.ok b)` of `target`
+  with `R a b`;
+* when `source` is undefined behaviour (`some .ub`), `target` must succeed (i.e. not be `none`), but may
```
nchappe: I believe this case is not currently handled.

Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -0,0 +1,111 @@
...
+  be either `some .ub` or `some (.ok _)`;
+* when `source` failed interpretation, `target` is unconstrained.
+-/
+def Interp.isRefinedBy (R : α → α → Prop) (source target : Interp α) : Prop :=
+  ∀ a, source = some (.ok a) → ∃ b, target = some (.ok b) ∧ R a b
```
nchappe: I would typically expect the `forall` to be on the target side in the definition of a refinement relation.
Otherwise, you could have some results on the target side that are not related to any source result.
(Here it still works as is because of determinism.)

Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -0,0 +1,111 @@
...
+  isTopLevel : op.getParentOp! ctx = some moduleOp
+
+/--
+The module `mod₁` (in `ctx₁`) is *refined by* the module `mod₂` (in `ctx₂`) when every top-level
+`func.func` of `mod₁` is refined, as a function, by a top-level `func.func` of `mod₂` that carries
+the same symbol name.
+-/
+def OperationPtr.isModuleRefinedBy (mod₁ : OperationPtr) (ctx₁ : WfIRContext OpCode)
+    (mod₂ : OperationPtr) (ctx₂ : WfIRContext OpCode) : Prop :=
+  ∀ (func₁ : OperationPtr) (func₁In : func₁.InBounds ctx₁.raw) (name : StringAttr),
```
nchappe, line 112: As is, there could be more functions defined in the target module. I don't think it's a major problem for now but you could perhaps add a comment about that?

Veir/Interpreter/Refinement/Basic.lean
math-fehr: Should be good now

Veir/Interpreter/Refinement/Basic.lean
math-fehr: Fixed

Veir/Interpreter/Refinement/Basic.lean
math-fehr: Fixed

Veir/Interpreter/Refinement/Basic.lean
math-fehr: Added a comment

nchappe (approved):

PR #704, opened 2026-05-30
feat(refinement): add reflexivity and transitivity lemmas for module refinements

math-fehr:
Those are simple lemmas that are needed when reasoning about refinements

### DIFF
--- a/Veir/Data/Refinement.lean
+++ b/Veir/Data/Refinement.lean
@@ -28,3 +28,11 @@ theorem isRefinedBy_eq {w : Nat} (i i' : Veir.Data.LLVM.Int w) :
             | .val v, .val v' => v = v'
             | .val _, .poison => False
             | .poison, _ => True) := by rfl
+
+@[simp, grind .]
+theorem isRefinedBy_refl {w : Nat} (i : Veir.Data.LLVM.Int w) : i ⊒ i := by
+  grind [isRefinedBy, cases Veir.Data.LLVM.Int]
+
+theorem isRefinedBy_trans {w : Nat} {i j k : Veir.Data.LLVM.Int w}
+    (h₁ : i ⊒ j) (h₂ : j ⊒ k) : i ⊒ k := by
+  grind [isRefinedBy, cases Veir.Data.LLVM.Int]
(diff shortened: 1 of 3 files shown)


Veir/Data/Refinement.lean
```diff
@@ -28,3 +28,11 @@ theorem isRefinedBy_eq {w : Nat} (i i' : Veir.Data.LLVM.Int w) :
             | .val v, .val v' => v = v'
             | .val _, .poison => False
             | .poison, _ => True) := by rfl
+
+@[simp, grind .]
+theorem isRefinedBy_refl {w : Nat} (i : Veir.Data.LLVM.Int w) : i ⊑ i := by
+  grind [isRefinedBy, cases Veir.Data.LLVM.Int]
+
+theorem isRefinedBy_trans {w : Nat} {i j k : Veir.Data.LLVM.Int w}
+    (h1 : i ⊑ j) (h2 : j ⊑ k) : i ⊑ k := by
```
regehr: I don't understand the role of h1 and h2 here, but the theorem looks good and useful

regehr (approved):
this all LGTM!

Veir/Data/Refinement.lean
math-fehr: So having `f : x -> y` and `f (h : x) : y` is equivalent, it's just to add a name to the lemma

ineol: Please use _1 and _2 for the hypotheses

ineol (approved):
Except for the hypothesis names LGTM

tobiasgrosser (approved):
This LGTM (modulo @ineol's comment).

math-fehr:
I removed reflexivity for function and module refinements for now.
The reason is that we do not yet have a proof that all operations "behave well", in the sense that a refinement of their input creates a refinement of their output.

GZGavinZhao (approved):
Nit: for the PR title, the changes fit `feat` better instead of `chore`, recommend changing it to `feat(refinement)`.

tobiasgrosser:
> Nit: for the PR title, the changes fit `feat` better instead of `chore`, recommend changing it to `feat(refinement)`.

Done.

PR #705, opened 2026-05-31
chore(IR): move lemma to the correct file

math-fehr:
We keep our `inBounds` lemmas about the `Rewriter` in their file directly, instead of the GetSet file.


tobiasgrosser (approved):
That's trivially OK.

PR #706, opened 2026-05-31
feat(IR): add inBounds lemmas for the WfRewriter

math-fehr:
These lemmas existed for the `Rewriter`, but were never ported to the `WfRewriter`.


tobiasgrosser (approved):
Nice!!

PR #708, opened 2026-05-31
feat(LLVM): add first-class `#llvm.fastmath` attribute

tobiasgrosser:
We add a first-class `#llvm.fastmath<>` attribute, replacing the previously used `fast`, `nnan`, `ninf`, and `nsz` unit attributes. This aligns our treatment of fastmath properties with `mlir-opt`. Concluding a series of patches, we can now also add `MLIR_ROUNDTRIP` to `Test/LLVM/identity.mlir`.

In contrast to MLIR, we do not store the `fast` flag separately, but always treat it as the combination of `nnan`, `ninf`, and `nsz`, which is canonical. Whenever all three options are set we print `fast` as the fast math mode, which is a simplification MLIR itself does not do. However, mlir folds `nnan`, `ninf`, and `nsz` into `fast` when both are present.

### DIFF
--- a/Veir/Parser/AttrParser.lean
+++ b/Veir/Parser/AttrParser.lean
@@ -163,17 +163,17 @@ def parseOptionalDenseArrayAttr : AttrParserM (Option DenseArrayAttr) := do
 
 def parseFastMathFlag : AttrParserM FastMathFlagsAttr := do
   if (← parseOptionalKeyword "fast".toByteArray) then
-    return { fast := true, nnan := false, ninf := false, nsz := false }
+    return { nnan := true, ninf := true, nsz := true }
   else if (← parseOptionalKeyword "nnan".toByteArray) then
-    return { fast := false, nnan := true, ninf := false, nsz := false }
+    return { nnan := true, ninf := false, nsz := false }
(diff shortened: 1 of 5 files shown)


Veir/Parser/AttrParser.lean
```diff
@@ -260,6 +290,10 @@ partial def parseOptionalDialectAttr : AttrParserM (Option Attribute) := do
   let startPos ← getPos
   let dialectName ← parseOptionalPrefixedKeyword .hashIdent
   let some dialectName := dialectName | return none
+
+  if dialectName = "llvm.fastmath".toByteArray then do
+    return ← parseFloatFastMathFlagsAttr
+
```
tobiasgrosser, line 295: @math-fehr, I am not sure this is the right place to hook this into the parser. A careful review would be appreciated.

regehr (approved):
cool! this all LGTM

math-fehr (approved):

Veir/Parser/AttrParser.lean
math-fehr: I think this is the right place, we haven't really done this for the other attributes, but we should move them there.

PR #711, opened 2026-05-31
feat(lit): run binaries without going through lake

tobiasgrosser:
Call `lake build` before running the individual test cases and then directly run `.lake/build/bin/veir-opt`. This avoids potential race conditions triggered by multiple `lake build`. It also accelerates the individual test cases as we only check a single time (for all test cases) if the binary must be rebuild.

This improves testing time on a recent Mac Book Pro (M4) from `21.83s` to `2.19s`, a speedup of about `10x`. Our GitHub CI accelerates from `1m22s` to `19s`.


math-fehr (approved):

regehr (approved):
LGTM!

PR #714, opened 2026-05-31
feat(LLVM): support llvm.unreachable

regehr:
`llvm.unreachable` comes up fairly commonly in code that we get via `clang` -> `mlir-translate`. luckily it has an extremely easy semantics in veir :)

### DIFF
--- a/Test/Interpreter/LLVM/unreachable.mlir
+++ b/Test/Interpreter/LLVM/unreachable.mlir
@@ -1,0 +1,10 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// Executing `llvm.unreachable` is immediate undefined behaviour.
+"builtin.module"() ({
+  "func.func"() <{sym_name = "main"}> ({
+    "llvm.unreachable"() : () -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Undefined behavior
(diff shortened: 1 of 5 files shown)


tobiasgrosser (approved):

Test/Interpreter/LLVM/unreachable.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// Executing `llvm.unreachable` is immediate undefined behaviour.
+"builtin.module"() ({
+  "func.func"() <{sym_name = "main"}> ({
+    "llvm.unreachable"() : () -> ()
```
math-fehr, line 6:
```suggestion
    "llvm.unreachable"() : () -> ()
    "func.return"() : () -> ()
```
Otherwise this will fail once we check for terminators

math-fehr (approved):
All good, just one comment on the test

Test/Interpreter/LLVM/unreachable.mlir
regehr: argh I applied this without thinking carefully enough. I backed it out-- we're already checking for terminators, and `llvm.unreachable` is one of those (and I added it to the list as part of this patch)

PR #715, opened 2026-05-31
refactor(dominance): clean up of dominance analysis unit tests

axelcool1234:
Also removed dominance analysis' `getDoms?` since no one will use it.

What was cleaned up:
 - Created a private type `ExpectedBlockDominance` so that dominance tests aren't passing an `Array (String × Option (Array String))` which has no readable semantic meaning. Instead, it now passes a `structure` with a block label and its expected dominators.
    - Consequently, I am not passing around `String`s and `Array String`s into my helpers - I just pass `ExpectedBlockDominance` and `Array ExpectedBlockDominance` directly.
 - On a similar note, I originally passed the `RecoveredNames.blocks` into my helpers - so their type signature often had `HashMap String BlockPtr` - also not as readable. I decided to just pass `RecoveredNames` directly. 
 - Many functions had nested `match` and if statements. I flattened them.
 - Many of these functions take in `IRContext` and `DataFlowContext`, but the ordering of these parameters were random. I normalized them (`IRContext` and `DataFlowContext` are the last parameters, which is what all of my functions do in the dataflow analysis framework)
 - I changed some of the comments for functions to be more concise.
 - `recoverNames?` now returns an exception if it fails.

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -2,70 +2,61 @@ import UnitTest.DataFlowFramework.Helpers
 
 import Veir.Analysis.DataFlow.DominanceAnalysis
 
-open Std (HashMap HashSet)
+open Std (HashSet)
 open Veir
+
+/-- Expected dominance information for one named block in the test harness. -/
+private structure ExpectedBlockDominators where
+  name : String
+  dominators : HashSet String
 
 /--
 Compare one expected dominator label against the observed dominance information.
 
-This checks `dominates`, `properlyDominates`, and membership in the dominator
-set returned by `BlockPtr.getDoms?`.
+This is the completeness half of the reachable block check: every expected
+dominator must be observed.
 -/
 private def compareExpectedDominator
-    (name : String)
+    (expected : ExpectedBlockDominators)
     (block : BlockPtr)
-    (observedDoms : HashSet BlockPtr)
+    (recovered : RecoveredNames)
+    (expectedDom : String)
     (dfCtx : DataFlowContext)
-    (irCtx : IRContext OpCode)
-    (blockMap : HashMap String BlockPtr)
-    (expectedDom : String) : MismatchReport :=
-  match blockMap[expectedDom]? with
-  | none =>
-    #[s!"dominators {name}: missing block label {expectedDom}"]
-  | some expectedBlock =>
-    Id.run do
-      let shouldProperlyDom := expectedDom ≠ name
-      let mut report := #[]
-      if !Veir.DominanceAnalysis.dominates expectedBlock block dfCtx irCtx then
-        report := report.push s!"dominators {name}: missing expected dominator {expectedDom}"
-      if !observedDoms.contains expectedBlock then
-        report := report.push
-          s!"dominators {name}: BlockPtr.getDoms? missing expected dominator {expectedDom}"
-      if Veir.DominanceAnalysis.properlyDominates expectedBlock block dfCtx irCtx ≠ shouldProperlyDom then
-        report := report.push
-          s!"dominators {name}: unexpected properlyDominates result for {expectedDom}"
-      report
+    (irCtx : IRContext OpCode) : MismatchReport := Id.run do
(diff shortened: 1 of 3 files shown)


regehr (approved):
@math-fehr should take a quick look but these changes seem fine to me Axel

axelcool1234:
Yeah I mostly just flattened a lot of functions (i.e. avoid nested if statements and matches), made their parameters match (some helpers took `IRContext` first, some last when it came to parameter ordering). I also made a simple `structure` type just for tests so that it's more readable. Simple stuff like that

tobiasgrosser:
Thank you for your change. This PR is reasonably small. However, it is not clear how and why there are changes. Can you explain in the PR body what was undesirable and how your cleanup addresses this. A good PR message should be such that (a) I should be convinced by the PR message that this change is desirable and (b) I should be able to check that the code changes implement exactly what is promised in the PR message.

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -2,70 +2,62 @@ import UnitTest.DataFlowFramework.Helpers
 
 import Veir.Analysis.DataFlow.DominanceAnalysis
 
-open Std (HashMap HashSet)
 open Veir
 
+/-- Expected dominance information for one named block in the test harness. -/
+private structure ExpectedBlockDominance where
+  name : String
+  dominators? : Option (Array String)
```
math-fehr: Why is this an array instead of a hashset? Does the order of dominators matter here?

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -2,70 +2,62 @@ import UnitTest.DataFlowFramework.Helpers
 
 import Veir.Analysis.DataFlow.DominanceAnalysis
 
-open Std (HashMap HashSet)
 open Veir
 
+/-- Expected dominance information for one named block in the test harness. -/
+private structure ExpectedBlockDominance where
```
math-fehr:
```suggestion
private structure ExpectedBlockDominators where
```
Maybe?

UnitTest/DataFlowFramework/Helpers.lean
```diff
@@ -136,24 +131,25 @@ Recover block and SSA value maps by pairing MLIR source names with IR traversal
 def recoverNames?
     (top : OperationPtr)
     (irCtx : IRContext OpCode)
-    (mlir : String) : Option RecoveredNames :=
+    (mlir : String) : Except String RecoveredNames := do
   let blockLabels := blockLabelsFromMlir mlir
```
math-fehr, line 135: Probably rename this to `recoverNames` since this doesn't return an option anymore?

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -2,70 +2,62 @@ import UnitTest.DataFlowFramework.Helpers
...
-    (dfCtx : DataFlowContext)
-    (irCtx : IRContext OpCode)
     (observedName : String)
-    (observedBlock : BlockPtr) : MismatchReport :=
-  Id.run do
+    (observedBlock : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : MismatchReport := Id.run do
+    let some expectedDoms := expected.dominators?
+      | return #[]
```
math-fehr: You use 4 character indentation here instead of 2 (which we use everywhere else)

UnitTest/DataFlowFramework/Dominance.lean
math-fehr: Is there any reason to have this being an `Option`? It seems you always expect *something*, so this should be probably just an `Array String`.

math-fehr (commented):
Thanks, I have some small stylistic comments, but otherwise sholud be good!

UnitTest/DataFlowFramework/Dominance.lean
axelcool1234: It does not, good suggestion!

axelcool1234:
@math-fehr Done!

math-fehr (approved):

PR #716, opened 2026-05-31
feat(dominance): immediate dominator API

axelcool1234:
This creates a new file `IR/Dominance.lean` which will eventually be populated with other functions meant for users to call when querying about dominance information. To start this off, I went with the simplest one - `immediateDominator?` is a wrapper for `getIDoms?`, which no one should use beyond `DominanceAnalysis` internally. 

I also added in checks in the dominance unittests for immediate dominators.

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -1,6 +1,7 @@
 import UnitTest.DataFlowFramework.Helpers
 
 import Veir.Analysis.DataFlow.DominanceAnalysis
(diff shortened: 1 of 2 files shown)


axelcool1234:
cc @math-fehr still cant add reviewers

axelcool1234:
cc @regehr

math-fehr (approved):

math-fehr:
It's fine for me to merge, unless you want john to look at it first? Otherwise feel free to just push the merge button (I think you should be able to now?)

PR #720, opened 2026-05-31
feat(dominance): Add dominance API for blocks

axelcool1234:
Similar to MLIR, this is region-aware (i.e., you can ask for the dominance relation between two blocks from differing regions). I created private `InsertPoint` dominance functions which the public block dominance functions use. Operation dominance functions will use them too eventually. I also added tests for dominance with regions.

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -321,18 +321,17 @@ def testDomIfLoopIf : String :=
   do not dominate each other.
 -/
 def testDomNestedRegions : String :=
-  run
-    "\"builtin.module\"() ({\n\
-    ^bb0:\n\
-      \"test.test\"() ({\n\
-      ^bb1:\n\
-        \"test.test\"() : () -> ()\n\
-      }) : () -> ()\n\
-      \"test.test\"() ({\n\
-      ^bb2:\n\
-        \"test.test\"() : () -> ()\n\
-      }) : () -> ()\n\
-    }) : () -> ()"
+  run r#""builtin.module"() ({
+^bb0:
+  "test.test"() ({
+  ^bb1:
+    "test.test"() : () -> ()
+  }) : () -> ()
+  "test.test"() ({
+  ^bb2:
+    "test.test"() : () -> ()
+  }) : () -> ()
+}) : () -> ()"#
     #[ { name := "bb0", dominators := { "bb0" },        iDom := "bb0" }
      , { name := "bb1", dominators := { "bb0", "bb1" }, iDom := "bb1" }
      , { name := "bb2", dominators := { "bb0", "bb2" }, iDom := "bb2" }
@@ -354,20 +353,19 @@ def testDomNestedRegions : String :=
           └───────┘
 -/
 def testDomDiamondNestedJoin : String :=
-  run
-    "\"builtin.module\"() ({\n\
-^bb0:\n\
-  \"test.test\"() [^bb1, ^bb2] : () -> ()\n\
-^bb1:\n\
-  \"test.test\"() [^bb3] : () -> ()\n\
-^bb2:\n\
-  \"test.test\"() [^bb3] : () -> ()\n\
-^bb3:\n\
-  \"test.test\"() ({\n\
-^bb4:\n\
-    \"test.test\"() : () -> ()\n\
-  }) : () -> ()\n\
-}) : () -> ()"
+  run r#""builtin.module"() ({
+^bb0:
+  "test.test"() [^bb1, ^bb2] : () -> ()
+^bb1:
+  "test.test"() [^bb3] : () -> ()
+^bb2:
+  "test.test"() [^bb3] : () -> ()
+^bb3:
+  "test.test"() ({
+^bb4:
+    "test.test"() : () -> ()
+  }) : () -> ()
+}) : () -> ()"#
     #[ { name := "bb0", dominators := { "bb0" },               iDom := "bb0" }
      , { name := "bb1", dominators := { "bb0", "bb1" },        iDom := "bb0" }
      , { name := "bb2", dominators := { "bb0", "bb2" },        iDom := "bb0" }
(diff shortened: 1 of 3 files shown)


regehr:
I feel like I'm losing the thread here. 
I get that you're adding the block interface.
in the meantime, what's going on with immediateDominator? is that going to go away? be updated? stay like it is?
I guess it doesn't matter for this PR, but I don't know who needs immediateDominator? anyhow

axelcool1234:
> I feel like I'm losing the thread here. I get that you're adding the block interface. in the meantime, what's going on with immediateDominator? is that going to go away? be updated? stay like it is? I guess it doesn't matter for this PR, but I don't know who needs immediateDominator? anyhow

I was going to keep it - but if you're convinced it won't be useful we can also remove it later.

regehr:
> > I feel like I'm losing the thread here. I get that you're adding the block interface. in the meantime, what's going on with immediateDominator? is that going to go away? be updated? stay like it is? I guess it doesn't matter for this PR, but I don't know who needs immediateDominator? anyhow
> 
> I was going to keep it - but if you're convinced it won't be useful we can also remove it later.

I'm certainly not convinced that it won't be useful-- but also I have no use for it in mind. is this something you've seen in other compilers? or in SSA algorithms?

axelcool1234:
> > > I feel like I'm losing the thread here. I get that you're adding the block interface. in the meantime, what's going on with immediateDominator? is that going to go away? be updated? stay like it is? I guess it doesn't matter for this PR, but I don't know who needs immediateDominator? anyhow
> > 
> > 
> > I was going to keep it - but if you're convinced it won't be useful we can also remove it later.
> 
> I'm certainly not convinced that it won't be useful-- but also I have no use for it in mind. is this something you've seen in other compilers? or in SSA algorithms?

It's used as a way to efficiently store dominance chains for blocks. Instead of storing an entire set of dominators for a block, you store just the immediate dominator. If you want to see all dominators for a block, you can just query for immediate dominator over and over. But as for how it's used in optimizations? I'm not sure.

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,127 @@ public section
...
+open Std (HashSet)
+
+/--
+Normalize an insertion point into `region` by walking outward through enclosing
+operations until the point lies in a block directly contained in `region`.
+
+If `point` already lies in `region`, it is returned unchanged. If the walk
+escapes the IR hierarchy before reaching `region`, return `none`.
+-/
+private partial def normalizeInsertPoint?
```
math-fehr: I think here we should not add `?` at the end of the function.
My understanding is that we add `?` only when we can have a non-`?` version, and where `?` dynamically check the hypotheses that are missing in the `?`?
@ineol @luisacicolini probably know more about these conventions!

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,127 @@ public section
...
+private partial def dominatesWithinRegion
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  if dominator = block then
+    true
+  else
+    match block.getIDom? dfCtx irCtx with
+    | some idom => idom ≠ block && dominatesWithinRegion dominator idom dfCtx irCtx
+    | none => false
```
math-fehr:
```suggestion
    let some idom := block.getIDom? dfCtx irCtx | return none
    idom ≠ block && dominatesWithinRegion dominator idom dfCtx irCtx
```
and adding `Id.run` at the beginning of the function?

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,127 @@ public section
...
+  let pointInRegion? :=
+    match point.block! irCtx with
+    | none => none
+    | some pointBlock =>
+        if (pointBlock.get! irCtx).parent = dominatorRegion then
+          point
+        else
+          -- Scoot up the point's region tree until we find a location in the
+          -- dominator's region that encloses it.  If this fails, then we know 
+          -- there is no dominance relation.
```
math-fehr:
```suggestion
          -- dominator's region that encloses it.  If this fails, then we know 
          -- there is no dominator doesn't properly dominates point.
```
Because it could be the other way around

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,127 @@ public section
...
+          normalizeInsertPoint? dominatorRegion point irCtx
+  let some point := pointInRegion?
+    | return false
+  let some pointBlock := point.block! irCtx
+    | return false
```
math-fehr: Can you replace this with
```
let some point:= normalizeInsertPoint? dominatorRegion point irCtx
  | return false 
let some pointBlock := point.block! irCtx
  | return false
```
?

I think it should do the same thing, and is much simpler?

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,127 @@ public section
...
+
+  if dominator = point then
+    return true
```
math-fehr: Just so you know, when we expose this entire function, we should add an argument `enclosingOpOk`, and this part should return `enclosingOpOk` instead of `true`.
Because while for operation dominance this is fine, for value dominance we want this to return `false` instead.

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,127 @@ public section
...
+  if dominatorBlock = pointBlock then
+    match dominator, point with
+    | .before dominatorOp, .before op =>
+        return dominatesWithinBlock dominatorOp op irCtx
+    | .before _, .atEnd _ =>
+        return true
+    | .atEnd _, _ =>
+        return false
```
math-fehr: Can you move this into a new function `InsertPoint.dominatesWithinBlock`?
This will remove a bit the complexity from this function, and makes things easier to understand I feel.

Veir/IR/Dominance.lean
```diff
@@ -19,6 +140,27 @@ def immediateDominator?
     (irCtx : IRContext OpCode) : Option BlockPtr :=
   block.getIDom? dfCtx irCtx
 
+/--
+Dominance query between two blocks.
```
math-fehr:
```suggestion
Dominance query between two blocks, where a block dominates itself.
```

Veir/IR/Dominance.lean
```diff
@@ -19,6 +140,27 @@ def immediateDominator?
...
+-/
+def dominates
+    [FactSpec .dominator]
+    (dominator block : BlockPtr)
+    (dfCtx : DataFlowContext)
+    (irCtx : IRContext OpCode) : Bool :=
+  (InsertPoint.atStart! dominator irCtx).dominates (InsertPoint.atStart! block irCtx) dfCtx irCtx
+
+/--
+Proper version of `BlockPtr.dominates`.
```
math-fehr:
```suggestion
Dominance query between two blocks, where a block does not dominates itself.
```

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -308,6 +308,104 @@ def testDomIfLoopIf : String :=
...
+    ^bb0:\n\
+      \"test.test\"() ({\n\
+      ^bb1:\n\
+        \"test.test\"() : () -> ()\n\
+      }) : () -> ()\n\
+      \"test.test\"() ({\n\
+      ^bb2:\n\
+        \"test.test\"() : () -> ()\n\
+      }) : () -> ()\n\
+    }) : () -> ()"
```
math-fehr: I'm sorry to never have questioned this, but you can have multi-lines strings in Lean.
For instance something like this:
```
r#"This is
a multi line string quoted"#
```
Can you change it for the tests you just added, and change it for the others in a different PR?

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -308,6 +308,104 @@ def testDomIfLoopIf : String :=
...
+^bb2:\n\
+      \"test.test\"() : () -> ()\n\
+    }) : () -> ()\n\
+  }) : () -> ()\n\
+}) : () -> ()"
+    #[ { name := "bb0", dominators := { "bb0" },               iDom := "bb0" }
+     , { name := "bb1", dominators := { "bb0", "bb1" },        iDom := "bb1" }
+     , { name := "bb2", dominators := { "bb0", "bb1", "bb2" }, iDom := "bb2" }
+     ]
+
```
math-fehr: Can you add this test, I think it should be interesting since you don't have any loop test yet?
You have a diamond `0, 1, 2, 3`, and `3` also points back to `0`.
Then, you have an exit block `4` from `2`.

math-fehr (commented):
Nice! I added a few comments, but mostly about style and code organization, otherwise everything else looks good!

Veir/IR/Dominance.lean
axelcool1234: To make sure I understand - that's because a value outside an enclosed MLIR region cannot be referenced inside that enclosed region right? Because it's passed in as a basic block argument? @math-fehr

axelcool1234:
Alright, thanks for your suggestions @math-fehr!

Veir/IR/Dominance.lean
math-fehr: It's because of this example:
```
%x = myop.withregion {
  %y = arith.add %x, %x
}
```
Here, `myop.withregion` dominates `arith.add` because `myop.region` is entered before `arith.add` is entered. However, `x` doesn't exist until after the region is executed, so `arith.add` is not allowed to have `x` as a parameter.

So, to do this check, you need to check that the operation producing the value dominates the operation using it, AND that the operation using it is not inside the producing operation region.

math-fehr (approved):
Nice, that looks good for me, let's wait until tomorrow if anyone has a comment, otherwise let's merge this!

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -308,6 +308,133 @@ def testDomIfLoopIf : String :=
...
+  │   │  └───┘  │
+  │ ┌─▼─┐     ┌─▼─┐
+  │ │ 1 │     │ 2 ├──┐
+  │ └─┬─┘     └─┬─┘  │
+  │   │  ┌───┐  │  ┌─▼─┐
+  │   └──► 3 ◄──┘  │ 4 │
+  │      └─┬─┘     └───┘
+  └────────┘
+-/
+def testDomDiamonLoop: String :=
```
axelcool1234: Spelling mistake here 💀

Veir/IR/Dominance.lean
```diff
@@ -6,6 +6,132 @@ public section
...
+/--
+Proper dominance query between two insertion points.
+
+If `point` lies in a nested region outside the region containing `dominator`, it
+is normalized into the dominator's region by replacing it with the enclosing
+operation position. Once both insertion points lie in the same region,
+dominance is decided by same block insertion point order or by block
+dominance.
+-/
+private def properlyDominates
```
ineol: If you remove the first line testing for `dominator = point`, does this function compute
(improper) dominance? If so, maybe defining proper dominance from that
function instead of defining dominance form proper dominance could make sense?

ineol (approved):
LGTM!

axelcool1234:
Alright this is ready for merging :) @math-fehr

PR #721, opened 2026-05-31
feat(dominance): Add dominance API for operations

axelcool1234:


regehr:
@axelcool1234 this is good, but I think what I would like is a way to walk blocks operations in dominance order and then also walk operations in dominance order, is that something you can give me? for example, for CSE I want this ordering in order to miss fewer CSE opportunities. basically, we want the table of available expressions to contain as much context as possible when it reaches each operation. does that make sense?

I also think that I want `InsertPoint.dominates`. if I said earlier that I did not, then apologies

axelcool1234:
> @axelcool1234 this is good, but I think what I would like is a way to walk blocks operations in dominance order and then also walk operations in dominance order, is that something you can give me? for example, for CSE I want this ordering in order to miss fewer CSE opportunities. basically, we want the table of available expressions to contain as much context as possible when it reaches each operation. does that make sense?
> 
> I also think that I want `InsertPoint.dominates`. if I said earlier that I did not, then apologies

So the dominance chain is like a singly linked list. You've got your block and its immediate dominator. You go up the dominance chain (and thus through the linked list) by getting that immediate dominator's immediate dominator, etc. 

For ops in the same block, obviously dominance between them is just going from the top of the block to its terminator. But for ops between blocks, you need to traverse from the _dominated_ up its dominator chain and see if you hit the SUPPOSED _dominator_.

So I guess my question is: can you do CSE by checking ops' dominators? I can provide an iterator for an ops' or blocks' dominator chain (i.e. things that dominate it), but the other way around is more complicated - as again this is a singly linked list rather than doubly linked list. If you need to traverse an op/block's _dominated_ list, I can probably change my dominance analysis to make it doubly linked.

axelcool1234:
Also the InsertPoint API was privated at the request of @math-fehr. I can make it public again in another PR.

axelcool1234:
> If you need to traverse an op/block's _dominated_ list, I can probably change my dominance analysis to make it doubly linked. 

@regehr my very first dominance analysis, before it was attached to the dataflow framework, was doubly linked. I can implement this again if you need it, it'll be a little tricky but it's no problem 😎

math-fehr (commented):
Nice!
What happens when an operation is inside another operation?
I don't know if you recall, but this is something that is controllable in MLIR, because sometimes you want it to dominate, but sometimes you don't (it depends if you are talking about result values, or execution of the operation).

Can you add this boolean on the `properlyDominates` version of the operation? For dominates, we don't need to, because we always want to dominate operation inside your region.

Also, can you add tests for these? ;)

axelcool1234:
> Nice! What happens when an operation is inside another operation? I don't know if you recall, but this is something that is controllable in MLIR, because sometimes you want it to dominate, but sometimes you don't (it depends if you are talking about result values, or execution of the operation).
> 
> Can you add this boolean on the `properlyDominates` version of the operation? For dominates, we don't need to, because we always want to dominate operation inside your region.

Do you mind if I do this in another PR? I added unittests though.

axelcool1234:
Review needed here, @math-fehr - I appreciate it as always :pray:

math-fehr (commented):
Nice! Can you just add one more test where you have at least 2 basic blocks in the same region? Then I think that should be good (I don't have any comment on the code otherwise!)

axelcool1234:
> Nice! Can you just add one more test where you have at least 2 basic blocks in the same region? Then I think that should be good (I don't have any comment on the code otherwise!)

Done! Thanks @math-fehr

math-fehr (approved):
Okay, thanks a lot!

PR #722, opened 2026-05-31
style(dominance): simplified name in dominance unittests

axelcool1234:
Just a simple rename, looks better.

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -9,7 +9,7 @@ open Veir
 /-- Expected dominance information for one named block in the test harness. -/
 private structure ExpectedBlockDominators where
   name : String
-  dominators : HashSet String
+  doms : HashSet String
   iDom : String
 
 /-- Expected dominance result for one pair of named operations. -/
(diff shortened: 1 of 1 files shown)


axelcool1234:
@math-fehr I had this commit ready to go for a bit. Simple rename. Although I'm not sure what's simpler than "dominator" and "dominated" for op tests, so I left that.

math-fehr:
I have to say, I think the previous names actually made more sense? At least, having these shorter names makes it less readable for people that are just looking at the code for the first time?

axelcool1234:
> I have to say, I think the previous names actually made more sense? At least, having these shorter names makes it less readable for people that are just looking at the code for the first time?

If I added comments explaining each field would that be okay?

math-fehr:
I think for these things it's still better to not require the user to use his IDE to understand at a glance right?
Alternative question, why do you want to change these names?

axelcool1234:
> I think for these things it's still better to not require the user to use his IDE to understand at a glance right?
> Alternative question, why do you want to change these names?

@math-fehr Well I have `iDom` as a field, so I thought `doms` would fit. One or the other is not named correctly. Want me to change that to `immediateDominator`?

math-fehr (approved):
Thanks!

PR #723, opened 2026-06-01
fix(ci): only store benchmarks results for PR from the same repo

tobiasgrosser:
The relevant condition was previously reverted by accident. This is now fixed.


axelcool1234:
LGTM. Let's hope that fixes it

axelcool1234:
I don't have permission to "approve" a PR yet.

PR #724, opened 2026-06-01
first-class support for LLVM void type

regehr:
this is a preliminary patch to actually support LLVM's void type (instead of leaving it unregistered) prior to typechecking function return operations against functions' declared return types

### DIFF
--- a/Veir/Parser/AttrParser.lean
+++ b/Veir/Parser/AttrParser.lean
@@ -341,6 +341,18 @@ partial def parseOptionalLocationAttr : AttrParserM (Option Attribute) := do
   return some (LocationAttr.mk body)
 
 /--
+  Parse an LLVM void type `!llvm.void`, if present.
+-/
+partial def parseOptionalLLVMVoidType : AttrParserM (Option TypeAttr) := do
+  let token ← peekToken
+  let .exclamationIdent := token.kind | return none
+  let input := (← getThe ParserState).input
+  let typeName := { token.slice with start := token.slice.start + 1 }.of input
+  if typeName ≠ "llvm.void".toByteArray then return none
+  let _ ← consumeToken
+  return some LLVM.VoidType.mk
+
+/--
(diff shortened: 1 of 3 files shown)


Veir/Parser/AttrParser.lean
```diff
@@ -340,6 +340,18 @@ partial def parseOptionalLocationAttr : AttrParserM (Option Attribute) := do
   parsePunctuation ")"
   return some (LocationAttr.mk body)
 
+/--
+  Parse an LLVM void type `!llvm.void`, if present.
```
axelcool1234, line 344:
```suggestion
Parse an LLVM void type `!llvm.void`, if present.
```

axelcool1234: Is there a format of having two spaces here? I see that @tobiasgrosser also did this - however he accepted my review to delete the spaces.

axelcool1234 (commented):
I can't see anything wrong from my limited understanding.

Veir/Parser/AttrParser.lean
regehr: most of the comments I've seen have the two spaces, so I'm doing it as well. let's try to be uniform about this stuff!

axelcool1234: Welp, all of my code comments are wrong then. Woops!

tobiasgrosser: I think we have a mix of these. We should research what the `Lean4` codebase does and the mathlib community does. Then have a mega-cleanup commit that explains what we want to use and why. This will resolve it once and for all.

tobiasgrosser (approved):
Nice. Let's see what @math-fehr thinks.

Veir/Parser/AttrParser.lean
axelcool1234: Great idea

nchappe (approved):

math-fehr (approved):

PR #727, opened 2026-06-01
chore(ir): add get-set lemmas for WfRewriter.insertOp?

math-fehr:
These are all simple lemmas that are ported from `Rewriter` to `WfRewriter`


tobiasgrosser (approved):
These make sense to me.

ineol (approved):
should it be `wfRewriter` in the lemma names?

PR #728, opened 2026-06-01
chore(interp): replace poison checks with `isPoison` def

luisacicolini:
In the interpreter, we replace the explicit checks for poison with the proper `isPoison` definition, to avoid duplicating code.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -5,6 +5,7 @@ import Veir.IR.WellFormed
 import Veir.PatternRewriter.Basic
 import Veir.Data.Comb.Basic
 import Veir.Data.LLVM.Int.Basic
+import Veir.Data.LLVM.Int.Bitblast
 import Veir.Data.RISCV.Reg.Basic
 import Veir.Data.HW.Basic
 import Veir.Data.Casting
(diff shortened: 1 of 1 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -530,18 +531,8 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
...
-      else if v' = -1 then
-        match lhs with
-        | .poison => Interp.ub
-        | .val v =>
-          if v = BitVec.intMin bw then Interp.ub
-          else return (#[.int bw (LLVM.Int.sdiv lhs rhs properties.exact)], mem, none)
-      else
-        return (#[.int bw (LLVM.Int.sdiv lhs rhs properties.exact)], mem, none)
+    if (LLVM.Int.sdiv lhs rhs).isPoison then Interp.ub
+    else return (#[.int bw (LLVM.Int.sdiv lhs rhs properties.exact)], mem, none)
```
tobiasgrosser:
```suggestion
    let res := LLVM.Int.sdiv lhs rhs properties.exact
    
    if res.isPoison then
      return Interp.ub
    
    return (#[.int bw res], mem, none)
```

Nice. You probably want to factor out the `LLVM.Int.sdiv` call.

luisacicolini:
after chatting with @math-fehr I realized this PR is wrong and I shall close it. 
In particular, we want to return poison (and not UB) in cases where rhs is `val` (different than -1) and lhs is `poison`. With the `isPoison` condition, we would return `UB` in this case, and this is incorrect.

tobiasgrosser:
> after chatting with @math-fehr I realized this PR is wrong and I shall close it. In particular, we want to return poison (and not UB) in cases where rhs is `val` (different than -1) and lhs is `poison`. With the `isPoison` condition, we would return `UB` in this case, and this is incorrect.

Right. I missed this myself. Sorry for the oversight.

PR #730, opened 2026-06-01
feat: add an API that can be queried to see if an operation has side effects

regehr:
Note that there was some polymorphic use of `test.test` which required patching Test/Passes/DCE/basic.mlir regardless of whether it is side-effecting or not. I decided to make it side-effecting.

### DIFF
--- a/Veir/GlobalOpInfo.lean
+++ b/Veir/GlobalOpInfo.lean
@@ -8,6 +8,7 @@ public import Veir.Dialects.ModArith.OpInfo
 public import Veir.Dialects.Cf.OpInfo
 public import Veir.Dialects.Comb.OpInfo
 public import Veir.Dialects.HW.OpInfo
+public import Veir.IR.Basic
 
 namespace Veir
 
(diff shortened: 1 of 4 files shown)


Veir/GlobalOpInfo.lean
```diff
@@ -324,3 +325,32 @@ def OpCode.isTerminator (opCode : OpCode) : Bool :=
...
+    | .builtin .unregistered
+    | .func .func
+    | .llvm .func
+    | .llvm .module_flags
+    | .llvm .store
+    | .riscv .ld
+    | .riscv .sd
+    | .test .test
+    | .hw .module => true
+    | _ => false
```
tobiasgrosser: Should this potentially be in the dialect-specific opinfo files? I am currently on a cab, but look at where getProperties is defined for the various ops.

Veir/GlobalOpInfo.lean
```diff
@@ -324,3 +325,32 @@ def OpCode.isTerminator (opCode : OpCode) : Bool :=
...
+
+  Also see:
+  https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/
+-/
+def OperationPtr.hasSideEffects (op : OperationPtr) (ctx : IRContext OpCode) : Bool :=
+  match op.getOpType! ctx with
+  | .llvm .load =>
+    (op.getProperties! ctx (.llvm .load)).volatile_
+  | opCode =>
+    opCode.isTerminator || match opCode with
```
axelcool1234: Is it typical to have the `match` on the same line like this?

axelcool1234 (commented):
I couldn't really find anything from my limited perspective :)

Veir/GlobalOpInfo.lean
regehr: I agree-- let me see what I can do

math-fehr: I think we should define interfaces properly at some point, not sure if here or later though!
But yeah, passing it to the `OpType` would probably be the way, just make sure that the default is `true` instead of `false`!

Veir/GlobalOpInfo.lean
math-fehr: I think it's what's expected, otherwise you would indent twice?

regehr:
@math-fehr @tobiasgrosser I refactored so that the default for every operation is to be side-effecting, but then we whitelist a collection that are pure, this should be safer.

I looked at factoring this out into per-dialect and there are other properties (isTerminator, for example) that will be involved in that, so I propose we take this patch as-is, and then separately work on a nicer mechanism for factoring dialect-specific information into their own pieces (and presumably this will eventually all be done declaratively)

tobiasgrosser (approved):
This LGTM. I am happy to factor out `hasSideEffect` myself.

Veir/GlobalOpInfo.lean
```diff
@@ -324,3 +325,40 @@ def OpCode.isTerminator (opCode : OpCode) : Bool :=
...
+  | .riscv _ => false
+  -- For LLVM we enumerate the pure ops
+  | .llvm .mlir__constant
+  | .llvm .and | .llvm .or | .llvm .xor
+  | .llvm .add | .llvm .sub | .llvm .mul
+  | .llvm .sdiv | .llvm .udiv | .llvm .srem | .llvm .urem
+  | .llvm .shl | .llvm .lshr | .llvm .ashr
+  | .llvm .icmp | .llvm .select
+  | .llvm .trunc | .llvm .sext | .llvm .zext
+  | .llvm .alloca | .llvm .getelementptr
```
math-fehr: LLVM alloca should have a side-effect!

Veir/GlobalOpInfo.lean
```diff
@@ -324,3 +325,40 @@ def OpCode.isTerminator (opCode : OpCode) : Bool :=
...
+  let opCode := op.getOpType! ctx
+  if opCode.isTerminator then true else
+  match opCode with
+  -- These dialects are pure
+  | .arith _ | .comb _ | .mod_arith _ | .datapath _ => false
+  | .builtin .unrealized_conversion_cast => false
+  | .hw .constant => false
+  -- RISC-V is pure register arithmetic except the memory ops
+  | .riscv .ld | .riscv .sd => true
+  | .riscv _ => false
```
math-fehr, line 350: This is a bit dangerous, because we might forget to add the new operations there. I guess it's fine for now though, since this will be refactored soon enough.

math-fehr (approved):
Nice! Just fix the llvm alloca before it though ;)

Veir/GlobalOpInfo.lean
regehr: are you sure about that? would would be the observable effect of an alloca when nobody observes its return value?

Veir/GlobalOpInfo.lean
regehr: I agree this is shaky, but as you say let's fix it a different way, thanks!
I'd like to start working on little "side effect free" theorems about instructions but I'm not sure how to go about it

PR #731, opened 2026-06-01
fix(llvm): fix notation for isRefinedBy

math-fehr:
The notation was mirrored, so it was technically incorrect.

This direction is consider the correct one because the target is a _refinement_ of the semantics (so less behavior). Also, source stays on the left side, and target on the right side, because you compile the source to the target.


tobiasgrosser (approved):
This is OK by me, but we should run this by @luisacicolini, @bollu, @alexkeizer, and everyone else both to double-check, but also to ensure we are consistent in the future.

Finally, can you add one or two sentences giving an explanation of why this is the right direction, e.g., why the direction of the less-equal connects well with the definition of a refinement.

nchappe (approved):

bollu:
LGTM

ineol (approved):
Thank you @math-fehr!

PR #735, opened 2026-06-02
test(LLVM/fastntt): update `fastNTT` test to use `long` and inline everything

luisacicolini:
We update the fastNTT test to use `long` and always inline, such that eventually we'll be able to apply our riscv instruction selection pass.


tobiasgrosser (approved):
@luisacicolini, this LGTM. However, can you add the command lines that you used to show how this got generated. Also, I wonder if it would make sense to commit the .c code as well? Such that we can reproduce this.

Bonus points if this has been generated by `vcc`. 😄

nchappe (approved):
Thanks! One has to look at the changes in #458 to get the whole picture. I agree with Tobias that some instructions to reproduce it would be nice.

GZGavinZhao (commented):
Nit: recommend changing PR title to `test(LLVM/fastntt)` instead of `chore(test)`. `chore` is used for stuff that's not code changes ([small guide](https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13#types)).

luisacicolini:
thank you all! I added the C code to the comment - I am not sure it makes a lot of sense to have a `.c` file itself in the code, unless we do the lowering every time (?) but let me know what you think and, if you would rather a `.c` file itself, where do you think it should live.

luisacicolini:
On the generation with `vcc` - I don't think I understand what that means @tobiasgrosser?

luisacicolini:
nvm @math-fehr enlightened me, I will fix it with `vcc` and push the c file

luisacicolini:
done - `fastntt.c` now lives in the `Vcc` tests, and we are testing the result in `LLVM/Test`.

Test/Vcc/fastntt.c
```diff
@@ -0,0 +1,208 @@
...
+           r = inverse ? r * 2 : m / 2
+         end
+        where bflyOp is one of:
+        bflyCT(A, B, root, cmod):
+            (A + root * B % cmod, A - root * B % cmod)
+    
+        bflyGS(A, B, root, cmod):
+            (A + B % cmod, (A - B) * root % cmod)
+    Reference: 
+    https://github.com/google/heir/blob/1210ad37dc9531d6e60d8ddbce81dbd79f7626a6/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp#L1060
```
luisacicolini:
```suggestion
    Reference: https://github.com/google/heir/blob/1210ad37dc9531d6e60d8ddbce81dbd79f7626a6/lib/Dialect/Polynomial/Conversions/PolynomialToModArith/PolynomialToModArith.cpp#L1060 
```

tobiasgrosser:
That's fine with me. Let's give it some time tonight and then merge it tomorrow, if no new concerns are raised.

math-fehr:
Can we remove the files in `Output/`? You can probably add `Test/**/Output/**` in the gitignore I guess?

luisacicolini:
@math-fehr I removed the tests for `fastntt` and added to the gitignore, but there are already some files for [demorgan](https://github.com/opencompl/veir/tree/main/Test/Vcc/Output) - should we remove them in another PR?

math-fehr:
I think we should remove them as well! In this PR or in another one, as you wish!
Btw, right now they seem to have been modified in your PR?

luisacicolini:
removed everything, will merge once CI passes - thanks everyone for the comments and reviews!

PR #737, opened 2026-06-03
test: pass unregistered dialect flag to mlir-opt in MLIR_UNREGISTERED_ROUNDTRIP

GZGavinZhao:
Current behavior: `MLIR_UNREGISTERED_ROUNDTRIP` actually doesn't pass `--allow-unregistered-dialect` on the command-line to `mlir-opt`. @math-fehr is this intended?


tobiasgrosser:
AFAIU, this is intentional. The dialects should already be available in `mlir-opt`, but are likely not yet implemented in `veir-opt`.

PR #738, opened 2026-06-03
test(parsing,cuda_tile): correctly test function type parsing with parentheses

### LINKED ISSUE #675: Parser bug: <> not parsable inside attributes
Found in #673. Namely, the following is not parsable:
```
"func.func"() <{function_type = !cuda_tile.ptr<i1> -> ()}> ({
  // ...
)}
```
From due to the `<>` from `!cuda_tile.ptr<i1>`. This is a parser bug that should be fixed.

GZGavinZhao:
Fixes #675.

My claim about the intended behavior in #675 is actually incorrect. Per the MLIR syntax:
```
function-type     ::= (type | type-list-parens) `->` (type | type-list-parens)
```
So it's actually correct that `!cuda_tile.ptr<i1> -> ()` is not parse-able! Instead, it should be written as `(!cuda_tile.ptr<i1>) -> ()`. I fixed this test, and also added two tests that confirm we accept the latter (with parentheses) and reject the former (without parentheses). The positive test runs `MLIR_ROUNDTRIP` in addition to `VEIR_ROUNDTRIP` to asserting that `mlir-opt` has the same behavior.

### DIFF
--- a/Test/CudaTile/types.mlir
+++ b/Test/CudaTile/types.mlir
@@ -1,9 +1,6 @@
 // RUN: VEIR_ROUNDTRIP
 "builtin.module"() ({
-  // TODO(#675): the proper type should be without quotes, i.e.
-  // `!cuda_tile.ptr<i1> -> ()`. Our parser cannot handle this right now, it's a
-  // bug that should be fixed.
-  "func.func"() <{function_type = "!cuda_tile.ptr<i1> -> ()", sym_name = "main"}> ({
+  "func.func"() <{function_type = (!cuda_tile.ptr<i1>) -> (), sym_name = "main"}> ({
     ^bb0(%arg0: !cuda_tile.ptr<i1>):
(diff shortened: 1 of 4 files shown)


tobiasgrosser (commented):
Hi @GZGavinZhao, this looks reasonable. However, I am a bit surprised that `cude_tile`, as a type, is supported by MLIR upstream? I would have expected for it to only work in an NVIDIA-specific setting.

GZGavinZhao:
`cuda_tile` is indeed not supported by MLIR upstream, so the `cuda_tile` test still uses `VEIR_ROUNDTRIP`. For the two additional tests that check behavior equivalence with `mlir-opt`, I used `i32`, so I was able to use `MLIR_ROUNDTRIP`. The goal is to test the parentheses, not the `!cuda_tile` part of the syntax, so the type we use inside the parentheses doesn't really matter.

tobiasgrosser (approved):
Got it. This is great. Thank you for adding the test coverage.

PR #739, opened 2026-06-03
test: add MLIR_INVALID that asserts mlir-opt failure

GZGavinZhao:
This is useful for parser tests that make sure `veir-opt` doesn't accept programs that `mlir-opt` doesn't accept. 

Complements #738 by using `MLIR_INVALID` to verify that `mlir-opt` doesn't accept unparenthesized function types as well.


tobiasgrosser:
This looks reasonable. However, I would prefer to have this stacked the other way around. Or rather, have #740 merged without this change and then commit this as cleanup. This will ensure this PR is merged with test coverage. As it stands, the changes in this PR do lack test coverage in this very PR.

tobiasgrosser (approved):

PR #740, opened 2026-06-03
test(parsing): verify that mlir-opt rejects unparenthesized function types

GZGavinZhao:
Fully resolve the shenanigans surrounding the syntax of function types that grew out of #675. Complements #738. Depends on #739.


tobiasgrosser:
This should probably be merged with the previous PR to ensure that `MLIR_INVALID` is indeed tested.

PR #743, opened 2026-06-03
feat: parse/print support for func.call and llvm.call

regehr:
hopefully just some boring bookkeeping


tobiasgrosser (approved):
LGTM. Let's see what @math-fehr says.

Test/Func/call.mlir
```diff
@@ -0,0 +1,17 @@
+// RUN: VEIR_ROUNDTRIP
+
```
math-fehr:
```suggestion
// RUN: VEIR_ROUNDTRIP
// RUN: MLIR_ROUNDTRIP
```

Test/LLVM/call.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: VEIR_ROUNDTRIP
```
math-fehr, line 1:
```suggestion
// RUN: VEIR_ROUNDTRIP
// RUN: MLIR_ROUNDTRIP
```

math-fehr (approved):
Nice!
Is there a reason that both operations have the same properties? In particular, `func.call` `callee` is never indirect, but still has an option, should we remove the option part?

regehr:
hi @math-fehr I've refactored so that func.func and llvm.func each have their own properties structure. this increases code size a bit, but then func.func does not have to deal with the optional call target. let me know which you prefer! I think this new version is totally fine

regehr:
converting to draft for now until I think about some things

regehr:
ok @math-fehr I had to make one of the tests `MLIR_UNREGISTERED_ROUNDTRIP`, it seems like the changes to support that one truly roundtripping through MLIR are more than I want to put in this PR, but I can work on that next. this is ready for another quick look.

Test/LLVM/call.mlir
```diff
@@ -0,0 +1,22 @@
...
+// RUN: MLIR_UNREGISTERED_ROUNDTRIP
```
math-fehr, line 2: Is something missing to make this an `MLIR_ROUNDTRIP`?

Test/LLVM/call_indirect.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: VEIR_ROUNDTRIP
```
math-fehr, line 1:
```suggestion
// RUN: VEIR_ROUNDTRIP
// RUN: MLIR_ROUNDTRIP
```

math-fehr (approved):
Thanks! I just added two comments, but otherwise if `MLIR_ROUNDTRIP` doesn't work just add `MLIR_UNREGISTERED_ROUNDTRIP` on the other one and feel free to merge it!

nchappe (approved):
Maybe the title could clarify that it's about parsing, otherwise this looks good!

Test/LLVM/call.mlir
regehr: yes, we're missing:
#llvm.cconv<…>
#llvm.linkage<…>
#llvm.tailcallkind<…>

I started to add those and maybe I did a bad job but it seemed to be more code than we wanted in this PR. I could take this on as a follow-on

regehr:
ok thank you @nchappe and @math-fehr !

I added MLIR_UNREGISTERED_ROUNDTRIP as requested. the changes for MLIR_ROUNDTRIP exceed the bounds of this PR, in my view, but I can work on that next.

I also changed the PR title

PR #746, opened 2026-06-04
feat(interpreter): add `VariableState.setVar`

math-fehr:
This is similar to `VariableState.setVar?` but requires a proof that the runtime value conforms to the type.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -158,7 +158,9 @@ def InterpreterState.empty (ctx : WfIRContext OpInfo) : InterpreterState ctx :=
   { variables := .empty ctx, memory := .empty }
 
 /--
-  Set the value of a variable.
+  Set the runtime value of a variable.
+  This function dynamically checks that the runtime value conforms to the variable type, and
+  return `none` otherwise.
 -/
 def VariableState.setVar? (state : VariableState ctx) (var : ValuePtr)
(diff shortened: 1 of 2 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -170,6 +172,18 @@ def VariableState.setVar? (state : VariableState ctx) (var : ValuePtr)
   else
     none
 
+/--
+  Set the runtime value of a variable.
+  This function requires the proof that the runtime value conforms to the variable type.
```
axelcool1234: Being nitpicky here: should probably say "a proof" rather than "the proof"

luisacicolini (approved):

Veir/Interpreter/Basic.lean
math-fehr: Yep, fixed it!

Veir/Interpreter/Basic.lean
```diff
@@ -170,6 +172,18 @@ def VariableState.setVar? (state : VariableState ctx) (var : ValuePtr)
   else
     none
 
+/--
+  Set the runtime value of a variable.
+  This function requires a proof that the runtime value conforms to the variable type.
+-/
+def VariableState.setVar (state : VariableState ctx) (var : ValuePtr)
+    (val : RuntimeValue) (h : val.Conforms (var.getType! ctx.raw))
```
```suggestion
    (val : RuntimeValue) (h : val.Conforms (var.getType! ctx.raw) := by grind)
```
?

ineol (approved):

PR #752, opened 2026-06-04
feat(EDyadic.Round): add directional rounders (RTN, RTP, RTZ)

bollu:
## Summary
Stage 6 of the EDyadic rounding stack. **Stacked on top of #751** (`round-edyadic-position`).

Adds three IEEE-754 rounding modes that round based on direction'.

- **`computeRoundRTP`** (round toward `+∞`): always pick `upper`.
- **`computeRoundRTN`** (round toward `−∞`): always pick `lower`.
- **`computeRoundRTZ`** (round toward zero): pick the approximant whose magnitude is smaller: i.e. `lower` for nonneg `x`, `upper` for neg `x`.

PR #753, opened 2026-06-04
feat(EDyadic.Round): add nearest rounders (RNE, RNA)

bollu:
## Summary
Stage 7 of the EDyadic rounding stack. **Stacked on top of #752** (`round-edyadic-modes-directional`).

Adds the round-to-nearest IEEE-754 modes.

- **`computeRoundRNE`** (round-to-nearest, tie to even): picks `upper` if strictly above the midpoint; at a tie picks whichever of `lower` / `upper` has an even significand; otherwise picks `lower`.
- **`computeRoundRNA`** (round-to-nearest, tie away from zero): picks the closer side, and at a tie picks `upper` for nonneg `x` and `lower` for neg `x`, so away from zero in both cases.

### DIFF
--- a/Veir/Data/FP/EDyadic/Round.lean
+++ b/Veir/Data/FP/EDyadic/Round.lean
@@ -313,7 +313,7 @@ private def computeRoundRNE
   else if computeIsLowerHalf sign mag k prec e then
     computeLower sign mag k prec e s
   else
-    -- Defensive default; unreachable given the partition of cases.
+    -- Unreachable given the partition of cases, default `computeLower`
     computeLower sign mag k prec e s
 
 /-- Round-to-nearest, tie away from zero. -/
--- a/Veir/Data/FP/EDyadic/Round.lean
+++ b/Veir/Data/FP/EDyadic/Round.lean
@@ -293,6 +293,55 @@ private def computeRoundRTZ
   if sign then computeUpper sign mag k prec e s
   else computeLower sign mag k prec e s
 
+/-! ## Nearest rounders: RNE, RNA, and the per-mode dispatcher
+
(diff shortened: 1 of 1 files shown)


Veir/Data/FP/EDyadic/Round.lean
```diff
@@ -293,6 +293,55 @@ private def computeRoundRTZ
...
+  if ! computeIsLowerHalf sign mag k prec e && ! computeIsTieBreak mag k prec e then
+    computeUpper sign mag k prec e s
+  else if computeIsTieBreak mag k prec e && computeIsUpperEven sign mag k prec then
+    computeUpper sign mag k prec e s
+  else if computeIsTieBreak mag k prec e && computeIsLowerEven sign mag k prec then
+    computeLower sign mag k prec e s
+  else if computeIsLowerHalf sign mag k prec e then
+    computeLower sign mag k prec e s
+  else
+    -- Defensive default; unreachable given the partition of cases.
```
luisacicolini: can we have a clearer docstring?

bollu: Not sure what a clearer docstring will be like? Can you suggest one?

bollu: I want to say that "the else branch is unreachable, but lean wants us to have something, so we just do `lower`.

luisacicolini:
```suggestion
    -- Unreachable given the partition of cases, default `computeLower`
```
this sounds clearer to me? not a big change but the `defensive` confused me

luisacicolini (approved):

PR #757, opened 2026-06-05
feat(parser): add locations on error thrown by `MlirParser.lean`

math-fehr:
These are the last errors in `MlirParser.lean` that were missing locations.


tobiasgrosser (approved):
Nice!

luisacicolini (approved):

PR #758, opened 2026-06-05
chore!: rename `!reg` -> `!riscv.reg`

nchappe:
to be consistent with xDSL


tobiasgrosser (approved):
LGTM.

nchappe:
Also added a trivial attribute parsing test for riscv.reg as I just noticed that we did not have one.

PR #759, opened 2026-06-05
test: relax fastNTT test so it passes on MacOS

regehr:
clang for MacOS and clang for Linux inject some different metadata into the LLVM module they emit. this was causing fastNTT to fail on my Mac.

this PR relaxes the checks so that we only check for a common subset. 

another way to fix this would be to insert Linux-specific and OSX-specific CHECK: lines into the test case, but that feels like overkill to me.


regehr:
also should we try to add an OSX CI runner? that would stop this sort of thing from happening

tobiasgrosser (approved):

AlexanderViand:
> also should we try to add an OSX CI runner? that would stop this sort of thing from happening

I opened an issue for this to see what people think: #853

PR #761, opened 2026-06-05
chore(test): add function return types

regehr:
ok this is a huge, boring PR that only touches the Test directory. what it does is get the Veir code base ready for a follow-on PR that checks return operations against the declared type of each function. there should be no functional change here.


math-fehr (approved):
Nice, thanks a lot!

PR #762, opened 2026-06-05
feat(llvm): support cconv and linkage attributes

regehr:
this is supposed to be a boring PR that simply supports a couple of LLVM attributes. this is on the path to roundtripping LLVM-dialect MLIR from `clang` without requiring our `--allow-unregistered-dialect` flag


math-fehr (approved):
Nice, thanks!

PR #763, opened 2026-06-06
feat(interpreter): add sorried lemma about interpretOp monotonicity

math-fehr:
This is a lemma that we need to prove that a local rewrite correctness extends to the entire module. For now, this lemma is sorried.


regehr (approved):
ok... I think r.1 is the results of the operation and r.2 is the memory and control flow state.. so one of these has to be refined and the other must be strictly equal. 

so this is basically a result about the compositionality of refinement?

I think this all makes sense but please document whatever is important about this for your readers!

tobiasgrosser (approved):

PR #764, opened 2026-06-06
feat(ir): Add lemmas about `getResults!` and `idxOf`

math-fehr:
This is a simple useful lemma that's used when you want to check if a value is a result of an operation in proofs. You can write `v \in op.getResults! ctx`, then write `v.idxOf (op.getResults! ctx)` to get its index, instead of having to match on `v`.


regehr (approved):
LGTM but maybe add a comment saying in what circumstances this should be useful

tobiasgrosser (approved):

ineol (approved):

PR #769, opened 2026-06-06
feat: typechecking of return operations against functions' declared types

regehr:
This closes the loop with my big PR yesterday that added return types to lots of functions in our `Test` directory, by actually checking that these types agree with return statements inside functions.

The way I did this here is very simple: when we reach a return operation, we walk up the tree to find the return type of the enclosing function. It would likely be more efficient to remember the function return type when we first encounter it and then thread that through the verifier, but when I started to do this it felt quite invasive, whereas the current approach here is not invasive.

### DIFF
--- a/Test/Verifier/func_return_type_mismatch.mlir
+++ b/Test/Verifier/func_return_type_mismatch.mlir
@@ -7,4 +7,4 @@
   }) : () -> ()
 }) : () -> ()
 
-// CHECK: Return operand 0 type does not match the function's declared result type
+// CHECK: func.return operand 0 type does not match the function's declared result type
--- a/Test/Verifier/func_return_type_mismatch.mlir
+++ b/Test/Verifier/func_return_type_mismatch.mlir
@@ -1,0 +1,10 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{function_type = (i32) -> i32, sym_name = "f"}> ({
+  ^bb0(%arg0: i1):
+    "func.return"(%arg0) : (i1) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Return operand 0 type does not match the function's declared result type
--- a/Veir/Verifier.lean
+++ b/Veir/Verifier.lean
@@ -10,51 +10,67 @@ import Veir.ForLean
 namespace Veir
 
 /--
-  Find the declared result types of the function that encloses `op`,
-  or fail if the function signature cannot be determined.
+  Walk up from `op` (a return-like terminator named `opName`) to the
+  operation that encloses its parent region, i.e. the enclosing function
+  operation.
 -/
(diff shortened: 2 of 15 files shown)


tobiasgrosser (approved):
That looks reasonable to me. 🙏 Let's see what @math-fehr says.

Test/Verifier/func_return_type_mismatch.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{function_type = (i32) -> i32, sym_name = "f"}> ({
+  ^bb0(%arg0: i1):
+    "func.return"(%arg0) : (i1) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Return operand 0 type does not match the function's declared result type
```
math-fehr:
```suggestion
// CHECK: `func.return` operand 0 type does not match the function's declared result type
```
To be more explicit?
Btw, I'm taking note on that and will work on adding locations to these errors whenever I have free time

Veir/Verifier.lean
```diff
@@ -9,6 +9,53 @@ import Veir.ForLean
 
 namespace Veir
 
+/--
+  Find the declared result types of the function that encloses `op`,
+  or fail if the function signature cannot be determined.
+-/
+def OperationPtr.getEnclosingFunctionOutputs (op : OperationPtr) (ctx : WfIRContext OpCode) :
```
math-fehr: I feel that maybe we should have two different functions here?
One for `func.func` and one for `llvm.func`? Just so we check that a `llvm.return` is in a `llvm.func`, and a `func.return` is in a `func.func`?

Veir/Verifier.lean
```diff
@@ -9,6 +9,53 @@ import Veir.ForLean
...
+-/
+def OperationPtr.verifyReturnTypes (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) : Except String PUnit := do
+  let outputs ← op.getEnclosingFunctionOutputs ctx
+  if op.getNumOperands ctx.raw opIn ≠ outputs.size then
+    throw s!"Expected {outputs.size} return operand(s)"
+  let opTypes := op.getOperandTypes! ctx.raw
+  for i in [0:outputs.size] do
+    if (opTypes[i]!).val ≠ outputs[i]! then
+      throw s!"Return operand {i} type does not match the function's declared result type"
```
math-fehr: Here, we should probably return the operation name instead of `Return`?

math-fehr (commented):
Nice! I just have a comment on the error messages, and maybe that we should check that `llvm.return` and `func.return` are only in a `llvm.func` and an `func.func`?

regehr:
@math-fehr ok I think I've addressed all of your feedback!
the stricter checking (not allowing a func.func to have an llvm.return and vice-versa) forced some tests to be trivially updated.

math-fehr (approved):
all good, thanks!

PR #770, opened 2026-06-06
fix(llvm): flip order of operands to llvm.store

regehr:
it should be `llvm.store %val, %ptr`, not `llvm.store %ptr, %val`
as documented here:
https://mlir.llvm.org/docs/Dialects/LLVM/#llvmstore-llvmstoreop

### DIFF
--- a/Test/Interpreter/LLVM/getelementptr.mlir
+++ b/Test/Interpreter/LLVM/getelementptr.mlir
@@ -8,7 +8,7 @@
       %off1 = "llvm.mlir.constant"() <{ "value" = 1 : i64 }> : () -> i64
       %ptr1 = "llvm.getelementptr"(%array, %off1) <{ elem_type = i64, rawConstantIndices = array<i32: -2147483648>}> : (!llvm.ptr, i64) -> !llvm.ptr
       %val1 = "llvm.mlir.constant"() <{ "value" = 256 : i64 }> : () -> i64
-      "llvm.store"(%ptr1, %val1) : (!llvm.ptr, i64) -> ()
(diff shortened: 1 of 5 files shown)


tobiasgrosser (approved):

tobiasgrosser:
I am surprised there is not test using `LLVM_ROUNDTRIP`. At best we add one and also add an `MLIR_ROUNDTRIP` line.

regehr:
> I am surprised there is not test using `LLVM_ROUNDTRIP`. At best we add one and also add an `MLIR_ROUNDTRIP` line.

let me add that.

I'm confused how things like my sudoku solver work, given that it performs plenty of `llvm.store` operations.

my thought was simply to fix this, but if you'd like me to look into this more deeply, I can do that

regehr:
ok this is added. it does not look like any existing MLIR_ROUNDTRIP test contained a store.

I looked at the sudoku solver and it contains 8 stores. they simply pass through Veir, completely incorrectly! if we had tried to interpret it, it would have crashed badly, but we can't since it contains a lot of function calls. if we had any optimization passes that did non-trivial changes to store operations then that would have stopped it from passing through in a way that would still work.

regehr:
also, the fact that we were able to import MLIR from clang and it didn't fail a typecheck indicates to me that we're seriously missing some typechecks in the verifier

tobiasgrosser:
> also, the fact that we were able to import MLIR from clang and it didn't fail a typecheck indicates to me that we're seriously missing some typechecks in the verifier

We do. You joined at a time where we barely managed to parse anything. We should add more typechecks.

PR #771, opened 2026-06-06
feat(LLVM): add refinement monotonicity theorems

tobiasgrosser:
These will help to prove `LLVM.interpretOp'`

### DIFF
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -399,101 +399,97 @@ theorem getValue_select {w : Nat} (x y : Int w) (c : Int 1) (h : (select c x y).
   simp [select, Id.run]
   grind
 
-theorem isRefineBy_add {w : Nat} (x₁ x₂ y₁ y₂ : Int w)
+theorem add_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w)
     (h₁ : x₁ ⊒ y₁) (h₂ : x₂ ⊒ y₂) (nsw nuw : Bool) :
     add x₁ x₂ nsw nuw ⊒ add y₁ y₂ nsw nuw := by
   grind
 
-theorem isRefineBy_sub {w : Nat} (x₁ x₂ y₁ y₂ : Int w)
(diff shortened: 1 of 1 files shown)


math-fehr (approved):
Nice!
@luisacicolini or @ineol probably know what names we should use there? For instance, we do not mention `isRefinedBy` anywhere, and maybe we should?

luisacicolini:
maybe `isRefinedBy_add`? This would be more consistent with the usual conventions imho, although I don't have a strong opinion here and overall LGTM

luisacicolini (approved):

ineol:
The [naming convention document](https://github.com/leanprover/lean4/blob/master/doc/std/naming.md) says: mono 	Theorems of the form a R b → f a R f b, where R is a transitive relation 	List.countP_mono_left.

the relation is implicit here so I think `op_mono` is fine?

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -398,3 +398,98 @@ theorem getValue_select {w : Nat} (x y : Int w) (c : Int 1) (h : (select c x y).
     (select c x y).getValue h = if _ : c.getValue = 1#1 then x.getValue else y.getValue := by
   simp [select, Id.run]
   grind
+
+theorem add_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w)
+    (h₁ : x₁ ⊒ y₁) (h₂ : x₂ ⊒ y₂) (nsw nuw : Bool) :
+    add x₁ x₂ nsw nuw ⊒ add y₁ y₂ nsw nuw := by
```
ineol, line 404: maybe leave the parameters of the operations implicit?
```suggestion
theorem add_mono {w : Nat} {x₁ x₂ y₁ y₂ : Int w} {nsw nuw : Bool}
    (h₁ : x₁ ⊒ y₁) (h₂ : x₂ ⊒ y₂) :
    add x₁ x₂ nsw nuw ⊒ add y₁ y₂ nsw nuw := by
```

ineol (approved):

PR #773, opened 2026-06-07
feat(llvm): add some more attributes

regehr:
ok, this adds a few more LLVM attributes, and finally we can start to disallow unregistered stuff in some of the roundtrip testing.

the tradeoff here is that to keep things simple, I've represented the attribute payloads as strings. we can either merge this PR as-is, and then go parse these payloads later on when we actually need to do that. or else, I could make those changes now. but my view is that it's fine to leave them as strings for now.


tobiasgrosser (approved):
LGTM

PR #775, opened 2026-06-07
feat(riscv): remove the requirement that riscv_cf operations have the next block as a fallthrough

regehr:
the plan is to handle this later in the pipeline, after isel

### DIFF
--- a/Veir/Verifier.lean
+++ b/Veir/Verifier.lean
@@ -1784,9 +1784,6 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
       throw "Expected 0 regions"
     if op.getNumSuccessors ctx.raw opIn ≠ 1 then
       throw "Expected 1 successor"
-    if let some block := (op.get ctx.raw opIn).parent then
-      if (block.get! ctx.raw).next ≠ some (op.getSuccessor! ctx.raw 0) then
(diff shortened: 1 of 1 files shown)


math-fehr:
I think this maybe need a bigger discussion, or at least I'm not aware that we reached a conclusion on this? My understanding is that we should keep this restriction, but we should instead add an intermediary dialect?

tobiasgrosser (approved):

tobiasgrosser:
We should check with @nchappe.

tobiasgrosser:
I am not sure we need an intermediate dialect. We may need to have a pass that creates this invariant at some point but I am not sure we need it in the verifier. Maybe an attribute that states that this property is preserved/expected?

regehr:
algorithmically, I think the transformation is relatively straightforward, but maybe not as easy as I was originally thinking. for example, it is common to see multiple branches to the same target. (this even happens with the silly little prime number example in a different PR of mine today!) some code duplication is required to enforce the invariant.

maybe it would make sense to have isel perform the BB reordering after it has finished its work? but this might be putting too much work into isel?

anyhow-- in the short term, before I can get any real work done here, we either need the BB ordering requirement to go away, or else I need to be allowed to do the necessary reordering myself. otherwise, no real code will go through `isel-riscv`.

nchappe (approved):

PR #776, opened 2026-06-07
feat(riscv): add the other four riscv conditional branch instructions

regehr:


math-fehr (commented):
This looks good to me, besides the fact that this doesn't check that the next block is the passthrough block.
I think we should discuss this because this will change quite a bit how we will design our backend and our proofs

regehr:
> This looks good to me, besides the fact that this doesn't check that the next block is the passthrough block. I think we should discuss this because this will change quite a bit how we will design our backend and our proofs

my view is that enforcing that requirement at isel time is too early: at this point we're likely to be introducing/eliminating basic blocks. for example, if we want to perform isel for an `llvm.select` for a riscv target that lacks `cmov`, then we're likely to do this using branching. for example, LLVM currently does this:
https://gcc.godbolt.org/z/vx3Wd8Gef

and if we're introducing new blocks during isel, then maintaining this invariant is going to be a pain!

seems much better to not have this requirement at this point, and then after isel (and maybe after regalloc), do a BB rectification pass that puts things into a correct, final order

tobiasgrosser (approved):
Nice.  I like this one.

luisacicolini (approved):
lgtm, thank you!!

nchappe (approved):

regehr:
ok looks I have enough approvals to merge this, thanks!

I have added the BB successor checks to the new instructions, so that everything stays consistent.

PR #777, opened 2026-06-07
fix(interpreter): skip over LLVM module flags when looking for an entry point

regehr:
The module flags emitted by LLVM were breaking our entry point scan, they need to be skipped over.

Also, add a test case ensuring that we can directly interpret MLIR that arrives through the clang->mlir-translate path.

### DIFF
--- a/Test/Vcc/primes.c
+++ b/Test/Vcc/primes.c
@@ -24,6 +24,7 @@ int main(void) {
         n++;
     }
 
+    // main's return value is 29, the 10th prime number
     return last_prime;
 }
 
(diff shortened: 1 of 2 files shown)


math-fehr (approved):
Nice! I am wondering if we should just instead remove the check that toplevel operations are in the module? That would make things much easier to deal with

regehr:
> Nice! I am wondering if we should just instead remove the check that toplevel operations are in the module? That would make things much easier to deal with

I don't have a preference! let me know which you prefer

tobiasgrosser (commented):
Nice. I like this end-to-end testing. 🙏

regehr:
I'm going to just merge this-- but I'm totally fine with any revisions of the code @math-fehr

PR #778, opened 2026-06-07
feat(llvm): support llvm tail call kind attribute

regehr:
add one more attribute that is blocking us from changing `MLIR_UNREGISTERED_ROUNDTRIP` to `MLIR_ROUNDTRIP`

the next blocker is `!llvm.struct<(i32, f32)>` which probably requires some real thought and planning to support! so I'm not touching that for now.


math-fehr (approved):
Nice!

PR #784, opened 2026-06-08
feat(refinement): Use ControlFlow refinement instead of equality

math-fehr:
The previous use of equality in `interpretOp'_monotone` was making this lemma False for terminators.


nchappe (approved):

regehr (approved):
this is cool!
I am enjoying watching how the lower-level refinement results are starting to compose into higher-level structure

ineol (approved):

PR #786, opened 2026-06-08
feat(RISCV): proof strategy and tactic for isel patterns

luisacicolini:
We introduce a new `refine_bv_decide` tactic to automate the verification of instruction selection patterns' refinement relation. To do this, we needed some additional lemmas in our simp-sets, in particular regarding the unfolding of the casting ops (`toInt`, `toReg`). 

We show how the tactic works in golfing the current instruction selection proofs. 

As a next step, I shall use the same strategy to prove the refinement relation for all the instruction selection patterns.

### DIFF
--- a/Veir/Data/Casting.lean
+++ b/Veir/Data/Casting.lean
@@ -17,7 +17,6 @@ public section
 /--
   Cast `LLVM.Int` to `RISCV.Reg`.
 -/
-@[grind]
 def LLVM.Int.toReg (i : Veir.Data.LLVM.Int w) : Veir.Data.RISCV.Reg :=
   match i with
   | .poison => .mk 0#64
--- a/Veir/Data/Casting.lean
+++ b/Veir/Data/Casting.lean
@@ -30,12 +30,12 @@ def LLVM.Int.toReg (i : Veir.Data.LLVM.Int w) : Veir.Data.RISCV.Reg :=
 def RISCV.Reg.toInt (r : Veir.Data.RISCV.Reg) (w : Nat) : Veir.Data.LLVM.Int w :=
   Veir.Data.LLVM.Int.val (BitVec.zeroExtend w r.val)
 
-@[llvm_toBitVec, grind =]
+@[reg_toBitVec, grind =]
 theorem toInt_eq_val {r : Veir.Data.RISCV.Reg} :
     (RISCV.Reg.toInt r 64).getValue = r.val := by
   simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.getValue]
 
-@[llvm_toBitVec, grind =]
+@[reg_toBitVec, grind =]
 theorem toInt_isPoison {r : Veir.Data.RISCV.Reg} :
     (RISCV.Reg.toInt r w).isPoison = false := by
   simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.isPoison]
--- a/Veir/Data/Casting.lean
+++ b/Veir/Data/Casting.lean
@@ -35,12 +35,12 @@ theorem toInt_eq_val {r : Veir.Data.RISCV.Reg} :
   simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.getValue]
 
 @[reg_toBitVec, grind =]
-theorem toInt_isPoison {r : Veir.Data.RISCV.Reg} :
+theorem toInt_isPoison {w : Nat} {r : Veir.Data.RISCV.Reg} :
     (RISCV.Reg.toInt r w).isPoison = false := by
   simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.isPoison]
 
 @[llvm_toBitVec, grind =]
-theorem val_toReg {i : Veir.Data.LLVM.Int w} :
+theorem val_toReg {w : Nat} {i : Veir.Data.LLVM.Int w} :
     (LLVM.Int.toReg i).val = if h : i.isPoison = true then 0#64 else i.getValue.zeroExtend 64 := by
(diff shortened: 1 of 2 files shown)


Veir/Data/Casting.lean
```diff
@@ -22,5 +26,24 @@ def LLVM.Int.toReg (i : Veir.Data.LLVM.Int w) : Veir.Data.RISCV.Reg :=
...
+    (RISCV.Reg.toInt r 64).getValue = r.val := by
+  simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.getValue]
+
+@[reg_toBitVec, grind =]
+theorem toInt_isPoison {r : Veir.Data.RISCV.Reg} :
+    (RISCV.Reg.toInt r w).isPoison = false := by
+  simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.isPoison]
+
+@[llvm_toBitVec, grind =]
+theorem val_toReg{i : Veir.Data.LLVM.Int w} :
```
math-fehr:
```suggestion
theorem val_toReg {i : Veir.Data.LLVM.Int w} :
```

math-fehr (approved):
Perfect, thanks

Veir/Data/Casting.lean
luisacicolini: addressed

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -11,30 +15,29 @@ import Std.Tactic.BVDecide
 
 namespace Veir.Data.RISCV
 
+/--
+  A tactic to unfold the semantics of operations on `LLVM.Int` and on `RISCV.Reg` to
+  bitvectors, such that `bv_decide` can prove the correctness of the refinement relation
+  over instruction selection patterns.
+-/
+macro "refine_bv_decide" : tactic =>
+  `(tactic| ((try simp only [llvm_toBitVec, reg_toBitVec]; try simp [LLVM.Int.getValue_eq_getValueD]; try bv_decide)))
```
ineol, line 24: you can also use `simp -failIfUnchanged` but I don't really know why there is an option for it specifically.

ineol (approved):

Veir/Data/Casting.lean
```diff
@@ -22,5 +25,24 @@ def LLVM.Int.toReg (i : Veir.Data.LLVM.Int w) : Veir.Data.RISCV.Reg :=
...
   Veir.Data.LLVM.Int.val (BitVec.zeroExtend w r.val)
+
+@[reg_toBitVec, grind =]
+theorem toInt_eq_val {r : Veir.Data.RISCV.Reg} :
+    (RISCV.Reg.toInt r 64).getValue = r.val := by
+  simp [RISCV.Reg.toInt, Veir.Data.LLVM.Int.getValue]
+
+@[reg_toBitVec, grind =]
+theorem toInt_isPoison {r : Veir.Data.RISCV.Reg} :
+    (RISCV.Reg.toInt r w).isPoison = false := by
```
tobiasgrosser:
```suggestion
theorem toInt_isPoison {w : Nat} {r : Veir.Data.RISCV.Reg} :
    (RISCV.Reg.toInt r w).isPoison = false := by
```

Veir/Data/Casting.lean
```diff
@@ -22,5 +25,24 @@ def LLVM.Int.toReg (i : Veir.Data.LLVM.Int w) : Veir.Data.RISCV.Reg :=
 /--
   Cast `RISCV.Reg` to `LLVM.Int`.
 -/
+@[grind]
 def RISCV.Reg.toInt (r : Veir.Data.RISCV.Reg) (w : Nat) : Veir.Data.LLVM.Int w :=
   Veir.Data.LLVM.Int.val (BitVec.zeroExtend w r.val)
+
+@[reg_toBitVec, grind =]
+theorem toInt_eq_val {r : Veir.Data.RISCV.Reg} :
+    (RISCV.Reg.toInt r 64).getValue = r.val := by
```
tobiasgrosser, line 34:
```suggestion
theorem toInt_eq_val {w : Nat} {r : Veir.Data.RISCV.Reg} :
    (RISCV.Reg.toInt r w).getValue = r.val := by
```

Should this be bitwidth-generic?

tobiasgrosser (approved):
Beautiful, thank you, @luisacicolini!

Veir/Data/Casting.lean
luisacicolini: no, because `r.val` is [`BitVec 64`](https://github.com/opencompl/veir/blob/5f58a633c746f5af6c549e17e698f70ce17611b5/Veir/Data/RISCV/Reg/Basic.lean#L14), so it would not type check

Veir/Data/Casting.lean
luisacicolini: addressed

Veir/Passes/InstructionSelection/Proofs.lean
luisacicolini: do you think it would be better? otherwise, for [consistency](https://github.com/opencompl/veir/blob/main/UnitTest/Bitblasting/Bitblasting.lean) with the other tactic I would avoid the flag

Veir/Data/Casting.lean
tobiasgrosser: OK.

Veir/Passes/InstructionSelection/Proofs.lean
tobiasgrosser: I think this is fine. If we want to go to `failIfUnchanged` we probably want a zulip discussion education us why it is useful?

PR #798, opened 2026-06-08
fix(refinements): allow `Interp` to refine `none` to everything

math-fehr:
Currently, if the interpretation fails in either the source or the target program, there is no refinement relation between them.

With this PR, a failure in the source program interpretation implies that any target program refines the source.


ineol (approved):

tobiasgrosser:
Post-merge review, this is exactly what we want. It helps #774.

PR #800, opened 2026-06-08
test(riscv): two actual programs, as test cases

regehr:
this adds two versions of a silly program that returns the value of the 10th prime number

one of them is literally translated from `clang -O3` output, and this test passes

the other is literally translated from `clang -O0` output, and it does not even roundtrip without allowing unregistered operations.

@nchappe if you feel like adding `riscv.lw` and `riscv.sw`, that would fix this. or perhaps I can do it if it's easy? let me know.

once VeIR supports all of the instructions used in this test, it will still be interesting getting it to interpret, since it creates a stack frame. we'll at least need to ensure that `SP` is initialized, I'm not sure what else.

### DIFF
--- a/Test/Interpreter/RISCV_programs/primes_O0.mlir
+++ b/Test/Interpreter/RISCV_programs/primes_O0.mlir
@@ -7,17 +7,14 @@
 //
 // RUN: VEIR_UNREGISTERED_ROUNDTRIP
 
-// CHECK: "func.func"() <{"function_type" = (!riscv.reg, !riscv.reg, !riscv.reg) -> !riscv.reg, "sym_name" = "main"}>
+// CHECK: "func.func"() <{"function_type" = (!riscv.reg) -> !riscv.reg, "sym_name" = "main"}>
 
 "builtin.module"() ({
-  // main(sp, ra, s0_caller) -- the architectural registers read on entry.
-  "func.func"() <{sym_name = "main", function_type = (!riscv.reg, !riscv.reg, !riscv.reg) -> !riscv.reg}> ({
+  // main(s0) -- frame base pointer for the stack spill slots (the prologue/epilogue
+  // that would set this up and save/restore sp/ra are elided at this abstraction level).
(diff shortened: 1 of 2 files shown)


regehr:
actually, I've removed the prologue and epilogue-- these don't belong at this abstraction level anyway.
even so, the stack accesses are register-indirect, so we'll have to set that up to get it to interpret.

tobiasgrosser (approved):
Very cool. Let's see what @nchappe and others say. In particular, is `riscv.alloc` abstraction may become relevant.

nchappe:
As one of the tests explicitly manipulates SP, it should probably use named registers, e.g. `!riscv.reg<sp>` (which we cannot parse or interpret for now).
An alternative would be to use `riscv_stack.alloca` instead, as mentioned by Tobias. Interpreter support for it will hopefully be merged in the coming days.

As for lw and sw, it's kind of the same as ld/sd, it wouldn't be hard to support them, but it may be easier to just change the source to use 64-bit values instead, as we did for fastntt.

regehr:
> As one of the tests explicitly manipulates SP, it should probably use named registers, e.g. `!riscv.reg<sp>` (which we cannot parse or interpret for now). An alternative would be to use `riscv_stack.alloca` instead, as mentioned by Tobias. Interpreter support for it will hopefully be merged in the coming days.
> 
> As for lw and sw, it's kind of the same as ld/sd, it wouldn't be hard to support them, but it may be easier to just change the source to use 64-bit values instead, as we did for fastntt.

nice! 
I propose this-- we merge this PR as-is, with one of the tests being non-executable, just a ROUNDTRIP test.
then, as the support you mention lands in subsequent days, I'll update the test that uses memory to use your code.

regehr:
@nchappe if you don't mind, I'll add the missing smaller loads and stores. as you say, this'll be easy given the support you already wrote! I was going to do that tonight but I got distracted by the missing support for immediate offsets.

nchappe (approved):
This works for me!

PR #801, opened 2026-06-08
feat(verifier): strengthen checks

regehr:
this PR brings a bunch of operation-level typechecks to our verifier. so for example it becomes impossible for a `select` to return a not-i1 or an add to take an `i8` and an `i9`.

the error messages are pretty good-- they mention the instruction. but we badly need line number reporting! 

in a follow-on PR I'll try to write a bunch of theorems about these operations that we can use later in the interpreter to try to prove that it does not fail to interpret a verified module!


tobiasgrosser (approved):
I am happy with this.

math-fehr (approved):
Very cool!

PR #805, opened 2026-06-09
fix(isel): swap args' order in isel patterns

luisacicolini:
because in the [riscv semantics](https://github.com/opencompl/veir/blob/main/Veir/Data/RISCV/Reg/Basic.lean) the order of operands is reversed (for consistency with their location in the actual instruction address), we need to keep this ordering in mind when writing instruction selection patterns.


regehr (approved):
LGTM!

PR #806, opened 2026-06-09
fix(riscv): args' order in `mul` semantics

luisacicolini:
fix the ordering of the operands in the semantics of the `mul` operation, for consistency with all other operations this is swapped.


regehr (approved):
LGTM!

PR #809, opened 2026-06-09
feat: Parse and verify RISCV stack operations

nchappe:
This departs a bit from https://github.com/xdslproject/xdsl/pull/6086 as alloca expects a value_type attribute to determine the size of the allocation, and it returns a RISCV register.

### DIFF
--- a/Test/RISCV_Stack/identity.mlir
+++ b/Test/RISCV_Stack/identity.mlir
@@ -1,0 +1,18 @@
+// RUN: VEIR_ROUNDTRIP
+
+"builtin.module"() ({
+  "func.func"() <{function_type = () -> (), sym_name = "main"}> ({
+    ^bb0():
+      %1 = "riscv_stack.alloca"() <{"value_type" = i64, "alignment" = 8 : i64}> : () -> !riscv.reg
+      "func.return"() : () -> ()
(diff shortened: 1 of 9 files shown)


Test/RISCV_Stack/identity.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: VEIR_UNREGISTERED_ROUNDTRIP
```
regehr: since you've added these instructions, shouldn't this be `VEIR_ROUNDTRIP`?

regehr:
I'm onboard with `alloca` here, but I don't understand why we would want stack-specific loads and stores. In general, a pointer can refer either to the stack, or the heap, or globals -- so it's not going to be easy to actually use these in practice, except in restricted situations. I would expect that the RISC-V memory model is a flat 64-bit address space and then normal loads and stores can go anywhere inside it.

math-fehr (approved):

math-fehr:
But let's wait for @regehr to approve it once the comments are addressed

nchappe:
Updated as discussed, and the stack slot type becomes unnecessary.

regehr (approved):
LGTM!

PR #810, opened 2026-06-09
feat(risvc): add riscv Zbb instructions orc.b and rev8

regehr:


luisacicolini (approved):
Looks great, thanks a lot!

PR #811, opened 2026-06-09
feat(ISel): prove correctness of selection patterns

luisacicolini:
Prove the correctness of all basic instruction selection patterns. Introduce `toInt_getValue` and update the `simp` set in the `refine_bv_decide` tactic to avoid a conversion `toNat` that was problematic (``BitVec.extractLsb_toNat`).
We do not prove the lemmas around sext/zext/trunc as we need to be careful with the widths we support for those ops.

In this PR, I am also fixing a mismatch in the `Reg` lemmas that was a consequence of #806.

### DIFF
--- a/Veir/Passes/InstructionSelection/Proofs.lean
+++ b/Veir/Passes/InstructionSelection/Proofs.lean
@@ -15,23 +15,6 @@ import Std.Tactic.BVDecide
 
 namespace Veir.Data.RISCV
 
-@[simp]
-theorem BitVec.ult_one_iff_of_lt {lhs : BitVec w} (h : 0 < w) :
-    (lhs.ult 1#w) = decide (lhs = 0#w) := by
-  simp only [BitVec.ult_eq_decide, BitVec.toNat_ofNat, decide_eq_decide]
-  rw [Nat.mod_eq_of_lt (by grind)]
-  simp [← BitVec.toNat_inj]
-
-@[simp]
-theorem BitVec.ult_zero_false {lhs : BitVec w} :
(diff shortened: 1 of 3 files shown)


tobiasgrosser (approved):
Very cool. Thank you, @luisacicolini. 🙏

math-fehr (approved):
Thanks!

regehr:
> oh wow this is awesome!
> does this also work for GEP?

It should. However, for this we may need an `LLVM.Int.getelementptr` function, which does not use `toNat`. 

I merged this PR in case you want to experiment with `getelementptr`.

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -39,5 +39,155 @@ theorem constant_refinement {v : Int}:
...
 theorem add_refinement {x y : LLVM.Int 64} :
     (Data.LLVM.Int.add x y) ⊒
-      (RISCV.Reg.toInt (Data.RISCV.add (LLVM.Int.toReg x) (LLVM.Int.toReg y)) 64) := by
+      (RISCV.Reg.toInt (Data.RISCV.add (LLVM.Int.toReg y) (LLVM.Int.toReg x)) 64) := by
+  refine_bv_decide
+
+/--
+  Prove the correctness of the `and` lowering pattern.
+-/
+theorem and_refinement{x y : LLVM.Int 64} :
```
tobiasgrosser, line 48:
```suggestion
theorem and_refinement {x y : LLVM.Int 64} :
```

We missed a whitespace here.

PR #813, opened 2026-06-09
feat(ir): add `BlockPtr.operationList` get-set lemmas for `Rewriter`

math-fehr:
This PR adds new get-set lemmas about `BlockPtr.operationList`. These lemmas are very useful to reason at a high-level about operations. For instance, in the rewriter, this should allow us to prove that a rewrite pattern transform a block `beforeOps ++ #[op] ++ afterOps` into `beforeOps ++ newOps ++ afterOps`. This will allow us to reason at the level of operation arrays, rather than doubly linked lists.

While this is a large PR, these lemmas have almost the same structure, starting with an `BlockPtr.OpChain` proof, then a `BlockPtr.operationList` proof.

### DIFF
--- a/Veir/IR/InBounds.lean
+++ b/Veir/IR/InBounds.lean
@@ -249,6 +249,10 @@ theorem OpOperandPtr.dealloc.inBounds_dealloc_genericPtr {ptr : GenericPtr} :
     ptr.InBounds (OperationPtr.dealloc op' ctx inBounds) := by
   grind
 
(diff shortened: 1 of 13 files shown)


tobiasgrosser (approved):

regehr (approved):
this is very cool!

luisacicolini (approved):
I only checked for style and proof conventions, but LGTM!

PR #814, opened 2026-06-09
feat(ir): add two lemmas about operation results

math-fehr:
One is that all results of an operation are in bounds, and the second one is a useful lemma to do a case analysis on.


tobiasgrosser (approved):

regehr (approved):
LGTM!

luisacicolini (approved):

PR #815, opened 2026-06-10
feat(verifier): add more checks to the verifier

regehr:
I understand that `Test/Verifier/llvm_constant_float_integer_width_mismatch.mlir` looks a little weird, the documented MLIR rules for this stuff are also a little weird. the behavior is the same as that of the upstream MLIR verifier. I think this stuff is correct. it's at least better than before :)


tobiasgrosser (approved):

math-fehr (approved):
Nice! (I really don't know why they have this weird verifier for llvm constants)

PR #818, opened 2026-06-10
feat(isel): convert llvm branches to riscv branches

tobiasgrosser:


regehr:
nice!

if I understand correctly, an llvm.icmp + llvm.cond_br is going to turn into two separate comparisons: one from the icmp, the second the bne that you're creating here. will we try to absorb the extra comparison using a RISC-V combine later on?
or will a subsequent version of this code try to absorb the llvm.icmp into the code that it generates in the first place?

regehr:
I looked through the current output of this pass on fastntt and it looks really good!

Veir/Passes/InstructionSelection/RISCV64Branches.lean
```diff
@@ -0,0 +1,105 @@
...
+namespace Veir
+
+set_option warn.sorry false in
+def convertBranch (ctx : WfIRContext OpCode) (op : OperationPtr)
+    (block : BlockPtr) : ExceptT String IO (WfIRContext OpCode) := do
+  let mut c := ctx
+
+   -- Check if the terminator operations can be converted to RISCV branches. If
+   -- not, we exit early and do not convert this predecessor block.
+   if op.getOpType! c.raw != .llvm .br &&
```
math-fehr: These two lines seems they are not indented by the same amount of characters?

Veir/Passes/InstructionSelection/RISCV64Branches.lean
```diff
@@ -0,0 +1,105 @@
...
+      op.getOpType! c.raw != .llvm .cond_br then do
+     return c
+
+   let mut some ip := InsertPoint.after? op c.raw | return c
+   let mut casts : Array (OperationPtr) := #[]
+
+   for i in List.reverse (List.range (op.getNumOperands! c.raw)) do
```
math-fehr: Instead of doing the `reverse` and `InsertPoint.after?`, you can use `InsertPoint.before` and it will insert things in the same order, so you don't need the `reverse`.

Veir/Passes/InstructionSelection/RISCV64Branches.lean
```diff
@@ -0,0 +1,105 @@
...
+   else if op.getOpType! c = OpCode.llvm .cond_br then do
+     let condProps : CondBrProperties := op.getProperties! c
+       (OpCode.llvm .cond_br)
+     let props : RISCVBrProperties := ⟨condProps.operandSegmentSizes⟩
+
+     WfRewriter.createOp c (.riscv_cf .bnez) #[]
+       (casts.map (fun cast => cast.getResult 0))
+       (op.getSuccessors! c.raw) #[] props ip sorry sorry sorry sorry
+   else
+     none | return c
```
math-fehr: Would it be nicer to maybe do the `let ... :=` inside the `if/else`?
I feel it's a bit weird having `none | return c` at the `else` part here

Veir/Passes/InstructionSelection/RISCV64Branches.lean
```diff
@@ -0,0 +1,105 @@
...
+
+   c := c'
+
+   if h : op.getNumRegions! c.raw = 0 && !op.hasUses! c.raw then
+     c := WfRewriter.eraseOp c op (by grind) (by grind) (sorry)
```
math-fehr: Should we return `none` otherwise? Otherwise we might end up with 2 terminator operations back to back.

math-fehr (approved):
Nice! I added some style comments, but otherwise the algorithm looks good.

Veir/Passes/InstructionSelection/RISCV64Branches.lean
tobiasgrosser: Fixed.

Veir/Passes/InstructionSelection/RISCV64Branches.lean
tobiasgrosser: I could use `InsertPoint.after`, but the `reverse` is still needed as otherwise the test case fails.

Veir/Passes/InstructionSelection/RISCV64Branches.lean
tobiasgrosser: Done.

Veir/Passes/InstructionSelection/RISCV64Branches.lean
tobiasgrosser: Not sure how I return none. Also, there else case should not arise. I just added this if to see if the sorrys can be eliminated.

Veir/Passes/InstructionSelection/RISCV64Branches.lean
tobiasgrosser: Any ideas appreciated.

PR #828, opened 2026-06-10
fix: make sure propositions are defined with `theorem`

math-fehr:
Some propositions where defined with `def`, which is now failing on newest Lean versions.


tobiasgrosser:
I merged a similar change with the latest lean update in #825. Closing this as already completed.

PR #830, opened 2026-06-10
feat(refinement): add lemmas about `Interp` refinements

math-fehr:
These are quite useful to avoid unfolding the `isRefinedBy` definition while doing proofs.


luisacicolini (approved):

regehr:
@math-fehr can you explain a little bit why `None` plays into the refinement relation at all?

regehr (approved):

math-fehr:
> @math-fehr can you explain a little bit why `None` plays into the refinement relation at all?

This is the pain point that I am currently trying to fix. The idea is that our interpreter can "fail" sometimes, because the input is incorrect. We don't really want to reason about it, but sadly because it is there we have to deal with it.
So the best we can do here (or so I thought) is to say that `none` is refined by anything. But what will happen is practice is that I am going to add a proof that `none` is never return by the interpreter on a verifying operation.

PR #831, opened 2026-06-10
chore(interpreter): add `OperationPtr.interpret` abbreviation

math-fehr:
This makes it easier to write proofs about it, otherwise it becomes a mouthful.


regehr (approved):
LGTM

PR #832, opened 2026-06-10
feat(riscv): check that imm12 arguments are in bounds

regehr:
you might think that checking a signed range for riscv.sltiu (which is an unsigned comparison) is wrong, but I am emulating the behavior of other risc-v assemblers, such as LLVM's

there are other immediates that should be checked, this is just a start

### DIFF
--- a/Test/RISCV/ld-offset-oob.mlir
+++ b/Test/RISCV/ld-offset-oob.mlir
@@ -1,4 +1,5 @@
 // RUN: not veir-opt %s 2>&1 | filecheck %s
+// MLIR_INVALID
 
 "builtin.module"() ({
   "func.func"() <{sym_name = "main", function_type = () -> !riscv.reg}> ({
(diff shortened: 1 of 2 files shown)


Test/RISCV/ld-offset-oob.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
```
math-fehr, line 1: Gavin added `MLIR_INVALID` that checks that MLIR fails on this file, can you add it here as well?

math-fehr (approved):
ew, I guess MLIR/LLVM go on the assumption that any integer is signless by default, so they always put the min the signed range during printing/parsing?

Test/RISCV/ld-offset-oob.mlir
regehr: nice! I hadn't seen that. done. we should probably add this elsewhere too.

nchappe (approved):
This looks good, but it would probably be possible to get the operation names for free using op.getOpType.opName

regehr:
> This looks good, but it would probably be possible to get the operation names for free using op.getOpType.opName

nice. I've done that now!

luisacicolini (approved):
amazing, thank you!

PR #833, opened 2026-06-10
feat(interpreter): add lemmas about `VariableState` and `Conforms`

math-fehr:
These are proofs that getting values and operands from a `VariableContext` conform.


regehr (approved):
LGTM!

luisacicolini (approved):
lgtm, thank you!

PR #836, opened 2026-06-10
feat(riscv_cf): add `beqz` and `bnez` instructions

tobiasgrosser:
We add support for the pseudo operations `beqz` and `bneq`. Both take one conditional operand that is compared against zero, which makes for an easy lowering from LLVM's conditional branch operation.

### DIFF
--- a/Veir/Verifier.lean
+++ b/Veir/Verifier.lean
@@ -2220,9 +2220,9 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
     if op.getNumSuccessors ctx.raw opIn ≠ 2 then
       throw "Expected 2 successors"
     let sizes := (op.getProperties! ctx.raw (OpCode.riscv_cf .cbr)).operandSegmentSizes
-    if _ : sizes.values.size ≠ 4 then
+    if _ : sizes.values.size ≠ 3 then
       throw "Expected 1 operand plus 2 variadic operands"
-    if sizes.values[0]! ≠ 1 || sizes.values[1]! ≠ 1 then
+    if sizes.values[0]! ≠ 1 then
       throw "Expected 1 operand plus 2 variadic operands"
     pure ()
   /- RISCV Stack -/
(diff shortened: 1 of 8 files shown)


Veir/Verifier.lean
```diff
@@ -2093,6 +2093,19 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
...
+      throw "Expected 0 results"
+    if op.getNumRegions ctx.raw opIn ≠ 0 then
+      throw "Expected 0 regions"
+    if op.getNumSuccessors ctx.raw opIn ≠ 2 then
+      throw "Expected 2 successors"
+    let sizes := (op.getProperties! ctx.raw (OpCode.riscv_cf .cbr)).operandSegmentSizes
+    if _ : sizes.values.size ≠ 3 then
+      throw "Expected 1 operand plus 2 variadic operands"
+    if sizes.values[0]! ≠ 1 then
+      throw "Expected 1 operand plus 2 variadic operands"
```
luisacicolini, line 2237: shouldn't this error message be different than the one above? it might be confusing, they seem to check different things

luisacicolini (approved):
Reviewed mostly for style, lgtm except for the comment I left

nchappe (commented):
Isn't this pseudo-instruction called bnez?

nchappe:
For context, I had initially implemented only bne and beq as I was expecting cond_br lowering to always give beqz or bnez (if the branch targets end up swapped), which are beq and bne with a zero-register operand. I feel like support for named registers, especially the zero register, would make things easier for lowering to RISCV.

Veir/Verifier.lean
tobiasgrosser: Updated. Thank you for the comment.

tobiasgrosser:
> For context, I had initially implemented only bne and beq as I was expecting cond_br lowering to always give beqz or bnez (if the branch targets end up swapped), which are beq and bne with a zero-register operand. I feel like support for named registers, especially the zero register, would make things easier for lowering to RISCV.

Thank you for the context. I now add `beqz` and `bnez`.

With respect to zero-register support, maybe Sasha can share with us at some point how he implements this in xDSL? I would love our designs to be aligned.

tobiasgrosser:
@nchappe, can I get a final 👍

nchappe (approved):

PR #837, opened 2026-06-10
fix(riscv): immediates' width in verifier for `lui`, `auipc`

luisacicolini:


Veir/Verifier.lean
```diff
@@ -162,6 +162,14 @@ def OperationPtr.verifyIntegerExtTypes (op : OperationPtr) (ctx : WfIRContext Op
   else
     pure ()
 
+def OperationPtr.verifyRISCVneg (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) (imm : Int) : Except String PUnit :=
+  if 1048575 < imm then
```
regehr: is `imm` an `LLVM.Int` or a Lean `Int`? if the latter, then it would be good to also check >= 0

regehr:
except for the one possible thing, this is perfect!

Veir/Verifier.lean
luisacicolini: added extra check, as it's a Lean `Int` :)

luisacicolini:
thanks @regehr, all addressed!

Veir/Verifier.lean
```diff
@@ -162,6 +162,14 @@ def OperationPtr.verifyIntegerExtTypes (op : OperationPtr) (ctx : WfIRContext Op
   else
     pure ()
 
+def OperationPtr.verifyRISCVneg (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) (imm : Int) : Except String PUnit :=
+  if 0 ≤ imm ∧ 1048575 < imm then
+    let instrName := String.fromUTF8! (op.getOpType ctx.raw opIn).name
+    throw s!"{instrName} immediate out of bounds: must fit in a signed 20-bit field."
```
nchappe: typo: unsigned

Veir/Verifier.lean
```diff
@@ -162,6 +162,14 @@ def OperationPtr.verifyIntegerExtTypes (op : OperationPtr) (ctx : WfIRContext Op
   else
     pure ()
 
+def OperationPtr.verifyRISCVneg (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) (imm : Int) : Except String PUnit :=
+  if 0 ≤ imm ∧ 1048575 < imm then
```
nchappe: this should probably be `if imm < 0 \/ 1048575 < imm then`
(and maybe explicitly write it as 2^20-1 for clarity?)

nchappe (commented):
Nice, just two typos

Veir/Verifier.lean
luisacicolini: fied

Veir/Verifier.lean
luisacicolini: addressed

luisacicolini:
all addressed - thanks for the review! Will merge once green

PR #839, opened 2026-06-11
feat(interpreter): add useful `ValueMapping` methods

math-fehr:
These methods are useful to talk about the effect of a value mapping on an operation.  `ValueMapping.applyToArray` applies the mapping to an entire array (which is often used on operation operands).

`ValueMapping.ReflectsResults` is used to show that `op` results map directly to `op'`, and that all values mapping to `op'` results are exactly the results of `op`. This is necessary to prove some of the refinement proofs later on.

### DIFF
--- a/Veir/Interpreter/Refinement/Basic.lean
+++ b/Veir/Interpreter/Refinement/Basic.lean
@@ -144,6 +144,21 @@ def OperationPtr.isModuleRefinedBy (mod₁ : OperationPtr) (ctx₁ : WfIRContext
 abbrev ValueMapping (ctx ctx' : WfIRContext OpInfo) : Type :=
   {v : ValuePtr // v.InBounds ctx.raw} → {v : ValuePtr // v.InBounds ctx'.raw}
 
+/-- Apply the value mapping to an array of values with separately their bounds information. -/
+def ValueMapping.applyToArray {ctx ctx' : WfIRContext OpInfo} (mapping : ValueMapping ctx ctx')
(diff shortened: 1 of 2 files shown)


Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -144,6 +144,22 @@ def OperationPtr.isModuleRefinedBy (mod₁ : OperationPtr) (ctx₁ : WfIRContext
 abbrev ValueMapping (ctx ctx' : WfIRContext OpInfo) : Type :=
   {v : ValuePtr // v.InBounds ctx.raw} → {v : ValuePtr // v.InBounds ctx'.raw}
 
+/-- Apply the value mapping to an array of values with separately their bounds information. -/
+def ValueMapping.applyToArray {ctx ctx' : WfIRContext OpInfo} (mapping : ValueMapping ctx ctx')
+    (vals : Array ValuePtr) (valsIn : ∀ v ∈ vals, v.InBounds ctx.raw := by grind) : Array ValuePtr :=
+  vals.attach.map (fun ⟨v, hv⟩ => (mapping ⟨v, valsIn v hv⟩).val)
+
+/--
+`mapping` *reflects* `op'`'s result pointers back to `op`'s: the only value it sends onto `op'`'s
```
ineol: I'd say something like
```
mapping *relfects* op''s results if the ...
```

ineol (approved):
I don't have a lot of context but this seems like reasonable definitions to me.

PR #841, opened 2026-06-11
feat(interpreter): Prove that refinement is preserved when interpreting the same operation

math-fehr:
This is a major step in lifting the proof of refinement of a local rewrite to the proof of refinement of the entire module. The proof strategy to do this rely on simulation, and the lemma we had here proves that the simulation of any operation (besides the one being deleted) on the left program can be matched by a simulation of the same operation (besides rewriting of its operands) on the right program.

### DIFF
--- a/Veir/Interpreter/Lemmas.lean
+++ b/Veir/Interpreter/Lemmas.lean
@@ -1,12 +1,13 @@
 import Veir.Interpreter.Basic
 import Veir.Dominance
 import Veir.Interpreter.Refinement.Basic
+import Veir.Interpreter.Refinement.Lemmas
 import Veir.Verifier
 
 namespace Veir
 
 variable {OpInfo : Type} [HasOpInfo OpInfo]
-variable {ctx : WfIRContext OpInfo}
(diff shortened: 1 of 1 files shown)


Veir/Interpreter/Lemmas.lean
```diff
@@ -1,13 +1,14 @@
...
+import Veir.Interpreter.Refinement.Lemmas
 import Veir.Verifier
 
 namespace Veir
 
 variable {OpInfo : Type} [HasOpInfo OpInfo]
-variable {ctx : WfIRContext OpInfo}
-variable {varState varState' : VariableState ctx}
+variable {ctx ctx' : WfIRContext OpInfo}
+variable {varState varState'  : VariableState ctx}
```
tobiasgrosser:
```suggestion
variable {varState varState' : VariableState ctx}
```

tobiasgrosser (approved):
Nice. Let's say what @ineol and @nchappe say.

regehr (approved):
this LGTM, though I definitely don't understand everything!

nchappe (approved):

PR #844, opened 2026-06-11
feat: Handle poison load/stores

nchappe:
Memory is extended with a second ByteArray that serves as a poison mask.
RISCV memory accesses ignore the poison mask.

A limitation is that poison pointer reads become null addresses,
as we cannot represent poison addresses for now.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -135,13 +135,16 @@ end RuntimeValue
 @[ext]
 structure MemoryState where
   contents : ByteArray
+  poisonMask : ByteArray
 
-def MemoryState.empty : MemoryState :=
-  { contents := (ByteArray.emptyWithCapacity 1024).extend 8 0xff }
+def MemoryState.empty : MemoryState := {
+  contents := (ByteArray.emptyWithCapacity 1024).extend 8 0xff,
+  poisonMask := (ByteArray.emptyWithCapacity 1024).extend 8 0xff
+}
 
 def MemoryState.ensureSize (mem : MemoryState) (size : Nat) : MemoryState :=
   if mem.contents.size < size then
-    { contents := mem.contents.extend (size - mem.contents.size) 0 }
+    ⟨mem.contents.extend (size - mem.contents.size) 0, mem.poisonMask.extend (size - mem.poisonMask.size) 0xff⟩
   else
     mem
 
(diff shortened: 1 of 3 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -135,13 +135,16 @@ end RuntimeValue
 @[ext]
 structure MemoryState where
   contents : ByteArray
+  poisonMask : ByteArray
```
math-fehr, line 138: Can you add a documentation comment on `poisonMask` to say that the `1` correspond to the poison bits?

math-fehr, line 138: Should you add the proof that `contents.size = poisonMask.size`? I think that should be fairly easy to add, and that would help a bit when writing proofs about it?

tobiasgrosser (approved):
This is great, thank you!

Veir/Interpreter/Basic.lean
math-fehr, line 138: Another question, should we have a proof that if  `poisonMask` has a `1`, then `contents` must be `0` at that place? @tobiasgrosser had this at some point so it is easier to define equality with these things, not sure what you think?

Veir/Interpreter/Basic.lean
```diff
@@ -368,6 +387,22 @@ def MemoryState.load (state : MemoryState) (addr size : UInt64)
   else
     Interp.ub
 
+/--
+  Check if any bit at the given memory address is poison.
+  Yields UB if the access is out of bounds.
```
math-fehr: Can you document that `size` represents a byte size, not a bit size?

Veir/Interpreter/Basic.lean
```diff
@@ -330,16 +333,32 @@ instance : Inhabited (Interp α) := ⟨(none : Option (UBOr α))⟩
...
+    let poison := ByteArray.replicate val.size 0
+    return ⟨val.copySlice 0 state.contents addr.toNat val.size false,
+      poison.copySlice 0 state.poisonMask addr.toNat val.size false⟩
+  else
+    Interp.ub
+
+/--
+  Poison the given number of bytes, starting from the given address in memory.
+  Yields UB if the access is out of bounds.
+-/
```
math-fehr, line 365: Can you document that `n` represents a byte size?

math-fehr (commented):
I added some comments, notably on the MemoryState about some fields that we may want to add as well to make this "safer".
Am I understanding correctly that everything here is operating per byte, and later on you will add support for non-aligned loads-stores?

Veir/Interpreter/Basic.lean
nchappe: added as a field

nchappe: I think there was some debate on whether that was the right thing to do for the int type? (but it ultimately evolved into a variant, which made this irrelevant)

I'm not familiar enough with lean equality to have a clear opinion. I've not changed it for now, I expect the proof would be slightly annoying but not hard, if somebody has strong feelings about it I can take a look.

nchappe:
Rebased and addressed comments.
As is, everything is per-byte. We could support the byte type (then raw loads and stores would probably work with byte type values), but it's still accessing full bytes of memory.

math-fehr (approved):

PR #848, opened 2026-06-11
feat(riscv): make sure riscv instructions operate on riscv registers

regehr:
I think this is the right invariant for riscv dialect operations-- easy!!!


luisacicolini (approved):
great, lgtm!

tobiasgrosser (approved):

PR #850, opened 2026-06-11
fix(arith): IntegerOverflowFlagsAttr as in MLIR

### LINKED ISSUE #826: `arith` overflow options fail to round-trip through `mlir-opt`
For example, in veir's `Test/Arith/identity.mlir`
```llvm
%addi_nsw = "arith.addi"(%5, %5) <{"overflowFlags" = 1 : i32}> : (i32, i32) -> i32
``` 
is not valid MLIR syntax and `mlir-opt` will complain about:
```
Test/Arith/identity.mlir:8:19: error: 
  invalid properties {overflowFlags = 1 : i32} for op arith.addi: 
    invalid attribute `overflowFlags` in property conversion: 1 : i32
      %addi_nsw = "arith.addi"(%5, %5) <{"overflowFlags" = 1 : i32}> : (i32, i32) -> i32
```
See https://godbolt.org/z/ordM3Exqf

Instead, the overflow flags should be specified as 
```llvm
%addi_nsw = "arith.addi"(%5, %5) <{overflowFlags = #arith.overflow<nsw>}> : (i32, i32) -> i32
```

This is an issue for using veir in-line with an existing MLIR-based compiler, so I'd like to fix it, but I'm not sure what the best way is:

1. Simply hack the printer/parser  for arith ops so that it produces the expected string
2. Introduce an actual `#arith.overflow` attribute like in MLIR, switch the arith ops over from a simple i32 attr to using the new attribute, then add the printer/parser for that attribute.

I'm leaning towards (2), but I'm wondering if that might break things / there was a conscious choice not to do that in the first place?

AlexanderViand:
Fixes #826  - veir's handling of nsw/nuw integer overflow flags did not match upstream MLIR, so IR with these flags would no longer-roundtrip.

This PR updates it from
```llvm
%addi_nsw = "arith.addi"(%5, %5) <{"overflowFlags" = 1 : i32}> : (i32, i32) -> i32
```
to
```llvm
%addi_nsw = "arith.addi"(%5, %5) <{overflowFlags = #arith.overflow<nsw>}> : (i32, i32) -> i32
```
by replacing the i32 `IntegerAttr` with `ArithIntegerOverflowFlagsAttr` mirroring MLIR.

To make reviewing easier, this is split into two commits: [the first one](https://github.com/opencompl/veir/pull/850/changes/9664e8b82b5e234fc94b0be9fe1c19e7fbad75e8) does all lean-side changes, [the second one](https://github.com/opencompl/veir/pull/850/changes/9853dc98a7e14c79052631714521c4228b84599c) applies the new syntax to the many *.mlir test files that use the overflow flags.

### DIFF
--- a/Test/Arith/identity.mlir
+++ b/Test/Arith/identity.mlir
@@ -1,4 +1,5 @@
 // RUN: VEIR_ROUNDTRIP
+// RUN: MLIR_ROUNDTRIP
 
 "builtin.module"() ({
   "func.func"() <{function_type = () -> (), sym_name = "main"}> ({
--- a/Test/Arith/identity.mlir
+++ b/Test/Arith/identity.mlir
@@ -5,9 +5,9 @@
     ^4():
       %5 = "arith.constant"() <{ "value" = 13 : i32 }> : () -> i32
       %addi = "arith.addi"(%5, %5) : (i32, i32) -> i32
-      %addi_nsw = "arith.addi"(%5, %5) <{"overflowFlags" = 1 : i32}> : (i32, i32) -> i32
-      %addi_nuw = "arith.addi"(%5, %5) <{"overflowFlags" = 2 : i32}> : (i32, i32) -> i32
-      %addi_nsw_nuw = "arith.addi"(%5, %5) <{"overflowFlags" = 3 : i32}> : (i32, i32) -> i32
+      %addi_nsw = "arith.addi"(%5, %5) <{"overflowFlags" = #arith.overflow<nsw>}> : (i32, i32) -> i32
+      %addi_nuw = "arith.addi"(%5, %5) <{"overflowFlags" = #arith.overflow<nuw>}> : (i32, i32) -> i32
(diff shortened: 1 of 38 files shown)


AlexanderViand:
The initial version of this PR changed the printing for both `llvm` and `arith` because in MLIR both use an `IntegerOverflowAttr` internally, but for `llvm` MLIR's printing/parsing converts it back to/from the `i32` form so veir was right there already.

AlexanderViand:
@tobiasgrosser should be good to go now :)

Test/Arith/identity.mlir
tobiasgrosser, line 1: Can you add an `// RUN: MLIR_ROUNDTRIP` here?

tobiasgrosser (approved):
Nice

Test/Arith/identity.mlir
AlexanderViand: Oh, nice, that'd have probably caught this issue earlier - do I then need to do some lit.cfg stuff to mark it as requiring mlir-opt?

AlexanderViand: Nevermind, found the definition of MLIR_ROUNDTRIP and it's guarded by a check for mlil-opt already

PR #845, opened 2026-06-11
feat(fastntt): add `main`

luisacicolini:
Add a main to fastNTT tests with some constant numbers. The `LLVM/fastNTT` test is updated accordingly, copied from vcc's output.

PR #852, opened 2026-06-11
fix(fastntt): re-relax fastNTT test so it passes on MacOS

AlexanderViand:
#845 made the fastNTT test fail on macOS (which generates different module flags), this PR re-applies the relaxations from #759 

I also had to relax the `alignment` on `llvm.alloca` to get it to pass, not sure if that's something the test was intentionally checking for?


luisacicolini:
oopsie, sorry, I had totally lost this thread of failed tests. LGTM, thanks! Imho we should be fine without the alignment (for the purpose of the current test)

math-fehr (approved):

PR #854, opened 2026-06-11
feat(dataflow): sparse forward dataflow analysis api

axelcool1234:
Any transfer function that works on `OperationPtr` will use this API. Simply implementing a (transfer) function of the form `OperationPtr -> DataFlowContext -> IRContext OpCode -> DataFlowContext` gets you a sparse forward analysis plugged into the dataflow framework for free. This API handles most of the stuff the implementer of a transfer function shouldn't concern themselves with.

### DIFF
--- a/Veir/Analysis/DataFlow/SparseAnalysis.lean
+++ b/Veir/Analysis/DataFlow/SparseAnalysis.lean
@@ -80,9 +80,7 @@ private def getSuccessorOperand?
     | _ =>
       panic! "SparseForwardDataFlowAnalysis.getSuccessorOperand?: non-branch op"
 
-/--
-Conservatively treat blocks as live when no liveness facts exist.
--/
+/-- Conservatively treat blocks as live when no liveness facts exist. -/
 private def isBlockLive
     (block : BlockPtr)
     (dfCtx : DataFlowContext)
@@ -105,7 +103,7 @@ private def isEdgeLive
   let _ := dfCtx
   true
 
-/-- Subscribe to block liveness updates if dead code analysis is registered. -/
+/-- No-op when no liveness analysis is registered. -/
 private def subscribeToBlockLiveness
     (analysisKind : AnalysisKind)
     (block : BlockPtr)
@@ -116,7 +114,7 @@ private def subscribeToBlockLiveness
   let _ := irCtx
   dfCtx
 
-/-- Subscribe to edge liveness updates if dead code analysis is registered. -/
+/-- No-op when no liveness analysis is registered. -/
 private def subscribeToEdgeLiveness
     (analysisKind : AnalysisKind)
     (edge : CFGEdge)
--- a/Veir/Analysis/DataFlow/SparseAnalysis.lean
+++ b/Veir/Analysis/DataFlow/SparseAnalysis.lean
@@ -1,6 +1,7 @@
 module
 
 public import Veir.Analysis.DataFlow.SparseFact
+public import Veir.Interfaces.ControlFlowInterfaces
 
 public section
 
@@ -44,41 +45,6 @@ def joinAndPropagate
     return dfCtx
   dfCtx.modifyFactAndPropagate kind (.ValuePtr target) 
     (SparseFact.setLatticeElement · newValue, true) irCtx
-
-/--
-Return whether the given operation is a branch op.
--/
-private def isBranchOp
-    (op : OperationPtr)
-    (irCtx : WfIRContext OpCode) : Bool :=
-  -- TODO: Replace this `.test .test` check once VeIR has proper branch ops.
-  match (op.get! irCtx.raw).opType with
-  | .test .test =>
-    true
-  | _ =>
-    false
-
-/--
-Return the SSA value forwarded to the given successor's block argument, if any.
--/
-private def getSuccessorOperand?
-    (op : OperationPtr)
-    (successorIndex : Nat)
-    (argumentIndex : Nat)
-    (irCtx : WfIRContext OpCode) : Option ValuePtr :=
-  if successorIndex >= op.getNumSuccessors! irCtx.raw then
-    panic! s!"SparseForwardDataFlowAnalysis.getSuccessorOperand?: successor index {successorIndex} out of range"
-  else
-    match (op.get! irCtx.raw).opType with
-    -- TODO: Replace this `.test .test` check once VeIR has proper branch ops.
-    -- `successorIndex` will become relevant then.
-    | .test .test =>
-      if argumentIndex < op.getNumOperands! irCtx.raw then
-        some (op.getOperand! irCtx.raw argumentIndex)
-      else
-        none
-    | _ =>
-      panic! "SparseForwardDataFlowAnalysis.getSuccessorOperand?: non-branch op"
 
 /-- Conservatively treat blocks as live when no liveness facts exist. -/
 private def isBlockLive
@@ -174,14 +140,20 @@ private def visitBlock
       continue
 
     -- Check if we can reason about the dataflow from the predecessor.
-    if !isBranchOp predecessorOp irCtx then
+    if !(predecessorOp.getOpType! irCtx.raw).isTerminator then
       for target in block.getArguments! irCtx.raw do
         dfCtx := joinAndPropagate kind target ⊤ dfCtx irCtx
       return dfCtx
 
+    let some successorOperands :=
+        BranchOpInterface.getSuccessorOperands? predecessorOp predUse.index irCtx.raw
+      | for target in block.getArguments! irCtx.raw do
+          dfCtx := joinAndPropagate kind target ⊤ dfCtx irCtx
+        return dfCtx
+
     for i in [0:block.getNumArguments! irCtx.raw] do
       let arg := block.getArgument i
-      match getSuccessorOperand? predecessorOp predUse.index i irCtx with
+      match successorOperands[i]? with
       | some operand =>
         -- Add the current block start program point as a dependency of the
         -- predecessor block's successor operand lattice state, so this block
--- a/Veir/Analysis/DataFlow/SparseAnalysis.lean
+++ b/Veir/Analysis/DataFlow/SparseAnalysis.lean
@@ -1,0 +1,371 @@
+module
+
+public import Veir.Analysis.DataFlow.SparseFact
+
+public section
+
+namespace Veir
+
+namespace SparseForwardDataFlowAnalysis
+
+variable {kind : FactKind} {Domain : Type}
+
+-- TODO: When this is verified, we will need something stronger than this for `Domain`
+variable [Top Domain] [Bot Domain] [Join Domain] [DecidableEq Domain]
+
+/--
+The transfer function signature used for custom sparse analyses.
+
+The framework handles operand subscriptions, invokes this hook with the current
+operand lattice elements, and then joins any returned updates into the result
+facts. Returning `none` for a result means the transfer contributes no new fact
+for that result.
+-/
+abbrev VisitOperationFn (Domain : Type) :=
+  OperationPtr -> Array Domain -> WfIRContext OpCode -> Array (Option Domain)
+
+/--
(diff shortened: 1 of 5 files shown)


Veir/Analysis/DataFlow/SparseAnalysis.lean
```diff
@@ -0,0 +1,365 @@
...
+  dfCtx.modifyFactAndPropagate kind (.ValuePtr target) 
+    (SparseFact.setLatticeElement · newValue, true) irCtx
+
+/--
+Return whether the given operation is a branch op.
+-/
+private def isBranchOp
+    (op : OperationPtr)
+    (irCtx : IRContext OpCode) : Bool :=
+  -- TODO: Replace this `.test .test` check once VeIR has proper branch ops.
```
regehr: there are plenty of branch ops in Veir right now!
but you do not want to be making these decisions here, take a look at `Veir/GlobalOpInfo.lean` which has some helpers for answering questions like this, that might be the place to put this information

Veir/Analysis/DataFlow/SparseAnalysis.lean
```diff
@@ -0,0 +1,365 @@
...
+private def getSuccessorOperand?
+    (op : OperationPtr)
+    (successorIndex : Nat)
+    (argumentIndex : Nat)
+    (irCtx : IRContext OpCode) : Option ValuePtr :=
+  if successorIndex >= op.getNumSuccessors! irCtx then
+    panic! s!"SparseForwardDataFlowAnalysis.getSuccessorOperand?: successor index {successorIndex} out of range"
+  else
+    match (op.get! irCtx).opType with
+    -- TODO: Replace this `.test .test` check once VeIR has proper branch ops.
```
regehr: also here-- I'm not exactly sure what you're doing, but don't hard-code information about dialects here if you can help it

Veir/Analysis/DataFlow/SparseAnalysis.lean
```diff
@@ -0,0 +1,365 @@
...
+    -- `successorIndex` will become relevant then.
+    | .test .test =>
+      if argumentIndex < op.getNumOperands! irCtx then
```
regehr: do you need this dynamic check? if I understand correctly, the called is making sure this is in-bounds, you should be able to attach a little proof of that

Veir/Analysis/DataFlow/SparseAnalysis.lean
```diff
@@ -0,0 +1,365 @@
...
+    let region := regionPtr.get! irCtx
+    if let some firstBlock := region.firstBlock then
+      for arg in firstBlock.getArguments! irCtx do
+        dfCtx := joinAndPropagate kind arg ⊤ dfCtx irCtx
+
+  initializeRecursively kind analysisKind visitOperationImpl top dfCtx irCtx
+
+/--
+Visit an insertion point. If this is at beginning of block and all
+control flow predecessors or callsites are known, then the arguments
```
regehr: arguments'

regehr:
@math-fehr should this be using WfIRContext instead of IRContext?

regehr:
if I understand correctly, an unrecognized branch causes the analysis to start putting tops into successor blocks. but maybe it would be better to error out with a message that the analysis is unable to continue? I feel like we don't want to paper over something that's a serious problem

Veir/Analysis/DataFlow/SparseAnalysis.lean
regehr: branch ops include cf.br, cf.cond_br, llvm.br, llvm cond_br, and also some RISC-V branches

regehr:
I also had claude review this PR and it came up with a list of stuff to look at-- I don't want to paste it here, please do this on your own

math-fehr:
> @math-fehr should this be using WfIRContext instead of IRContext?

That would be better indeed! I'm not sure it's strictly necessary, but it might definitely bite us later if we don't do this.
We can fix this in a later PR though, but we should do it soon-ish.


Otherwise, I'll just wait that axel fixes your changes and take a look at claude comments, so then I can review it when it's more ready!

axelcool1234:
@math-fehr @regehr Sounds good guys, thanks for the review. I might only be able to get back to this next week. I'm wrapping up my internship (I have two presentations to prepare for this weekend and do this coming week). I really appreciate your patience, I know I have taken forever on this stuff. I look forward to finally getting this into VeIR :)

Veir/Analysis/DataFlow/SparseAnalysis.lean
axelcool1234: Here, I implemented an interface: https://github.com/opencompl/veir/pull/1237

axelcool1234:
@math-fehr I just fixed all of John's concerns over hard coded branch op checking now that I got the interface in. This is up for reviewing again :)

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
```diff
@@ -0,0 +1,341 @@
+module
+
+public import Veir.Analysis.DataFlow.SparseFact
+public import Veir.Interfaces.ControlFlowInterfaces
+
+public section
+
+namespace Veir
+
+namespace SparseForwardDataFlowAnalysis
```
regehr, line 11: maybe change the file name to match this.
alternatively, you could put both forward and backward analyses in this same file, but probably best not to do that

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
```diff
@@ -0,0 +1,341 @@
...
+-- TODO: When this is verified, we will need something stronger than this for `Domain`
+variable [Top Domain] [Bot Domain] [Join Domain] [DecidableEq Domain]
+
+/--
+The transfer function signature used for custom sparse analyses.
+
+The framework handles operand subscriptions, invokes this hook with the current
+operand lattice elements, and then joins any returned updates into the result
+facts. Returning `none` for a result means the transfer contributes no new fact
+for that result.
```
regehr, line 24: we really want to permit a transfer function to return none? what's the argument for this?

regehr (commented):
if nothing imports your new files do they even get compiled? if you're not ready for real clients yet maybe fix this by putting a trivial analysis into a unit test?

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
axelcool1234: Yeah I was planning on having both in the same file. But you're right, I could just have them in separate files.

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
axelcool1234: If there's nothing to update in the lattice, it returns none. Example:
in constant propagation, when trying to infer the abstract value of `x` in `x = y + z`, if either `y` or `z` or both are bottom, we return none. This is essentially the same as returning bottom in this case (this is because everything not explicitly stored in the framework's product lattice is by default bottom), but returning none is more efficient as that allows the dataflow framework to do nothing for that abstract value, i.e. to not explicitly store bottom which is what everything is by default, but without taking up actual space in the product lattice - it is implicitly bottom rather than explicitly. Due to the dataflow framework being monotone, this is sound (bottom is the lowest element in the lattice, so we will never have a situation where returning bottom turns a non-bottom value back to bottom vs returning none keeping it where it is, since this'd violate monotonicity) assuming our transfer functions are correct (once this is verified this'll be ensured). We _could_ just always return something in our abstract domain, but making it `Option` allows the framework to decide to not update anything instead, which is valid since abstract values change in a non-decreasing manner (so it can stay where it is). 

I hope this word vomit made sense. For constant propagation, the only time `none` is returned for an abstract value, it means it is already `bottom`, because if that weren't the case we are being unsound somewhere.

axelcool1234:
> if nothing imports your new files do they even get compiled? if you're not ready for real clients yet maybe fix this by putting a trivial analysis into a unit test?

My own constant propagation transfer function is a client right now for the sparse analysis API. I could have that be the trivial analysis. It is its own PR though (https://github.com/opencompl/veir/pull/660) - about 130 lines (which will significantly lower once I hook it up to your folder). Do you want me to do this @regehr ? Add the contents of that PR to this one and then update it to use your folding logic? That way it has real code using it along with tests.

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
axelcool1234:
```
                    ┌─────────────┐
                    │      ⊤      │
                    │   unknown   │
                    └──────┬──────┘
                           │
                           │  less information
                           │
                    ┌──────▼──────┐
                    │  Constant   │
                    │             │
                    └──────┬──────┘
                           │
                           │  less information
                           │
                    ┌──────▼──────┐
                    │      ⊥      │
                    │uninitialized│
                    └─────────────┘
```
 ⊥  <=  Constant  <=  ⊤
 
 
 When constant propagation starts, everything starts as ⊥. So when we see `x = ⊥ + Constant`, we return none for `x`, as the only thing `x` could possibly be is `⊥` if we are actually sound. Because if we weren't, and lets say `x` was a constant, then we are violating monotonicity because ⊥ <= Constant and returning none is the same as returning bottom in this case! We wouldn't be non-decreasing!

axelcool1234: Essentially, returning none is okay when we already know the value the abstract value is and know that we are keeping it where it is. In constant propagation's case, we know that `x` is bottom so there's no need to update it to bottom again. We can just not update it, which is more efficient as it is not explicitly stored as such. It is an optimization

axelcool1234: If this didn't make sense (I hope it did), please tell me and I'll try and explain it again in a different way

regehr: thanks. I think this makes sense. I thought you were allowing the actual transfer functions to return None which seemed wrong, but as an internal optimization on top of standard transfer functions this seems fine.

regehr:
> > if nothing imports your new files do they even get compiled? if you're not ready for real clients yet maybe fix this by putting a trivial analysis into a unit test?
> 
> My own constant propagation transfer function is a client right now for the sparse analysis API. I could have that be the trivial analysis. It is its own PR though (#660) - about 130 lines (which will significantly lower once I hook it up to your folder). Do you want me to do this @regehr ? Add the contents of that PR to this one and then update it to use your folding logic? That way it has real code using it along with tests.

no, let's keep this PR simple. 
but in general we should be super careful in contributing files that are disconnected from the veir build! these might not even compile and CI can't see this.

regehr (approved):

Veir/Analysis/DataFlow/SparseAnalysis.lean
```diff
@@ -0,0 +1,341 @@
...
+  dfCtx
+
+/-- No-op when no liveness analysis is registered. -/
+private def subscribeToEdgeLiveness
+    (analysisKind : AnalysisKind)
+    (edge : CFGEdge)
+    (dfCtx : DataFlowContext) : DataFlowContext :=
+  let _ := analysisKind
+  let _ := edge
+  dfCtx
```
axelcool1234: These were left conservative as liveness analysis wasn't in VeIR. I'm gonna change this.

axelcool1234:
@math-fehr @regehr This is ready I believe

ineol:
@axelcool1234 CI is failing with: warning: Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean is an orphan and is not compiled

regehr (approved):

regehr:
LGTM!

PR #856, opened 2026-06-11
feat(dataflow): liveness domain

axelcool1234:
This abstract domain will be used for dead code analysis.

### DIFF
--- a/Veir/Analysis/DataFlow/Domains/LivenessDomain.lean
+++ b/Veir/Analysis/DataFlow/Domains/LivenessDomain.lean
@@ -1,0 +1,107 @@
+module
+
+public import Veir.Analysis.DataFlow.Domains.AbstractDomain
+
+public section
+
+namespace Veir
+
+/-!
+# Liveness domain
+
+Instantiation of `AbstractDomain` with a two point lattice used by dead code analysis.
+`dead` denotes an unreachable program point or edge, while `live` denotes an executable one.
+-/
+
+/-- Abstract values used to track executability. -/
+inductive Liveness where
+  | dead
+  | live
+deriving BEq, DecidableEq, TypeName
+
+namespace Liveness
+
+/-- Defines the ordering of abstract values in the liveness domain. -/
+def le (x y : Liveness) : Prop :=
+  match x, y with
+  | .dead, _ => True
+  | .live, .live => True
+  | .live, .dead => False
+
+instance : LE Liveness where
+  le := le
+
+@[simp] theorem le_def (a b : Liveness) : (a ≤ b) ↔ le a b := Iff.rfl
+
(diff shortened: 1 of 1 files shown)


axelcool1234:
Interesting to note: A lot of the theorems for `LivenessDomain` are the same (in the body) as those of `ConstantDomain`. There's probably something to that right?

axelcool1234:
@regehr Does this look like a reasonable domain for dead code analysis? It's basically just top and bottom - two element lattice.

Veir/Analysis/DataFlow/Domains/LivenessDomain.lean
```diff
@@ -0,0 +1,114 @@
...
+
+public section
+
+namespace Veir
+
+/-!
+# Liveness domain
+
+Instantiation of `AbstractDomain` with a two point lattice used by dead code analysis.
+`dead` denotes an unreachable program point or edge, while `live` denotes an executable one.
```
math-fehr:
```suggestion
`dead` denotes an unreachable program point or edge, while `live` denotes a reachable one.
```

Veir/Analysis/DataFlow/Domains/LivenessDomain.lean
```diff
@@ -0,0 +1,114 @@
...
+  | .live, .dead => False
+
+instance : LE Liveness where
+  le := le
+
+@[simp] theorem le_def (a b : Liveness) : (a ≤ b) ↔ le a b := Iff.rfl
+
+@[simp, grind .]
+theorem le_top (a : Liveness) : a ≤ .live := by
+  cases a <;> trivial
```
math-fehr: This simp should be removed, otherwise all your other `<=` simp will never trigger because `<=` is eliminated and transformed to `le`.

math-fehr (approved):
Looks good to me, I would just remove the `simp` from `le_def`, and change one comment, then let's wait for john review!

axelcool1234:
> Looks good to me, I would just remove the `simp` from `le_def`, and change one comment, then let's wait for john review!

If I remove `simp` from `le_def` the theorems fail, just like with constant domain

math-fehr:
Yes, but the point is that you call `simp [le_def]` in the proofs, so that they take care of it.
The idea is that marking it `@[simp]` means that this gets unfolded automatically by *everyone*, while here you just want to unfold it locally for your proofs

axelcool1234:
> Yes, but the point is that you call `simp [le_def]` in the proofs, so that they take care of it. The idea is that marking it `@[simp]` means that this gets unfolded automatically by _everyone_, while here you just want to unfold it locally for your proofs

Ohhhhhh! Thanks for the clarification! I'll fix ConstantDomain to do this too in another PR.

regehr (approved):
LGTM!

PR #858, opened 2026-06-12
feat(dataflow): Fix `propagate` to take in `LatticeAnchor`

axelcool1234:
Since facts do not store their anchors (see https://github.com/opencompl/veir/commit/1db68dc83b0fb242386bfb3838a5077dfdb69048), `propagate` needs to take it as a parameter.

PR #859, opened 2026-06-12
feat(dataflow): sparse fact api

axelcool1234:
The following is a type that builds off of the generic `Fact` type. You can define "sparse facts" - facts that attach to `Value`s, using this type. It'll handle notifying dependents and subscribed analyses for you when it changes.

### DIFF
--- a/Veir/Analysis/DataFlow/Facts.lean
+++ b/Veir/Analysis/DataFlow/Facts.lean
@@ -83,6 +83,13 @@ structure RegionMetadataPayload where
   postOrderIndex : HashMap BlockPtr Nat := {}
 
 /--
+A sparse dataflow fact payload for one abstract domain.
+-/
+structure SparsePayload (Domain : Type) where
+  useDefSubscribers : Array AnalysisKind := #[]
+  latticeElement : Domain
+
+/--
 The fact specific data stored for each fact kind.
 -/
 @[expose] def FactPayload : FactKind → Type
@@ -94,8 +101,8 @@ A dataflow fact stored by the framework.
 
 Each fact associates with a lattice anchor (some location in the program), has
 an array of dependents (other facts that "depend" on this fact's current state in
-some fashion), has an array of analysis subscribers (similar to dependents except
-it's entire analyses that depend on this fact's current state), and has the fact
+some fashion), has an array of analysis subscribers (similar to dependents except 
+it's entire analyses that depend on this fact's current state), and has the fact 
 specific payload determined by its `FactKind`.
 -/
 structure Fact (kind : FactKind) where
--- a/Veir/Analysis/DataFlow/Facts.lean
+++ b/Veir/Analysis/DataFlow/Facts.lean
@@ -92,12 +92,15 @@ The fact specific data stored for each fact kind.
 /--
 A dataflow fact stored by the framework.
 
-Each fact associates with a lattice anchor (some location in the program), has 
+Each fact associates with a lattice anchor (some location in the program), has
 an array of dependents (other facts that "depend" on this fact's current state in
-some fashion), and has the fact specific payload determined by its `FactKind`.
+some fashion), has an array of analysis subscribers (similar to dependents except
+it's entire analyses that depend on this fact's current state), and has the fact
+specific payload determined by its `FactKind`.
 -/
 structure Fact (kind : FactKind) where
   dependents : Array WorkItem := #[]
+  subscribers : Array AnalysisKind := #[]
   payload : FactPayload kind
 
 namespace Fact
@@ -113,6 +116,15 @@ Add one dependent work item to the fact.
 -/
 def addDependent (fact : Fact kind) (workItem : WorkItem) : Fact kind :=
   fact.setDependents (fact.dependents.push workItem)
+
+/--
+Subscribe one analysis to changes of this fact.
+-/
+def subscribe (fact : Fact kind) (analysisKind : AnalysisKind) : Fact kind :=
+  if fact.subscribers.contains analysisKind then
+    fact
+  else
+    { fact with subscribers := (fact.subscribers.push analysisKind) }
 
 /--
 Enqueue all dependents of this fact.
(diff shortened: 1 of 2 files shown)


Veir/Analysis/DataFlow/Facts.lean
```diff
@@ -92,13 +98,15 @@ The fact specific data stored for each fact kind.
...
 
-Each fact carries a lattice anchor (some location in the program this fact
-associates with), an array of dependents (other facts that "depend" on this fact's
-current state in some fashion), and the fact specific payload determined by its
-`FactKind`.
+Each fact carries a lattice anchor (which is some location in the program), 
+an array of dependents (other facts that "depend" on this fact's current 
+state in some fashion), an array of analysis subscribers (similar to dependents
+execpt it's entire analyses that depend on this fact's current state), and the 
+fact specific payload determined by its `FactKind`.
```
ineol: Facts don't carry lattice anchors, I think I removed them a while back but forgot to change the doc :(

axelcool1234: @ineol I'll fix that comment in this PR: https://github.com/opencompl/veir/pull/858

Veir/Analysis/DataFlow/SparseFact.lean
```diff
@@ -0,0 +1,83 @@
...
+      for analysisKind in state.subscribers do
+        match InsertPoint.after? user irCtx with
+        | some point =>
+          dfCtx := dfCtx.enqueue (point, analysisKind)
+        | none =>
+          pure ()
+      maybeUse := (use.get! irCtx).nextUse
+  | _ =>
+    pure ()
+  dfCtx
```
axelcool1234, line 55: @math-fehr Here is an example of `subscribers` being used. For any fact that attaches to SSA `Value`s, it enqueues into the worklist all of its uses for each subscribed analysis whenever said SSA fact changes. That's because its uses will be affected by that change. Analyses concerned about said change of this fact will be subscribed so that it can run its transfer function on the uses.

axelcool1234:
Cc @math-fehr @regehr @ineol this is the first step towards supporting transfer functions on SSA values, which is what John is mostly concerned with. I need some review of this code, thank you 🙏

Veir/Analysis/DataFlow/SparseFact.lean
```diff
@@ -0,0 +1,83 @@
+module
+
+public import Veir.Analysis.DataFlowFramework
+public import Veir.Analysis.DataFlow.Domains.AbstractDomain
+
+public section
+
+namespace Veir
+
+class SparseFactSpec (kind : FactKind) (Domain : outParam Type) where
```
math-fehr, line 14: Can you add documentation to this class?
Also, given that this is only a single element that is a proof, is there really a need for it to be a class?
Maybe @ineol knows what's best practice here

Veir/Analysis/DataFlow/SparseFact.lean
```diff
@@ -0,0 +1,83 @@
...
+  payloadEq : FactPayload kind = SparsePayload Domain
+
+namespace SparseFact
+
+variable {kind : FactKind} {Domain : Type}
+variable [SparseFactSpec kind Domain]
+
+private theorem payloadEq : FactPayload kind = SparsePayload Domain :=
+  SparseFactSpec.payloadEq (kind := kind)
```
math-fehr: What is the point of this theorem, compared to just using the class here?

Veir/Analysis/DataFlow/SparseFact.lean
```diff
@@ -0,0 +1,83 @@
...
+def propagate (state : Fact kind) (anchor : LatticeAnchor) 
+  (dfCtx : DataFlowContext) (irCtx : IRContext OpCode) : DataFlowContext := Id.run do
+  let mut dfCtx := { dfCtx with workList := state.enqueueDependents dfCtx.workList }
+  match anchor with
+  | .ValuePtr ssaValue =>
+    let mut maybeUse := ssaValue.getFirstUse! irCtx
+    while let some use := maybeUse do
+      let user := (use.get! irCtx).owner
+      for analysisKind in state.subscribers do
+        match InsertPoint.after? user irCtx with
```
math-fehr: You should swap the `for` and the `match` here.

Veir/Analysis/DataFlow/SparseFact.lean
```diff
@@ -0,0 +1,83 @@
...
+        | some point =>
+          dfCtx := dfCtx.enqueue (point, analysisKind)
+        | none =>
+          pure ()
```
math-fehr: Also, should the for loop and enqueue thing be moved to a separate function outside this file? This seems very reusable?

math-fehr (commented):
Sorry for the delay!
I just added some small comments, but otherwise I think that's good!

Veir/Analysis/DataFlow/SparseFact.lean
axelcool1234: Good catch, not sure why I did that. Thanks.

Veir/Analysis/DataFlow/SparseFact.lean
axelcool1234: It actually is reused. The propagate function is given to `FactSpec`. Every `SparseFact` gets this `propagate` via the `FactSpec` instance for free.

Veir/Analysis/DataFlow/SparseFact.lean
axelcool1234: There will be more sparse fact types that use it (such as a type for known bits). Also, by registering a `SparseFactSpec` for your fact type, you get `FactSpec` for free. It's all you need to do as a user creating a new fact type for your analysis in the dataflow framework.

axelcool1234: This is more obvious in the PR that directly uses this: https://github.com/opencompl/veir/pull/854. It takes in a `SparseFactSpec` instead of a `FactSpec`.

math-fehr: Okay, then can you add this in the documentation, so that it is clear what it is meant to be?

math-fehr (approved):
Thanks, just add the documentation for the class and then I'll merge it!

PR #860, opened 2026-06-12
feat(dataflow): fact subscribers

axelcool1234:
This adds an extra field to `Fact` for analyses subscribed to it. This is much broader than a `Fact f`'s `dependents`, which are for other `Fact`s that need to be notified when `f` changes. Subscribers can be used to notify an arbitrary number of `Fact`s of the subscribed analysis. This is used in both dead code analysis and sparse analyses like constant propagation analysis.

### DIFF
--- a/Veir/Analysis/DataFlow/Facts.lean
+++ b/Veir/Analysis/DataFlow/Facts.lean
@@ -92,13 +92,15 @@ The fact specific data stored for each fact kind.
 /--
 A dataflow fact stored by the framework.
 
-Each fact carries a lattice anchor (some location in the program this fact
-associates with), an array of dependents (other facts that "depend" on this fact's
-current state in some fashion), and the fact specific payload determined by its
-`FactKind`.
+Each fact carries a lattice anchor (which is some location in the program), 
+an array of dependents (other facts that "depend" on this fact's current 
+state in some fashion), an array of analysis subscribers (similar to dependents
+execpt it's entire analyses that depend on this fact's current state), and the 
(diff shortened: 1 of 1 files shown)


math-fehr:
I am quite confused by your description, do you have an example on what would be the difference between subscribers and dependents?

axelcool1234:
> I am quite confused by your description, do you have an example on what would be the difference between subscribers and dependents?

@math-fehr I don't blame you I had trouble differentiating the two at first as well. Here's a good example: in dead code analysis, when a block goes from dead to live in the lattice, it enqueues the block and every single operation within that block for every analysis that's subscribed to it.

Now, you _could_ have only `dependents`. But it's be an insane amount of `dependents` for some analyses. `subscribers` provide more flexibility and prevents this space explosion.

math-fehr:
So what would subscribers do in that case?

axelcool1234:
> So what would subscribers do in that case?

@math-fehr for every subscribed analysis, it enqueues the pair (subscribed analysis, block) and (subscribed analysis, operation) for every operation in the block.

axelcool1234:
> > So what would subscribers do in that case?
> 
> @math-fehr for every subscribed analysis, it enqueues the pair (subscribed analysis, block) and (subscribed analysis, operation) for every operation in the block.

Instead of having every single one of these pairs in `dependents`, which could be literally hundreds, you just have an analysis subscribe and it does this double for loop when dead code changes.

axelcool1234:
Closed in favor of https://github.com/opencompl/veir/pull/859

PR #861, opened 2026-06-12
feat(riscv): introduce riscv.addi when possible

regehr:
also tweaked the signature of `matchLi` since this saves a line of code in both of its current clients, hopefully that's a good general improvement


math-fehr (approved):
Cool!

luisacicolini (approved):
lgtm, thank you!

PR #862, opened 2026-06-12
feat(riscv): combines introducing the remaining commutative imm12 instructions

regehr:
ok I think all of these combines are fine... they're all commutative so should be no risk here. `addiw` is a bit of an outlier since it's a 32-bit operation, but I don't think this matters here!


tobiasgrosser (commented):
How cool. 🙏

luisacicolini (approved):
lgtm, thank you!

regehr:
I have updated these to use LocalRewritePattern, everything else unchanged. I'll leave this here for a bit in case anyone wants to look again, but I don't think that's necessary!

luisacicolini:
chatted with @math-fehr, all is good

PR #863, opened 2026-06-12
fix(dataflow): removed bad simp from constant domain

axelcool1234:
Did as described here: https://github.com/opencompl/veir/pull/856#issuecomment-4694289583
But for constant domain.


math-fehr (approved):
Thanks!

PR #864, opened 2026-06-12
chore(veiropt): tighten up command line parsing

regehr:
I spent a while trying to figure out why `veiropt --passes=cse` wasn't doing anything. it's because we were silently dropping unrecognized flags. let's tighten that up so it doesn't trip up anyone else. solution is the standard one: parsed options are consumed, and we error out if anything is left at the end.


axelcool1234 (commented):
LGTM

tobiasgrosser (approved):

math-fehr (approved):
Nice, thanks!

PR #865, opened 2026-06-13
feat(rewriter): Require inBounds argument to rewrite patterns

math-fehr:
A lot of sorries cannot be removed in our rewrite patterns because rewrite patterns do not take `op.InBounds ctx` proofs as arguments.

This PR adds the `inBounds` argument in rewrite patterns, though it does not remove any sorries yet.


tobiasgrosser (approved):
Nice.

PR #867, opened 2026-06-13
feat(cse): use dominance to make CSE work across blocks

regehr:
ok it looks like this bit is actually load-bearing:
```
  -- `domCtx` keeps a stable handle on the original context after
  -- `ctx` is shadowed by the mutable copy.
  let domCtx := ctx.raw
```
this is worrisome to me because I don't understand the consequences


regehr:
NB: this PR cannot be merged until dominance is robust with respect to deleting instructions

axelcool1234:
I'll make sure to unblock this soon John 😎

Veir/IR/Dominance.lean
```diff
@@ -183,6 +183,28 @@ def properlyDominates
...
+    let mut blocks := #[]
+    if let some metadata := region.getRegionMetadataFact? dfCtx irCtx then
+      blocks := (metadata.postOrderIndex.toArray.qsort (·.2 > ·.2)).map (·.1)
+    for block in blocks do
+      let mut currentOp := (block.get! irCtx.raw).firstOp
+      while let some innerOp := currentOp do
+        ops := ops.push innerOp
+        ops := ops ++ innerOp.opsInDominanceOrder dfCtx irCtx
+        currentOp := (innerOp.get! irCtx.raw).next
+  return ops
```
math-fehr, line 206: If you want to make this efficient (as much as with an iterator), what you can do is pass `ops` as an argument, and only append to it.
That way you never do these costly array concatenations.

Veir/IRNesting.lean
```diff
@@ -236,4 +237,18 @@ partial def RegionPtr.isAncestorOf
...
+region is one of the isolated operation's direct regions; different regions of
+the same isolated operation are separate scopes.
+-/
+partial def RegionPtr.nearestIsolatedScope? {OpInfo : Type} [HasOpInfo OpInfo]
+    (region : RegionPtr) (ctx : IRContext OpInfo) : Option RegionPtr := do
+  let parentOp ← (region.get! ctx).parent
+  if HasOpInfo.isIsolatedFromAbove (parentOp.get! ctx).opType then
+    return region
+  let parentRegion ← parentOp.getParentRegion! ctx
+  parentRegion.nearestIsolatedScope? ctx
```
math-fehr: Can you move this in `HasOpInfo`?

math-fehr (approved):
Very cool! I just had two small comments, feel free to merge once they are addressed

tobiasgrosser (approved):
Very cool. Ready-to-merge after @math-fehr's comments are addressed.

Veir/IR/Dominance.lean
regehr: I had claude do some benchmarking of your version and it helps only minimally. something else in CSE is dominanting runtime. I propose not making this change and then we try to speed things up in a separate profiling run later.

PR #870, opened 2026-06-13
feat(rewriter): add `!` version of `WfRewriter` functions

math-fehr:
These functions do not require to provide any proofs and instead check dynamically every required proofs.

There is currently no proofs provided for these new functions.


tobiasgrosser (approved):
Nice.

PR #872, opened 2026-06-15
feat(riscv): move constant arguments into the immediate field of shifts, when possible

regehr:
these have all been proved! but the vibe-proofs are not something we can upstream. Mathieu has looked at them :). so we contribute only the code and wait proofs will wait until the infrastructure is ready.


luisacicolini:
LGTM, thanks! We shall do some refactoring after we have canonicalization to avoid the double check on both operands, but happy to merge as it is for now

luisacicolini (approved):

PR #873, opened 2026-06-15
feat(interpreter): Add InterpreterState.DefinesDominating

math-fehr:
This proposition asserts that the variable state should contain all variables that are dominating the current location. From this definition, we can now have a proof that `interpretOp` is never return `none` whenever the state is `DefinesDominating`

### DIFF
--- a/Veir/Interpreter/EquationLemma.lean
+++ b/Veir/Interpreter/EquationLemma.lean
@@ -139,3 +139,89 @@ theorem interpretOp_equationLemmaAt {ctx : WfIRContext OpCode} {opInBounds} {sta
     · grind [interpretOp_equationHolds_other]
     · grind
   · grind [interpretOp_equationHolds_self]
+
+/-- An interpreter state satisfies the `DefinesDominating` invariant at a program point if it
+defines all values that dominate that program point. This should be satisfied by any state in the
+interpreter. -/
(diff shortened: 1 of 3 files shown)


luisacicolini (approved):

Veir/Interpreter/EquationLemma.lean
```diff
@@ -139,3 +139,78 @@ theorem interpretOp_equationLemmaAt {ctx : WfIRContext OpCode} {opInBounds} {sta
...
   · grind [interpretOp_equationHolds_self]
+
+/-- An interpreter state satisfies the `DefinesDominating` invariant at a program point if it
+defines all values that dominate that program point. This should be satisfied by any state in the
+interpreter. -/
+def InterpreterState.DefinesDominating {ctx : WfIRContext OpCode} (state : InterpreterState ctx)
+    (location : InsertPoint) (_locInBounds : location.InBounds ctx.raw := by grind) : Prop :=
+  ∀ (value : ValuePtr) (_valueInBounds : value.InBounds ctx.raw),
+  value.dominatesIp location ctx →
+  ∃ val, state.variables.getVar? value = some val
```
ineol:
```suggestion
  (state.variables.getVar? value).isSome
```

ineol: and also in the sequel? or does it not work well with grind?

ineol (approved):

PR #878, opened 2026-06-15
refactor: promote `llvm_bv_decide` to a non-test module.

alexkeizer:
This PR moves `llvm_bv_decide` from a test file into `LLVM.Int.Simp`, which seemed appropriate given the relevant simpset was defined there.

Unfortunately, the `getValue_eq_getValueD` lemma isn't yet available in this file, so I introduce a new `llvm_toBitVec_post` lemma for the second `simp`. This extra simpset will be useful when we wish to tighten the simp-step. It currently includes the default global simpset, and indeed some of the tests break if we tighten the simp call to *just* the `getValueD` lemma. In a future PR, we should figure out which simp-lemmas are actually needed for these tests, add them to `llvm_toBitVec_post` and then tighten the `simp` (having a dependency on the default simpset seems undesirable).

Futhermore, I split the tactic into a `llvm_bv_normalize` tactic which does just the simping, and a `llvm_bv_decide` tactic which normalizes (using `llvm_bv_normalize`) and then calls `bv_decide`, for easier debugging.

Finally, I removed usages of `try` from the tactic. `try` silently swallows errors, making debugging harder. Instead, I use `simp -failIfUnchanged` to not throw an error when simp didn't change the goal, and `<;>` to not throw an error if a goal was solved by an earlier tactic in the sequence.


luisacicolini:
Thank you! this lgtm, can you also refactor the proofs in `https://github.com/opencompl/veir/blob/main/Veir/Passes/InstructionSelection/Proofs.lean` if possible? I'm afraid that the ashr/sra proof might break, but everything else should be okay

tobiasgrosser (commented):
Nice. Let's address @luisacicolini's comments. From my side, all is good.

alexkeizer:
> Thank you! this lgtm, can you also refactor the proofs in `https://github.com/opencompl/veir/blob/main/Veir/Passes/InstructionSelection/Proofs.lean` if possible? I'm afraid that the ashr/sra proof might break, but everything else should be okay

First impression on opening this file: the div/rem refinement theorems at the bottom show `unsolved goals` locally, despite CI seeming to be happy. Is my local setup broken, or is this file just not excersised in CI?

Second question: what refactor do you want? Should we unify the `refine_bv_decide` tactic from that file into the `llvm_bv_decide` I define in this PR? Perhaps the tactic ought to be named `veir_bv_decide` to be more general?

alexkeizer:
I moved all the risc-v bitblasting normalization lemmas into the same simpset, renamed the tactic to `veir_bv_decide` to reflect the more general nature, and used it for the proofs in the linked file, as per the commit:
````quote
The RISC-V instruction selection proofs also had a separate bitblasting preprocessing, which this commit now unifies into a single tactic. Since this tactic is no longer llvm-specific, but also incorporates RISC-V semantics, I renamed the tactics to `veir_bv_{normalize,decide}`.

Similarly, I renamed the backing simpsets to `veir_bv_normalize`(`_post`), moving all simp lemmas from both `llvm_toBitVec` and `riscv_toBitVec` into this shared preprocessing simpsets.

Finally, I've dropped the `refine_bv_decide` tactic in the InstructionSelection proofs, since this job is now done by the common `veir_bv_decide` tactic.
````

One caveat: as mentioned above, 4 lemmas from the InstructionSelection Proofs file do not solve on my local machine, but they didn't solve either before my changes so I assume it's orthogonal.

@luisacicolini since it's a relatively big change, please give it another review!

PS:
> I'm afraid that the ashr/sra proof might break

Yeah, I had to remove the `toNat` lemma from the second stage normalization; this is a good argument for why the global simpset really shouldn't be in there!

luisacicolini:
Sorry only saw this now (please ping me on zulip in case I miss these messages). I also realized the proofs were broken, #888 fixes them - shall we merge this first and I will adapt #888 afterwards?

luisacicolini (approved):
lgtm, thank you!

PR #880, opened 2026-06-16
feat: Add improved Lean compiler reduceArity pass

ineol:
This will be useful for the byte-level IR representation later.

### DIFF
--- a/ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
+++ b/ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
@@ -1,0 +1,315 @@
+/-
+Copyright (c) 2022 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Leonardo de Moura, Leo Stefanesco
+-/
+module
+
+prelude
+public import Lean.Compiler.LCNF.Internalize
+import Lean.Compiler.LCNF.PrettyPrinter
+import Lean.Compiler.LCNF.Toposort
+
+open Std (HashMap HashSet)
+
+public section
+
+
+namespace Lean.Compiler.LCNF.Extras
+/-!
+# Function arity reduction
+
+This module finds "used" parameters in a declaration, and then
+create an auxiliary declaration that contains only used parameters.
+For example:
+```
+def f (x y : Nat) : Nat :=
+  let _x.1 := Nat.add x x
+  let _x.2 := Nat.mul _x.1 _x.1
+  _x.2
+```
+is converted into
+```
+def f._redArg (x : Nat) : Nat :=
+  let _x.1 := Nat.add x x
+  let _x.2 := Nat.mul _x.1 _x.1
+  _x.2
+def f (x y : Nat) : Nat :=
+  let _x.1 := f._redArg x
+  _x.1
+```
+Note that any `f` full application is going to be inlined in the next `simp` pass.
+
+This module does not have similar support for mutual recursive applications.
+We assume this limitation is irrelevant in practice.
+
+New features:
+- handle recursive functions (cf https://github.com/leanprover/lean4/issues/13355)
+  ```
+    def ff (x y : Nat) : Nat := 
+      match x with
+      | 0 => 0 
+      | x + 1 => 1 + ff x (x + y)
+  ```
+- add track dependencies for cases with a single alternative
+- errors out if a parameter whose name starts with `GHOST` remains in a function
+whose name ends with Impl.
+-/
+
+namespace FindUsed
+
+/-- The context keeps track of the name and parameters of the function we are
+  compiling to handle recursive calls. -/
+structure Context where
+  decl : Decl .pure
+  params : FVarIdSet
+
+/-- Denotes that rhs depends on lhs. -/
+structure Equation where
+  lhs : Array FVarId
+  rhs : FVarId
+
+structure State where
+  decls : HashMap FVarId (Array FVarId)
+  equations : HashMap FVarId (Array FVarId)
+  facts : HashSet FVarId
+
+def printEqs (s : State) : CompilerM Unit := do
+  trace[Compiler.reduceArity2] "Before: "
+  trace[Compiler.reduceArity2] "Facts"
+  for f in s.facts do
+    trace[Compiler.reduceArity2] "{mkFVar f}"
+  trace[Compiler.reduceArity2] "Equations"
+  for (var, vars) in s.equations do
+    trace[Compiler.reduceArity2] "{mkFVar var} --> {vars.toList.map mkFVar}"
+
+abbrev FindUsedM := ReaderT Context <| StateRefT State CompilerM
+
+/-- Record the fact that `rhs` depends on `lhs` to the system. -/
+def emitEq (lhs : Array FVarId) (rhs : FVarId) : FindUsedM Unit :=
+  modify (fun s => {s with 
+    equations := s.equations.insert rhs <| (s.equations.getD rhs #[]) ++ lhs})
+
+/-- Record the fact that `rhs` is needed. -/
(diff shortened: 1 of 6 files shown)


Veir/IR/DeallocLemmas.lean
```diff
@@ -9,6 +9,7 @@ import Veir.IR.Fields
 import Veir.IR.GetSet
 import Veir.IR.Grind
 import Veir.IR.InBounds
+import ExArray.Ranges
```
math-fehr: Is there a reason we need this here?

Veir/IR/Basic.lean
```diff
@@ -1,7 +1,7 @@
 module
 
 import Std.Data.HashMap
-import Veir.Prelude
+public import Veir.Prelude
```
math-fehr, line 4: Why is this needed here?

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
```diff
@@ -0,0 +1,316 @@
+/-
+Copyright (c) 2022 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Leonardo de Moura, Leo Stefanesco
+-/
```
math-fehr, line 5: Why do we have this license thing here, but not in any other file?

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
```diff
@@ -0,0 +1,316 @@
...
+```
+is converted into
+```
+def f._rarg (x : Nat) : Nat :=
+  let _x.1 := Nat.add x x
+  let _x.2 := Nat.mul _x.1 _x.1
+  _x.2
+def f (x y : Nat) : Nat :=
+  let _x.1 := f._rarg x
+  _x.1
```
math-fehr:
```suggestion
def f._redArg (x : Nat) : Nat :=
  let _x.1 := Nat.add x x
  let _x.2 := Nat.mul _x.1 _x.1
  _x.2
def f (x y : Nat) : Nat :=
  let _x.1 := f._redArg x
  _x.1
```
from what I understand from the code?

ExArray/lean/ExArray/Ranges.lean
```diff
@@ -11,7 +11,7 @@ These are the only ranges we will use in this file
 abbrev CORange := Rco Nat
 
 @[grind=]
-theorem mem_range_nat (i: Nat) (r: CORange) : (i ∈ r) ↔ r.lower ≤ i ∧ i < r.upper := by
+theorem mem_range_nat' (i: Nat) (r: CORange) : (i ∈ r) ↔ r.lower ≤ i ∧ i < r.upper := by
```
math-fehr: I guess there was a name clash somewhere?

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
```diff
@@ -0,0 +1,316 @@
...
+whose names are of the form `GHOST...`. -/
+def checkParamsForGhosts (params : Array (Param .pure)) (decl : Decl .pure) : CompilerM Unit := do
+  let .str _ name := decl.name | return
+  unless name.endsWith "Impl" do return
+  for param in params do
+    trace[Compiler.reduceArity2] "vetting {param.binderName}"
+    if let .str _ s := param.binderName then
+      if s.startsWith "GHOST" then
+        logError  m!"Ghost parameter {param.binderName} has not been eliminated
+        in {decl.name}!"
```
math-fehr: This message will contain whitespaces as well right?

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
```diff
@@ -0,0 +1,316 @@
...
+New features:
+- handle recursive functions (cf https://github.com/leanprover/lean4/issues/13355)
+  ```
+    def ff (x y : Nat) : Nat := 
+      match x with
+      | 0 => 0 
+      | x + 1 => 1 + ff x (x + y)
+  ```
+- add track dependencies for cases with a single alternative
+- errors out if a paramter whose name starts with `GHOST` remains in a function
```
math-fehr:
```suggestion
- errors out if a parameter whose name starts with `GHOST` remains in a function
```

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
```diff
@@ -0,0 +1,316 @@
...
+    for (a, i) in args.zipIdx do
+      if used[i]! = true then
+        argsNew := argsNew.push a
+    return .jmp fvarId argsNew
+
+end ReduceArity
+
+open FindUsed ReduceArity Internalize
+
+/-- Emit an error in case a the function is called `...Impl` and still has arguments
```
math-fehr:
```suggestion
/-- Emit an error in case the function is called `...Impl` and still has arguments
```

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
```diff
@@ -0,0 +1,316 @@
...
+    unless declName == (← read).decl.name do return code.updateLet! decl (← reduce k)
+    let liveVars := (← read).liveVars
+    let params := (← read).decl.params
+    let mut argsNew := #[]
+    for param in params, arg in args do
+      if liveVars.contains param.fvarId then
+        argsNew := argsNew.push arg
+      -- over application
+      for arg in args[params.size...*] do
+        argsNew := argsNew.push arg
```
math-fehr: I don't claim I understand this code that much, but here you are having 3 nested for loops? Is there a reason why `arg` is iterated on twice in a nested way?

math-fehr (commented):
Nice! I added mostly style comments, otherwise  I don't claim I understand most of this code (only what it does)!

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
ineol: I guess since it's derived from Lean's source code we're supposed to keep this notice?

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
ineol: Indeed, it was a typo from the lean compiler :)

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
ineol: it was an indentation problem!

ExArray/lean/ExArray/Ranges.lean
ineol: It's a more general question: right now this new compiler pass is not used to compile veir, since it's not really needed. And this is to avoid a clash with a lemma in exarray. I don't know what's the best course of action now

Veir/IR/Basic.lean
ineol: this is to propagate the usage of the new compiler pass in case it's imported from prelude, but we're not doing it, but it's nice to be able to test easily...

ExArray/lean/ExArray/Ranges.lean
math-fehr: Maybe we can rename it to `CORange.mem_range_nat` to avoid the clash?

ExArray/lean/ExArray/CompilerExtras/ReduceArity.lean
math-fehr: Because this is a copy + fix of lean source code?

ineol: Yeah exactly

PR #882, opened 2026-06-16
fix(riscv): matching imm12 binops

regehr:
fix `imm12` binops matching according to canonicalization


luisacicolini:
lgtm, thank you!

luisacicolini (approved):

PR #883, opened 2026-06-16
feat(riscv): introduce slti and sltiu

regehr:
also add proofs of these. the proofs are for the specific rewrite, not for the code performing the rewrite, which must wait for us to build out some other infrastructure first.


luisacicolini (approved):
lgtm, thank you!

PR #884, opened 2026-06-16
feat: Add xDSL-style typing support for allocated registers

nchappe:
Only the x0-x31 names are supported for now.


luisacicolini (approved):
amazing, thank you!

PR #886, opened 2026-06-16
Feat: chacha20 c impl. incl. current mlir output

salinhkuhn:
This PR adds the test case for ChaCha20 (RFC 8439)  similar in spirit to fastntt.c. 
  The implementation is structured in three inlined layers ( quarter_round, chacha20_block, chacha20_xor ->  following RFC 8439 and cross-checked against the reference ph4r05/py-chacha20poly1305 chacha.py.


Test/Vcc/chacha20.c
```diff
@@ -0,0 +1,122 @@
+// RUN: ./Tools/vcc -O --emit-mlir %s -o %t.mlir
+// RUN: tbd 
+/*
+    An implementation of the ChaCha20 stream cipher (RFC 8439, see bellow).
```
luisacicolini:
```suggestion
    An implementation of the ChaCha20 stream cipher (RFC 8439, see below).
```

luisacicolini:
Thank you for the PR :) 

could you look at https://github.com/opencompl/veir/pull/735? 
Ideally, we would like to have a c file compiled with vcc _and_ add the result of this compilation (which should be LLVM IR) also to the LLVM test cases, so we can test it with veir-opt :) and if some ops are not parsed correctly please do open an issue!

Test/Vcc/chacha20.c
```diff
@@ -0,0 +1,122 @@
+// RUN: ./Tools/vcc -O --emit-mlir %s -o %t.mlir
+// RUN: tbd
```
luisacicolini: here we probably want to run the same vcc tests as the other two tests - then in the Test/LLVM folder we want to test the actual llvm ir to be parsed with veir-opt

Test/Vcc/chacha20.c
```diff
@@ -0,0 +1,122 @@
...
+
+    uint8_t ciphertext[114];
+    chacha20_xor(key, nonce, 1, plaintext, ciphertext, 114);
+
+    // Return the first four ciphertext bytes (so basically the start of the first block) which we will check against the RFC 8439 expected value (0x6e2e359a).
+    // Currently "restricted" to return only the first four bytes to obtain a 32bit value.
+    return ((int)ciphertext[0] << 24) | ((int)ciphertext[1] << 16) | ((int)ciphertext[2] << 8) | (int)ciphertext[3];
+}
+
+// CHECK: Program output: #[0x6e2e359a#32] (CURRENTLY NOT USED BUT WOULD BE THE VALUE TO CHECK AGAINST THE RFC EXPECTED OUTPUT WHEN INTERPRETATION WORKS)
```
luisacicolini: to check the interpretation we can also add a test in test/interpreter :)

Test/Vcc/chacha20.c
salinhkuhn: adressed (but output might need to optimised or me not running the entire encryption loop or us just accepting the longer code)

Test/Vcc/chacha20.c
```diff
@@ -0,0 +1,1038 @@
...
+        size_t chunk = len - done;
+        if (chunk > 64)
+            chunk = 64;
+        for (size_t i = 0; i < chunk; i++)
+            out[done+i] = in[done+i] ^ block[i];
+        done += chunk;
+    }
+}
+
+int main() {
```
luisacicolini: could you remove the main? I suspect it might be the reason we still have `alloca` in the mlir (similarly to what happened for fastntt). It is already enough to have an instruction-selected kernel for the algorithm (see #890 for reference)

luisacicolini:
Thank you! Left one more comment. After that is addressed, we should check whether we can instruction-select it (as we did in #890 for fastntt). Also, could you remove the `.md` empty file? :)

salinhkuhn:
> Thank you! Left one more comment. After that is addressed, we should check whether we can instruction-select it (as we did in #890 for fastntt). Also, could you remove the `.md` empty file? :)

done now, looks like we still have quite a few allocas and geps left. I assume it is because the algorithm uses local arrays. The build fails because this PR should be stacked on #911 .
@tobiasgrosser  @luisacicolini

luisacicolini (approved):
lgtm, thank you!

PR #888, opened 2026-06-16
fix(riscv): isel proofs

luisacicolini:
I found out that the instruction selection proofs were in fact not proven - the `contextual` enables `simp` to get some further lemmas from the hypotheses and solve them.


alexkeizer:
The more concerning part to me is that CI didn't flag the broken proofs. Is it deliberate that these files are not built in CI, or is this a bug?

tobiasgrosser:
> The more concerning part to me is that CI didn't flag the broken proofs. Is it deliberate that these files are not built in CI, or is this a bug?

We should make sure this file is build as part of `lake build`. Do we need to include this somewhere?

luisacicolini:
Added! My bad, I had forgotten to import them.

math-fehr (approved):

PR #889, opened 2026-06-17
feat(vcc): optionally run llvm's O2 passes

regehr:


tobiasgrosser (approved):
Nice. Maybe we should disable vectorisation in O2?

regehr:
> Nice. Maybe we should disable vectorisation in O2?

good catch! done.

luisacicolini (approved):

PR #895, opened 2026-06-18
fix(riscv): remove commutative match 

luisacicolini:
Because we now canonicalize constants to be on the rhs of arithmetic operations, we can simplify the combines based on the assumption that this canonicalization has happened.


tobiasgrosser (approved):
This is beautiful. Thank you for catching it and resolving it.

PR #898, opened 2026-06-19
feat(riscv): zext/sext/trunc proofs for legal types

luisacicolini:
We add the remaining correctness proofs for the isel patterns of sext/zext/trunc types for legal types within RV64 
and update the tests accordingly. We also check the condition on the relation between the width of the input and output types (e.g., for sext, the return type's width must be greater than the operand's). 

Legal extensions are: i16 -> i64, i16 -> i32, i32 -> i64, while truncation is always legal (as described in the [legalizer](https://github.com/llvm/llvm-project/blob/16a0a1042f7e4e5a0c667096fcdeb5803e06d120/llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp#L171-L179)).

The patterns are based on the inspection of [llc](https://github.com/opencompl/lean-mlir/pull/1981)'s behaviour.


tobiasgrosser:
Very cool.

I am curious, where did you derive these pattern from? Are these from LLVM? I would appreciate on paragraph in the PR message how `sext`/`zext` is lowered and why power-of-two types are the right `legal` types. Does the LLVM backend mark these specific types also as `legal`?

luisacicolini:
thank you for the review @tobiasgrosser, addressed

Veir/Passes/InstructionSelection/RISCV64.lean
```diff
@@ -11,6 +11,13 @@ namespace Veir
 
 /-! # Lowering Patterns -/
 
+/-- Extension operations (`sext`/`zext`) in RISC-V 64 are only legal
+  from `i16` to `i64`, from `i16` to `i32`, and from `i32` to `i64`.
+  See: https://github.com/llvm/llvm-project/blob/16a0a1042f7e4e5a0c667096fcdeb5803e06d120/llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp#L171-L179
+-/
+def isLegalExtOpWidth (w : Nat) : Bool :=
+  w = 16 ∨ w = 32
```
regehr, line 19: I don't understand the rationale for not supporting sext/zext from 8 bits -- there are RISC-V instructions for these operations, I think?

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -283,3 +283,147 @@ theorem orcb_refinement_y7 {z : LLVM.Int 64} :
           (Data.RISCV.orcb (LLVM.Int.toReg
             (Data.LLVM.Int.and z (LLVM.Int.constant 64 0x8080808080808080)))) 64 := by
   veir_bv_decide
+
+/--
+  Prove the correctness of the `zext` lowering pattern `i8` -> `i16`
```
regehr, line 288: totally up to you if you want to do it, but this code might be shorter if you define a generic [sz]ext_x_y theorem and then instantiate it many times

regehr (approved):
see comments, but LGTM!

Veir/Passes/InstructionSelection/RISCV64.lean
luisacicolini: afaiu from the [legalizer](https://github.com/llvm/llvm-project/blob/16a0a1042f7e4e5a0c667096fcdeb5803e06d120/llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp#L171-L179) they are not legal (?), am I misunderstanding?

tobiasgrosser:
Let's merge this for now. :-)

PR #900, opened 2026-06-20
feat: custom syntax a la TableGen

GZGavinZhao:
For CI and visibility...

### DIFF
--- a/VeirOpt.lean
+++ b/VeirOpt.lean
@@ -49,6 +49,8 @@ structure VeirOptArgs where
   passes : PassPipeline OpCode
   /-- Whether to accept ops/types/attrs from unregistered dialects. -/
   allowUnregisteredDialect : Bool
+  /-- Whether to print operations using their custom (pretty) assembly format. -/
+  pretty : Bool
 
 /--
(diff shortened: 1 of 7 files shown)


VeirOpt.lean
```diff
@@ -49,6 +49,8 @@ structure VeirOptArgs where
   passes : PassPipeline OpCode
   /-- Whether to accept ops/types/attrs from unregistered dialects. -/
   allowUnregisteredDialect : Bool
+  /-- Whether to print operations using their custom (pretty) assembly format. -/
+  pretty : Bool
```
tobiasgrosser, line 53: We probably want to make `--pretty` the default and use MLIR's `-mlir-print-op-generic` to print the generic syntax, no?

GZGavinZhao: Yes, in a separate PR I want to remove the `--pretty` the default and update all the tests to use the new syntax.

The current PR still makes printing the generic syntax the default because otherwise there would be a diff for every test file due to the new syntax. I imagine that'd cause massive conflicts for everyone else's PRs, so I'm deferring flipping that switch to a future PR.

PR #901, opened 2026-06-20
feat(riscv): select riscv.andn, riscv.orn, riscv.xnor

regehr:
a few little things to note about this:
- I noticed that the unrealized conversion casts were making the code annoying so I made some helpers to add those with less boilerplate
- since the LLVM dialect code has passed our verifier, checking only one of (input1, input2, result) for width 64 suffices
- I moved our existing `matchNot` function to a shared file. note that it already expected the literal constant to be on the RHS, even before we had a canonicalizer


luisacicolini (approved):
lgtm, thank you!

PR #903, opened 2026-06-21
feat(riscv): tighten up the riscv-dialect verifier for branch operations

regehr:
it was possible to crash the interpreter by parsing in MLIR containing malformed riscv-dialect branches. this tightens that up as much as possible

as a side effect, the riscv branch part of the verifier becomes more concise


tobiasgrosser (approved):
Nice.

luisacicolini:
this LGTM, maybe let's wait for @nchappe to take a look?

regehr:
(I should add that the reason I started to work on this is that I think I saw a bug that will trip over these rules, but I don't want to debug it by looking at interpreter crashes!)

nchappe (approved):

PR #904, opened 2026-06-22
fix(riscv): pick the right operand segment sizes for riscv_cf.bnez and riscv_cf.beqz

regehr:
also add some tests with asymmetical operandSegmentSizes, which is what was required to trigger the bug


nchappe (approved):

tobiasgrosser (approved):
Nice. Let's say what @nchappe says.

PR #905, opened 2026-06-22
fix(riscv): fix a crash during branch isel

regehr:
I think this might be the right fix, but I'm not at all sure. the problem is that `convertBranch` mutates the IR causing a data structure invalidation bug, causing `isel-br-riscv64` to crash with an OOB index, for the provided test case, which simply has two branches both going to the same BB


tobiasgrosser:
Thank you.

PR #907, opened 2026-06-22
feat(llvm): add min and max operations to the LLVM dialect

regehr:
I think (hope) this is pretty boring code!

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -692,6 +692,26 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
     if h: bw' ≠ bw then none else
     let rhs := rhs.cast (by simp at h; exact h)
     return (#[.int bw (LLVM.Int.xor lhs rhs)], mem, none)
+  | .intr__smax => do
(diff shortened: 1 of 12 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -692,6 +692,26 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : HasDialectOpInfo.proper
     if h: bw' ≠ bw then none else
     let rhs := rhs.cast (by simp at h; exact h)
     return (#[.int bw (LLVM.Int.xor lhs rhs)], mem, none)
+  | .intr__smax => do
```
luisacicolini, line 695: is there a reason for the double `__`? Otherwise, we usually just have one `_`

luisacicolini (approved):
left a small comment, other than that lgtm!

Veir/Interpreter/Basic.lean
regehr: this comes from `mkOpName` from `Veir/Meta/OpCode.lean`, it uses the double underscore itself, it looked like to work around some Lean issues with dots in identifiers, maybe -- I didn't totally get it but it's already there and we see it in `mlir__constant` for example in the code base already

PR #910, opened 2026-06-22
fix(riscv): some operands were out of order in riscv isel, fix this and put better tests on them

regehr:
sorry for the somewhat alarming tests here, but they do all fail without the corresponding code changes


luisacicolini (approved):
lgtm, thanks for catching this! The reason for this is that in the semantics we invert the ordering of operands, for consistency with the acual bits in the instruction, and so I thought I would swap the operands in the isel pass too, while in fact the interpreter swaps them already. Maybe we should just revert all orderings and remove the swapping in the first place.

regehr:
hi @luisacicolini do you want to take over this PR? I would say just keep the test cases and then do whatever looks right to the code. 

I think we should probably have similar side-by-side pre- and post-isel interpreter tests for all of the other instructions, not just these non-commutative ones

I had seen the operand-swapping and thought it looked good too! I only saw this when I started passing post-isel code to the interpreter

luisacicolini:
I think we can merge this and as you say have the same kind of testing for all ops - if the tests are fine I trust them. we might want to refactor this at some point but not a priority

PR #913, opened 2026-06-23
fix(riscv): icmp operands swapped

regehr:


tobiasgrosser (approved):
Let's see what @luisacicolini says.

PR #915, opened 2026-06-23
feat(riscv): select more zbb operations

regehr:
ok, sorry, this is a big PR, but it's pretty much just filling in the blanks between the LLVM and RISC-V levels, it's not that interesting and I've done a pretty good job testing, I hope.

even so, this PR has left out some stuff like the general case for funnel shift, which I hope we do not run into in our tests.

I am really enjoying tests like this one that insist that the interpreter result is the same before and after isel!
Test/Passes/InstructionSelection/RISCV64/fshl_exec.mlir

something else I did is introduce `.ll` files as something `lit` recognizes, so we can test the path all the way from LLVM IR -> RISC-V. later I will add some `llubi` tests to check that their semantics matches ours.

the proof parts are all vibe coded, sorry, hope they're OK


luisacicolini (approved):
lgtm, thank you! for the funnel shift proofs I suspect we could build some more library and automate it more - which would be ideal - but I am happy to merge this as it is, I can work on the proof refactoring in a subsequent PR.

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -482,3 +482,145 @@ theorem trunc_refinement_16_8 {x : LLVM.Int 16} :
...
+  `bv_decide`.
+-/
+
+/--
+  Prove the correctness of the `fshl` rotate lowering: a funnel shift with
+  identical data operands is a rotate-left, lowered to `rol`.
+-/
+theorem fshl_rol_refinement {a c : LLVM.Int 64} :
+    (Data.LLVM.Int.fshl a a c) ⊒
+      (RISCV.Reg.toInt (Data.RISCV.rol (LLVM.Int.toReg c) (LLVM.Int.toReg a)) 64) := by
```
tobiasgrosser, line 534: Should this not work via: veir_bv_decide? Also, for all remaining proofs in this file?

tobiasgrosser (approved):

Veir/Passes/InstructionSelection/Proofs.lean
regehr: ah, I had asked claude to do that and here's what it said:
```
  The funnel-shift rotate lowerings cannot use `veir_bv_decide` directly: its
  simp normalization rewrites the BitVec shift in the funnel-shift semantics into
  a *symbolic* `Nat` shift (`c.toNat % w`) on the `2*w`-bit concatenation, which
  `bv_decide` cannot bit-blast. Instead we reduce the refinement to a pure-BitVec
  goal by hand (discharging the poison cases from the non-poison hypothesis on the
  result, and keeping the shift amounts as `BitVec`s) and finish with bare
  `bv_decide`.
```

regehr: I did check that it was correct that veir_bv_decide can't do it

tobiasgrosser: Let's do this in a follow-up.

PR #917, opened 2026-06-24
fix(llvm): make veir_bv_decide more robust

tobiasgrosser:
Do not use an open simp-set, but instead explicitly specify the theorems that we need. This allows us to bitblast more rewrite rules.

Initially, we used:

> simp +contextual -failIfUnchanged [veir_bv_normalize_post, -BitVec.extractLsb_toNat]

which pulled in the default simp-set, which sometimes turns a bitvector expression into an expression over Nat. Even though we dropped one of the offending theorems, this was not enough and other theorems introduced non-bitblastable expressions. We now tighten it to:

> simp only +contextual -failIfUnchanged [veir_bv_normalize_post]

plus an explicit list of attributes added to `veir_bv_normalize` and `veir_bv_normalize_post`. This will make our simplifications more controllable and immediately allows us to use `veir_bv_normalize` on a set of new theorems that were not bitblastable without this improved preprocessing.

### DIFF
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -12,6 +12,19 @@ namespace Veir.Data.LLVM.Int
 
 public section
 
+attribute [veir_bv_normalize] Bool.false_eq_true false_and or_self decide_false dite_eq_ite Bool.if_false_right
+  Bool.decide_or Bool.decide_eq_true Bool.and_true Bool.or_eq_false_iff implies_true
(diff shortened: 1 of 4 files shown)


tobiasgrosser:
This cleans up some proofs introduced in #915.

regehr:
amazing!!
I need a little time to understand it though

regehr (approved):
ok I think I understand! LGTM

tobiasgrosser:
> ok I think I understand! LGTM

I also updated the commit message.

PR #919, opened 2026-06-24
fix(isel): preserve the block argument type

salinhkuhn:
This PR ensures block argument types are preserved through the branch lowering pass. Previously every block argument was cast back to i64 and that broke later ops that then mixed i64 and i32 operands. We now cast each argument back to its original type.


tobiasgrosser:
Hi @salinhkuhn, this is great. Can you add a test case?

salinhkuhn:
> Hi @salinhkuhn, this is great. Can you add a test case?

I pushed one similar to `br.mlir`

@luisacicolini

Test/Passes/InstructionSelection/RISCV64/br_block_arg_width.mlir
```diff
@@ -0,0 +1,22 @@
...
+"builtin.module"() ({
+    "func.func"() <{"function_type" = (i32, i64) -> (i32), "sym_name" = "block_arg"}> ({
+    ^bb0(%a: i32, %n: i64):
+        "llvm.br"(%a) [^bb1] : (i32) -> ()
+
+    ^bb1(%b: i32):
+        %c = "llvm.add"(%b, %b) : (i32, i32) -> i32
+        "func.return"(%c) : (i32) -> ()
+    }) : () -> ()
+}) : () -> ()
```
tobiasgrosser: Nice. Can you add this to `Test/Passes/InstructionSelection/RISCV64/br.mlir`?

tobiasgrosser (approved):
Otherwise, this looks good. Let's wait for a second 👍 from @luisacicolini or @regehr.

Test/Passes/InstructionSelection/RISCV64/br_block_arg_width.mlir
salinhkuhn: done

tobiasgrosser:
Nice. This LGTM.

PR #921, opened 2026-06-24
feat(llvm): support i32 for `constant` and arithmetic operations.

salinhkuhn:
This PR adds support for the llvm `i32` types needed in Chacha20.
Namely, it adds: i32 constant, add, lshr, or, shl, sub, xor, zext `i8` to `i32`, sext `i8` to `i64`  and their corresponding proof and registration into the rewriter.

Support for load and store will be added in a follow-up PR.

### DIFF
--- a/Veir/ForLean.lean
+++ b/Veir/ForLean.lean
@@ -424,3 +424,4 @@ namespace Rat
 def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
+
--- a/Veir/ForLean.lean
+++ b/Veir/ForLean.lean
@@ -2,7 +2,7 @@ module
 
 public import Std.Data.ExtHashSet
 public import Std.Data.HashMap
-public import Veir.Meta.BVNormalizeSimp -- registers the `veir_bv_normalize` simp attr
+public import Veir.Meta.BVNormalizeSimp
 import all Init.Data.Array.Basic -- unfold [Array.popWhile] in Array.getElem?_popWhile_of_false
 public section
 
--- a/Veir/ForLean.lean
+++ b/Veir/ForLean.lean
@@ -2,6 +2,7 @@ module
 
 public import Std.Data.ExtHashSet
 public import Std.Data.HashMap
+public import Veir.Meta.BVNormalizeSimp -- registers the `veir_bv_normalize` simp attr
 import all Init.Data.Array.Basic -- unfold [Array.popWhile] in Array.getElem?_popWhile_of_false
 public section
 
@@ -407,6 +408,14 @@ theorem umulOverflow_comm {w : Nat} (x y : BitVec w) :
     x.umulOverflow y = y.umulOverflow x := by
   grind [BitVec.umulOverflow]
 
+@[veir_bv_normalize]
+theorem setWidth_ofInt_32_64 (v : Int) :
+    BitVec.setWidth 32 (BitVec.ofInt 64 v) = BitVec.ofInt 32 v := by
+  rw [← BitVec.toInt_inj]
+  simp only [BitVec.toInt_setWidth, BitVec.toNat_ofInt, Nat.reducePow, Int.cast_ofNat_Int, Int.ofNat_toNat,
+    BitVec.toInt_ofInt]
+  grind
+
 end BitVec
 
 namespace Rat
@@ -415,4 +424,3 @@ namespace Rat
 def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
-
--- a/Veir/Passes/InstructionSelection/Proofs.lean
+++ b/Veir/Passes/InstructionSelection/Proofs.lean
@@ -32,6 +32,18 @@ theorem poisonConst_refinement :
   veir_bv_decide
 
 /--
+  Prove the correctness of the `constant` lowering pattern at i32, currently using grind tactic.
+-/
+theorem constant_refinement_32 {v : Int} :
+    (LLVM.Int.constant 32 v) ⊒ (RISCV.Reg.toInt (Data.RISCV.li (BitVec.ofInt 64 v)) 32) := by
+  have h : BitVec.setWidth 32 (BitVec.ofInt 64 v) = BitVec.ofInt 32 v := by
+      rw [← BitVec.toInt_inj]
+      simp only [BitVec.toInt_setWidth, BitVec.toNat_ofInt, Nat.reducePow, Int.cast_ofNat_Int, Int.ofNat_toNat,
+       BitVec.toInt_ofInt]
+      grind
+  veir_bv_decide
+
+/--
   Prove the correctness of the `add` lowering pattern.
 -/
 theorem add_refinement {x y : LLVM.Int 64} :
--- a/Veir/Passes/InstructionSelection/Proofs.lean
+++ b/Veir/Passes/InstructionSelection/Proofs.lean
@@ -40,6 +40,15 @@ theorem add_refinement {x y : LLVM.Int 64} :
   veir_bv_decide
 
 /--
+  Prove the correctness of the `add` lowering pattern at i32, which selects the
+  32-bit `addw`.
+-/
+theorem addw_refinement {x y : LLVM.Int 32} :
+    (Data.LLVM.Int.add x y) ⊒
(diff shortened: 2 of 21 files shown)


salinhkuhn:
@luisacicolini @tobiasgrosser any opinions on this ?

tobiasgrosser:
There is a conflict on this PR. Can you rebase to main?

tobiasgrosser (approved):
Otherwise, this LGTM.

regehr:
Sarah this looks great!
I think there's some risk here around the fact that instructions like `addw` perform sign extension instead of zero extension.
I think you'll want some test cases like this to make sure you're always doing the right thing:
```llvm
  define i64 @f1(i32 %x, i32 %y) {
    %z = add i32 %x, %y
    %r = sext i32 %z to i64
    ret i64 %r
  }

  define i64 @f2(i32 %x, i32 %y) {
    %z = add i32 %x, %y
    %r = zext i32 %z to i64
    ret i64 %r
  }

  define i1 @add32_slt_var(i32 %x, i32 %y, i32 %limit) {
    %z = add i32 %x, %y
    %c = icmp slt i32 %z, %limit
    ret i1 %c
  }

  define i1 @add32_ult_var(i32 %x, i32 %y, i32 %limit) {
    %z = add i32 %x, %y
    %c = icmp ult i32 %z, %limit
    ret i1 %c
  }
```

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -24,6 +24,18 @@ theorem constant_refinement {v : Int}:
...
+/--
+  Prove the correctness of the `constant` lowering pattern at i32 (similar to the `i64` case above), however currently using the grind tactic.
+-/
+theorem constant_refinement_32 {v : Int} :
+    (LLVM.Int.constant 32 v) ⊒ (RISCV.Reg.toInt (Data.RISCV.li (BitVec.ofInt 64 v)) 32) := by
+  have h : BitVec.setWidth 32 (BitVec.ofInt 64 v) = BitVec.ofInt 32 v := by
+      rw [← BitVec.toInt_inj]
+      simp only [BitVec.toInt_setWidth, BitVec.toNat_ofInt, Nat.reducePow, Int.cast_ofNat_Int, Int.ofNat_toNat,
+       BitVec.toInt_ofInt]
+      grind
```
tobiasgrosser: Why is this not decided via `veir_bv_decide`? In fact, you should probably double-check that you use `RISCV.Reg.toInt` and `LLVM.Int.toReg`.

Let's see what @luisacicolini says.

math-fehr:
I had to cancel the CI as one test is in an infinite loop, do you also have this locally?

Veir/Passes/InstructionSelection/Proofs.lean
```diff
@@ -31,6 +31,18 @@ theorem poisonConst_refinement :
     (LLVM.Int.poison (w := _)) ⊒ (RISCV.Reg.toInt (Data.RISCV.li (BitVec.ofInt 64 0)) 64) := by
   veir_bv_decide
 
+/--
+  Prove the correctness of the `constant` lowering pattern at i32 (similar to the `i64` case above), however currently using the grind tactic.
+-/
+theorem constant_refinement_32 {v : Int} :
+    (LLVM.Int.constant 32 v) ⊒ (RISCV.Reg.toInt (Data.RISCV.li (BitVec.ofInt 64 v)) 32) := by
+  have h : BitVec.setWidth 32 (BitVec.ofInt 64 v) = BitVec.ofInt 32 v := by
```
luisacicolini: can we have this as a separate lemma? Ideally, it should be added to [`ForLean`](https://github.com/opencompl/veir/blob/main/Veir/ForLean.lean), and tagged `veir_bv_normalize`.

Veir/Passes/InstructionSelection/Proofs.lean
luisacicolini: afaiu we need it because bv_decide gest stuck when thinking about `ofInt`. The lemma should belong in the `veir_bv_normalize` simp set imho.

regehr: I can't do better than what's here-- I'll move this to the other file as Luisa suggests and let someone smarter than me improve the proof!

Veir/Passes/InstructionSelection/Proofs.lean
regehr: ok, wow, trying to do this gets me in over my head in Lean-land. 
can I land this patch and leave making the lemma available to bv_decide to someone who's better at this?

Veir/Passes/InstructionSelection/RISCV64.lean
```diff
@@ -12,12 +12,12 @@ namespace Veir
...
 
-/-- Extension operations (`sext`/`zext`) in RISC-V 64 are only legal
-  from `i16` to `i64`, from `i16` to `i32`, and from `i32` to `i64`.
+/-- Extension operations (`sext`/`zext`) in RISC-V 64 are legal from `i8`, `i16`, and
+  `i32` source widths (`zext.b`/`zext.h`/`zext.w`, `sext.b`/`sext.h`/`sext.w`).
   See: https://github.com/llvm/llvm-project/blob/16a0a1042f7e4e5a0c667096fcdeb5803e06d120/llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp#L171-L179
 -/
 def isLegalExtOpWidth (w : Nat) : Bool :=
-  w = 16 ∨ w = 32
+  w = 8 ∨ w = 16 ∨ w = 32
```
luisacicolini, line 20: why do we need 8?

Veir/Passes/InstructionSelection/Proofs.lean
luisacicolini: sure, I can do that in the next PR

Veir/Passes/InstructionSelection/RISCV64.lean
salinhkuhn: chacha uses i8 integer for the state

Veir/Passes/InstructionSelection/Proofs.lean
salinhkuhn: @luisacicolini  I did it in a commit on this PR now, let me know what you think ?

salinhkuhn:
> Sarah this looks great! I think there's some risk here around the fact that instructions like `addw` perform sign extension instead of zero extension. I think you'll want some test cases like this to make sure you're always doing the right thing:
> 
> ```llvm
>   define i64 @f1(i32 %x, i32 %y) {
>     %z = add i32 %x, %y
>     %r = sext i32 %z to i64
>     ret i64 %r
>   }
> 
>   define i64 @f2(i32 %x, i32 %y) {
>     %z = add i32 %x, %y
>     %r = zext i32 %z to i64
>     ret i64 %r
>   }
> 
>   define i1 @add32_slt_var(i32 %x, i32 %y, i32 %limit) {
>     %z = add i32 %x, %y
>     %c = icmp slt i32 %z, %limit
>     ret i1 %c
>   }
> 
>   define i1 @add32_ult_var(i32 %x, i32 %y, i32 %limit) {
>     %z = add i32 %x, %y
>     %c = icmp ult i32 %z, %limit
>     ret i1 %c
>   }
> ```

@regehr Okay good idea, thanks. I added a test case in 'add_ext_exec.mlir' (similar to `bit_intrinsics_exec.mlir`), is that how you imagined it ? I don't support `icmp i32` at the moment since I focused on implementing what Chacha needs, so I only implemented the upper two cases.

Veir/ForLean.lean
```diff
@@ -2,6 +2,7 @@ module
 
 public import Std.Data.ExtHashSet
 public import Std.Data.HashMap
+public import Veir.Meta.BVNormalizeSimp -- registers the `veir_bv_normalize` simp attr
```
```suggestion
public import Veir.Meta.BVNormalizeSimp

File: Veir/ForLean.lean [hunk 5]
```diff
@@ -415,4 +424,3 @@ namespace Rat
 def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
-
```
luisacicolini, line 418: unrelated?

luisacicolini (approved):
left two more comments, once those are addressed it's ready to go :)

PR #925, opened 2026-06-25
chore(riscv): refactor introduction of immediate instructions to be part of isel instead of combines

regehr:
This comes from a discussion with Luisa about how to make our code mirror LLVM's backend structure more closely.

In this PR you can see references to specific parts of LLVM!

there's a change it here that looks unrelated, an NFC signature change to `matchAnd` in `Veir/Passes/Matching.lean`, but it is needed in order to give a uniform signature to a bunch of `match` function that I compose in `Veir/Passes/InstructionSelection/RISCV64Sdag.lean`. hopefully this is OK with folks!


luisacicolini (approved):
that's perfect - and it's great to have the references to llvm, thank you!

PR #926, opened 2026-06-26
chore(verifier): factor out some redundancy

regehr:
this is a big patch, but mechanical and NFC. the redundancy in the verifier was bothering me, and I think this'll make it more tractable to work with in the future.


tobiasgrosser:
That's great. 🙏

PR #928, opened 2026-06-26
feat(riscv): isel for `llvm.mlir.poison`

luisacicolini:
instruction selection of `llvm.mlir.poison` to `li 0`


regehr (approved):
this LGTM!

this code reminds me that I've been meaning to introduce an "is this an i64" helper, which will save 1 line in a large number of matching patterns

PR #930, opened 2026-06-26
fix(isel): poison instruction selection rule 

luisacicolini:
`llvm.mlir.poison` in LLVM is `i1`, not `i64`, and we previously could not lower it. I am therefore removing all checks on the width of the poison, such that it can always be lowered.

### DIFF
--- a/Test/Passes/InstructionSelection/RISCV64/poison.mlir
+++ b/Test/Passes/InstructionSelection/RISCV64/poison.mlir
@@ -2,11 +2,9 @@
 
 "builtin.module"() ({
     "func.func"()  <{function_type = () -> ()}> ({
-        %one = "llvm.mlir.poison"() : () -> i64
-        %two = "llvm.mlir.poison"() : () -> i32
+        %one = "llvm.mlir.poison"() : () -> i1
         // CHECK: [[A:%.*]] = "riscv.li"() <{"value" = 0 : i64}> : () -> !riscv.reg
-        // CHECK-NEXT: %{{.*}} = "builtin.unrealized_conversion_cast"([[A]]) : (!riscv.reg) -> i64
-        // CHECK-NEXT: %{{.*}} = "llvm.mlir.poison"() : () -> i32
+        // CHECK-NEXT: %{{.*}} = "builtin.unrealized_conversion_cast"([[A]]) : (!riscv.reg) -> i1
         "func.return"() : () -> ()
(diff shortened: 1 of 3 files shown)


Test/Passes/InstructionSelection/RISCV64/poison.mlir
```diff
@@ -2,11 +2,9 @@
 
 "builtin.module"() ({
     "func.func"()  <{function_type = () -> ()}> ({
-        %one = "llvm.mlir.poison"() : () -> i64
-        %two = "llvm.mlir.poison"() : () -> i32
+        %one = "llvm.mlir.poison"() : () -> i1
         // CHECK: [[A:%.*]] = "riscv.li"() <{"value" = 0 : i64}> : () -> !riscv.reg
-        // CHECK-NEXT: %{{.*}} = "builtin.unrealized_conversion_cast"([[A]]) : (!riscv.reg) -> i64
-        // CHECK-NEXT: %{{.*}} = "llvm.mlir.poison"() : () -> i32
+        // CHECK-NEXT: %{{.*}} = "builtin.unrealized_conversion_cast"([[A]]) : (!riscv.reg) -> i1
```
tobiasgrosser, line 7: Are we loosing test coverage here?

tobiasgrosser (approved):
Otherwise, this LGTM.

regehr (approved):
@luisacicolini I don't understand the comment that `llvm.mlir.poison` is `i1` -- it looks to me to be polymorphic in the sense that you can create a poison value of any type. and that seems to be what your code is saying here. I just don't understand what `i1` has to do with it?

but LGTM

luisacicolini:
@regehr I specified that because in all occurrences I ran into the type of `poison` was `i1` - if it's polymorphic we can just remove the comment.

Test/Passes/InstructionSelection/RISCV64/poison.mlir
luisacicolini: no, the previous lowering was too restrictive

luisacicolini:
merged, I can't properly run our eval without this commit.

tobiasgrosser:
You should still add test cases where poison is i64 and i32. :-) It is indeed polymorphic.

luisacicolini:
@tobiasgrosser added here: https://github.com/opencompl/veir/pull/938

PR #932, opened 2026-06-26
fix(sqlite): parse zeroed array initializers

### LINKED ISSUE #779: sqlite blocker: zeroed array initializers
below is a minimal reproducer for an MLIR feature that we're not supporting. I have no idea how many VeIR-blockers are in sqlite but this is the first one that `vcc` runs into

```
~/Downloads/sqlite-amalgamation-3530200$ cat repro.ll             
@g = global { i32, [4 x i8] } { i32 1, [4 x i8] zeroinitializer }
~/Downloads/sqlite-amalgamation-3530200$ ~/veir/Tools/vcc repro.ll
vcc: lake failed:
/var/folders/dl/7j43h1fx0pdczx6mx5cjxsp80000gn/T/tmphhrmr4f9/repro.mlir:4:43: error: attribute expected
    %1 = "llvm.mlir.constant"() <{value = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
                                          ^
```

sishtiaq:
Parse dense<0>:tensor<4xi8> as UnregisteredAttrBody for now. 
This fixes #779.

### DIFF
--- a/Test/Builtin/parse_zeroed_array_initializers.mlir
+++ b/Test/Builtin/parse_zeroed_array_initializers.mlir
@@ -1,0 +1,8 @@
+// RUN: VEIR_ROUNDTRIP
+// RUN: MLIR_ROUNDTRIP
+
+"builtin.module"() ({
+    %1 = "llvm.mlir.constant"() <{value = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
+    // CHECK: "llvm.mlir.constant"() <{"value" = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
+}) : () -> ()
+
--- a/Veir/Parser/AttrParser.lean
+++ b/Veir/Parser/AttrParser.lean
@@ -667,6 +667,30 @@ partial def parseType (errorMsg : String := "type expected") : AttrParserM TypeA
   | none => throwAtCurrentPos errorMsg
 
 /--
+  Parse a dense elements attribute, if present.
+  Its syntax is `dense<value> : type`, e.g. `dense<0> : tensor<4xi8>`.
+  Both the value body and the type are stored as raw strings pending full
+  tensor/vector type support.
+-/
+partial def parseOptionalDenseElementsAttr : AttrParserM (Option DenseElementsAttr) := do
+  if !(← parseOptionalKeyword "dense".toByteArray) then
+    return none
+  parsePunctuation "<"
+  let value ← parseUnregisteredAttrBody
+  parsePunctuation ">"
+  parsePunctuation ":"
+  -- tensor<4xi8> is an unregistered type: capture it as a balanced string
+  -- using the same scheme as parseOptionalDialectAttr's fallback case.
+  let typeStartPos ← getPos
+  parseKeyword "tensor".toByteArray
+  parsePunctuation "<"
+  let _ ← parseUnregisteredAttrBody
+  let typeEndPos := (← peekToken).slice.stop
+  parsePunctuation ">"
+  let typeStr := (Slice.mk typeStartPos typeEndPos).of (← getThe ParserState).input
+  return some (DenseElementsAttr.mk value (String.fromUTF8! typeStr))
+
+/--
   Parse an entry in an attribute dictionary, which has the form `name = value`
   or the shorthand `name` (equivalent to `name = unit`).
 -/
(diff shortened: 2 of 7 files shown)


Veir/Parser/AttrParser.lean
```diff
@@ -666,6 +666,30 @@ partial def parseType (errorMsg : String := "type expected") : AttrParserM TypeA
...
+  if !(← parseOptionalKeyword "dense".toByteArray) then
+    return none
+  parsePunctuation "<"
+  let value ← parseUnregisteredAttrBody
+  parsePunctuation ">"
+  parsePunctuation ":"
+  -- tensor<4xi8> is an unregistered type: capture it as a balanced string
+  -- using the same scheme as parseOptionalDialectAttr's fallback case.
+  let typeStartPos ← getPos
+  parseKeyword "tensor".toByteArray
```
Chobbes, line 685: Couldn't this be vector too?

Test/LLVM/parse_zeroed_array_initializers.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: VEIR_ROUNDTRIP
+
+"builtin.module"() ({
+    %1 = "llvm.mlir.constant"() <{value = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
+    // CHECK:     "llvm.mlir.constant"() <{"value" = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
```
tobiasgrosser:
```suggestion
    // CHECK: "llvm.mlir.constant"() <{"value" = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
```

tobiasgrosser (commented):
Nice!

There is a `.DS_Store` file that should be removed.

There are also two more TODOs to resolve. Otherwise, this looks good. 

We probably want to have feedback from @math-fehr, too.

Chobbes (commented):
This is looking good to me. I did have a question about this being fixed to tensor and not vector as well?

Test/Builtin/parse_zeroed_array_initializers.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: VEIR_ROUNDTRIP
+
+"builtin.module"() ({
+    %1 = "llvm.mlir.constant"() <{value = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
+    // CHECK:     "llvm.mlir.constant"() <{"value" = dense<0> : tensor<4xi8>}> : () -> !llvm.array<4 x i8>
+}) : () -> ()
+
```
tobiasgrosser, line 8: Also, I do not think we want our test cases in application-specific folders? Can we add this to `Test/LLVM/*`?

regehr (commented):
please remove `Veir/.DS_Store` from the PR!

tobiasgrosser:
@sishtiaq, there are still two TODO's in the code. Are you still working on this or do you propose to merge with these two todos. In case of the latter, can you add a comment explaining your plan.

Veir/Parser/AttrParser.lean
sishtiaq: `tensor` for now as that's what sqlite3.c throws out. But good point, leaving as an issue for later: https://github.com/opencompl/veir/issues/940.

sishtiaq:
> @sishtiaq, there are still two TODO's in the code. Are you still working on this or do you propose to merge with these two todos. In case of the latter, can you add a comment explaining your plan.

I've removed the TODOs from the code. Instead, there are now issues https://github.com/opencompl/veir/issues/941 and https://github.com/opencompl/veir/issues/942 for them instead.

tobiasgrosser (approved):
@math-fehr, can you have a look?

Test/Builtin/parse_zeroed_array_initializers.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: VEIR_ROUNDTRIP
```
math-fehr, line 1: I believe we should put this in `Test/Builtin` rather than `Test/LLVM` (`dense` is in the builtin dialect technically).

Also, can you add `MLIR_ROUNDTRIP` as well in a second line? So we can check against MLIR.

math-fehr (approved):
All good to me besides where the test file should be, I can merge this after it's moved!

Test/Builtin/parse_zeroed_array_initializers.mlir
sishtiaq: Moved and added.

PR #935, opened 2026-06-26
feat(verifier): generalize OperationPtr.Verified.arith_addi for other integer binops

regehr:
@math-fehr there's no rush on this at all, but when you have time I'd like to get a bit of feedback whether I'm on the right track here-- just a step towards the results we'll need in order to discharge the big top-level sorry in the verifier


math-fehr (approved):
I think that makes sense to me, thanks!

PR #936, opened 2026-06-27
feat(rewriter): RAUW + erase is safe

regehr:
Mathieu if you have time to give me feedback on this I would appreciate it.
I was going to do more work on Verifier.lean but it seems like maybe starting here is more useful, and then I can go back to the verifier lemmas once I know more about what exactly they must provide


Veir/PatternRewriter/Basic.lean
```diff
@@ -215,6 +215,29 @@ def replaceValue (rewriter: PatternRewriter OpInfo) (oldVal newVal: ValuePtr)
...
+    (hop : op.InBounds rewriter.ctx.raw)
+    (hres : (op.getResult 0 : ValuePtr).InBounds rewriter.ctx.raw)
+    (hnew : newValue.InBounds rewriter.ctx.raw)
+    (hne : (op.getResult 0 : ValuePtr) ≠ newValue)
+    (hone : op.getNumResults! rewriter.ctx.raw = 1)
+    (hregions : op.getNumRegions! rewriter.ctx.raw = 0) :
+    ∃ rewriter',
+      (rewriter.replaceValue (op.getResult 0 : ValuePtr) newValue hne hres hnew).eraseOp op
+        = some rewriter' :=
+  ⟨_, rfl⟩
```
math-fehr: Can you move these lemmas to a new file `PatternRewriter/Lemmas.lean` (unless one exist already?)

math-fehr: Well, I realize from your theorem that my definition of `eraseOp` returned an option for no reasons ^^'
`eraseOp` should always succeed, and your proof is essentially not even using your hypotheses to be proven.
I think we should then rather have a PR that removes the Option from `eraseOp` here, and that should make things easier!

Veir/Rewriter/WfRewriter/GetSet.lean
```diff
@@ -875,6 +876,55 @@ theorem OperationPtr.getResultTypes!_WfRewriter_replaceValue :
...
+  fun_induction WfRewriter.replaceValue <;>
+    grind [Id.run, ValuePtr.hasUses!_def, ValuePtr.getFirstUse!_eq_getFirstUse]
+
+/-- After replacing the (sole) result of `op` with another value, `op` no longer has any uses. -/
+@[grind =]
+theorem OperationPtr.hasUses!_WfRewriter_replaceValue_getResult0
+    {op : OperationPtr} {newValue : ValuePtr}
+    {hres : (op.getResult 0 : ValuePtr).InBounds ctx.raw}
+    {hnew : newValue.InBounds ctx.raw}
+    {hne : (op.getResult 0 : ValuePtr) ≠ newValue}
```
math-fehr: I believe you can just remove these parameters, as they can be inferred from the goal. That should make this a bit nicer to read

Veir/Rewriter/WfRewriter/GetSet.lean
```diff
@@ -875,6 +876,55 @@ theorem OperationPtr.getResultTypes!_WfRewriter_replaceValue :
...
+    (hres : (op.getResult 0 : ValuePtr).InBounds ctx.raw)
+    (hnew : newValue.InBounds ctx.raw)
+    (hne : (op.getResult 0 : ValuePtr) ≠ newValue)
+    (hone : op.getNumResults! ctx.raw = 1)
+    (hregions : op.getNumRegions! ctx.raw = 0) :
+    let ctx' :=
+      WfRewriter.replaceValue ctx (op.getResult 0 : ValuePtr) newValue hne hres hnew
+    op.InBounds ctx'.raw ∧
+    op.getNumRegions! ctx'.raw = 0 ∧
+    op.hasUses! ctx'.raw = false := by
```
math-fehr: I'm not sure we overall need this theorem?
Mostly because each component is already a simp/grind lemma?

Veir/Rewriter/WfRewriter/GetSet.lean
```diff
@@ -875,6 +876,55 @@ theorem OperationPtr.getResultTypes!_WfRewriter_replaceValue :
...
+      (WfRewriter.replaceValue ctx (op.getResult 0) newValue hne hres hnew).raw = false := by
+  rw [OperationPtr.hasUses!_eq_false_iff_hasUses!_getResult_eq_false]
+  intro index hindex
+  have hnum : op.getNumResults!
+      (WfRewriter.replaceValue ctx (op.getResult 0) newValue hne hres hnew).raw = 1 := by
+    rw [OperationPtr.getNumResults!_WfRewriter_replaceValue]; exact hone
+  have hidx0 : index = 0 := by omega
+  subst hidx0
+  exact ValuePtr.hasUses!_WfRewriter_replaceValue_oldValue
+
```
math-fehr: Can you add these to a new `Rewriter/WfRewriter/Lemmas.lean`?
I'm trying to have `GetSet.lean` being very mechanical, and this is slightly outside of its scope!

math-fehr (commented):
Thanks! I left some comments, but otherwise I feel that's going to be quite useful.
also, as I mentionned in the comments, I feel we should just fix `eraseOp`!

Veir/PatternRewriter/Basic.lean
regehr: this became totally unneeded after other changes you made in July!

regehr:
hi @math-fehr -- sorry for the long delay getting back to this one.

I've tried to address all of your concerns, and then I've also tried to push this result a little bit further than I had before. the concrete accomplishment now is that this comment is proved rather than assumed:
```
-- All results of `op` have been replaced above, so `op` is dead and can be erased.
```

PR #938, opened 2026-06-28
fix(isel): polymorphism of `poison`

luisacicolini:
Test lowering of poison constant with `i32` and `i64` type and fix the matching that yielded an index out of bound error whenever invoked.


tobiasgrosser (approved):
Perfect!

PR #944, opened 2026-06-28
feat(verifier): check dominance

regehr:
ok this is a substantial tooling up of our verifier. the patch got bigger and messier than I wanted, but I think it's all good changes.
the main thing this does is attempts to make our verifier return the same verdict as MLIR's verifier, for a variety of interesting cases. so there are lots of new tests! and also there's a new test predicate `VEIR_MLIR_SAME_VERDICT` that insists that either mlir-opt and veir-opt both accept or both reject a test case. 
I had to change a bunch of unrelated test cases-- these are ones that are rejected by upstream MLIR, and now are rejected by Veir.
the other goal of this patch is to hopefully set us up to de-axiomitize `Veir/Dominance.lean`. our goal is that this verifier acts as a dynamic checker for freshly parsed code, and after that all transformations must prove that their output is verified if their input is.


Veir/IR/Dominance.lean
axelcool1234, line 154: Probably mention graph region semantics here? Similar to:
https://mlir.llvm.org/doxygen/classmlir_1_1DominanceInfo.html#a9b5b6aadbdcfc3be72d934d7d0432602

Veir/IR/Dominance.lean
```diff
@@ -144,7 +174,7 @@ def immediateDominator?
 /--
 Dominance query between two blocks, where a block dominates itself.
 -/
-def dominates
+def dominatesByAnalysis
```
axelcool1234, line 189: Why the renames? I think the previous names are fine still

axelcool1234 (commented):
My first skim of this PR. Will give another review round when I'm not busy :)

Veir/IR/Dominance.lean
regehr: name collision between your version of dominance and @math-fehr's!

axelcool1234: > name collision between your version of dominance and @math-fehr's!

Ah, right.

axelcool1234:
For the dominance analysis changes, LGTM :)

Test/Verifier/dominance_back_edge.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+// RUN: VEIR_MLIR_SAME_VERDICT
```
math-fehr: Gavin added `MLIR_INVALID`, maybe we should use that instead of adding `VEIR_MLIR_SAME_VERDICT`?

Test/Verifier/dominance_graph_intra_ok.mlir
```diff
@@ -0,0 +1,22 @@
...
+"builtin.module"() ({
+  "func.func"() <{function_type = () -> (), sym_name = "m"}> ({
+  ^entry():
+    %v = "test.test"() : () -> i64
+    "test.test"() ({
+    ^g0():
+      %a = "test.test"() : () -> i64
+    ^g1():
+      "test.test"(%a, %v) : (i64, i64) -> ()
+    }) : () -> ()
```
math-fehr, line 24: I do not understand why this test actually passes?
You mention in the comment that `test.test` is a graph region, but in that case graph regions only have a single block!
In general, any block that is part of a multi-block region should have a terminator

Test/Verifier/dominance_graph_region_ok.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: veir-opt %s | filecheck %s
+// RUN: VEIR_MLIR_SAME_VERDICT
+
+// Dominance is only required in SSACFG regions. The body of an unregistered op
+// is a graph region, where a value defined in one block may be used in another
```
math-fehr: It is not actually true that the body of an unregistered op is a graph region.
Because graph regions have only a single block, this means that whenever any operation has a region with multiple blocks, the region needs to be a SSACFG region.

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -34,11 +52,14 @@ private def compareExpectedDominator
     (irCtx : IRContext OpCode) : MismatchReport := Id.run do
   let some expectedBlock := recovered.blocks[expectedDom]?
     | return #[s!"dominators {expected.name}: missing block label {expectedDom}"]
-  let shouldProperlyDom := expectedDom ≠ expected.name
+  -- In a graph region every dominator (including the block itself) properly
+  -- dominates the block; in an SSACFG region a block does not properly dominate
+  -- itself.
```
math-fehr, line 61: Where does this rule come from?
My understanding is that "properly dominates" means dominates and not equal?

Veir/IR/Dominance.lean
```diff
@@ -38,10 +52,14 @@ private partial def BlockPtr.dominatesWithinRegion
...
-    let some idom := block.getIDom? dfCtx irCtx | return false
-    idom ≠ block && dominatesWithinRegion dominator idom dfCtx irCtx
+    return true
+  -- In a graph region there is no block ordering, so every block dominates every
+  -- other block in the region.
+  if let some region := (block.get! irCtx).parent then
+    if !region.hasSSADominanceByKind irCtx then
+      return true
+  let some idom := block.getIDom? dfCtx irCtx | return false
+  return idom ≠ block && dominatesWithinRegion dominator idom dfCtx irCtx
```
math-fehr, line 67: There can only be a single block in a graph region, so this function should probably be unchanged?

Veir/IR/Dominance.lean
```diff
@@ -116,14 +138,29 @@ private def dominates
...
     (dfCtx : DataFlowContext)
     (irCtx : IRContext OpCode) : Bool :=
-  dominator ≠ point && dominator.dominates point dfCtx irCtx
+  if dominator = point then
+    match dominator.block! irCtx with
+    | some block =>
+      match (block.get! irCtx).parent with
+      | some region => !region.hasSSADominanceByKind irCtx
+      | none => false
+    | none => false
```
math-fehr: Should this code be factored somewhere? You also use the exact same one earlier

Veir/Verifier.lean
```diff
@@ -953,14 +954,125 @@ def RegionPtr.getRegionKind (region : RegionPtr) (ctx : WfIRContext OpCode) : Re
...
+  MLIR's generic verifier and catches malformed successor-bearing unknown/test
+  ops even when they are not registered terminators.
 -/
 def OperationPtr.verifyTerminatorPosition (op : OperationPtr) (ctx : WfIRContext OpCode)
     (opIn : op.InBounds ctx.raw) : Except String PUnit := do
   let operation := op.get ctx.raw opIn
   if operation.opType.isTerminator && operation.next.isSome then
     throw "Expected a terminator to be the last operation of its block"
+  if op.getNumSuccessors ctx.raw opIn ≠ 0 && operation.next.isSome then
+    throw "operation with block successors must terminate its parent block"
```
math-fehr, line 74: Only terminators can have successors. So maybe there is a stronger check to be made somewhere about this?

Veir/Verifier.lean
```diff
@@ -953,14 +954,125 @@ def RegionPtr.getRegionKind (region : RegionPtr) (ctx : WfIRContext OpCode) : Re
...
+def OperationPtr.verifyOperandIsolation (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) : Except String PUnit := do
+  match (op.get ctx.raw opIn).parent with
+  | none => pure ()
+  | some useBlock =>
+    match (useBlock.get! ctx.raw).parent with
+    | none => pure ()
+    | some useRegion =>
+      let instrName := String.fromUTF8! (op.getOpType ctx.raw opIn).name
+      for i in [0:op.getNumOperands ctx.raw opIn] do
```
math-fehr, line 129:
```suggestion
      for i in 0...op.getNumOperands ctx.raw opIn do
```
I believe this is the standard notation for ranges (unless it is ..).
This avoids allocating a list

Veir/Verifier.lean
```diff
@@ -975,25 +1087,135 @@ def BlockPtr.verifyTerminator (block : BlockPtr) (ctx : WfIRContext OpCode)
...
+  -- (`mlir/lib/IR/Verifier.cpp`): the region's entry block is always reachable,
+  -- and every other block is reachable iff the region's dominator tree says so.
+  -- A block carries a dominator fact in exactly those cases, so fact presence is
+  -- the executable form of `isReachableFromEntry`.
+  --
+  -- A consequence, shared with MLIR, is that a use hidden in a *non-entry* block
+  -- of a graph region with no intra-region control-flow edges is unreachable and
+  -- therefore not checked, even if it captures a value that does not dominate the
+  -- graph-owning operation. (A capture used in the graph region's *entry* block
+  -- is still checked, because the entry block is always reachable.)
```
math-fehr: I would argue that if an operation is unreachable, then its operands are dominated by the value definitions. This is because dominance is defined as "all path that goes to A goes first to B", but here since there is no path, it is trivially correct!

This has the same effect (that we don't need to check unreachable blocks), but the reasoning is different (and will be reflected in the proof).

Veir/Verifier.lean
```diff
@@ -975,25 +1087,135 @@ def BlockPtr.verifyTerminator (block : BlockPtr) (ctx : WfIRContext OpCode)
...
+  that each step lines up with a clause of the definition, which is what makes the
+  implication "the check succeeds ⇒ `ctx.Dom`" clear.
+-/
+def WfIRContext.verifyDominance
+    (ctx : WfIRContext OpCode) (top : OperationPtr) : Except String Unit := do
+  if _ : top.InBounds ctx.raw then
+    let some dfCtx := fixpointSolve top #[DominanceAnalysis] ctx.raw
+      | throw "dominance analysis did not terminate"
+    ctx.raw.forOpsDepM fun op opIn =>
+      op.verifyOperandDominance ctx dfCtx opIn
```
math-fehr: This will actually break if we have any operation that wasn't deallocated, nor inserted in the module.
We can keep it as-is here, but we'll have to fix it at some point (hopefully after I add walkers to walk the IR a bit more nicely).

math-fehr (commented):
Very cool! I added a few comments, but otherwise the design seems great!
Quick question, it seems you are saying graph regions can sometimes have multiple blocks (like in a unregistered operation). Is this something from MLIR?

Also, for next time, it would probably have been easier for me if some parts of the PR (like the additions of verifiers about terminators, not necessarily dominance) would have been split. But no need for this time since I reviewed it!

Veir/Verifier.lean
axelcool1234: Woah I had no idea this syntax was available!

Veir/Verifier.lean
axelcool1234: Is this guaranteed considering we use verified ops and blocks?

UnitTest/DataFlowFramework/Dominance.lean
axelcool1234: @regehr A better comment would bring up the fact the graph region is one block, similar to this description for proper dominance. I believe this also answers your question @math-fehr :
https://mlir.llvm.org/doxygen/classmlir_1_1DominanceInfo.html#af425eb9134e887aa56ec1169819d261e

regehr:
> Also, for next time, it would probably have been easier for me if some parts of the PR (like the additions of verifiers about terminators, not necessarily dominance) would have been split. But no need for this time since I reviewed it!

yeah, sorry, scope creep. I'll try not to do that again.

regehr:
> Very cool! I added a few comments, but otherwise the design seems great! Quick question, it seems you are saying graph regions can sometimes have multiple blocks (like in a unregistered operation). Is this something from MLIR?

yes, this is permitted in MLIR, but it's kind of weird.
see for example `mlir/lib/IR/RegionKindInterface.cpp` which makes `mlir::mayBeGraphRegion()` return true for every unregistered operation, causing dominance checking to be skipped. it's sort of weird mess. did I answer your question?

PR #945, opened 2026-06-28
feat(riscv): lower post-isel risc-v dialect code to MIR and use LLVM to regalloc and generate asm

regehr:
I had intended this as a proof of concept that we would not merge, but it ends up that it gives us a very powerful oracle for external validation of our compiler: the `spike` RISC-V simulator. I don't want to leave this test oracle laying on the ground, so I suggest that we merge this even if we don't mention it in our paper submission.

this is all vibe-coded, but the vibe code lives in separate files and we can simply delete this once we have working regalloc, BB ordering, etc. in Veir.


tobiasgrosser (approved):
I love it. Especially, as we can then reuse LLVM's register allocator, which is both more generic and avoids any differences in the register allocation algorithm. Very cool.

luisacicolini (approved):
lgtm, thanks a lot. It's super helpful to have this!

PR #946, opened 2026-06-29
feat(printer): declarative assembly-format DSL + custom (pretty) printing

GZGavinZhao:
Add a TableGen-style declarative assembly-format DSL for printing operations in their custom (pretty) syntax, currently gated behind the `--pretty` flag.

Currently three examples are given:
- `func.return` and `func.call` using the declarative syntax, mirroring the TableGen's declarative syntax.
- `func.func` uses a dedicated printer hook, mirroring MLIR's `hasCustomAssemblyFormat`.

Parsing of pretty syntax is added in a follow-up; this PR only changes printing, so the default (generic) output and all existing tests are unaffected.

~~TBD: document where exactly our format is different from the upstream TableGen syntax.~~ Documented in the header comment of `Veir/AssemblyFormat.lean`. Almost all of the difference comes from VeIR not having named operands/results for operations. To achieve true 1-1 mapping, we need some form of build-time (ideally run-time as well) map from an operation's field names to index/type.

### DIFF
--- a/Test/Func/pretty_print.mlir
+++ b/Test/Func/pretty_print.mlir
@@ -1,0 +1,23 @@
+// Custom (pretty) printing of the func dialect via the declarative assembly
+// format. The input is in generic form; with --pretty the func ops are printed
+// in their custom syntax.
+
+// RUN: veir-opt --pretty %s | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{function_type = (i32) -> i32, sym_name = "callee"}> ({
+  ^bb0(%arg0: i32):
+    "func.return"(%arg0) : (i32) -> ()
+  }) : () -> ()
+  "func.func"() <{function_type = (i32) -> i32, sym_name = "caller"}> ({
+  ^bb0(%arg0: i32):
+    %0 = "func.call"(%arg0) <{callee = @callee}> : (i32) -> i32
+    "func.return"(%0) : (i32) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK:      func.func @callee(%arg{{[0-9]+_[0-9]+}}: i32) -> i32 {
+// CHECK-NEXT:   func.return %arg{{[0-9]+_[0-9]+}} : i32
+// CHECK:      func.func @caller(%arg{{[0-9]+_[0-9]+}}: i32) -> i32 {
+// CHECK:        %{{[0-9]+}} = func.call @callee(%arg{{[0-9]+_[0-9]+}}) : (i32) -> i32
+// CHECK-NEXT:   func.return %{{[0-9]+}} : i32
--- a/Test/Func/pretty_print.mlir
+++ b/Test/Func/pretty_print.mlir
@@ -1,7 +1,7 @@
-// Custom (pretty) printing of `func.func` via the `HasCustomPrinting`
-// interface. The input is in generic form; with --pretty the `func.func` ops
-// are printed in their custom syntax (mirroring MLIR's hasCustomAssemblyFormat)
-// while everything else keeps the generic form.
+// Custom (pretty) printing via the `HasCustomPrinting` interface. The input is
+// in generic form; with --pretty, `func.func` uses its hand-written printer
+// (mirroring MLIR's hasCustomAssemblyFormat) and `func.return`/`func.call` use
+// their declarative assembly formats; everything else keeps the generic form.
 
 // RUN: veir-opt --pretty %s | filecheck %s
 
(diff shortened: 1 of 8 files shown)


tobiasgrosser (approved):
Nice. Let's see what @math-fehr says.

AlexanderViand (commented):
This looks really cool 👍 

One (hopefully small) request: could you add an `MLIR_ROUNDTRIP`-style test that checks that the pretty-printed syntax veir-opt produces matches the real mlir-opt pretty print (i.e., each one can parse what the other prints, and the textual representation stays stable in the process)?

EDIT: Actually, I just tried the mlir->veir direction manually (`mlir-opt Test/Func/pretty_syntax.mlir | lake exe veir-opt`) and I think in order to make this work you'd also need to add pretty-printing for `builtin.module`:
```
<stdin>:1:1: error: operation expected
module {
^
```

So maybe this is more of a follow-up than something to deal with in this PR

GZGavinZhao:
As I mentioned in the last meeting, I would prefer having full delcarative ODS/TableGen custom syntax support rather than implementing a large-in-syntax-coverage-but-small-in-usage-coverage subset. With this current subset, most of the custom syntax in MLIR we cannot just copy over; they need to be either rewritten to fit the subset of custom syntax that we support (e.g. `func.call` implemented here) or straight up un-expressable (e.g. `cf.cond_br`). I have an idea of how to do this, will submit a draft PR early next week.

math-fehr:
Wait, so do you mean that you don't think you can iteratively add feature to that particular DSL such that you could cover all of TableGen DSL?
Because I believe that the long term solution is to have a DSL, plus an escape hatch in Lean code. I don't think we want to prevent escape hatches at all.

GZGavinZhao:
I'll resolve the conflicts and clean it up later today before requesting a review.

AlexanderViand:
> I'll resolve the conflicts and clean it up later today before requesting a review.

@GZGavinZhao Thanks! Do you consider this ready to review now?

math-fehr (commented):
Do you think you could split this PR further? Currently it introduce both special printing for `func.func`, and the assembly-format DSL. I think they should both be separate PRs.

Also, I'm wondering if we can make this somehow more separate from the main `Printer.lean` file. Could you design this so that there is:
1) An entry point outside of the `Printer.lean` file such that we can register the printing function? For instance just a parameter of `Printer.lean` that contains a map between operations to printing functions. There might be some difficulties because of partial fixpoints, let's see.
2) Then on top of this build the assembly format DSL, so that it generates the function that can then be passed to the printer.

I think that way we have a solid fundation!

GZGavinZhao:
Sorry for the delay. This idea sounds good! I'll split out the code so that registering non-generic printing for operations is a separate file. Then, completely hand-written custom printing and `assemblyFormat` custom printing all go through this non-generic printing code path.

GZGavinZhao:
Over the weekend I'll rebase this PR on top of #1236.

Test/Func/pretty_print.mlir
```diff
@@ -1,7 +1,7 @@
-// Custom (pretty) printing of `func.func` via the `HasCustomPrinting`
-// interface. The input is in generic form; with --pretty the `func.func` ops
-// are printed in their custom syntax (mirroring MLIR's hasCustomAssemblyFormat)
-// while everything else keeps the generic form.
+// Custom (pretty) printing via the `HasCustomPrinting` interface. The input is
+// in generic form; with --pretty, `func.func` uses its hand-written printer
+// (mirroring MLIR's hasCustomAssemblyFormat) and `func.return`/`func.call` use
+// their declarative assembly formats; everything else keeps the generic form.
 
 // RUN: veir-opt --pretty %s | filecheck %s
```
tobiasgrosser, line 6: I feel we would want to flip the option name and mirror mlir with `-mlir-print-op-generic does`. Happy to do this in a follow-up PR, where you just drop the `--pretty` from all these test cases and add `-mlir-print-op-generic` to all the other ones (including macros). You probably want to open on zulip a conversation on how you plan to transition towards the new `pretty` syntax.

Veir/AssemblyFormat.lean
```diff
@@ -0,0 +1,449 @@
...
+[`assemblyFormat`](https://mlir.llvm.org/docs/DefiningDialects/Operations/#declarative-assembly-format),
+for describing the *custom* (pretty) syntax of an operation. The intent is that
+a format string written for MLIR's `assemblyFormat` can be reused in VeIR with
+little to no change.
+
+The DSL is interpreted at runtime: `Format.parse` turns a format string into a
+`Format` AST, and `formatPrinter` compiles that AST into a
+`Veir.Printer.CustomPrinter` (printing everything after the operation's name
+prefix). The generated printers are registered dialect-locally: each dialect
+declares its format strings in its `Printing.lean` (e.g.
```
tobiasgrosser, line 16:
```suggestion
declares its format strings in its `Printing.lean` (e.g.,
```

tobiasgrosser (approved):
Very nice! Please wait for @math-fehr's in-detail feedback!

PR #949, opened 2026-06-29
feat(isel): isel patterns for `i1 -> i64`

luisacicolini:
adding the sext/zext isel patterns `i1 -> i64`
see: https://godbolt.org/z/eGqqW6fsh, https://godbolt.org/z/KhePdTKj6
since these are a bit different than the existing ones I put them in the sdag set, but don't have a strong opinion here - @regehr do you have any thoughts?


tobiasgrosser (approved):
Nice. Let's see what @regehr says. 🥲

regehr (approved):
Luisa this LGTM.

you could make the isel patterns a bit more concise by using the castToReg and replaceWithReg helpers that I had added, but this isn't a big deal

PR #950, opened 2026-06-29
feat(mir): add load and stores to MIRPrinter

luisacicolini:
Enable printing of ld/sd operations when using llvm's regallocator


regehr (approved):
LGTM!

PR #951, opened 2026-06-30
fix(parser): allow omitting `!llvm.` prefix for nested `!llvm.struct`

GZGavinZhao:
LLVM dialect has a custom directive `PrettyLLVMType` that allows one to write types from the LLVM dialect without `!llvm.` prefix when it's already nested inside another LLVM-dialect type. For example, `!llvm.array<2 x struct<...>>` is the same as `!llvm.array<2 x !llvm.struct<...>>`.

This convenience syntax is only available for types wrapped inside the `PrettyLLVMType` custom directive. In practice, this convenience syntax is available wherever a nested type is allowed, so we can always turn it on when we know we're parsing inside a `!llvm.` type. We're already doing this for some LLVM-dialect types like `!llvm.array`; this PR just enables this convenience syntax for `!llvm.struct` as well.

Printing without `!llvm.` is not supported for now, always prints the full `!llvm.struct` regardless of nesting status, same behavior as `!llvm.aray`. 

Note that since this shorthand is valid for any LLVM-dialect type, our `parseLLVMType` is incomplete (only handles 4 out of the 10 LLVM-dialect types). This needs to be fixed in future PRs. Tracked in #952.

### DIFF
--- a/UnitTest/AttrParser.lean
+++ b/UnitTest/AttrParser.lean
@@ -232,6 +232,36 @@ macro "#assert " e:term : command =>
 /-! ## LLVM Byte type -/
 #assert expectSuccessType "!llvm.byte<64>" (LLVM.ByteType.mk 64)
 
+/-! ## LLVM Struct type (parsed opaquely; see `parseOptionalLLVMStructType`)
+
+  The struct type is handled by a dedicated parser that accepts both the
+  standalone `!llvm.struct<...>` form and the bare `struct<...>` form used when a
+  struct is nested inside another LLVM type (e.g. an array element). Both forms
+  are normalized to the full `!llvm.struct<...>` spelling, and neither requires
+  `allowUnregisteredDialect` (like `!llvm.array`). The struct name, fields, and
+  packed flag are *not* modeled structurally — they survive only as text. -/
+
+-- Standalone struct: both forms parse identically, with or without the flag.
+#assert expectSuccessType "!llvm.struct<(i32, f32)>"
+  ⟨UnregisteredAttr.mk "!llvm.struct<(i32, f32)>" true, by grind⟩ false
+#assert expectSuccessType "!llvm.struct<(i32, f32)>"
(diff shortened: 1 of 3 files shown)


GZGavinZhao:
Wait, let me move the comment about the convenience syntax to `parseLLVMType`. This way it's clearer since omitting `!llvm.` applies to all the LLVM-dialect types.

GZGavinZhao:
Done.

tobiasgrosser (approved):
Very beautiful. Let's see if @regehr or @math-fehr have further input.

regehr (approved):
this LGTM, thanks!

UnitTest/AttrParser.lean
```diff
@@ -229,6 +229,36 @@ macro "#assert " e:term : command =>
...
 #assert expectSuccessAttr "!llvm.array<2 x !llvm.array<3x i64>>" (LLVM.ArrayType.mk 2 $ LLVM.ArrayType.mk 3 $ IntegerType.mk 64)
 
+/-! ## LLVM Struct type (parsed opaquely; see `parseOptionalLLVMStructType`)
+
+  The struct type is handled by a dedicated parser that accepts both the
+  standalone `!llvm.struct<...>` form and the bare `struct<...>` form used when a
+  struct is nested inside another LLVM type (e.g. an array element). Both forms
+  are normalized to the full `!llvm.struct<...>` spelling, and neither requires
+  `allowUnregisteredDialect` (like `!llvm.array`). The struct name, fields, and
+  packed flag are *not* modeled structurally — they survive only as text. -/
```
sishtiaq, line 242: Maybe file an issue for that.

sishtiaq (approved):
Maybe file an issue to properly define the struct, etc types. https://github.com/opencompl/veir/issues/952 hints at this too, but only talks about parsing.

UnitTest/AttrParser.lean
GZGavinZhao: Good point! Opened #959.

PR #953, opened 2026-06-30
feat(llvm): support `lw/sw` and `lb/sb` for i8/i32 load/stores

salinhkuhn:
This PR adds `lw/sw`(i32)  as well as `lb/sb` (i8)  lowerings.

 On proofs: These aren't fully proven yet but it is analogous to the rest of memory (i64). Proves and follow-up exec test as a  lightweight check will be added in the future + someone else working on the memory proofs.

Where these lowering come from: LLVM's RISC-V backend selects integer loads/stores by byte width — i8/i32/i64 → LB/LW/LD and SB/SW/SD (it only uses the zero-extending LBU/LWU when the loaded value is explicitly zero-extended). We mirror that with the plain signed loads and I think we should implement the LBU/LWU lowering as a rewrite in the future? Because it needs to also account for having a zero-extension after the load which can be pattern-matched.

### DIFF
--- a/Test/Passes/InstructionSelection/RISCV64/load.mlir
+++ b/Test/Passes/InstructionSelection/RISCV64/load.mlir
@@ -9,5 +9,25 @@
         // CHECK-NEXT: {{.*}} = "builtin.unrealized_conversion_cast"({{.*}}) : (!riscv.reg) -> i64
         "func.return"() : () -> ()
     }) : () -> ()
-    
+
+    // i32 load lowers to `riscv.lw`
+    "func.func"()  <{function_type = (!llvm.ptr) -> ()}> ({
+    ^bb0(%a: !llvm.ptr):
+        %val = "llvm.load"(%a) : (!llvm.ptr) -> i32
+        // CHECK:      {{.*}} = "builtin.unrealized_conversion_cast"({{.*}}) : (!llvm.ptr) -> !riscv.reg
+        // CHECK-NEXT: {{.*}} = "riscv.lw"({{.*}}) <{"value" = 0 : i64}> : (!riscv.reg) -> !riscv.reg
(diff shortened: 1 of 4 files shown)


luisacicolini (approved):
lgtm, thank you!

tobiasgrosser:
This does not come with rewrites proven? Should we add these, too? Also, I would appreciate 1-2 sentences on if/how LLVM lowers these.

No need to block this on complex changes, but we should add at least some information on why this is correct for somone who does not know the LLVM backends.

salinhkuhn:
> This does not come with rewrites proven? Should we add these, too? Also, I would appreciate 1-2 sentences on if/how LLVM lowers these.
> 
> No need to block this on complex changes, but we should add at least some information on why this is correct for somone who does not know the LLVM backends.

Yes I can add an explanation. However, with regard to adding proven rewrites:  AFAIU we don't have them for i64 either, and someone is working on the proofs currently? 
I tried to find some proof for the `i64` and I understood that currently we are checking correctness via execution and there is no proper proof ?
How should I do the i32/i8 rewrites, do you mean writing such tests?

tobiasgrosser:
Then add the explanation for now to the PR message before merging. We can do the other proofs in a follow-up commit.

tobiasgrosser (approved):

regehr (approved):
this LGTM!

salinhkuhn:
> This does not come with rewrites proven? Should we add these, too? Also, I would appreciate 1-2 sentences on if/how LLVM lowers these.
> 
> No need to block this on complex changes, but we should add at least some information on why this is correct for somone who does not know the LLVM backends.

Done

salinhkuhn:
> Then add the explanation for now to the PR message before merging. We can do the other proofs in a follow-up commit.



> Then add the explanation for now to the PR message before merging. We can do the other proofs in a follow-up commit.

Done

PR #964, opened 2026-07-01
feat(riscv): more isel rules for i32

regehr:
this is a complement to Sarah's recent patch, it's sort of a mechanical addition of a bunch of isel rules for i32, so we can hopefully lower more real applications. I hope it's boring code! some of it such as the 32-bit `bitreverse` is a little bit exciting but I've tested it fairly thoroughly and also the refinement proof goes.

one nasty thing here is a little patch to the Veir->MIR printer, which was emitting incorrect MIR code for 32->64 conversions, yikes! these need to be zero extensions.

this patch has been fuzzed and seems solid.


luisacicolini (approved):
lgtm, thank you!

PR #968, opened 2026-07-01
feat(Rewriter): introduce eraseOp!/replaceValue! and clear remaining pass sorries

tobiasgrosser:
- Adds `PatternRewriter.eraseOp!` and `.replaceValue!`, mechanical panicking wrappers over the existing `WfRewriter.eraseOp!`/`replaceValue!`.
- Uses them (alongside the earlier `createOp!`/`replaceOp!`) to clear the remaining sorries in `InstCombine`, `ModArithToArith`, `MIRCombinesVeir`, `CastsReconciliation`, `DCE`, `RISCVCombines/Combine`, and `CSE` (which calls `WfRewriter` directly).
- `RISCV64Branches.lean` is intentionally left untouched: its `let some x := ... | return c` fallbacks around these calls are a genuine graceful-skip path rather than an always-true proof obligation, so converting to `!` would silently turn skips into panics.

### DIFF
--- a/Veir/Passes/CSE.lean
+++ b/Veir/Passes/CSE.lean
@@ -127,7 +127,6 @@ def key? (ctx : IRContext OpCode) (op : OperationPtr) : Option Key := do
       return ordinaryKey ctx op kind
   | _ => none
 
-set_option warn.sorry false in
 /-- Perform CSE on a single BB: Walk the operations, building up a
     hash of available values. For any operation whose value is already
     available, replace it with the earlier one. -/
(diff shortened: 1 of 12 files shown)


regehr (approved):
LGTM!!

math-fehr:
> RISCV64Branches.lean is intentionally left untouched: its let some x := ... | return c fallbacks around these calls are a genuine graceful-skip path rather than an always-true proof obligation, so converting to ! would silently turn skips into panics.

I disagree with this, we should also convert the ones in `RISCV64Branches.lean`. I believe the LLM thought the `return c` was different for some reasons, but it should panic in this case.

tobiasgrosser:
This is meanwhile upstream in various different ways.

PR #969, opened 2026-07-01
feat: sha 256 block compression function analog to chacha and fastntt

salinhkuhn:
This PR adds C code for the SHA-256 block compression function  (core of SHA-256)  together with the test cases that verify our RISC-V lowering of it. It follows the same self-contained style as fastntt.c / chacha20.c (so a single exported function over caller-provided buffers, no local arrays or globals).

### DIFF
--- a/Test/Passes/InstructionSelection/RISCV64/sha256.mlir
+++ b/Test/Passes/InstructionSelection/RISCV64/sha256.mlir
@@ -288,16 +288,12 @@
 // CHECK-NEXT:         %[[v864:[0-9]+]] = "riscv.sextw"(%[[v863]]) : (!riscv.reg) -> !riscv.reg
 // CHECK-NEXT:         %[[v859:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[varg7_1]]) : (!llvm.ptr) -> !riscv.reg
 // CHECK-NEXT:         %[[v861:[0-9]+]] = "riscv.sh2add"(%[[v864]], %[[v859]]) : (!riscv.reg, !riscv.reg) -> !riscv.reg
-// CHECK-NEXT:         %[[v862:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v861]]) : (!riscv.reg) -> !llvm.ptr
-// CHECK-NEXT:         %[[v856:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v862]]) : (!llvm.ptr) -> !riscv.reg
-// CHECK-NEXT:         %[[v857:[0-9]+]] = "riscv.lw"(%[[v856]]) <{"value" = 0 : i64}> : (!riscv.reg) -> !riscv.reg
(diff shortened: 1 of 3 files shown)


Test/Passes/InstructionSelection/RISCV64/sha256.mlir
```diff
@@ -0,0 +1,903 @@
...
+// CHECK-NEXT:         %[[v900:[0-9]+]] = "riscv.li"() <{"value" = 32 : i32}> : () -> !riscv.reg
+// CHECK-NEXT:         %[[v901:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v900]]) : (!riscv.reg) -> i32
+// CHECK-NEXT:         %[[v898:[0-9]+]] = "riscv.li"() <{"value" = 11 : i32}> : () -> !riscv.reg
+// CHECK-NEXT:         %[[v899:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v898]]) : (!riscv.reg) -> i32
+// CHECK-NEXT:         %[[v896:[0-9]+]] = "riscv.li"() <{"value" = 25 : i32}> : () -> !riscv.reg
+// CHECK-NEXT:         %[[v897:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v896]]) : (!riscv.reg) -> i32
+// CHECK-NEXT:         %[[v894:[0-9]+]] = "riscv.li"() <{"value" = -1 : i32}> : () -> !riscv.reg
+// CHECK-NEXT:         %[[v895:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v894]]) : (!riscv.reg) -> i32
+// CHECK-NEXT:         %[[v892:[0-9]+]] = "riscv.li"() <{"value" = 2 : i32}> : () -> !riscv.reg
+// CHECK-NEXT:         %[[v893:[0-9]+]] = "builtin.unrealized_conversion_cast"(%[[v892]]) : (!riscv.reg) -> i32
```
tobiasgrosser, line 253: There is a huge number of unrealized_conversion_casts. Why have these not been eliminated?

tobiasgrosser:
Very cool. I let @luisacicolini do the full review.

Test/Passes/InstructionSelection/RISCV64/sha256.mlir
salinhkuhn: good point, we never added them for `i32` and `ptr` `reg` `ptr`. will look into it now but the `reg` `i32` `reg` cast could be tricky.

salinhkuhn: Added it in #970

regehr (approved):

regehr:
this LGTM but I think it should land only after #970 has landed and the tests are updated to not contain all the casts

Test/Passes/InstructionSelection/RISCV64/sha256.mlir
salinhkuhn: I updated it and removed all the casts via a rewrite wherever we had a cast in the direction of i32 to reg. The remaining casts are in exactly the opposite direction, and we currently can't remove those.

PR #970, opened 2026-07-02
 feat: add `i32` and `ptr` round trip casting

salinhkuhn:
This PR adds the checks that allow to fold conversion casts from 'i32' to 'reg' back to 'i32' as well as 'ptr' to 'reg' to 'ptr' (i64) and reverse 'reg' to 'ptr' to 'reg'.

Currently fails the fastest and chacham checks because their CHECK lines still have the explicit conversion cast not folded away.

### DIFF
--- a/Test/Passes/CastsReconciliation/basic.mlir
+++ b/Test/Passes/CastsReconciliation/basic.mlir
@@ -108,6 +108,71 @@
         // CHECK-NEXT:   "test.test"([[ARG]], [[ARG]]) : (i64, i64) -> ()
         "func.return"() : () -> ()
     }) : () -> ()
+
+  ^10():
+    "func.func"()  <{function_type = (i32) -> ()}> ({
+      ^1(%0 : i32):
+        // i32 -> reg -> i32 
+        %1 = "builtin.unrealized_conversion_cast"(%0) : (i32) -> !riscv.reg
+        %2 = "builtin.unrealized_conversion_cast"(%1) : (!riscv.reg) -> i32
+        "test.test"(%2) : (i32) -> ()
+        // CHECK:        ^{{.*}}([[ARG:%.*]] : i32):
+        // CHECK-NEXT:   "test.test"([[ARG]]) : (i32) -> ()
+        "func.return"() : () -> ()
+    }) : () -> ()
+
+  ^11():
+    "func.func"()  <{function_type = (!riscv.reg) -> ()}> ({
+      ^1(%0 : !riscv.reg):
+        // reg -> i32 -> reg : NOT IMPLEMENT YET SO SHOULD NOT BE FOLDED AWAY.
+        %1 = "builtin.unrealized_conversion_cast"(%0) : (!riscv.reg) -> i32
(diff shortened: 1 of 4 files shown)


Test/Passes/CastsReconciliation/basic.mlir
```diff
@@ -108,6 +108,71 @@
...
+        "test.test"(%2) : (i32) -> ()
+        // CHECK:        ^{{.*}}([[ARG:%.*]] : i32):
+        // CHECK-NEXT:   "test.test"([[ARG]]) : (i32) -> ()
+        "func.return"() : () -> ()
+    }) : () -> ()
+
+  ^11():
+    "func.func"()  <{function_type = (!riscv.reg) -> ()}> ({
+      ^1(%0 : !riscv.reg):
+        // reg -> i32 -> reg : NOT IMPLEMENT YET SO SHOULD NOT BE FOLDED AWAY.
```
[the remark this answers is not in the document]
tobiasgrosser: I would just say "should not be folded away." Moving a RISCV register to an i32 register will lose information, no?

Its great that you added this example to demonstrate we do not fold this cast away.

Veir/Passes/CastsReconciliation/Reconciliation.lean
```diff
@@ -14,6 +14,25 @@ def isRiscvRegToI64Cast (inputType interType : TypeAttr): Bool :=
...
+  `toReg` zero-extends into the 64-bit register and `toInt 32` takes the low 32
+  bits.
+-/
+def isRiscvRegToI32Cast (inputType interType : TypeAttr): Bool :=
+ match inputType.val, interType.val with
+  | .integerType x, .registerType _ => x.bitwidth = 32
+  | _, _ => false
+
+/-!
+  Same for `!llvm.ptr` and `!riscv.reg` (analog to 'RiscvRegToI64Cast')
```
tobiasgrosser: Here we need to be a bit careful, as this is only correct if llvm.ptr is 64-bit wide, no? We can have this for now, but maybe add a comment that this only works in this specific case. We may need to think about how to actually prove this.

Veir/Passes/CastsReconciliation/Reconciliation.lean
```diff
@@ -14,6 +14,25 @@ def isRiscvRegToI64Cast (inputType interType : TypeAttr): Bool :=
   | .integerType x, .registerType _ => x.bitwidth = 64
   | _, _ => false
 
+/-!
+  Same for `!riscv.reg` and `i32`, but only the `i32 -> reg -> i32` direction:
+  `toReg` zero-extends into the 64-bit register and `toInt 32` takes the low 32
+  bits.
+-/
```
tobiasgrosser, line 19: A comment should be independent, as the context may change in the future. Can you drop the `Same for ...` part and make it independent?

tobiasgrosser (commented):
Nice. This looks good. I would suggest you update the other test cases that need updating, and we merge this before #969.

regehr (approved):
this LGTM once Tobias's comments are resolved.
I think we should get this merged as soon as possible so we can look at the next set of issues that come up!

PR #973, opened 2026-07-02
feat: add `veir_bv_normalize` annotations to new Byte API

alexkeizer:
This PR adds the appropriate annotations to make `Byte.cast`, `allPoison`, trunc` and `isRefinedBy` bitblastable (simply by unfolding the definitions).

I also refactored the definition of `lshr` so that it is immediately bitblastable, instead of having to write a separate unfolding lemma. This new definition is otherwise equivalent to the previous one.


nchappe (approved):

tobiasgrosser (approved):
LGTM. Thank you!

PR #974, opened 2026-07-03
feat(riscv): better codegen for conditionals

regehr:
this PR does a few things:
- add a combine that looks for uses of a register that has `li 0` in it, and switch this over to a use of the RISC-V zero register
- veir2mir support for the zero register
- better lowerings for several `icmp` forms. notably, f1-f4 in `Test/Passes/InstructionSelection/RISCV64/icmp_imm.mlir` codegen then same using `veir` as `llc`!
- relax the verifier a bit, so it doesn't complain about uses of the zero register

these changes make our lowering of `fastntt` several instructions smaller!


luisacicolini (approved):
lgtm!

tobiasgrosser (approved):

PR #978, opened 2026-07-03
fix: make `createOp!` return an `Option` when allocation fails

math-fehr:
`createOp!` was previously panicking whenever allocation failed. This meant that we couldn't really reason about `createOp!`, as we cannot reason whether an allocation would panic or not.


tobiasgrosser:
Should we then not introduce a function `createOp?`

math-fehr:
> Should we then not introduce a function `createOp?`

Not here I would say, `createOp?` would be one that returns `none` if any of the required hypotheses would not hold. Here we still do a dynamic check, just not for the allocation which can fail (but that we cannot really reason about).

tobiasgrosser (approved):
OK, then let's get this in.

PR #979, opened 2026-07-03
feat(refinement): add `InterpreterState.isRefinedByAt` relation

math-fehr:
It differs to `isRefinedBy` by requiring values between states to refine each others only if they are dominated by the "refinement point".

This new definition will superceed `isRefinedBy`, as it can handle things like looks that `isRefinedBy` cannot handle (as some "old" values stay in the state).


regehr (approved):
LGTM, this seems quite understandable and good

PR #981, opened 2026-07-03
feat(parser): support forward references to SSA values

GZGavinZhao:
Implements MLIR parser support for SSA forward references.

The parser now:
- creates parser-local placeholders for SSA operands referenced before definition
- follows MLIR's approach by using temporary `builtin.unrealized_conversion_cast` operation results as placeholders, rather than adding a new SSA value kind
- resolves placeholders when the real SSA definition is parsed
- reports unresolved or type-incompatible forward references with source locations

Added coverage for three cases:
- valid SSA forward reference across blocks
- missing forward block reference in `[]` successor syntax
- parser unit coverage for forward SSA references, including multi-result references and error cases

### DIFF
--- a/UnitTest/DataFlowFramework/Helpers.lean
+++ b/UnitTest/DataFlowFramework/Helpers.lean
@@ -27,7 +27,7 @@ def parseTopLevelOp (s : String) : Except String (OperationPtr × MlirParserStat
     | throw "internal error: failed to create IR context"
   let parserState ← (ParserState.fromInput s.toByteArray).mapError toString
   let (op, mlirState, _) ←
-    ((parseOp none).run (MlirParserState.fromContext ctx) parserState).mapError toString
+    (Veir.Parser.parseTopLevelOp.run (MlirParserState.fromContext ctx) parserState).mapError toString
   pure (op, mlirState)
 
 /--
--- a/UnitTest/MlirParser.lean
+++ b/UnitTest/MlirParser.lean
@@ -14,7 +14,7 @@ def testParseOp (s : String) : IO Unit :=
   | some (ctx, _) =>
     match ParserState.fromInput (s.toByteArray) with
     | .ok parser =>
-      match (parseOp none).run (MlirParserState.fromContext ctx) parser with
+      match parseTopLevelOp.run (MlirParserState.fromContext ctx) parser with
       | .ok (op, state, _) => Printer.printOperation state.ctx op
       | .error err => .error (toString err)
     | .error err => .error (toString err)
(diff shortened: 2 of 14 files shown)


GZGavinZhao:
Bad linearity, some fixes incoming.

tobiasgrosser:
Nice!

> Bad linearity, some fixes incoming.

Looking forward to them.

regehr:
yay!! psyched for this to land

UnitTest/MlirParser.lean
```diff
@@ -280,3 +280,94 @@ def testParseOp (s : String) : IO Unit :=
...
+/--
+  error: use of undefined value %a
+-/
+#guard_msgs in
+#eval! testParseOp r#""builtin.module"() ({
+  "test.test"(%a) : (i32) -> ()
+  "test.test"() ({
+    %a = "test.test"() : () -> i32
+  }) : () -> ()
+}) : () -> ()"#
```
math-fehr, line 384: Can you add this test in an `.mlir` file? Just so we can see where the error location is?

Veir/Parser/MlirParser.lean
```diff
@@ -35,6 +35,29 @@ def BlockEntry.block : BlockEntry → BlockPtr
...
+  Bookkeeping for an SSA value that has been referenced before it is defined
+  (a "forward reference"). For each referenced result index, we hold a detached
+  single-result placeholder operation whose result stands in for the eventual
+  definition. When the definition is finally parsed, all uses of each
+  placeholder result are replaced by the real value (via `Rewriter.replaceValue`)
+  and the placeholder operation is erased, keeping the module well-formed by
+  construction.
+
+  This mirrors MLIR's `createForwardRefPlaceholder`, which likewise uses a throwaway
+  operation result (not a distinct kind of SSA value) as the placeholder.
```
math-fehr:
```suggestion
  This mirrors MLIR's `createForwardRefPlaceholder`, which likewise uses a throwaway
  operation result as the placeholder.
```
It's good for me to know I was wrong, but I don't think we should mention this in the documentation comments ^^

Veir/Parser/MlirParser.lean
```diff
@@ -184,6 +193,93 @@ def modifyContextM' (f : WfIRContext OpCode → MlirParserM (α × WfIRContext O
...
+  Resolve a forward-referenced value now that its real definition has been parsed.
+  For each result index that was referenced, replace all uses of the placeholder result
+  with the corresponding real value and erase the placeholder operation.
+-/
+def resolveForwardValue (name : ByteArray) (pos : Location) (fwd : ForwardValue)
+    (values : Array ValuePtr) : MlirParserM Unit := do
+  for (index, (placeholderOp, useLoc)) in fwd.placeholders do
+    let some realValue := values[index]?
+      | throw (({ msg := s!"definition of value %{String.fromUTF8! name} provides {values.size} results, but result #{index} was used",
+                  pos := some pos } : ParserError).addNote useLoc "value used here")
```
math-fehr, line 251: This error is not tested yet?

Veir/Parser/MlirParser.lean
```diff
@@ -600,6 +735,12 @@ partial def parseRegion : MlirParserM RegionPtr := do
     if let .ForwardDeclared _ forwardLoc := entry then
       throwAt forwardLoc s!"block %{String.fromUTF8! blockName} was used but never defined"
 
+  /- Check that all values forward referenced in this region were eventually defined. -/
+  let currentDepth := (← getThe MlirParserState).definitionsPerScope.size - 1
+  for (valueName, fwd) in (← getThe MlirParserState).forwardValues do
+    if fwd.scopeDepth = currentDepth then
+      throwAt fwd.loc s!"use of undefined value %{String.fromUTF8! valueName}"
```
math-fehr: This is actually too strict.
For example:
```
^entry:
  jmp ^bb1
^bb0:
  my.scope() {
    test.test(%x) : i32
  }
^bb1:
  %x = arith.constant 3 : i32
  jmp ^bb0
```

Here `x` is forward declared, but it can still be defined later in the program, outside of its scope, in a larger scope.

So you shouldn't actually check that values are defined or not. This is done in MLIR in `IsolatedFromAbove` scopes, but since we don't handle this we don't have to do this here. We should only check this at the most toplevel parsing function.

Veir/Parser/MlirParser.lean
```diff
@@ -184,6 +193,93 @@ def modifyContextM' (f : WfIRContext OpCode → MlirParserM (α × WfIRContext O
...
+    modifyContextM fun ctx => do
+      let ⟨hOldIn⟩ ← checkValueInBounds placeholderValue ctx.raw
+      let ⟨hNewIn⟩ ← checkValueInBounds realValue ctx.raw
+      let ⟨hNe⟩ ← checkValuesNe placeholderValue realValue
+      pure (WfRewriter.replaceValue ctx placeholderValue realValue hNe hOldIn hNewIn)
+    modifyContextM fun ctx => do
+      let ⟨hOpIn⟩ ← checkOpInBounds placeholderOp ctx.raw
+      let ⟨hNoRegions⟩ ← checkOpNoRegions placeholderOp ctx.raw
+      let ⟨hNoUses⟩ ← checkOpNoUses placeholderOp ctx.raw
+      pure (WfRewriter.eraseOp ctx placeholderOp hNoRegions (by simpa using hNoUses) hOpIn)
```
math-fehr:
```suggestion
      pure (WfRewriter.eraseOp ctx placeholderOp hNoRegions (by grind) hOpIn)
```
Or instead make `checkOpNoUses` have the exact same shape as what you are expecting in `eraseOp`

math-fehr (commented):
Nice! I added some comments inline, but otherwise I see two "bugs" here:

1) Right now, entry points in the parser (like `parseOp`) might let the forward declared values escape. So for instance if we parse `%x = my.op ...`, then `%x` is forward declared, but the operation is still returned to the user. So we should check in all public operations that the forward declared values are all gone.

2) This one is more subtle, but look at what happens when you parse
```
"builtin.module"() ({
  "test.use"(%a) : (i32) -> ()
  "test.wrapper"() ({
    "test.use"(%a) : (i32) -> ()
    %a = "test.def"() : () -> i32
  }) : () -> ()
  %a = "test.def"() : () -> i32
}) : () -> ()
```
Before, it was failing because `a` was redefined, but now it seems like it is making this program pass without an error (and also uses the second `%a` everywhere. Can you have this test and make sure it returns an error as well?

Veir/Parser/MlirParser.lean
GZGavinZhao: Updated `checkOpNoUses` to use the same type as the corresponding argument for `WfRewriter.eraseOp`.

UnitTest/DataFlowFramework/Helpers.lean
```diff
@@ -27,7 +27,7 @@ def parseTopLevelOp (s : String) : Except String (OperationPtr × MlirParserStat
     | throw "internal error: failed to create IR context"
   let parserState ← (ParserState.fromInput s.toByteArray).mapError toString
   let (op, mlirState, _) ←
-    ((parseOp none).run (MlirParserState.fromContext ctx) parserState).mapError toString
+    (Veir.Parser.parseTopLevelOp.run (MlirParserState.fromContext ctx) parserState).mapError toString
```
GZGavinZhao, line 30: Unfortunate name conflict :(

GZGavinZhao:
Rebased on `main` and addressed the comments as we decided to focus on parsing the scope correctly and don't worry about rejecting programs based on dominance (for now). Ready for a new round of review.

Veir/Parser/MlirParser.lean
```diff
@@ -628,4 +769,15 @@ partial def parseOptionalBlock (ip : BlockInsertPoint) : MlirParserM (Option Blo
 
 end
 
+/-- Check that all SSA values forward referenced while parsing were eventually defined. -/
+def checkNoUnresolvedForwardValues : MlirParserM Unit := do
+  for (valueName, fwd) in (← get).forwardValues do
+    throwAt fwd.loc s!"use of undefined value %{String.fromUTF8! valueName}"
+
+/-- Parse a top-level operation and report any unresolved SSA forward references. -/
+partial def parseTopLevelOp : MlirParserM OperationPtr := do
```
math-fehr, line 778:
```suggestion
def parseTopLevelOp : MlirParserM OperationPtr := do
```

math-fehr (approved):
Looks good for me, I just have one more requirement, can you make `parseOp` private now, so that all users are using `parseTopLevelOp`?

GZGavinZhao:
`parseOp` is now `private` and without `partial`. Ready for final review. Thanks!

math-fehr (approved):
Perfect, thanks for the quick turnaround!

PR #983, opened 2026-07-04
feat: Replace reg -> iX -> reg casts with zero extension

nchappe:
This is done using either dedicated zext instructions, or shift-left then shift-right.


tobiasgrosser (approved):

regehr (approved):
LGTM!

PR #986, opened 2026-07-04
chore(isel): add BitVec.udiv_one_shl to prove sorry

tobiasgrosser:


regehr (approved):
this is very cool!
it's super amazing to be have a solver that we can teach new tricks, so different than working with Z3

Veir/ForLean.lean
```diff
@@ -453,4 +463,3 @@ namespace Rat
 def twoPow (k : Int) : Rat := 2 ^ k
 
 end Rat
-
```
luisacicolini, line 456: unrelated?

luisacicolini (approved):
lgtm, thank you!

Veir/ForLean.lean
tobiasgrosser: Merged separately.

PR #963, opened 2026-06-30
chore (mod_arith): add mod_arith cast reconciliation

AlexanderViand:
Enables `builtin.unrealized_conversion_cast` reconciliation between `!mod_arith.int` and IntegerType.

PR #987, opened 2026-07-04
feat(riscv): convert function arguments / return values to riscv registers

regehr:
this PR makes the `reconcile-cast` pass turn function arguments and return values into RISC-V registers, when this is safe to do. this eliminates a lot of conversion casts! it only does this for `i32`, `i64`, and `ptr`.

this PR also touches Verifier.lean because there was some shared functionality that I wanted to reuse, and then also the return type checker had to be relaxed to permit a function to return the zero register.

I hope that between this PR and #983 we get rid of most or all conversion casts!

vsmith had to be updated to truncate the 64-bit register return value to 32 bits, when the original function returned i32.


tobiasgrosser (approved):

AlexanderViand:
@regehr 

While updating #963 to current main, I noticed that this riscv.reg rewrite is applying way too broadly and should imho not be part of the default cast-reconciliation.

As-is, it would turn 
```llvm
"func.func"()  <{function_type = (i32) -> (), sym_name = "f18"}> ({
      ^1(%0 : i32):
        %1 = "builtin.unrealized_conversion_cast"(%0) : (i32) -> !mod_arith.int<7 : i32>
        %2 = "builtin.unrealized_conversion_cast"(%1) : (!mod_arith.int<7 : i32>) -> i32
        "test.test"(%2) : (i32) -> ()
        "func.return"() : () -> ()
    }) : () -> ()
```
   
 into 
```llvm
    "func.func"() <{"function_type" = (!riscv.reg) -> (), "sym_name" = "f18"}> ({
      ^135(%arg135_0 : !riscv.reg):
        %152 = "builtin.unrealized_conversion_cast"(%arg135_0) : (!riscv.reg) -> i32
        "test.test"(%152) : (i32) -> ()
        "func.return"() : () -> ()
    }) : () -> ()
```

Which is not something we'd expect a generic/non-riscv-specific pass called `--reconcile-cast` to do (and breaks our lowering pipeline)

Would you be OK with moving this to a separate pass?

regehr:
> Would you be OK with moving this to a separate pass?

ah, yikes, yes-- I was considering this pass logically part of the RISC-V backend but if it's generic infrastructure then this is incorrect! I'll continue discussion over at your new PR

PR #989, opened 2026-07-04
feat: add lemmas about successful operation matching

math-fehr:
These lemmas provide the information gathered whenever a match was successful.


tobiasgrosser (approved):

Veir/Passes/Matching.lean
```diff
@@ -1,3 +1,4 @@
+
 import Veir.Pass
 import Veir.PatternRewriter.Basic
```
tobiasgrosser:
```suggestion
import Veir.Pass
import Veir.PatternRewriter.Basic
```

PR #991, opened 2026-07-05
chore(isel): add BitVec.sdiv_one_shl_of_smod_eq_zero 

tobiasgrosser:
Add `BitVec.sdiv_one_shl_of_smod_eq_zero`: an exact `sdiv` by `2^k` agrees with the
arithmetic shift `sshiftRight'` when there's no remainder to round. Extend
`veir_bv_normalize`'s post-phase discharger to bv_omega so the lemma's shift-range
side condition can be closed from the theorem's own ambient hypothesis, letting
`sdivPow2Exact_pos_refinement` and `sdivwPow2Exact_pos_refinement` close with a single
`veir_bv_decide` instead of timing out.

### DIFF
--- a/Veir/ForLean.lean
+++ b/Veir/ForLean.lean
@@ -442,6 +442,37 @@ theorem udiv_one_shl {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂) :
       rw [BitVec.toNat_twoPow_of_le hk, BitVec.toNat_ofNat, Nat.zero_mod])
     rw [htw, BitVec.udiv_zero, BitVec.ushiftRight_eq_zero hk]
 
+/-- The heterogeneous-width variant of `toInt_sshiftRight'`, which is only stated for a shift
+    amount of the same width as the shifted value. -/
+theorem toInt_sshiftRight'' {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂) :
+    (x.sshiftRight' k).toInt = x.toInt >>> k.toNat := by
+  rw [BitVec.sshiftRight_eq']
+  exact BitVec.toInt_sshiftRight
+
+/-- An exact `sdiv` by a power of two (i.e. one where the dividend has no fractional part to
+    round, witnessed by `x.smod (2^k) = 0`) agrees with the arithmetic shift `x.sshiftRight' k`.
+    `k` must leave room for a sign bit (`k.toNat + 1 < w`, i.e. `2^k` isn't `x`'s own `intMin`),
+    matching the range for which `sdivPow2Exact`/`sdivwPow2Exact` are selected. -/
+@[veir_bv_normalize_post]
+theorem sdiv_one_shl_of_smod_eq_zero {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂)
+    (hk : k.toNat + 1 < w₁) (h : x.smod ((1#w₁) <<< k) = 0#w₁) :
+    x.sdiv ((1#w₁) <<< k) = x.sshiftRight' k := by
+  have hy : ((1#w₁) <<< k).toInt = ((2 ^ k.toNat : Nat) : Int) := by
+    rw [BitVec.shiftLeft_eq', ← BitVec.twoPow_eq, BitVec.toInt_twoPow, if_neg (by omega),
+      if_neg (by omega)]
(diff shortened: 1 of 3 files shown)


Veir/ForLean.lean
```diff
@@ -442,6 +442,37 @@ theorem udiv_one_shl {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂) :
       rw [BitVec.toNat_twoPow_of_le hk, BitVec.toNat_ofNat, Nat.zero_mod])
     rw [htw, BitVec.udiv_zero, BitVec.ushiftRight_eq_zero hk]
 
+/-- The heterogeneous-width variant of `toInt_sshiftRight'`, which is only stated for a shift
+    amount of the same width as the shifted value. -/
+theorem toInt_sshiftRight'' {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂) :
+    (x.sshiftRight' k).toInt = x.toInt >>> k.toNat := by
+  rw [BitVec.sshiftRight_eq']
+  exact BitVec.toInt_sshiftRight
```
```suggestion
  simp
```

Veir/ForLean.lean
```diff
@@ -442,6 +442,37 @@ theorem udiv_one_shl {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂) :
...
+
+/-- An exact `sdiv` by a power of two (i.e. one where the dividend has no fractional part to
+    round, witnessed by `x.smod (2^k) = 0`) agrees with the arithmetic shift `x.sshiftRight' k`.
+    `k` must leave room for a sign bit (`k.toNat + 1 < w`, i.e. `2^k` isn't `x`'s own `intMin`),
+    matching the range for which `sdivPow2Exact`/`sdivwPow2Exact` are selected. -/
+@[veir_bv_normalize_post]
+theorem sdiv_one_shl_of_smod_eq_zero {w₁ w₂ : Nat} (x : BitVec w₁) (k : BitVec w₂)
+    (hk : k.toNat + 1 < w₁) (h : x.smod ((1#w₁) <<< k) = 0#w₁) :
+    x.sdiv ((1#w₁) <<< k) = x.sshiftRight' k := by
```
luisacicolini, line 459: afaiu this is not in normal form, which is:
```
x.sdiv (1#w₁ <<< k.toNat) = x.sshiftRight k.toNat
```

luisacicolini: I was trying to golf the proof a bit because there are a lot of hypotheses but couldn't come up with anything better

Veir/Meta/BVDecide.lean
```diff
@@ -14,11 +14,23 @@ Preprocessing happens in two stages because rewrites are generally
...
+
+Exception: conditional rewrites whose side condition only becomes available *after*
+unfolding poison-checks into a chain of hypotheses (e.g. `isPoison_sdiv`) must go in
+`veir_bv_normalize_post` instead, since only its `simp` call runs with `+contextual`
+(threading those hypotheses forward as it descends). Its `discharger := bv_omega` lets
+such a lemma's side condition draw on both the threaded hypotheses and any ambient
+local hypotheses (e.g. a range bound on a shift amount) that aren't part of that chain,
+since `bv_omega` scans the whole local context, not just the current subgoal. The
+first-phase `simp` is plain (no `+contextual`, no discharger), so such lemmas silently
+fail to fire there.
```
alexkeizer, line 26: I'm a bit confused, the comment explains that these lemmas are an exception and have to go in the second phase because only the second phase has a discharger, but it is exactly this PR that adds said discharger, so it could've just added the discharger in the first phase? In general, the comment is very wordy, but I can't really make sense of what it's saying.

Further, why do we actually need the `bv_omega` discharger here? I would not expect `veir_bv_normalize` to close goals, it's job is just to, well, normalize and then let `bv_decide` close the goals.

If the particular goal is not closable by bv_decide, we should either investigate if there's a missing normalization lemma that would let bv_decide close it, but if that's just fundamentally not possible, I'd say it ought such goals ought to be out-of-scope for `veir_bv_decide`

alexkeizer: If the problem is a time-out, let's report that upstream, and avoid the issue locally by just setting the time-out higher in `veir_bv_decide`. That feels much safer than relying on brittle magic dischargers

PR #993, opened 2026-07-05
feat(llvm): add 6 saturating operations + integer abs

regehr:
also vsmith support. not extensively tested yet, but it looks like there aren't any problems.


luisacicolini (approved):
perfect, thanks a lot!

PR #995, opened 2026-07-05
feat(riscv): Use `LocalRewritePattern` in SelectionDag lowerings

math-fehr:
This is a first step to be able to reason about these patterns.
`LocalRewritePattern` are more constrained `RewritePattern` that return a list of operations and a list of values. The operations are inserted in place of the operation that was matched, and then the values replace the original operation.


tobiasgrosser (approved):
This looks like a good mechanical change. Maybe explain in 3-4 sentences what the local rewrites do, otherwise, I glad to see it merged.

math-fehr:
Message updated!

PR #996, opened 2026-07-05
feat(riscv): lower LLVM saturating operations and integer abs to RISC-V

regehr:
very lightly tested, but the lowerings themselves follow LLVM closely so they should be OK

someone knowledgeable should look over the changes to `Veir/Data/LLVM/Int/Bitblast.lean` this is what claude wanted to do to make proofs work


luisacicolini (approved):
lgtm, including proofs, thanks!!!

PR #997, opened 2026-07-05
feat: lowering patterns for libdiv

luisacicolini:
Support i8 and i1 types in lowering `and`, `or`, `constant`. 
Support i8 type in lowering `ashr`, `icmp`. 
Support i1 type in lowering `select`, `sext to i32`.


tobiasgrosser:
Nice.

PR #998, opened 2026-07-06
feat(riscv): lower the funnel shift general case

regehr:
this adds funnel shift lowering to risc-v for the general case, for 32 and 64 bits.


tobiasgrosser (approved):
Very cool. Thank you, @regehr!

PR #999, opened 2026-07-06
chore(isel): rewrite `x.sdiv ((1#64) <<< k) ` to ift+shift

tobiasgrosser:
In case of signed divisions by (negative) one shifted to the left by `k`, we can find a division-free statement that uses just shifts on the RHS. The precise RHS differs based on various factors. Hence, we rewrite the RHS to an if statement.
In particular:

```lean
  if x.msb then
    if k = 0 then
      x.sshiftRight' k
    else if k ≠ (-1 : BitVec 6) then
      (x + (x.sshiftRight' (-1 : BitVec 6) >>> (-k))).sshiftRight' k
    else
      if x = BitVec.intMin 64 then 1#64 else 0#64
  else
    x.sshiftRight' k
```

While the RHS has several branches, `sdiv` is fully eliminated, which means this rewrite should terminate. We apply this rewrite in the post-normalization phase to not prevent simplifications on sdiv to trigger.

This also avoids the need of complex dischargers in `veir_bv_normalize`, which is something too fragile to use in a normalizer.

We add a similar rewrite for the `-1`. Together, they close the remaining open instruction selection proofs.

### DIFF
--- a/Veir/Meta/BVDecide.lean
+++ b/Veir/Meta/BVDecide.lean
@@ -14,32 +14,11 @@ Preprocessing happens in two stages because rewrites are generally
 easier to state with the dependently-typed `Int.getValue`, but `bv_decide` works
 better with the non-dependent `Int.getValueD`. Thus, `veir_bv_normalize_post`
 contains `Int.getValue_eq_getValueD`, which rewrites the former into the latter.
-
-Exception: conditional rewrites whose side condition only becomes available *after*
-unfolding poison-checks into a chain of hypotheses (e.g. `isPoison_sdiv`) must go in
-`veir_bv_normalize_post` instead, since only its `simp` call runs with `+contextual`
-(threading those hypotheses forward as it descends) and has a discharger that lets
-such a lemma's side condition draw on both the threaded hypotheses and any ambient
-local hypotheses (e.g. a range bound on a shift amount) that aren't part of that chain,
-by scanning the whole local context, not just the current subgoal, the way `bv_omega`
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
This keeps some divisions that are not eliminated, but extending the timeouts will allow them to be proven.

georgerennie (approved):
I think this is fine for now to me - we will want to replace these hardcoded widths with a more generic version subsequently (I have a local version of this I have mostly working, but given they aren't needed for the story right now I can wait to finish and merge them). Also we will want to reduce the timeout again I think lest VeIR take years to compile.

Veir/Meta/BVDecide.lean
```diff
@@ -14,22 +14,11 @@ Preprocessing happens in two stages because rewrites are generally
...
-since `bv_omega` scans the whole local context, not just the current subgoal. The
-first-phase `simp` is plain (no `+contextual`, no discharger), so such lemmas silently
-fail to fire there.
 -/
 @[expose] macro "veir_bv_normalize" : tactic =>
   `(tactic| ((
       simp -failIfUnchanged only [veir_bv_normalize] <;>
       simp +contextual -failIfUnchanged
-        (discharger := bv_omega)
         only [veir_bv_normalize_post, reduceIte])))
```
alexkeizer, line 22: ❤️

tobiasgrosser: Thank you for the review

PR #1001, opened 2026-07-06
feat(riscv): rewrites to remove sign/zero extends when we can prove this is safe

regehr:
with links into the LLVM backend.

some of these get subsumed and can be removed when we get known/demanded bits.


tobiasgrosser:
Nice!

PR #1015, opened 2026-07-08
feat(riscv): verify the rewrites in RISCV64Sdag.lean

regehr:
well, I think the interesting part of this PR is that claude seems to have created some useful infrastructure for matching multi-instruction patterns on the LHS. 

on the other hand there are some truly horrific-looking proofs in here. 

also @math-fehr I let claude rewrite your markdown file-- let me know if I should revert that.

### DIFF
--- a/Veir/Passes/InstructionSelection/RISCV64Sdag.lean
+++ b/Veir/Passes/InstructionSelection/RISCV64Sdag.lean
@@ -20,11 +20,19 @@ def singleSetBit (x : BitVec 64) : Option Int :=
 /-! # SelectionDAG Lowering Patterns  -/
 
 /--
-  `and x (not y)` -> `riscv.andn x y`. The `not` may appear on either operand.
+  Shared shape of the Zbb negated-operand lowerings (`andn`/`orn`/`xnor`): match a binary LLVM
+  op on `i64` one of whose operands is a `not` (`xor _, -1`), cast the plain operand `x` and the
+  `not`'s operand `y` to registers, apply the binary reg-reg riscv op `dst`, and cast the result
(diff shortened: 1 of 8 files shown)


Veir/Passes/InstructionSelection/RISCV64Sdag.lean
```diff
@@ -20,11 +20,19 @@ def singleSetBit (x : BitVec 64) : Option Int :=
...
 
 /--
-  `and x (not y)` -> `riscv.andn x y`. The `not` may appear on either operand.
+  Shared shape of the Zbb negated-operand lowerings (`andn`/`orn`/`xnor`): match a binary LLVM
+  op on `i64` one of whose operands is a `not` (`xor _, -1`), cast the plain operand `x` and the
+  `not`'s operand `y` to registers, apply the binary reg-reg riscv op `dst`, and cast the result
+  back to `i64`. Lowerings of this shape share the correctness proof of
+  `lowerBinopNotLocal_preservesSemantics`, so proving a new one only requires the per-op
+  data-level refinement lemmas.
 -/
```
math-fehr, line 27:
```suggestion
  op on `i64` one of whose operands is a `not` (`xor _, -1`), cast the plain operand `x` and the
  `not`'s operand `y` to registers, apply the binary reg-reg riscv op `dst`, and cast the result
  back to `i64`.
-/
```
For some reasons, LLMs are overeager to add extra information in documentation even though that's not that much relevant imo.

math-fehr (approved):
Nice! The parts outside of the `RewriteProofs/` looks good!
The lemmas and proofs are horrible yeah, the goal is to find a good API/DSL later on to make this nicer.

PR #1040, opened 2026-07-09
feat: prove correctness of immediate-form icmp instruction selection

tobiasgrosser:
Proves `PreservesSemantics` for `slti_local`: an `llvm.icmp x C` on `i64` (constant rhs fitting a signed 12-bit immediate) lowers to

- `slti`/`sltiu` for `slt`/`ult`,
- `xori (slti/sltiu _ C) 1` for `sge`/`uge` (sense-inverted),
- the off-by-one `x ≤ C ≡ x < C+1` form `slti/sltiu _ (C+1)` for `sle`/`ule` (`ule` guarded on `C ≠ -1`, where `C+1` would wrap).

This is the first `i1`-result lowering. New infrastructure:

- **`IsVerifiedICmp` / `Verified.llvm_icmp`** verifier bundle (`Veir/Verifier.lean`) — two equal-typed operands, `i1` result.
- **`matchIcmp_interpretOp_unfold`** — the one-step interpretation unfold specialised to `icmp` (dynamic operand-width recovery, fixed `i1` result value `LLVM.Int.icmp`).
- Six `slti_*`/`sltiu_*` data (refinement) lemmas, including the `C+1` overflow bridges for the `≤` predicates.

**Proof tooling.** The predicate-dispatch `match` must be reduced with `split` — `simp only []`/`dsimp` unfold `getType!`/`Option.bind` and leave the term ill-typed under instance transparency. The emit chain is then peeled with `rw [Option.bind_eq_some_iff]`-based bind extraction (which composes with `split`, unlike the `simp`-based `peel*` macros), factored into the file-local `sltiNonWrapArm`/`sltiWrapArm` arm macros (3-op and 4-op chains). Axioms pinned via `#guard_msgs` (standard sound set, no `sorryAx`).

Also wires `LowerSlti` into the `RewriteProofs` aggregator so it is built and checked by `lake build`.


tobiasgrosser:
Superseded: `main` already merged an independent, canonical proof of `slti_local_preservesSemantics` (with its own `IsVerifiedICmp`/`Verified.llvm_icmp` bundle, wired into the aggregator, and a `CommonMatchEqns` module that pre-generates the shared matcher auxiliaries). Closing this duplicate. The one thing main omitted — a `ProofStrategy.md` row for the slti lowering — will be folded into the build-wiring cleanup PR.

PR #1043, opened 2026-07-09
feat: prove correctness of the LLVM-dialect RISC-V combines

tobiasgrosser:
`RISCVCombines/Proofs.lean` covers the combines that rewrite already-selected `riscv` ops, as exact equalities over the total `Data.RISCV.Reg` type. The 63 combines that rewrite `llvm` ops had no data-level lemma at all — half of `Combine.lean` was unproven.

## Lemmas

New `RISCVCombines/LLVMProofs.lean`: one `veir_bv_decide` lemma per LLVM combine, named after the rule it justifies (so the naming convention that links `Proofs.lean` to `Combine.lean` extends to it).

LLVM ops carry poison, so the obligation is a refinement rather than an equality — the rewritten value may be *more* defined than the original, never less:

```lean
theorem AndSextSext {x y : Int 32} :
    and (sext x 64 h32_64) (sext y 64 h32_64) ⊒ sext (and x y) 64 h32_64 := by
  veir_bv_decide
```

`bv_decide` bitblasts and so cannot handle generic widths. These are the `i64` instantiation: `sext`/`zext` from `i32` to `i64`, `trunc` from `i64` to `i32`, and `i1` for `select` conditions and `icmp` results.

## 18 unsound combines, fixed

Writing the lemmas surfaced that 18 combines were unsound. Each copied a poison-producing flag off a matched op onto a **different** op it creates. Dropping a flag only removes poison and is always sound; transplanting one onto an op with a different poison condition is not.

- **`nsw` on a hoisted `shl`/`trunc`** — `AndShlShl`, `AndTruncTrunc`, and the `Or`/`Xor` variants.
- **`nuw`/`exact` on hoisted `Or`/`Xor`** — `X` alone can supply the discarded bit that poisons the created op while the source is fine. (`And` is safe: its bits are a subset of `Y`'s.)
- **`disjoint` on `or`** — a *second, independent* failure in `OrTruncTrunc`, `OrShlShl`, `OrLshrLshr`, `OrAshrAshr`, `OrAndAnd`: `X` and `Y` may overlap only in bits the hoisted op discards.
- **`nneg` on a hoisted `zext`** — `OrZextZext`, `XorZextZext`.
- **flag transplant onto an unrelated arithmetic op** — `sub_add_reg_x_sub_y_add_x`, `sub_add_reg_x_sub_x_add_y`, `add_shift`, `add_shift_commute`, `sub_one_from_sub_rw`.

Concrete counterexamples, verified by `#eval` (each prints `(source, target)` as `(value, poison)`):

| combine | flag | witness |
| --- | --- | --- |
| `AndShlShl` | `nsw` | `X = 2^62`, `Y = -2^62`, `Z = 1` → source `2^63`, target `poison` |
| `AndTruncTrunc` | `nsw` | `X = 2^31`, `Y = -1` → source `0x80000000`, target `poison` |
| `OrAndAnd` | `disjoint` | `X = 1`, `Y = -1`, `Z = -2` → source `-2`, target `poison` |
| `sub_one_from_sub_rw` | `nuw` | `A = 5`, `B = 3` → source `1`, target `poison` |

`Combine.lean` now clears the offending flag on the created op, dropping the minimum. `AndTruncTrunc` and `AndShlShl` still keep `nuw`, since the bits `X & Y` discards are a subset of `Y`'s. Each fix carries a comment naming its reason and pointing at its lemma; each lemma records the counterexample that motivated the clear.

With the created flag cleared, the flags read off *matched* ops decouple from it, so each lemma is stated with those fully free and only the created op carries a literal `false`:

```lean
and (shl x z s0 u0) (shl y z false u1) ⊒ shl (and x y) z false u1   -- would have been
and (shl x z s0 u0) (shl y z s1 u1)    ⊒ shl (and x y) z false u1   -- as committed
```

## Coverage

63 → 126 of the 127 registered rules. The one remaining, `li_zero_to_x0`, rewrites `li 0` to a read of `x0`; its obligation is a register-file fact, not a bitvector identity.

These are still standalone data-level identities — nothing yet ties `AndSextSext`-the-theorem to `AndSextSext`-the-rule, and `Combine.lean` still discharges its IR-bookkeeping obligations with `sorry`. Same gap as the existing `Proofs.lean`; wiring the two together is follow-up work.

## Testing

- All 63 lemmas are `sorry`-free and depend only on `propext`, `Classical.choice`, `Quot.sound` — no `sorryAx`, no `native_decide`.
- The existing 82 `RISCVCombines` tests use flag-free inputs; `veir-opt -p=riscv-combine` output is byte-identical across all 82 before and after, so the fix only affects flag-carrying IR.
- New `flag_drop_poison_safety.mlir` covers nine of the fixed rules and **fails against the pre-fix pass** (checked), including that `AndShlShl` turns `overflowFlags = 3` into `overflowFlags = 2` rather than dropping `nuw` too.
- Full suite green: 647 lit tests, `lake test`, and a warning-clean `lake build Veir` (touched modules also built individually).


regehr (approved):

tobiasgrosser:
Let's close this. We will reprove this when puddle is ready.

PR #1055, opened 2026-07-10
Combine proofs 

tobiasgrosser:


tobiasgrosser:
We will port these proofs to puddle. So lets close this PR.

PR #1060, opened 2026-07-10
feat(veir-opt): teach optimizer about collections of passes

regehr:
I would like the optimizer to understand about common groups of passes that are used together to achieve some effect, so we don't have to have different versions of this sitting around in different parts of the repo.

I think two things are up for debate here:
- what are the actual passes we want to invoke for "-O" and also for the RISC-V backend? I've taken some guesses and am open to feedback
- do we want this option to compose with other pass options, so that we can build larger pipelines? for now I've gone with "no" to keep things simple


tobiasgrosser (approved):
I love it. Let's see what @luisacicolini @math-fehr and @nchappe say.

math-fehr:
I'm not a big fan of this particular design, because as you said you cannot compose it with other groups.
Maybe it would be better to just create passes that are collection of passes, so that we can compose everything? Also, it should be fairly easy to implement

regehr:
> I'm not a big fan of this particular design, because as you said you cannot compose it with other groups. Maybe it would be better to just create passes that are collection of passes, so that we can compose everything? Also, it should be fairly easy to implement

take another look please @math-fehr ?

math-fehr:
Sorry, what I meant is having both `-p` and `-pgrp` in the same option (like `-p`).
So you would have `-p=canonicalize,O,canonicalize`.
I think it would make this more readable and easy to understand?

regehr:
@math-fehr ok look one more time please!

math-fehr (approved):
Thanks, sorry for the delay!

PR #1061, opened 2026-07-11
feat(riscv): Use `LocalRewritePattern` in RISC-V combines

tobiasgrosser:
Restate every rewrite pattern in the RISC-V combine pass as a
`LocalRewritePattern`: `foo_local` threads a `WfIRContext`, creates
detached operations, and returns the new operations together with the
replacement values, while `foo` becomes a thin
`RewritePattern.fromLocalRewrite` wrapper so the pattern list is
unchanged.

This removes all 39 `sorry`s from the pass, consolidating the proof
obligations into `fromLocalRewrite`, and is a first step towards


regehr (approved):
obviously I did not read every line of this, but the pattern looks good!

math-fehr (approved):
I haven't read every single line, but it seems consistent and makes sense

PR #1063, opened 2026-07-12
feat(riscv): Use `LocalRewritePattern` in the MIR combines

tobiasgrosser:
Restate every rewrite pattern in `MIRCombinesVeir` as a
`LocalRewritePattern`, as was already done for `same_val_zero_1`:
`foo_local` threads a `WfIRContext`, creates detached operations, and
returns the new operations together with the replacement values, while
`foo` becomes a thin `RewritePattern.fromLocalRewrite` wrapper so the
pattern list is unchanged.

This removes all `sorry`s from the pass, and is a first step towards


tobiasgrosser:
Merging this as trivial, as this is literally the same change as in #1061.

PR #1064, opened 2026-07-12
feat(instcombine): Use `LocalRewritePattern` in the instcombine pass

tobiasgrosser:
Restate every rewrite pattern in the instcombine pass as a
`LocalRewritePattern`: `foo_local` threads a `WfIRContext`, creates
detached operations, and returns the new operations together with the
replacement values, while `foo` becomes a thin
`RewritePattern.fromLocalRewrite` wrapper so the pattern list is
unchanged.

This removes all `sorry`s from the pass, and is a first step towards


tobiasgrosser:
I merged this as trivial, as this is structurally identical to #1061.

PR #1065, opened 2026-07-12
feat(riscv): Use the `!` rewriter functions in the branch lowering

tobiasgrosser:
Replace the proof-carrying `WfRewriter` calls in the branch instruction
selection with their panicking `!` variants, which check their
preconditions at runtime instead. This removes all 24 `sorry`s from the
pass.

Dropping the `sorry`s re-enables the unused-variable linter for these
declarations, which exposes that `convertBranch` never uses its `block`
argument; remove it.

### DIFF
--- a/Veir/Passes/InstructionSelection/RISCV64Branches.lean
+++ b/Veir/Passes/InstructionSelection/RISCV64Branches.lean
@@ -8,9 +8,8 @@ import Std
 
 namespace Veir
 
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
I would appreciate a comment on the block argument removal. @regehr.

math-fehr (approved):

PR #1066, opened 2026-07-12
feat(passes): Use the `!` rewriter functions in DCE, CSE and ModArithToArith

tobiasgrosser:
Replace the proof-carrying `replaceValue` and `eraseOp` calls in the
three passes with their panicking `!` variants, which check their
preconditions at runtime instead. This removes all `sorry`s from them.

Dropping the `sorry`s re-enables the unused-variable linter for these
declarations, which exposes that the patterns never use their
`opInBounds` argument; rename it to `_opInBounds`.


tobiasgrosser:
Merged as obvious, as this matches #1061.

PR #1068, opened 2026-07-12
feat(LLVM): prove monotonicity of the remaining `LLVM.Int` operations

tobiasgrosser:
`Bitblast` already proves that the arithmetic, bitwise, shift, cast, comparison
and select operations are monotone with respect to refinement. Extend this to the
operations that were still missing: the funnel shifts, the bit-counting and
bit-reversing intrinsics, the signed and unsigned min/max, `abs`, and the
saturating arithmetic.

Poison is refined by anything and a concrete value is refined only by itself, so
`grind` discharges each of them from the existing `isPoison`/`getValue` equations.
`bswap` has no width-generic `getValue` equation, so it argues by cases on the
source instead.

### DIFF
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -816,6 +816,89 @@ theorem isPoison_freeze {w : Nat} (x : Int w) :
   simp [freeze, isPoison, Id.run]
   grind
 
+
+theorem fshl_mono {w : Nat} (a₁ b₁ c₁ a₂ b₂ c₂ : Int w)
+    (ha : a₁ ⊒ a₂) (hb : b₁ ⊒ b₂) (hc : c₁ ⊒ c₂) :
+    fshl a₁ b₁ c₁ ⊒ fshl a₂ b₂ c₂ := by
+  grind
+
+theorem fshr_mono {w : Nat} (a₁ b₁ c₁ a₂ b₂ c₂ : Int w)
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
I am still working on a bitwidth-generic version of `bswap`.

tobiasgrosser:
> I am still working on a bitwidth-generic version of `bswap`.

Let's do this in a separate PR. This is ready-to-merge from my side.

math-fehr (approved):

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -816,6 +816,89 @@ theorem isPoison_freeze {w : Nat} (x : Int w) :
...
+/-- `bswap` has no width-generic `getValue` equation, so we argue by cases on the source instead:
+poison propagates, and a concrete source is refined only by itself. -/
+theorem bswap_mono {w : Nat} (x₁ x₂ : Int w) (h : x₁ ⊒ x₂) :
+    bswap x₁ ⊒ bswap x₂ := by
+  cases x₁ with
+  | poison => grind
+  | val v =>
+    have : x₂ = .val v := by
+      cases x₂ <;> grind [isRefinedBy]
+    subst this
```
alexkeizer, line 851: This can be golfed slightly, into:
```
 obtain rfl : x₂ = .val v := by
      cases x₂ <;> grind [isRefinedBy]
```

Not that we necessarily have to do so here, just a cute pattern to be aware of (`obtain rfl : _ := _` does the exact same as `have : _ := _` followed by `subst this`)

PR #1069, opened 2026-07-12
feat(interpret): prove `interpretOp'` monotone for the riscv dialect

tobiasgrosser:
RISC-V operands are registers, which carry no poison, so refinement on them is
equality: an operand array of registers is refined only by itself
(`RuntimeValue.eq_of_arrayIsRefinedBy_of_regs`). That discharges all 110 riscv
opcodes at once, without reasoning about any of them individually: either every
operand is a register, and then the refined operands are the original ones, so
both sides interpret to the very same result; or some operand is not a register,
and every opcode that reads its operands fails to interpret, while the opcodes
that ignore their operands (`li`, `lui`) again produce the very same result on
both sides.

Dispatch the `riscv` case of `interpretOp'_monotone` to it. The other dialects
remain `sorry`.

### DIFF
--- a/Veir/Interpreter/Lemmas.lean
+++ b/Veir/Interpreter/Lemmas.lean
@@ -541,14 +541,51 @@ theorem RuntimeValue.eq_of_arrayIsRefinedBy_of_regs {a b : Array RuntimeValue}
     (h : a ⊒ b) (hregs : ∀ v ∈ a, ∃ r, v = .reg r) : b = a := by
   grind [arrayIsRefinedBy, reg_of_isRefinedBy, Array.getElem_mem]
 
+@[grind =]
+theorem Interp.pure_def {α : Type} (a : α) : (pure a : Interp α) = some (.ok a) := rfl
+
+@[grind =]
+theorem Interp.bind_def {α β : Type} (x : Interp α) (f : α → Interp β) :
+    (x >>= f) = match x with
+      | none => none
+      | some .ub => some .ub
+      | some (.ok a) => f a := rfl
+
+/--
+A RISC-V operation that interprets successfully produces register results and no control flow
+action: a single register for the arithmetic and load opcodes, and no result at all for the stores.
+Note that the memory is *not* preserved -- loads grow it via `ensureSize` and stores write to it.
+-/
+theorem Riscv.interpretOp'_ok_results {vals : Array RuntimeValue} {mem' : MemoryState}
+    {act : Option ControlFlowAction}
+    (h : Riscv.interpretOp' opType properties resultTypes operands blockOperands mem
+      = some (.ok (vals, mem', act))) :
+    ((∃ r, vals = #[.reg r]) ∨ vals = #[]) ∧ act = none := by
+  cases opType <;> simp only [Riscv.interpretOp'] at h <;> grind
+
+/--
+A non-register operand is either fatal or irrelevant: every RISC-V opcode that reads its operands
+pattern-matches them as registers and fails to interpret otherwise, and the opcodes that ignore
+their operands (`li`, `lui`) interpret to the very same result whatever the operands are.
+-/
+theorem Riscv.interpretOp'_eq_none_or_eq_of_not_regs {operands operands' : Array RuntimeValue}
+    (hregs : ¬ ∀ v ∈ operands, ∃ r, v = .reg r) :
+    Riscv.interpretOp' opType properties resultTypes operands blockOperands mem = none ∨
+    Riscv.interpretOp' opType properties resultTypes operands blockOperands mem
+      = Riscv.interpretOp' opType properties resultTypes operands' blockOperands mem := by
+  cases opType <;>
+    simp only [Riscv.interpretOp'] <;>
+    first
+      | (right; trivial)
+      | (left; split <;> grind [Array.mem_def])
+
 /--
 `Riscv.interpretOp'` is monotone in its operands.
 
 RISC-V operands are registers, which carry no poison, so refinement on them is equality: either
 every operand is a register -- and then the refined operands are the original ones and both sides
-interpret to the very same result -- or some operand is not a register, and every opcode that
-reads its operands fails to interpret (opcodes that ignore their operands, such as `li`, again
-produce the very same result on both sides).
+interpret to the very same result -- or some operand is not a register, and
+`Riscv.interpretOp'_eq_none_or_eq_of_not_regs` applies.
 -/
 theorem Riscv.interpretOp'_monotone {operands operands' : Array RuntimeValue} :
     operands ⊒ operands' →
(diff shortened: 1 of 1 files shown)


Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,70 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
     (mem : MemoryState) :
   ∃ res, op.interpret ctx.raw operands mem = some res := by sorry
 
+/-- The relation `interpretOp'_monotone` establishes between two interpretation results. -/
+abbrev InterpretResultIsRefinedBy :
+    Array RuntimeValue × MemoryState × Option ControlFlowAction →
+    Array RuntimeValue × MemoryState × Option ControlFlowAction → Prop :=
+  fun r₁ r₂ => r₁.1 ⊒ r₂.1 ∧ r₁.2.1 = r₂.2.1 ∧
+    ControlFlowAction.optionIsRefinedBy r₁.2.2 r₂.2.2
```
math-fehr, line 653: This should probably be added in `Interpreter/Basic.lean` if possible. Unless there is a cycle of dependency from that?
In any case, since this is introduced, we should also use it in the other parts of the codebase I would say.

Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,70 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
...
+-/
+theorem RuntimeValue.eq_of_arrayIsRefinedBy_of_regs {a b : Array RuntimeValue}
+    (h : a ⊒ b) (hregs : ∀ v ∈ a, ∃ r, v = .reg r) : b = a := by
+  obtain ⟨hsize, helem⟩ := h
+  apply Array.ext hsize.symm
+  intro i hib hia
+  obtain ⟨r, hr⟩ := hregs a[i] (Array.getElem_mem hia)
+  have hrefines := helem i hia
+  rw [getElem!_pos a i hia, getElem!_pos b i hib, hr] at hrefines
+  rw [hr]
```
math-fehr: Is there a reason you use `rw` instead of `simp` here?
Also, how much of this can be replaced by `grind`?

Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,70 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
...
+/--
+A register runtime value can only be refined by itself, so operand arrays that consist purely of
+registers are refined only by themselves. This makes every dialect whose operands are registers
+(`riscv`, `riscv_cf`, `riscv_stack`, `rv64`) monotone for free: the refined operands are the
+original ones, so both sides interpret to the very same result.
+-/
+theorem RuntimeValue.eq_of_arrayIsRefinedBy_of_regs {a b : Array RuntimeValue}
+    (h : a ⊒ b) (hregs : ∀ v ∈ a, ∃ r, v = .reg r) : b = a := by
+  obtain ⟨hsize, helem⟩ := h
+  apply Array.ext hsize.symm
```
math-fehr: does the `ext` tactic works here?

Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,70 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
...
+reads its operands fails to interpret (opcodes that ignore their operands, such as `li`, again
+produce the very same result on both sides).
+-/
+theorem Riscv.interpretOp'_monotone {operands operands' : Array RuntimeValue} :
+    operands ⊒ operands' →
+    Interp.isRefinedBy InterpretResultIsRefinedBy
+      (Riscv.interpretOp' opType properties resultTypes operands blockOperands mem)
+      (Riscv.interpretOp' opType properties resultTypes operands' blockOperands mem) := by
+  intro h
+  by_cases hregs : ∀ v ∈ operands, ∃ r, v = .reg r
```
math-fehr, line 726: This proof strategy seems quite weird.
Here you are casing on the fact that the results of `interpretOp'` are registers or not.
Is there any case where this isn't the case? I feel like we should have a theorem that asserts this.

Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,70 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
...
+      simp only [Riscv.interpretOp'] <;>
+      first
+        | apply interpretResult_isRefinedBy_refl
+        | (split
+           · exfalso
+             rename_i heq
+             refine hregs (fun v hv => ?_)
+             have hv' : v ∈ operands.toList := hv.val
+             rw [heq] at hv'
+             simp at hv'
```
math-fehr: non-terminal `simp` here

Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,70 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
...
+             grind
+           · simp [Interp.isRefinedBy])
```
math-fehr: How many of the cases go through the first one, compared the second one?
If we move this to a theorem that the return value is a register, I feel this proof would become more readable

math-fehr (commented):
I had a few comments on the proof strategy, but nice otherwise!

Veir/Interpreter/Lemmas.lean
tobiasgrosser: Its now all grind.

tobiasgrosser: This is now all by `grind`.

Veir/Interpreter/Lemmas.lean
tobiasgrosser: This should now be clearer with helper theorems.

Veir/Interpreter/Lemmas.lean
```diff
@@ -516,6 +517,89 @@ theorem exists_interpretOp'_eq_some {ctx : WfIRContext OpCode} {op : OperationPt
...
+    (x >>= f) = match x with
+      | none => none
+      | some .ub => some .ub
+      | some (.ok a) => f a := rfl
+
+/--
+A RISC-V operation that interprets successfully produces register results and no control flow
+action: a single register for the arithmetic and load opcodes, and no result at all for the stores.
+Note that the memory is *not* preserved -- loads grow it via `ensureSize` and stores write to it.
+-/
```
math-fehr, line 688:
```suggestion
action: a single register for the arithmetic and load opcodes, and no result at all for the stores.
```
I am not sure why mentioning memory preservation here?

math-fehr (approved):
Just one comment on some documentation, but otherwise all good!

Issue #954, opened 2026-06-30
parser fails to parse non-UTF8 values.


sishtiaq:
When parsing `sqlite3`, we fail to parse the `sqlite3UpperToLower` table as it contains non-UTF8 values:

<img width="1504" height="215" alt="Image" src="https://github.com/user-attachments/assets/112157ea-6dc3-49b1-b7ad-60ca74624808" />

sishtiaq:
Fixed by #1072.

PR #1072, opened 2026-07-13
fix(parser): Parse more than utf8

sishtiaq:
fix(parse): Extend parser to parse more than UTF-8 (#954).

`sqlite3.c` has a char table that needs to be parsed. This PR allows non-UTF-8 to be parsed too. The three chars {\n,\t,\"} are still printed as human-readable, even if their corresponding hex codes are parsed. 

The main change is that `Parser.parseStringLiteral` now returns a `ByteArray`, rather than a `String`. Some of the previous uses (`AttrParser`, `MlirParser`)  converted to `Byte` anyway; these calculations have now gone. 
On the other hand, `UnitTest/Parser` has a wrapper to convert to `String` for comparison.

### DIFF
--- a/UnitTest/Parser.lean
+++ b/UnitTest/Parser.lean
@@ -238,83 +238,89 @@ end parseKeyword
 
 section parseStringLiteral
 
+private def parseStringLiteralStr : EStateM ParserError ParserState String :=
+  parseStringLiteral >>= fun b => return String.fromUTF8! b
+
+private def parseOptionalStringLiteralStr : EStateM ParserError ParserState (Option String) :=
+  parseOptionalStringLiteral >>= fun ob => return ob.map String.fromUTF8!
+
 /--
   info: "Success: hello world!"
 -/
 #guard_msgs in
-#eval testParser "\"hello world!\"" parseStringLiteral
+#eval testParser "\"hello world!\"" parseStringLiteralStr
 
 /-- info: <test>:1:1: error: string literal expected
 hello world!
 ^ -/
 #guard_msgs (whitespace := exact) in
-#eval testParser "hello world!" parseStringLiteral
+#eval testParser "hello world!" parseStringLiteralStr
 
 /--
   info: <unknown location>: error: expected '"' in string literal
 -/
 #guard_msgs in
-#eval testParser "\"unterminated string" parseStringLiteral
+#eval testParser "\"unterminated string" parseStringLiteralStr
 
 /--
   info: "Success: \n"
 -/
 #guard_msgs in
-#eval testParser "\"\\n\"" parseStringLiteral
+#eval testParser "\"\\n\"" parseStringLiteralStr
 
 /--
   info: "Success: hello\tworld"
 -/
 #guard_msgs in
-#eval testParser "\"hello\\tworld\"" parseStringLiteral
+#eval testParser "\"hello\\tworld\"" parseStringLiteralStr
 
 /--
   info: "Success: say \"hi\""
 -/
 #guard_msgs in
-#eval testParser "\"say \\\"hi\\\"\"" parseStringLiteral
+#eval testParser "\"say \\\"hi\\\"\"" parseStringLiteralStr
 
 /--
   info: "Success: back\\slash"
 -/
 #guard_msgs in
-#eval testParser "\"back\\\\slash\"" parseStringLiteral
+#eval testParser "\"back\\\\slash\"" parseStringLiteralStr
 
 /--
   info: "Success: A"
 -/
 #guard_msgs in
-#eval testParser "\"\\41\"" parseStringLiteral
+#eval testParser "\"\\41\"" parseStringLiteralStr
 
 /--
   info: "Success: a"
 -/
 #guard_msgs in
-#eval testParser "\"\\61\"" parseStringLiteral
+#eval testParser "\"\\61\"" parseStringLiteralStr
 
 /--
   info: "Success: *"
 -/
 #guard_msgs in
-#eval testParser "\"\\2a\"" parseStringLiteral
+#eval testParser "\"\\2a\"" parseStringLiteralStr
 
 /--
   info: "Success: *"
 -/
 #guard_msgs in
-#eval testParser "\"\\2A\"" parseStringLiteral
+#eval testParser "\"\\2A\"" parseStringLiteralStr
 
 /--
   info: "Success: O"
 -/
 #guard_msgs in
-#eval testParser "\"\\4f\"" parseStringLiteral
+#eval testParser "\"\\4f\"" parseStringLiteralStr
 
 /--
   info: "Success: O"
 -/
 #guard_msgs in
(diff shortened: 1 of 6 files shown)


UnitTest/Parser.lean
```diff
@@ -325,31 +331,40 @@ since single-byte values with a letter first digit (>= 0xA0) aren't valid UTF-8
...
+#eval testParser "\"hello world!\"" parseOptionalStringLiteralStr
+
+/-!
+{\t, \n, \"} and {\09, \0A, \22} are printed as {\t, \n, \"}.
+-/
+/--
+  info: "Success: (some (\t\n\"hello world\t\n\"))"
+-/
+#guard_msgs in
+#eval testParser "\"\\t\\n\\\"hello world\\09\\0A\\22\"" parseOptionalStringLiteralStr
```
math-fehr, line 368: I guess you should add a test here where we return `none` from `parseOptionalStringLiteralStr`?

Veir/IR/Attribute.lean
```diff
@@ -754,18 +754,37 @@ instance : ToString RegisterType where
...
-def escapeStringLiteral (s : String) : String := Id.run do
+-- def escapeStringLiteral (s : String) : String := Id.run do
+--   let mut result := ""
+--   for c in s.toList do
+--     if c == '\\' then result := result ++ "\\\\"
+--     else if c == '"' then result := result ++ "\\\""
+--     else if c == '\n' then result := result ++ "\\n"
+--     else if c == '\t' then result := result ++ "\\t"
+--     else result := result.push c
+--   return result
```
math-fehr:
```suggestion
```

Veir/IR/Attribute.lean
```diff
@@ -754,18 +754,37 @@ instance : ToString RegisterType where
...
+
+private def hexDigit (n : UInt8) : Char :=
+  if n < 10 then Char.ofNat (n.toNat + '0'.toNat)
+  else Char.ofNat (n.toNat - 10 + 'A'.toNat)
+
+def escapeStringLiteralBytes (b : ByteArray) : String := Id.run do
```
math-fehr:
```suggestion
def escapeStringLiteral (b : ByteArray) : String := Id.run do
```
I think `Bytes` is not needed since we have it as argument already?

Veir/Parser/MlirParser.lean
```diff
@@ -499,23 +499,24 @@ partial def parseOptionalOp (ip : Option InsertPoint) : MlirParserM (Option Oper
   let results ← parseOpResults
   let opNameStart ← getPos
   let some opName ← parseOptionalStringLiteral | return none
+  let opNameStr := String.fromUTF8! opName
```
math-fehr: Here this will fail if we give non-utf8 operation name.
This was before failing with an "internal error" message.
Can you use `fromUTF8?` here and fail with an error message as well?

Veir/IR/Attribute.lean
```diff
@@ -754,18 +754,37 @@ instance : ToString RegisterType where
 instance : ToString RegisterAttr where
   toString attr := s!"{attr.value} : !riscv.reg"
 
-def escapeStringLiteral (s : String) : String := Id.run do
+-- def escapeStringLiteral (s : String) : String := Id.run do
```
GZGavinZhao: Stray code?

GZGavinZhao: Nevermind, didn't read the PR description.

Veir/Parser/Parser.lean
```diff
@@ -283,28 +283,27 @@ private def processEscapes (input : ByteArray) (basePos : Nat) : Except ParserEr
...
     let processed ← ofExcept (processEscapes raw (token.slice.start.byteOffset + 1))
-    match String.fromUTF8? processed with
-    | some str => return some str
-    | none => throwAt token.slice.start "internal error: failed converting string literal"
+    return some processed
   | none => return none
 
 /--
   Parse a string literal.
   Raise an error if the next token is not a string literal.
```
math-fehr: You should also update this documentation AFAIK

math-fehr (commented):
Thanks a lot! I added a few comments, but otherwise all good.

Veir/IR/Attribute.lean
sishtiaq: Removed commented-out code.

UnitTest/Parser.lean
sishtiaq: You mean an expected negative test like `testParser "123"` `=` `none`? If so, done.

Veir/IR/Attribute.lean
sishtiaq: Yep, gone now.

Veir/IR/Attribute.lean
sishtiaq: Renamed.

Veir/Parser/MlirParser.lean
sishtiaq: Done. I've included `opName` in the exn message, though I'm not sure if lean can toString it?

Veir/Parser/Parser.lean
sishtiaq: Do you mean to say that the comment should say "byte" rather than "string"? If so, done.

sishtiaq:
I've addressed all comments @math-fehr @GZGavinZhao. If you can have a check, and put the PR in to the merge queue please.

math-fehr (approved):
Thanks, that looks good to me, I'll just wait for @GZGavinZhao to review it then I can merge it!

Veir/Parser/MlirParser.lean
```diff
@@ -499,23 +499,25 @@ partial def parseOptionalOp (ip : Option InsertPoint) : MlirParserM (Option Oper
   let results ← parseOpResults
   let opNameStart ← getPos
   let some opName ← parseOptionalStringLiteral | return none
+  let some opNameStr := String.fromUTF8? opName
+    | throwAt opNameStart s!"op '{opName}' not a valid UTF8 string."
```
GZGavinZhao: This would cause an invalid UTF8 error message to print the offending name as a decimal array: `op '[192, 128, 255]' not a valid UTF8 string.`. I think it would be slightly better UX if we do:
```
| throwAt opNameStart s!"op '{escapeStringLiteral opName}' is not a valid UTF-8 string."
```
so the error message looks like: `op '\C0\80\FF' is not a valid UTF-8 string.`

GZGavinZhao (approved):
Looks great! Thanks for implementing this. Just a small error message improvement suggestion, then should be good to merge!

Veir/Parser/MlirParser.lean
sishtiaq: Good point. Should use my own pretty-printer to, er, pretty-print.

sishtiaq: Now done.

math-fehr:
Thanks!

Issue #607, opened 2026-05-19
Add `FunctionOpInterface`


math-fehr:
We currently have both `func.func` and `llvm.func`, which both have a very similar interface.
It would be nice to have the equivalent of `FunctionOpInterface` in MLIR (https://mlir.llvm.org/doxygen/FunctionInterfaces_8h.html) that provides a set of useful functions to query information on these functions (probably not modifying them for now).

naveen-seth:
Closed by #1075

PR #1075, opened 2026-07-13
feat: add FunctionOpInterface

naveen-seth:
Addresses #607.
This adds `FunctionOpInterface`, which provides generic functions for interacting with function-like ops (currently `func.func` and `llvm.func`).

This is intended as an equivalent of MLIR's `FunctionOpInterface` (https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/FunctionInterfaces.td).

This initial PR limits itself to query functions and mostly only implements functions that can simplify some existing code.

This also fixes an issue where VEIR incorrectly accepted `func.func/llvm.func` operations without the required `function_type` attribute due to the previous handling of function arguments.

### DIFF
--- a/Veir/Interfaces/FunctionInterfaces.lean
+++ b/Veir/Interfaces/FunctionInterfaces.lean
@@ -94,7 +94,7 @@ def getArgumentTypes? (funcOp: OperationPtr) (raw: IRContext OpCode) :
   ft.inputs
 
 /-- Returns the number of function results. -/
-public def getNumResults? (funcOp : OperationPtr) (raw : IRContext OpCode) : Option Nat := do
+def getNumResults? (funcOp : OperationPtr) (raw : IRContext OpCode) : Option Nat := do
   rlet ft ← getFunctionType? funcOp raw
   ft.outputs.size
 
--- a/Veir/Interfaces/FunctionInterfaces.lean
+++ b/Veir/Interfaces/FunctionInterfaces.lean
@@ -110,6 +110,6 @@ def getResultTypes! (funcOp : OperationPtr) (raw : IRContext OpCode) : Array Att
 
 end FunctionOpInterface
 
-end -- public section
+end
 
 end Veir
--- a/Veir/Interfaces/FunctionInterfaces.lean
+++ b/Veir/Interfaces/FunctionInterfaces.lean
@@ -1,0 +1,102 @@
+module
+
+/- FunctionOpInterface
+
+  This file provides the `FunctionOpInterface` interface, which provides support
+  for interacting with operations that behave like functions.
+  Currently, this supports llvm.func and func.func.
+
+  Also see:
+  https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/FunctionInterfaces.td
+-/
+
+public import Veir.GlobalOpInfo
+public import Veir.IR.Fields
+
+namespace Veir
+
+public section
+
+namespace FunctionOpInterface
+
+/-- Returns the symbol name of the function. -/
+public def getSymName? (funcOp : OperationPtr) (raw : IRContext OpCode) : Option StringAttr :=
+  match funcOp.getOpType! raw with
+  | .func .func =>
+    (funcOp.getProperties! raw (.func .func) : FuncFuncProperties).sym_name
+  | .llvm .func =>
+    (funcOp.getProperties! raw (.llvm .func) : LLVMFuncProperties).sym_name
+  | _ => none
+
+/-- Returns the type of the function. -/
+public def getFunctionType? (funcOp : OperationPtr) (raw : IRContext OpCode) :
+    Option FunctionType := do
+  match funcOp.getOpType! raw with
+  | .func .func =>
+    let ta ← (funcOp.getProperties! raw (.func .func) : FuncFuncProperties).function_type
+    match ta.val with
+    | .functionType ft => some ft
+    | _ => none
+  | .llvm .func =>
+    let ta ← (funcOp.getProperties! raw (.llvm .func) : LLVMFuncProperties).function_type
+    match ta.val with
+    | .llvmFunctionType ft => some ft
+    | _ => none
+  | _ => none
+
+/-
+  Body Handling
+-/
+
+/-- Returns the region containing the body of this function. -/
+public def getFunctionBody (funcOp : OperationPtr) (raw : IRContext OpCode)
+    (opInBounds : funcOp.InBounds raw := by grind)
+    (hasRegion : 0 < funcOp.getNumRegions raw opInBounds := by grind) : RegionPtr :=
+  funcOp.getRegion raw 0 opInBounds hasRegion
+
+public def getFunctionBody! (funcOp : OperationPtr) (raw : IRContext OpCode) : RegionPtr :=
+  funcOp.getRegion! raw 0
+
+@[grind =_, eq_bang ←]
+theorem getFunctionBody!_eq_getFunctionBody {funcOp : OperationPtr} {raw : IRContext OpCode}
+    {opInBounds} (hasRegion : 0 < funcOp.getNumRegions raw opInBounds) :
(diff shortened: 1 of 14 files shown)


naveen-seth:
Hi, I've only implemented a small subset of the non-modifying functions in MLIR's `FunctionOpInterface` here. Some are easy to add (`isEmpty`, `getNumResults`, ...), but I wanted to get feedback first before adding many functions without finding any potential call sites in the existing code. I could do this in a follow-up or expand this PR.

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+  Body Handling
+-/
+
+/-- Returns the region containing the body of this function. -/
+public def getFunctionBody (funcOp : OperationPtr) (raw : IRContext OpCode)
+    (opInBounds : funcOp.InBounds raw := by grind)
+    (hasRegion : 0 < funcOp.getNumRegions raw opInBounds := by grind) : RegionPtr :=
+  funcOp.getRegion raw 0 opInBounds hasRegion
+
+public def getFunctionBody! (funcOp : OperationPtr) (raw : IRContext OpCode) : RegionPtr :=
```
math-fehr: Can you add documentation on this function as well?

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+
+grind_pattern getFunctionBody!_inBounds => (getFunctionBody! funcOp raw), raw.FieldsInBounds
+
+/-- Returns the first block in the body region. -/
+public def getFirstBlock? (funcOp : OperationPtr) (raw : IRContext OpCode) : Option BlockPtr :=
+  ((getFunctionBody! funcOp raw).get! raw).firstBlock
+
+/-- Returns the last block in the body region. -/
+public def getLastBlock? (funcOp : OperationPtr) (raw : IRContext OpCode) : Option BlockPtr :=
+  ((getFunctionBody! funcOp raw).get! raw).lastBlock
```
math-fehr: Can you rename `getFirstBlock` to `getEntryBlock`?
Also, I would remove `getLastBlock?`. It shouldn't have any big uses, and I think it would be better if users manually fetch it rather than exposing an API for that.
In general, the first block is important (as it is the entry block), but the last block doesn't have a particular meaning.

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/-
+  Argument and Result Handling
+-/
```
math-fehr:
```suggestion
/-!
## Argument and Result Handling
-/
```

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+  | .llvm .func =>
+    let ta ← (funcOp.getProperties! raw (.llvm .func) : LLVMFuncProperties).function_type
+    match ta.val with
+    | .llvmFunctionType ft => some ft
+    | _ => none
+  | _ => none
+
+/-
+  Body Handling
+-/
```
math-fehr:
```suggestion
/-!
## Body Handling
-/
```

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/- FunctionOpInterface
+
+  This file provides the `FunctionOpInterface` interface, which provides support
+  for interacting with operations that behave like functions.
+  Currently, this supports llvm.func and func.func.
+
+  Also see:
+  https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/FunctionInterfaces.td
+-/
```
math-fehr:
```suggestion
/-!
# FunctionOpInterface

This file provides the `FunctionOpInterface` interface, which provides support
for interacting with operations that behave like functions.
Currently, this supports llvm.func and func.func.

Also see:
https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/FunctionInterfaces.td
-/
```

The `/-!` comments are specific for markdown, they are quite nice

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/-
+  Argument and Result Handling
+-/
+
+/-- Returns the number of function arguments. -/
+public def getNumArguments? (funcOp : OperationPtr) (raw : IRContext OpCode) : Option Nat := do
+  rlet ft ← getFunctionType? funcOp raw
+  ft.inputs.size
+
```
math-fehr, line 89: Would it make sense to add `getArgumentTypes?` as well?
And `getNumResults?` too?

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,102 @@
...
+  Also see:
+  https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/FunctionInterfaces.td
+-/
+
+public import Veir.GlobalOpInfo
+public import Veir.IR.Fields
+
+namespace Veir
+
+public section
```
math-fehr, line 19: Because you have `public section` here, you can remove all `public` from the definitions below

Veir/Passes/CastsReconciliation/Reconciliation.lean
```diff
@@ -148,13 +149,11 @@ def coerceFunction (ctx : WfIRContext OpCode) (funcOp : OperationPtr) :
...
   let mut ctx := ctx
-  let region := funcOp.getRegion! ctx.raw 0
-  let some entry := (region.get! ctx.raw).firstBlock | return ctx
+  let some entry := FunctionOpInterface.getFirstBlock? funcOp ctx.raw | return ctx
   let returnCode := returnOpCodeFor (funcOp.getOpType! ctx.raw)
   -- Default the output types to the currently-declared ones, then flip coerced positions.
   -- This preserves non-integer results and `llvm.func`'s `void` return.
-  let declaredFt := (readFunctionType? ctx.raw funcOp).getD { inputs := #[], outputs := #[] }
-  let mut outputs : Array Attribute := declaredFt.outputs
+  let mut outputs : Array Attribute := (FunctionOpInterface.getResultTypes? funcOp ctx.raw).getD #[]
```
math-fehr: Would it make sense to add `getResultTypes!` as well for these kind of cases?
You don't have to do it in this PR btw, just some observation.

math-fehr (commented):
Really cool, this cleans quite a few parts in the project!
I just added some style/API comments, but otherwise the structure looks solid so far!

Veir/Interfaces/FunctionInterfaces.lean
naveen-seth: Added, thanks!

Veir/Passes/CastsReconciliation/Reconciliation.lean
naveen-seth: Added.

naveen-seth:
Thanks for reviewing! There were some failing tests where `func.func` was missing the `function_type` attribute. Previously they worked fine because the entry block arguments were used to infer the signature instead. 
I’ve added an invariant check to the verifier too.

Veir/Interfaces/FunctionInterfaces.lean
```diff
@@ -0,0 +1,115 @@
...
+  rlet ft ← getFunctionType? funcOp raw
+  ft.outputs
+
+/-- Returns the result types of the function. -/
+def getResultTypes! (funcOp : OperationPtr) (raw : IRContext OpCode) : Array Attribute :=
+  (getResultTypes? funcOp raw).get!
+
+end FunctionOpInterface
+
+end -- public section
```
luisacicolini:
```suggestion
end
```

luisacicolini (approved):
reviewed for style, lgtm

math-fehr (approved):
Nice, thanks a lot!

PR #1079, opened 2026-07-14
feat: Verify sym_name attribute on func.func and llvm.func

naveen-seth:
`func.func` and `llvm.func` require the `sym_name` attribute, but this is not currently enforced and many tests do not declare it.
This is now enforced by the verifier.

This PR is split into two commits to make review easier:
The first adds the check itself and tests for the expected error message.

The second updates existing tests that were missing `sym_name`.
This was automated with a script: functions are now named `foo`, `bar`, `baz`, or `f0`, `f1`, ... when there are more than three.

### DIFF
--- a/Test/Func/identity.mlir
+++ b/Test/Func/identity.mlir
@@ -2,7 +2,7 @@
 
 "builtin.module"() ({
   ^4():
-    "func.func"()  <{function_type = (i32) -> i32}> ({
+    "func.func"()  <{function_type = (i32) -> i32, sym_name="foo"}> ({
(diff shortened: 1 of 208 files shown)


luisacicolini (approved):

naveen-seth:
Hi, thanks for reviewing! I'm not able to add to the merge queue yet. Can you do this for me?

luisacicolini:
yes, I was waiting for @math-fehr to take a second look if possible

math-fehr (approved):
Thanks a lot!

PR #1081, opened 2026-07-15
fix(IR): RegionPtr.setParent should take an Option

ineol:


tobiasgrosser (approved):
Let's see what @math-fehr says.

math-fehr (approved):

PR #1082, opened 2026-07-15
feat(veir): make eraseOp keep going, if it makes more things dead

regehr:
the annoying thing here is how many tests it affects, but I think this is a good change. what we want is to not need to constantly run DCE, but rather only after we unavoidably create dead code.


Veir/IR/Basic.lean
```diff
@@ -1085,6 +1085,21 @@ theorem getProperties!_eq_of_OperationPtr_get!_eq {op : OperationPtr} :
...
+  NOTE: ¬ hasSideEffects does not imply that an operation is safe to
+        speculate. For that we also need it to never trigger immediate
+        UB. We'll have to deal with this later on.
+
+  Also see:
+  https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/
+-/
+def hasSideEffects (op : OperationPtr) (ctx : IRContext OpInfo) : Bool :=
+  let opType := op.getOpType! ctx
+  HasOpInfo.hasSideEffects opType (op.getProperties! ctx opType)
```
math-fehr: Can you move this to the new `Interfaces/` folder, to a new file?
I'm trying to have only pure data structure things in `IR/Basic.lean`

Also, I would probably change the documentation to not necessarily talk about DCE, but mentionning here that `hasSideEffects` means that the operation has no side-effects, besides `UB` (just so the code stays up to date at all time).

Veir/IR/OpInfo.lean
```diff
@@ -50,6 +50,13 @@ its properties
 class HasOpInfo (opCode: Type)
     extends Hashable opCode, Repr opCode, Inhabited opCode, HasDialectOpInfo opCode where
   moduleOpCode: opCode
+  /--
+  Whether an operation with this opcode and these properties may have effects
+  that make it ineligible for DCE and other transformations that add / remove /
+  rearrange instructions (terminators count as having effects). Defaults to
+  `true` for every opcode, which conservatively disables such transformations.
```
math-fehr: Same here about DCE

Veir/PatternRewriter/Basic.lean
```diff
@@ -209,28 +209,64 @@ def insertOp! (rewriter: PatternRewriter OpInfo) (op: OperationPtr) (ip : Insert
...
+  let mut worklist := rewriter.worklist.remove op
+  for operand in op.getOperands ctx hOp do
+    let some defOp := operand.getDefiningOp! ctx | continue
+    if useChainHasAtMostOneUserBesides ctx (operand.getFirstUse! ctx) op none 1_000_000_000 then
+      worklist := worklist.push defOp
+  return { rewriter with
     ctx := WfRewriter.eraseOp rewriter.ctx op opRegions opUses hOp,
     hasDoneAction := true,
-    worklist := rewriter.worklist.remove op,
+    worklist
```
math-fehr, line 258: Do you know the reason why we first walk the op operands, then we erase the operation?
Compared to first erasing the operation, then checking if some of the values are unused at that point?
That would make this code simpler

Veir/PatternRewriter/Basic.lean
```diff
@@ -209,28 +209,64 @@ def insertOp! (rewriter: PatternRewriter OpInfo) (op: OperationPtr) (ip : Insert
...
+  if hOp : op.InBounds rewriter.ctx.raw then
+    if opRegions : op.getNumRegions! rewriter.ctx.raw = 0 then
+      if opUses : !op.hasUses! rewriter.ctx.raw then
+        rewriter.eraseOp op opRegions opUses hOp
+      else
+        panic! "PatternRewriter.eraseOp! failed: operation has uses"
+    else
+      panic! "PatternRewriter.eraseOp! failed: operation has regions"
+  else
+    panic! "PatternRewriter.eraseOp! failed: operation is out of bounds"
```
math-fehr, line 275: What's the reason of changing this?
I would agree that I prefer this version (and same for all the other ops here), but I don't understand why you need this change

Veir/PatternRewriter/Basic.lean
```diff
@@ -375,8 +411,14 @@ private partial def RewritePattern.applyOnceInContext
     let (opOpt, newWorklist) := rewriter.worklist.pop
     let op := opOpt.get!
     rewriter := { rewriter with worklist := newWorklist }
-    if _ : op.InBounds rewriter.ctx.raw then
-      rewriter ← pattern rewriter op (by grind)
+    if hin : op.InBounds rewriter.ctx.raw then
+      -- Erase trivially dead operations directly, as in MLIR's greedy driver.
+      if hdead : op.getNumRegions! rewriter.ctx.raw = 0
+          ∧ !op.hasUses! rewriter.ctx.raw
+          ∧ !op.hasSideEffects rewriter.ctx.raw then
```
math-fehr: Should you add a `isTriviallyDead` function (that you could put in `Interface/` as well) for this?
This would also help me a lot since proving deletion of a `isTriviallyDead` operation would be a helper function.

math-fehr (commented):
Nice! I added a few style/organization comments, but otherwise all good!

Veir/IR/Basic.lean
regehr: not relevant right now but: I have an item on my TODO list to get the side effect zoo straightened out in Veir.

and separately, of course we want to delegate this code to dialect-specific specification mechanisms.

Veir/PatternRewriter/Basic.lean
regehr: I guess I'm just not seeing how that should work. isn't WfRewriter.eraseOp going to delete everything including the operand list? so after we do that how do we know which values to look at?

Veir/PatternRewriter/Basic.lean
regehr: well, the old `eraseOp!` went directly to `WfRewriter.eraseOp!`, bypassing the higher-level `eraseOp` in this file. so that part of the change is necessary for `eraseOp!` to get the new recursive behavior.

Veir/PatternRewriter/Basic.lean
regehr: done!

Veir/PatternRewriter/Basic.lean
```diff
@@ -216,28 +209,64 @@ def insertOp! (rewriter: PatternRewriter OpInfo) (op: OperationPtr) (ip : Insert
...
+      let useStruct := use.get! ctx
+      let owner := useStruct.owner
+      if owner = exceptOp ∨ otherUser = some owner then
+        useChainHasAtMostOneUserBesides ctx useStruct.nextUse exceptOp otherUser maxIteration
+      else if otherUser.isNone then
+        useChainHasAtMostOneUserBesides ctx useStruct.nextUse exceptOp (some owner) maxIteration
+      else
+        false
+    | none => true
+  | 0 => false
```
math-fehr: I think it would be probably better to return an `Option Bool` here, to indicate when the "fuel" expired.
In general, what would also work is using a `partial def` that returns the option, and that way you don't even have to provide a fuel, you can just use normal recursion.

math-fehr (approved):
Thanks, I just have one nit about how to define a function so that we can reason about it, but otherwise all good!

math-fehr (approved):

PR #1083, opened 2026-07-15
Update Lean to 4.32

ineol:
------

The gigantic proofs in `Veir/Passes/InstructionSelection/RewriteProofs/` break.


math-fehr:
Should be better now that these are removed?

math-fehr (approved):
Nice!

PR #1084, opened 2026-07-15
feat(veir): constant folding

regehr:
this is the code side of constant folding. there's a proof side as well that I'll submit separately, once we've agreed on a design.

how this works is: folding is a library that is invoked from `canonicalize`, but also available for other clients such as SCCP, which will get updated to use the new interface.

the folder itself uses the interpreter to do the folding, which I think is nice because it avoids creating a copy of the semantics.

the line between "folding" and "instcombine" is that the folder never creates new operations aside from constants. I think this is defensible and good. so it does fold `add x, 0 -> x`.

the folder is poison-respecting. it folds immediate UB to poison. I believe this is the correct design.


regehr:
note: I'm not trying to merge this PR, but rather will break off pieces of it to make code review easier

Issue #1074, opened 2026-07-13
`llvm.mlir.global` initializer should `return`


sishtiaq:
Veir's verifier rejects a `llvm.return` inside a `llvm.mlir.global`'s initializer region. 
The check is too strict: it should accept/require `llvm.return` inside either `llvm.func` or `llvm.mlir.global`.

This is a testcase from sqlite3.c:
```
// RUN: VEIR_UNREGISTERED_ROUNDTRIP

"builtin.module"() ({
  "llvm.mlir.global"() <{addr_space = 0 : i32, alignment = 1 : i64, constant, dso_local, global_type = !llvm.array<1 x struct<"struct.et_info", (i8, i8)>>, linkage = #llvm.linkage<internal>, sym_name = "fmtinfo", unnamed_addr = 0 : i64, visibility_ = 0 : i64}> ({
    %0 = "llvm.mlir.constant"() <{value = 10 : i8}> : () -> i8
    %1 = "llvm.mlir.constant"() <{value = 114 : i8}> : () -> i8
    %2 = "llvm.mlir.undef"() : () -> !llvm.struct<"struct.et_info", (i8, i8)>
    %3 = "llvm.insertvalue"(%2, %1) <{position = array<i64: 0>}> : (!llvm.struct<"struct.et_info", (i8, i8)>, i8) -> !llvm.struct<"struct.et_info", (i8, i8)>
    %4 = "llvm.insertvalue"(%3, %0) <{position = array<i64: 1>}> : (!llvm.struct<"struct.et_info", (i8, i8)>, i8) -> !llvm.struct<"struct.et_info", (i8, i8)>
    %5 = "llvm.mlir.undef"() : () -> !llvm.array<1 x struct<"struct.et_info", (i8, i8)>>
    %6 = "llvm.insertvalue"(%5, %4) <{position = array<i64: 0>}> : (!llvm.array<1 x struct<"struct.et_info", (i8, i8)>>, !llvm.struct<"struct.et_info", (i8, i8)>) -> !llvm.array<1 x struct<"struct.et_info", (i8, i8)>>
    "llvm.return"(%6) : (!llvm.array<1 x struct<"struct.et_info", (i8, i8)>>) -> ()
  }) : () -> ()
}) : () -> ()
```

sueszli:
Resolved by #1085

sishtiaq:
Thank you @sueszli!

sueszli:
[...]

PR #1085, opened 2026-07-15
fix(llvm): verify llvm.return in llvm.mlir.global initializers

sueszli:
Patches: https://github.com/opencompl/veir/issues/1074

> Veir's verifier rejects a llvm.return inside a llvm.mlir.global's initializer region.
> The check is too strict: it should accept/require llvm.return inside either llvm.func or llvm.mlir.global.


Test/LLVM/global_initializer.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: VEIR_UNREGISTERED_ROUNDTRIP
+
+// Reduced from sqlite3.c
+"builtin.module"() ({
```
sueszli, line 4: This is the exact snippet that @sishtiaq provided.

Happy to break down into a more deliberate test case or add `#guard` statements or `theorem`s.

sishtiaq: It'd be good to have this as a test (in `Test`), as you've added. 
But can also be a `#guard`-based unit test (in `UnitTest`?).

math-fehr: I would prefer keeping things in `.mlir` files as much as possible, as I find them easier to read (parsing an entire module is really ugly in our unit tests).

math-fehr:
Looks good to me, I would just probably add a failing test for `llvm.mlir.global`?

Veir/Verifier.lean
```diff
@@ -77,6 +73,37 @@ def OperationPtr.verifyLLVMReturnTypes (op : OperationPtr) (ctx : WfIRContext Op
...
+  Check an `llvm.return`'s operands against its enclosing `llvm.func` or `llvm.mlir.global`.
+-/
+def OperationPtr.verifyLLVMReturnTypes (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) : Except String PUnit := do
+  let enclosingOp ← op.getEnclosingFunctionOp ctx "llvm.return"
+  let badEnclosure : Except String PUnit :=
+    throw "Expected llvm.return to be enclosed by llvm.func or llvm.mlir.global"
+  match enclosingOp.getOpType! ctx.raw with
+  | .llvm .func => op.verifyLLVMFuncReturnTypes ctx opIn enclosingOp
+  | .builtin .unregistered =>
```
sishtiaq, line 101: Such a tedious way (`.builtin .unregistered`) to get the enclosing scope! Not your code, just how MLIR represents this construct.

sishtiaq (approved):

Test/LLVM/global_initializer.mlir
sueszli: Based on your PR approval, I assume that sticking to MLIR tests is fine.

Did I interpret this correctly? @sishtiaq

sueszli:
Thank you so much for the review @math-fehr @sishtiaq 

I'll add the negative test as well now

Veir/Verifier.lean
sueszli: I agree 😅

math-fehr (approved):
Thanks!

PR #1093, opened 2026-07-16
feat(dominance): add propositional CFG paths and reachability definitions

math-fehr:
This PR defines a propositional definition of paths through a region's CFG (`RegionPtr.Path`),
and the propositional definition of reachability for blocks inside a region
(`BlockPtr.ReachableFromEntry`).

These two definitions are central to the definition of dominance, as dominance
inside a region is defined in terms of paths from the region's entry block.


tobiasgrosser (approved):

math-fehr:
Closing because I'm changing the order/organization for upstreaming.

PR #1094, opened 2026-07-17
fix(parser): restore block state when parsing empty regions

### LINKED ISSUE #1000: bug in parser when parsing empty regions
In the parser, the lines:
```
  /- Case where there are no blocks inside the region. -/
  if (← parseOptionalPunctuation "}") then
    return region
```

skip this change:
```
  modifyThe MlirParserState fun s => {s with blocks := oldBlocks}
```
which should be called before we exit the region

sueszli:
Fixes #1000 (just a missing branch in parser)


math-fehr (approved):
Perfect, thanks a lot!

PR #1096, opened 2026-07-17
feat: add --disable-verifiers flag to veir-opt

### LINKED ISSUE #819: veir-opt should have an option `--disable-verifiers`
It is useful to print unverified IR while debugging.

naveen-seth:
Closes #819.
This allows printing unverified IR for debugging.

### DIFF
--- a/Test/test-verifier.mlir
+++ b/Test/test-verifier.mlir
@@ -1,10 +1,9 @@
 // RUN: not veir-opt %s 2>&1 | filecheck %s
+// RUN: veir-opt --disable-verifiers %s 2>&1 | filecheck --check-prefix=CHECK-NO-VERIFY %s
 
-// CHECK: Error verifying input program: Expected 1 region
 "builtin.module"() ({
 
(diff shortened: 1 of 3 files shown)


Test/test-verifier.mlir
```diff
@@ -4,3 +4,7 @@
 "builtin.module"() ({
 
 }, {}) : () -> ()
+
+// RUN: veir-opt --disable-verifiers %s 2>&1 | filecheck --check-prefix=CHECK-NO-VERIFY %s
```
math-fehr: Can you add this line at the top of the file as well?

math-fehr (approved):
Very cool! I just had one very nit comment on a test, but otherwise all good!

Test/test-verifier.mlir
naveen-seth: Thanks for reviewing!
Also moved the check line down to better match the other tests, hope that's right.

math-fehr: All good!

PR #1101, opened 2026-07-17
fix: properly handle FuncFuncProperties.function_type

naveen-seth:
In #1075, I added `function_type` to `FuncFuncProperties` to align it with `LLVMFuncProperties`. This was done because it simplifies the newly added `FunctionOpInterface` and because MLIR’s `FunctionOpInterface` treats function_type as a required ODS argument.

This wasn’t done properly: `function_type` remained duplicated in `FuncFuncProperties.extra`, and `FuncFuncProperties.function_type` was not written out in `Properties.toAttrDict`.
The duplication caused the bug to go undetected.


math-fehr (approved):
Oops, my bad I missed that!

PR #1105, opened 2026-07-18
fix(verifier): report operation name in failure messages

### LINKED ISSUE #976: Report name of operation in all verifier failure messages
During debugging yesterday, at one point we simply got "Expected 2 results" as output from veir-opt, with no indication _which op_ was expecting the two results. We found it eventually (there's only a handful that do) via Cmd+F in Verifier.lean, but it made me realise that, at a minimum, each verifier failure should report the operation name.

sueszli:
Closes #976

### DIFF
--- a/Veir/Verifier.lean
+++ b/Veir/Verifier.lean
@@ -1013,12 +1013,17 @@ public section
 -/
 def WfIRContext.verify (ctx : WfIRContext OpCode) : Except String Unit := do
   ctx.raw.forOpsDepM (fun op opIn => do
-    op.verifyLocalInvariants ctx opIn
-    if let .riscv _ := op.getOpType ctx.raw opIn then
-      op.verifyRISCVRegisterTypes ctx opIn
-    match (op.get ctx.raw opIn).parent with
-    | some _ => op.verifyTerminatorPosition ctx opIn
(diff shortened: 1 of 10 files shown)


Veir/Verifier.lean
```diff
@@ -1013,12 +1013,17 @@ public section
 -/
 def WfIRContext.verify (ctx : WfIRContext OpCode) : Except String Unit := do
   ctx.raw.forOpsDepM (fun op opIn => do
-    op.verifyLocalInvariants ctx opIn
-    if let .riscv _ := op.getOpType ctx.raw opIn then
-      op.verifyRISCVRegisterTypes ctx opIn
-    match (op.get ctx.raw opIn).parent with
-    | some _ => op.verifyTerminatorPosition ctx opIn
-    | none => pure ())
+    let opType := op.getOpType ctx.raw opIn
```
sueszli, line 1016: This is less error prone than adding the op name to each throw string across 130ish sites across Verifier.lean.

tobiasgrosser (approved):

math-fehr (approved):
Very cool, thanks a lot!

sueszli:
Thanks for reviewing!

PR #1109, opened 2026-07-22
chore: separation of matchers

luisacicolini:
This PR separates the matchers for the RISCV and LLVM dialect ops


math-fehr (approved):
Nice, thanks for cleaning this up!

PR #1110, opened 2026-07-22
feat(ir): Add `OperationPtr.getParentRegion!`

math-fehr:
This is a simple definition to grab a parent region without having to first grab a parent block. It is purely an API to make things simpler, not faster.

### DIFF
--- a/Veir/IR/Basic.lean
+++ b/Veir/IR/Basic.lean
@@ -2560,10 +2560,29 @@ end BlockOperandPtrPtr
 
 namespace OperationPtr
 
+/-- Return the region directly containing an operation, if one exists. -/
+def getParentRegion! (op : OperationPtr) (ctx : IRContext OpInfo) : Option RegionPtr := do
+  let block ← (op.get! ctx).parent
+  (block.get! ctx).parent
+
+@[grind =]
+theorem getParentRegion!_eq_some_iff {op : OperationPtr} :
+    op.getParentRegion! ctx = some region ↔
+      ∃ block,
+        (op.get! ctx).parent = some block ∧
+        (block.get! ctx).parent = some region := by
+  simp only [OperationPtr.getParentRegion!, bind, Option.bind]
+  grind
+
+theorem getParentRegion!_eq_some_of_parent_of_parent {op : OperationPtr} :
+    (op.get! ctx).parent = some block →
+    (block.get! ctx).parent = some region →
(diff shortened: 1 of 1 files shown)


luisacicolini (approved):

Veir/IR/Basic.lean
```diff
@@ -2560,10 +2560,28 @@ end BlockOperandPtrPtr
 
 namespace OperationPtr
 
+/-- Return the region directly containing an operation, if one exists. -/
+def getParentRegion! (op : OperationPtr) (ctx : IRContext OpInfo) : Option RegionPtr := do
+  rlet block ← (op.get! ctx).parent
+  (block.get! ctx).parent
```
```suggestion
  let block ← (op.get! ctx).parent
  (block.get! ctx).parent
```
I don't think we need to know anything inside the function?

Veir/IR/Basic.lean
```diff
@@ -2560,10 +2560,28 @@ end BlockOperandPtrPtr
...
+    (op.get! ctx).parent = some block →
+    (block.get! ctx).parent = some region →
+    op.getParentRegion! ctx = some region := by
+  grind [OperationPtr.getParentRegion!]
+
 @[expose]
 def getParentOp! (op : OperationPtr) (ctx : IRContext OpInfo) : Option OperationPtr := do
-  rlet block ← (op.get! ctx).parent
-  rlet region ← (block.get! ctx).parent
+  rlet region ← op.getParentRegion! ctx
```
ineol: same?

Veir/IR/Basic.lean
```diff
@@ -2572,7 +2590,7 @@ theorem getParentOp!_eq_some_iff {child parent : OperationPtr} {ctx : IRContext
         (child.get! ctx).parent = some block ∧
         (block.get! ctx).parent = some region ∧
         (region.get! ctx).parent = some parent := by
-  grind [OperationPtr.getParentOp!]
+  grind [OperationPtr.getParentOp!, OperationPtr.getParentRegion!]
```
ineol: it's not important but it's annoying that `getParentRegion!_eq_some_iff` is not enough to prove this without unfolding.

ineol (approved):
LGTM, just small nitpicks

Veir/IR/Basic.lean
math-fehr: Switching to `let`, this works now somehow. I have to unfold `bind` and `Option.bind` though.

PR #1111, opened 2026-07-23
feat(dominance): Define operation and block dominance

math-fehr:
These definitions follow compcertssa definition of dominance (https://compcertssa.gitlabpages.inria.fr/html/compcert.midend.Dom.html) but extended for MLIR's hierarchical dominance.

This PR only contains the definitions, and do not replace yet the `axiom` in `Dominance.lean`, nor introduce lemmas about it.

### DIFF
--- a/Veir/Dominance/Basic.lean
+++ b/Veir/Dominance/Basic.lean
@@ -1,0 +1,246 @@
+module
+
+public import Veir.Interfaces.RegionKindInterfaces
+public import Veir.IRNesting
+
+import Veir.IR.InBounds
+
+/-!
+# Dominance Definitions
+
+Core propositional definitions for control-flow paths, reachability, and dominance.
+
+The implementation of dominance in a region is based on CompcertSSA's definition
+(https://compcertssa.gitlabpages.inria.fr/html/compcert.midend.Dom.html). It is extended
+to support dominance between operations in different regions, following MLIR's informal
+definition.
+-/
+
+public section
+
+namespace Veir
+
+variable {OpInfo : Type} [HasOpInfo OpInfo]
+
+/-!
+## CFG Paths and Block Dominance
+
+Block dominance is defined in terms of CFG paths between blocks of the same region.
+A CFG path (`RegionPtr.Path`) is a nonempty list of blocks that traverse the CFG edges
+inside a region.
+
+From the definition of CFG paths, we define a dominance relation between blocks in a region
+(`BlockPtr.ProperlyDominatesInRegion`). In SSACFG regions, a block `A` properly dominates a block
+`B` (`BlockPtr.ProperlyDominatesInSSACFGRegion`) if every CFG path from the region's entry block to
+`B` contains `A`. In graph regions, all blocks properly dominate
+(`BlockPtr.ProperlyDominatesInGraphRegion`) each others. Proper dominance
+(`BlockPtr.ProperlyDominates`) is then defined as the
+property that every CFG path from the region's entry to a dominated block must contain the
+(distinct) dominator block.
+
+Proper dominance (`BlockPtr.ProperlyDominates`) in a region is then extended across regions by
+considering ancestors of the dominated block. If an ancestor of the dominated block is properly
+dominated in a region by the dominator block, then the dominator block properly dominates the
+dominated block. Optionally, a flag can allow the dominator block to be an ancestor of the dominated
+block.
+
+Dominance is then defined as the reflexive closure of proper dominance, allowing a block to dominate
+itself.
+-/
+
+/--
+A nonempty CFG path between two (possibly equal) blocks in `region`.
+
+The path is witnessed by a list of blocks, which is nonempty, and contains the
+source and target blocks as its first and last elements, respectively. Every block
+in the path has `region` as its parent, and every consecutive pair is related by
+a CFG edge.
+-/
+inductive RegionPtr.Path (region : RegionPtr) (ctx : WfIRContext OpInfo) :
+    BlockPtr → BlockPtr → List BlockPtr → Prop where
+  | Single {block : BlockPtr}
+      (parent : (block.get! ctx.raw).parent = some region) :
+      region.Path ctx block block [block]
+  | Cons {source next target : BlockPtr} {blocks : List BlockPtr}
+      (parent : (source.get! ctx.raw).parent = some region)
+      (successor : next ∈ source.getSuccessors! ctx.raw)
+      (tail : region.Path ctx next target blocks) :
+      region.Path ctx source target (source :: blocks)
+
+/--
+Proper dominance between `dominator` and `dominated` in a graph `region`.
+
+This is defined as the property that both blocks are in the same region, and that the region is a
+graph region. In practice, if the context is verified, this means that both blocks are also equal.
+-/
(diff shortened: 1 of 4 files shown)


Veir/Dominance/Basic.lean
```diff
@@ -0,0 +1,175 @@
...
+dominator and dominated blocks as its first and last elements, respectively. Every block
+in the path has `region` as its parent, and every consecutive pair is related by
+a CFG edge.
+-/
+inductive RegionPtr.Path (region : RegionPtr) (ctx : WfIRContext OpInfo) :
+    BlockPtr → BlockPtr → List BlockPtr → Prop where
+  | single {block : BlockPtr}
+      (parent : (block.get! ctx.raw).parent = some region) :
+      region.Path ctx block block [block]
+  | cons {dominator next dominated : BlockPtr} {blocks : List BlockPtr}
```
ineol: I don't think we should use dominator/dominated in this definition?

Veir/Dominance/Basic.lean
```diff
@@ -0,0 +1,175 @@
...
+  ∃ region,
+    (dominator.get! ctx.raw).parent = some region ∧
+    (dominated.get! ctx.raw).parent = some region ∧
+    ∀ entry blocks,
+      (region.get! ctx.raw).firstBlock = some entry →
+      region.Path ctx entry dominated blocks →
+      dominator ∈ blocks
+
+/-- Proper dominance between blocks. It is defined as dominance plus inequality. -/
+def BlockPtr.ProperlyDominates (dominator dominated : BlockPtr)
```
ineol: Maybe add "strict dominance" in the doc string in case a non llvm person greps for it?

luisacicolini (approved):
lgtm but I don't have enough context knowledge to have an opinion about the definitions themselves.

Veir/Dominance/Basic.lean
```diff
@@ -0,0 +1,175 @@
...
+    (ctx : WfIRContext OpInfo) : Prop where
+  | block {block : BlockPtr}
+      (dominance : dominator.DominatesInSSACFGBlock dominated block region ctx) :
+      DominatesInSSACFGRegion dominator dominated region ctx
+  | blocks {dominatorBlock dominatedBlock : BlockPtr}
+      (dominatorParent : (dominator.get! ctx.raw).parent = some dominatorBlock)
+      (dominatedParent : (dominated.get! ctx.raw).parent = some dominatedBlock)
+      (dominatorRegion : dominator.getParentRegion! ctx.raw = some region)
+      (dominatedRegion : dominated.getParentRegion! ctx.raw = some region)
+      (regionSSA : region.hasSSADominance ctx = true)
```
ineol: is it necessary?  I would have thought that since `region.hasSSADominance = false` implies a single block, `dominatorBlock.ProperlyDominates dominatedBlock` implies `region.hasSSADominance = true`

Veir/IR/InBounds.lean
```diff
@@ -27,6 +27,18 @@ variable {ctx ctx' : IRContext OpInfo}
...
+@[grind →]
+theorem OperationPtr.in_bounds_of_parent_eq_some
+    {op : OperationPtr} {block : BlockPtr} {ctx : IRContext OpInfo}
+    (parentEq : (op.get! ctx).parent = some block) :
+    op.InBounds ctx := by
+  apply Classical.byContradiction
+  intro opOutOfBounds
+  have opDefault : op.get! ctx = default := by grind
+  simp only [opDefault] at parentEq
+  contradiction
```
ineol:
```suggestion
  grind [Operation.default_parent_eq]
```
after adding `Operation.default_parent_eq` with the related lemma in `Basic.lean`.

ineol (approved):
Looks like sound defs to me!

Veir/Dominance/Basic.lean
math-fehr: So this is the case only if we have `ctx.Verified`. I felt that it was more precise with this definition, but I'm fine changing it to the definition where we have `ctx.Verified`.
In all case, I want these definitions to be entirely opaque, so that lemmas would be the one exposing these things.
Let me know if you prefer the `ctx.Verified` definition.

Veir/Dominance/Basic.lean
math-fehr: Oops, fixed here!

regehr:
I'm not sure about proper dominance in graph regions.

MLIR contrains this code in `Dominance.cpp`:
```c++
template <bool IsPostDom>
bool DominanceInfoBase<IsPostDom>::properlyDominatesImpl(
    Block *aBlock, Block::iterator aIt, Block *bBlock, Block::iterator bIt,
    bool enclosingOk) const {
  assert(aBlock && bBlock && "expected non-null blocks");

  // A block iterator (post)dominates, but does not properly (post)dominate,                                  
  // itself unless this is a graph region.                                                                    
  if (aBlock == bBlock && aIt == bIt)
    return !hasSSADominance(aBlock);
```

doesn't this make it return true for the reflexive, graph region case, whereas your code returns false?

regehr:
also maybe remove comments about strict dominance since MLIR seems to only talk about proper dominance

math-fehr:
> doesn't this make it return true for the reflexive, graph region case, whereas your code returns false?

Ohhh, I'm going to take a bit trying to understand that, I probably have misunderstood what proper dominance is, and I think I'll go read more info about it, thanks a lot for spotting it!
Probably proper dominance is not just `dominance and different` for graph regions then, and is something more subtle!

regehr:
> Ohhh, I'm going to take a bit trying to understand that, I probably have misunderstood what proper dominance is, and I think I'll go read more info about it, thanks a lot for spotting it! Probably proper dominance is not just `dominance and different` for graph regions then, and is something more subtle!

I'm not clear on why MLIR made this choice, though

regehr (approved):
besides the one thing, LGTM!

Veir/Dominance/Basic.lean
```diff
@@ -0,0 +1,242 @@
...
+-/
+inductive BlockPtr.ProperlyDominates (dominator dominated : BlockPtr) (ctx : WfIRContext OpInfo)
+    (enclosingOk : Bool) : Prop where
+  | Ancestor
+      (ancestor : (IRNode.block dominator).Ancestor (.block dominated) ctx)
+      (hNe : dominator ≠ dominated)
+      (hEnclosingOk : enclosingOk = true)
+  | AncestorDominatedInRegion (ancestor : BlockPtr) (region : RegionPtr)
+      (hAncestor : (IRNode.block ancestor).Ancestor (.block dominated) ctx)
+      (h : dominator.ProperlyDominatesInRegion ancestor region ctx)
```
math-fehr: @ineol, is there a way to have both a name for the argument `enclosingOk`, while still defining the argument as an index rather than a parameter (meaning the `Ancestor` case would specify `enclosingOk` as `true`, which should simplify casing?)

regehr (approved):
this LGTM!

Veir/Dominance/Basic.lean
ineol: did you try
```
inductive BlockPtr.ProperlyDominates ... : (enclosingOk : Bool) → Prop
```
or did I misunderstand your question?

math-fehr: No that's perfect, thanks!

PR #1117, opened 2026-07-23
feat(arith): remaining arith dialect interpreter semantics

regehr:
I think this is all correct, but of course we want to be careful since nothing here can be proved, since we lack a definitive formal version of `arith`

after this I'll do a pass that lowers arith to llvm (which would be an excellent place to test out Sid's width-independent proofs) with the goal of being able to end up at RISC-V all the way from ModArith.

### DIFF
--- a/Veir/Interpreter/Basic.lean
+++ b/Veir/Interpreter/Basic.lean
@@ -641,7 +641,124 @@ def Arith.interpretOp' (opType : Veir.Arith) (properties : HasDialectOpInfo.prop
     if h: bw' ≠ bw then none else
     let rhs := rhs.cast (by simpa using h)
     return (#[.int bw (LLVM.Int.select cond lhs rhs)], none)
-  | _ => none
+  | .cmpi => do
+    let [.int bw lhs, .int bw' rhs] := operands.toList | none
+    if h: bw' ≠ bw then none else
+    let rhs := rhs.cast (by simp at h; exact h)
+    -- `arith.cmpi` lowers to `llvm.icmp`; the arith and LLVM predicate encodings
+    -- coincide, so `properties.predicate` is used directly. Result is `i1`.
+    return (#[.int 1 (LLVM.Int.icmp lhs rhs properties.predicate)], none)
+  | .maxsi => do
+    let [.int bw lhs, .int bw' rhs] := operands.toList | none
+    if h: bw' ≠ bw then none else
+    let rhs := rhs.cast (by simp at h; exact h)
+    return (#[.int bw (LLVM.Int.smax lhs rhs)], none)
+  | .minsi => do
+    let [.int bw lhs, .int bw' rhs] := operands.toList | none
+    if h: bw' ≠ bw then none else
+    let rhs := rhs.cast (by simp at h; exact h)
+    return (#[.int bw (LLVM.Int.smin lhs rhs)], none)
+  | .maxui => do
+    let [.int bw lhs, .int bw' rhs] := operands.toList | none
+    if h: bw' ≠ bw then none else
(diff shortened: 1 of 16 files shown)


Veir/Interpreter/Basic.lean
```diff
@@ -641,7 +641,124 @@ def Arith.interpretOp' (opType : Veir.Arith) (properties : HasDialectOpInfo.prop
...
+    let [.int bw a, .int bw' b] := operands.toList | none
+    if h: bw' ≠ bw then none else
+    let b := b.cast (by simp at h; exact h)
+    -- Lowering (arith ExpandOps): `a == 0 ? 0 : ((a - 1) udiv b) + 1`. The
+    -- `udiv` makes a zero (or poison) divisor undefined behaviour, exactly as
+    -- for `arith.divui`.
+    match b with
+    | .poison => Interp.ub
+    | .val bv =>
+      if bv = 0 then Interp.ub
```
math-fehr: does `let .val bv := b | Interp.ub` works here instead of the match?

Veir/Interpreter/Basic.lean
```diff
@@ -641,7 +641,124 @@ def Arith.interpretOp' (opType : Veir.Arith) (properties : HasDialectOpInfo.prop
...
+    let zero : LLVM.Int bw := .val 0
+    let one : LLVM.Int bw := .val 1
+    let z := LLVM.Int.sdiv a b
+    let notExact := LLVM.Int.icmp a (LLVM.Int.mul z b) .ne
+    let signEqual := LLVM.Int.icmp (LLVM.Int.icmp a zero .slt) (LLVM.Int.icmp b zero .slt) .eq
+    let cond := LLVM.Int.and notExact signEqual
+    let result := LLVM.Int.select cond (LLVM.Int.add z one) z
+    -- UB gating mirrors `arith.divsi` (divide-by-zero, INT_MIN / -1).
+    match b with
+    | .poison => Interp.ub
```
math-fehr: same here

Veir/Interpreter/Basic.lean
```diff
@@ -641,7 +641,124 @@ def Arith.interpretOp' (opType : Veir.Arith) (properties : HasDialectOpInfo.prop
...
+    | .val bv =>
+      if bv = 0 then Interp.ub
+      else if bv = -1 then
+        match a with
+        | .poison => Interp.ub
```
math-fehr: same here

Veir/Interpreter/Basic.lean
```diff
@@ -641,7 +641,124 @@ def Arith.interpretOp' (opType : Veir.Arith) (properties : HasDialectOpInfo.prop
...
+    let negOne : LLVM.Int bw := .val (BitVec.allOnes bw)
+    let z := LLVM.Int.sdiv a b
+    let notExact := LLVM.Int.icmp a (LLVM.Int.mul z b) .ne
+    let signOpposite := LLVM.Int.icmp (LLVM.Int.icmp a zero .slt) (LLVM.Int.icmp b zero .slt) .ne
+    let cond := LLVM.Int.and notExact signOpposite
+    let result := LLVM.Int.select cond (LLVM.Int.add z negOne) z
+    -- UB gating mirrors `arith.divsi` (divide-by-zero, INT_MIN / -1).
+    match b with
+    | .poison => Interp.ub
+    | .val bv =>
```
math-fehr: same here

math-fehr (approved):
Very cool! I just added one nitpick on the code syntax, but besides that we can merge it!

regehr:
@math-fehr I went with a different refactoring than the one you suggested, but the effect should be the same!

math-fehr:
That looks much much nicer, thanks!

PR #1118, opened 2026-07-24
feat: verifier rejects arith and llvm operations that use or produce i0 values

regehr:


tobiasgrosser (approved):

math-fehr:
Hmm, while I agree that we should verify these things, I would argue that this verification should be done at the dialect level, rather than adding more cases to `verify`.
To be more precise, I want `verify` to work on `OpInfo` at some point, and that means that we should move away anything dialect-specific.
Do you think you could move these verifier to the dialect themselves?

regehr:
> Do you think you could move these verifier to the dialect themselves?

I understand what you want at a high level but I don't understand what this should look like in the code, since we don't really have dialect-specific verifiers yet. can you explain a bit more?

math-fehr:
Oh sorry, so what I mean is that we should move these to the operations themselves.
Probably like having a function like `checkIsNonNullIntegerType` and using it in every operation verifier?

regehr:
@math-fehr please take another quick look!

Veir/Verifier.lean
```diff
@@ -407,6 +407,25 @@ def OperationPtr.verifyRISCVRegisterTypes (op : OperationPtr) (ctx : WfIRContext
...
+  Whether `i0` is legal is a per-dialect policy, so this is called explicitly
+  from the verifier arm of each operation that forbids it (currently every
+  `arith` and `llvm` operation) rather than being applied dialect-wide.
+-/
+def OperationPtr.checkIsNonNullIntegerType (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) : Except String PUnit := do
+  let instrName := String.fromUTF8! (op.getOpType ctx.raw opIn).name
+  let opTypes := op.getOperandTypes! ctx.raw
+  for i in [0:opTypes.size] do
+    if let .integerType intType := (opTypes[i]!).val then
```
math-fehr, line 421:
```suggestion
  for opType in opTypes do
    if let .integerType intType := opType then
```

Veir/Verifier.lean
```diff
@@ -407,6 +407,25 @@ def OperationPtr.verifyRISCVRegisterTypes (op : OperationPtr) (ctx : WfIRContext
...
+      if intType.bitwidth = 0 then
+        throw s!"{instrName}: operand {i} has forbidden i0 type"
+  for i in [0:op.getNumResults ctx.raw opIn] do
+    if let .integerType intType := ((op.getResult i).get! ctx.raw).type.val then
```
math-fehr, line 425:
```suggestion
  for resType in op.getResultTypes! ctx.raw do
    if let .integerType intType := resType then
```

math-fehr (approved):
Thanks! If my modifications work, you can just merge it as well!

Veir/Verifier.lean
regehr: it did not work but it suggested an improvement anyway, thanks!

PR #1121, opened 2026-07-25
feat: a pass that lowers arith to llvm

regehr:
this contains a couple of suboptimal lowerings where we don't yet support LLVM's x.with.overflow intrinsics. I am working on adding those but it's going to take a while -- 3 or 4 PRs at least -- so I'd like to land this patch and then it'll be a simple fix to use the x.with.overflow intrinsics later on.

### DIFF
--- a/Veir/Passes/ArithToLLVM.lean
+++ b/Veir/Passes/ArithToLLVM.lean
@@ -1,0 +1,299 @@
+module
+
+public import Veir.Pass
+import Veir.PatternRewriter.Basic
+import Veir.Passes.Matching
+import Veir.Properties
+
+namespace Veir
+
+/-!
+  # ArithToLLVM pass
+
+  Lowers every operation of the `arith` dialect into the `llvm` dialect, leaving no
+  `arith` operation behind. This mirrors MLIR's `ArithToLLVM` conversion
+  (`mlir/lib/Conversion/ArithToLLVM/ArithToLLVM.cpp`).
+
+  Unlike MLIR, Veir has no `index` type and no separate LLVM integer type: both
+  dialects operate on the same signless `IntegerType`. The type conversion is
+  therefore the identity, and — unlike `ModArithToArith` — this pass needs no
+  `unrealized_conversion_cast` plumbing at all. Result and operand types are
+  carried over verbatim.
+
+  Each lowering is a direct transcription of the corresponding `arith` interpreter
+  case in `Veir/Interpreter/Basic.lean`, which defines the `arith` semantics in
+  terms of the same `LLVM.Int.*` primitives the target `llvm` ops evaluate to. The
+  three ops with no single LLVM target (`ceildiv*`, `floordivsi`) and the
+  two-result "extended" ops are expanded exactly as those interpreter cases (and
+  MLIR's `ExpandOps` / `MulIExtendedOpLowering`) prescribe.
+-/
+
+/-! ## Emission helpers -/
+
+/-- Emit `llvm.mlir.constant value : i<width>`. -/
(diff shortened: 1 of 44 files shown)


Veir/Passes/ArithToLLVM.lean
```diff
@@ -0,0 +1,292 @@
...
+  # ArithToLLVM pass
+
+  Lowers every operation of the `arith` dialect into the `llvm`
+  dialect. This mirrors MLIR's `ArithToLLVM` conversion
+  (`mlir/lib/Conversion/ArithToLLVM/ArithToLLVM.cpp`).
+
+  Since result and operand types are carried over verbatim, this is a fairly
+  easy lowering.
+
+  TODO: Use the LLVM dialect's x.with.overflow intrinsics once these
```
sishtiaq, line 20: I would already file an issue for this TODO. Then change to comment to point to the issue.

Veir/Passes/ArithToLLVM.lean
```diff
@@ -0,0 +1,292 @@
...
+/-! ## Emission helpers -/
+
+/-- Emit `llvm.mlir.constant value : i<width>`. -/
+def emitLLVMIntConst (rewriter : PatternRewriter OpCode) (value : Int) (width : Nat)
+    (ip : InsertPoint) : Option (PatternRewriter OpCode × ValuePtr) := do
+  let ty : IntegerType := IntegerType.mk width
+  let props : LLVMConstantProperties := { value := .integer (IntegerAttr.mk value ty) }
+  let (rewriter, op) ← rewriter.createOp! (.llvm .mlir__constant)
+    #[(ty : TypeAttr)] #[] #[] #[] props (some ip)
+  return (rewriter, (op.getResult 0 : ValuePtr))
```
math-fehr:
```suggestion
  return (rewriter, op.getResult 0)
```
Does this work by any chance? (I don't know)

Veir/Passes/ArithToLLVM.lean
```diff
@@ -0,0 +1,292 @@
...
+  if op.getOpType! ctx = .arith theArithOp then
+    if op.getNumOperands! ctx = 2 then
+      let operands := op.getOperands! ctx
+      let a := operands[0]!
+      let b := operands[1]!
+      let some width := intBitwidth rewriter a | return rewriter
+      let iN : TypeAttr := IntegerType.mk width
+      let i2N : TypeAttr := IntegerType.mk (2 * width)
+      let ip := InsertPoint.before op
+      let (rewriter, low) ← emitLLVMBin rewriter .mul { nsw := false, nuw := false } iN a b ip
```
math-fehr: You can probably avoid this multiplication by just using a truncation of `wideMul` instead?

Veir/Passes/ArithToLLVM.lean
regehr: done!

math-fehr (approved):
Nice! I just have 2 nit picks in the code, but otherwise this looks good!

Veir/Passes/ArithToLLVM.lean
math-fehr: Sorry I should have mentionned this, but since this work can you do the same modification on the other functions?

Veir/Passes/ArithToLLVM.lean
regehr: argh!!! I should have checked the codegen against C++ MLIR. 
now I believe they all match

regehr:
thanks for the detailed comments, @math-fehr -- all of them are addressed

math-fehr (approved):
Nice, thanks!

PR #1122, opened 2026-07-25
fix(verifier): name the operation in block terminator failures

### LINKED ISSUE #976: Report name of operation in all verifier failure messages

sueszli:
#1105 closed #976 but the solution was incomplete.

This is a continuation of that PR.

### DIFF
--- a/Test/Verifier/func_func_empty_block.mlir
+++ b/Test/Verifier/func_func_empty_block.mlir
@@ -6,4 +6,4 @@
   }) : () -> ()
 }) : () -> ()
 
-// CHECK: Expected the block to end in a terminator, but the block is empty
+// CHECK: func.func: Expected the block to end in a terminator, but the block is empty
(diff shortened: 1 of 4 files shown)


sueszli:
I noticed that naming is now done in two places (in the `Except.mapError` in `WfIRContext.verify` and in `named` here) which could drift.

Happy to unify here or in a follow-up.

But a shared helper may not be worth the loss of locality.

math-fehr (approved):
Nice, thanks for all these improvements!

sueszli:
Yay! Thanks for reviewing so quickly

PR #1126, opened 2026-07-25
fix(riscv): define pure operations using an allowlist rather than a blocklist

regehr:
this fixes an inconsistency where all loads other than `ld` were considered pure

### DIFF
--- a/Veir/GlobalOpInfo.lean
+++ b/Veir/GlobalOpInfo.lean
@@ -411,9 +411,37 @@ def OperationPtr.hasSideEffects (op : OperationPtr) (ctx : IRContext OpCode) : B
   | .arith _ | .comb _ | .mod_arith _ | .datapath _ => false
   | .builtin .unrealized_conversion_cast => false
   | .hw .constant => false
-  -- RISC-V is pure register arithmetic except the memory ops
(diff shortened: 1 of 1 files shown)


Veir/GlobalOpInfo.lean
```diff
@@ -411,9 +411,37 @@ def OperationPtr.hasSideEffects (op : OperationPtr) (ctx : IRContext OpCode) : B
   | .arith _ | .comb _ | .mod_arith _ | .datapath _ => false
   | .builtin .unrealized_conversion_cast => false
   | .hw .constant => false
-  -- RISC-V is pure register arithmetic except the memory ops
-  | .riscv .ld | .riscv .sd | .riscv .sw | .riscv .sh | .riscv .sb => true
-  | .riscv _ => false
+  -- Enumerate the pure subset of RISC-V
+  | .riscv .ld | .riscv .lw | .riscv .lwu
```
math-fehr: I think these first ones are meant to be `true` instead of `false`?

tobiasgrosser: Yes, loads read from stateful memory and are consequently not pure.

regehr:
fixed and ready for re-review, thanks!

math-fehr (approved):
Thanks!

PR #1129, opened 2026-07-26
feat: first part of constant folding

regehr:
ok the new thinking is a series of smaller PRs building up to constant folding. here's the first one, it materializes runtime constant values from IR constant values.


Veir/Fold/ConstantValue.lean
```diff
@@ -0,0 +1,58 @@
...
+namespace Veir
+
+/--
+  If `val` is defined by a supported constant-like operation, return the
+  runtime value it materializes.
+
+  The conversions match the interpretation of the corresponding operations in
+  `interpretOp'`. Since `mod_arith` has no interpreter semantics, its constants
+  are recognized only when they satisfy the canonical-residue invariant
+  expected by the `mod-arith-to-arith` lowering.
```
math-fehr:
```suggestion
  The conversions match the interpretation of the corresponding operations in
  `interpretOp'`.
```
I'm not sure why we had the mod_arith comment here?

Veir/Fold/ConstantValue.lean
```diff
@@ -0,0 +1,58 @@
...
+-/
+def ValuePtr.constantValue (val : ValuePtr) (ctx : IRContext OpCode) : Option RuntimeValue := do
+  let defOp ← val.getDefiningOp! ctx
+  match defOp.getOpType! ctx with
```
math-fehr: Quick question, is there a reason we are reimplementing these functions, rather than using the `interpretOp'` function directly? By that, I mean should we call `interpretOp'` here for each branch rather than reimplementing the function?

UnitTest/ConstantValue.lean
```diff
@@ -0,0 +1,59 @@
...
+  if modNeg.isSome then
+    return "accepted a negative mod_arith.constant"
+
+  return "ok"
+
+/--
+info: "ok"
+-/
+#guard_msgs in
+#eval! testConstantValue
```
math-fehr: A bit of a nit, but can we have one #guard per test, rather than one for all? That should make debugging easier in general?

math-fehr (commented):
Nice, I just left small comments!
One quick question otherwise, how would you feel moving this to `Interface/ConstantLike.lean`? AFAIK, this is exactly ConstantLike interface that exist in MLIR.

Veir/Fold/ConstantValue.lean
regehr: oh, wow, I didn't realize this could be so easy

regehr:
thanks @math-fehr ! ready for another quick look

math-fehr (approved):
Nice! Do you think we can move the `ConstantValue.lean` content to `ConstantLike.lean` in the Interfaces? I'm not sure if there will be some dependency cycle or not yet.
But to me the `ValuePtr.constantValue` is going to be useful in more places than the folder? I'm fine with either way if you prefer it in the folder, but I believe it will be more discoverable in ConstantLike

regehr:
@math-fehr yes makes sense-- implemented your list suggestion and queueing this up now

PR #1130, opened 2026-07-27
feat(arith): support arith.subui_extended

regehr:
this was left out for some reason


Veir/Passes/ArithToLLVM/Proofs.lean
```diff
@@ -89,6 +89,17 @@ theorem addui_extended_lowering (a b : Int 8) :
...
 
+/--
+The two results of `arith.subui_extended` are reproduced by an unflagged
+`llvm.sub` and `llvm.icmp ult` on the operands: an unsigned subtraction borrows
+exactly when `a <u b`.
+-/
+theorem subui_extended_lowering (a b : Int 8) :
+    (sub a b, usubOverflowFlag a b) =
+      let diff := sub a b
+      (diff, icmp a b .ult) := by
```
math-fehr:
```suggestion
      (sub a b, icmp a b .ult) := by
```
In general, it's quite bad to have `let` in theorems, so let's avoid them when we can

math-fehr (approved):
All good! Can you also add support for it in CSE? I think it will not work out of the box and consider it side-effectful?

regehr:
> All good! Can you also add support for it in CSE? I think it will not work out of the box and consider it side-effectful?

added!!

math-fehr (approved):
Makes sense to me!

PR #1131, opened 2026-07-27
feat: canonicalize mod_arith constants

regehr:
for purposes of CSE, lowering, etc. -- we want mod_arith constants to be canonical


Veir/Passes/Canonicalize/Proofs.lean
```diff
@@ -0,0 +1,32 @@
...
+/-!
+Correctness proofs for the `mod_arith` patterns in `Canonicalize.lean`.
+-/
+
+/--
+Replacing the value of `mod_arith.constant` from `value` to `value % modulus`
+preserves its value in `ℤ/modulusℤ`.
+-/
+theorem canonicalizeModArithConstant_correct (value modulus : Int) :
+    constant modulus value = constant modulus (value % modulus) := by
```
math-fehr:
```suggestion
@[simp, grind =]
theorem canonicalizeModArithConstant_correct (value modulus : Int) :
    constant modulus (value % modulus) = constant modulus value := by
```
I feel this direction makes more sense, and that allows you to use `simp` as well

Veir/Passes/Canonicalize.lean
```diff
@@ -9,10 +9,26 @@ namespace Veir
...
+  let some (_, props) := matchOp op rewriter.ctx (.mod_arith .constant) 0
+    | return rewriter
+  let resultType := (op.getResult 0 : ValuePtr).getType! rewriter.ctx.raw
+  let .modArithType modArithType := resultType.val
+    | return rewriter
+  let canonicalValue := props.value.value % modArithType.modulus.value
+  if canonicalValue == props.value.value then return rewriter
+  let canonicalProps : ModArithConstantProperties :=
+    { value := { props.value with value := canonicalValue } }
+  let (rewriter, newOp) ← rewriter.createOp! (.mod_arith .constant)
```
math-fehr: I guess here, instead of creating a new operation, you can use `rewriter.setProperties!` instead? Unless it doesn't exist yet?

math-fehr (approved):
Nice! I just have 2 nit comments, but otherwise looks good to me.
I would wait for @AlexanderViand comment though!

AlexanderViand (approved):
LGTM, thank you! 

One small note is that I'm not sure the current theorems over `Int` are what we'll need in the end. Most `mod_arith` transformations are trivial over $\mathbb{Z}$ and all the hard stuff only happens when you consider actual IntegerTypes that can over/underflow. But there's nothing wrong with having the "obvious" theorems first and worrying about the harder version later.

Veir/Passes/Canonicalize.lean
```diff
@@ -10,10 +10,24 @@ namespace Veir
...
 
+def canonicalizeModArithConstant (rewriter : PatternRewriter OpCode) (op : OperationPtr)
+    (_ : op.InBounds rewriter.ctx.raw) : Option (PatternRewriter OpCode) := do
+  let some (_, props) := matchOp op rewriter.ctx (.mod_arith .constant) 0
+    | return rewriter
+  let resultType := (op.getResult 0 : ValuePtr).getType! rewriter.ctx.raw
+  let .modArithType modArithType := resultType.val
+    | return rewriter
+  let canonicalValue := props.value.value % modArithType.modulus.value
+  if canonicalValue == props.value.value then return rewriter
```
math-fehr:
```suggestion
  if canonicalValue = props.value.value then return rewriter
```

math-fehr (approved):

PR #1135, opened 2026-07-27
feat(riscv): support volatile memory accesses

regehr:
previously we only tracked volatile through the LLVM dialect, meaning that DCE at the RISC-V level would remove memory accesses that had previously been volatile. this fixes that. it's super easy except that it touches a bunch of files.

the important test is `Test/Passes/DCE/riscv-load-volatile.mlir` which shows that a regular load without users is eliminated by DCE but a volatile load without users is not


luisacicolini (approved):
lgtm, thank you! I just have one question, in `volatile_` the underscore is necessary for MLIR compliance and we are calling the name `volatile_` as well (e.g. in the functions) for consistency - right?

regehr:
> lgtm, thank you! I just have one question, in `volatile_` the underscore is necessary for MLIR compliance and we are calling the name `volatile_` as well (e.g. in the functions) for consistency - right?

I believe this to be the case but really I just tried to copy the implementation of the LLVM dialect!

luisacicolini:
> > lgtm, thank you! I just have one question, in `volatile_` the underscore is necessary for MLIR compliance and we are calling the name `volatile_` as well (e.g. in the functions) for consistency - right?
> 
> I believe this to be the case but really I just tried to copy the implementation of the LLVM dialect!

sounds good, was asking out of curiosity :)

PR #1136, opened 2026-07-27
fix(vsmith): mirror a pass name change in LLVM

regehr:
this makes vsmith in risc-v mode keep working with latest LLVM


tobiasgrosser (commented):
Would this break older LLVM versions?

regehr:
> Would this break older LLVM versions?

yes but only for people running vsmith in risc-v mode, which I don't think includes anyone but me

luisacicolini (approved):

PR #1137, opened 2026-07-27
chore(riscv): normalize optimization pass ordering

regehr:
this makes both `vsmith` and `veir-opt -p=riscv` use the same pass ordering that we use in the paper evaluation repo


luisacicolini (approved):
thank you!!!

PR #1139, opened 2026-07-27
feat: towards support for globals

regehr:
this PR registers mlir__global and mlir__addressof, and then follows through with other code supporting this such as verifier support

the short-term goal (not nearly achieved yet) is to run `Test/LLVM/global_load_store.mlir` though the veir interpreter. this test simple loads from an i32 global, increments the value, and uses the result to store back to the global and also returns it from main().


Test/LLVM/addressof_invalid_result.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
```
math-fehr, line 1: Can you also add the command that checks that MLIR fails on this? I forgot the name but it's in the lit.cfg

Veir/Verifier.lean
```diff
@@ -1112,6 +1128,27 @@ private def WfIRContext.graphRegionsHaveAtMostOneBlock (ctx : WfIRContext OpCode
...
+      let symbolName := "@".toUTF8 ++ props.sym_name.value
+      if globals.contains symbolName then
+        let displayName := String.fromUTF8? symbolName |>.getD "<non-UTF8 global symbol>"
+        throw s!"llvm.mlir.global: duplicate global symbol '{displayName}'"
+      globals := globals.insert symbolName op
+  for op in ctx.raw.operations.keys do
+    if op.getOpType! ctx.raw = .llvm .mlir__addressof then
+      let props := op.getProperties! ctx.raw (.llvm .mlir__addressof)
+      if !globals.contains props.global_name.value.toUTF8 then
+        throw s!"llvm.mlir.addressof: symbol '{props.global_name.value}' does not name an llvm.mlir.global"
```
math-fehr, line 1195: We should figure a nice way to write this out at some point! Not in this PR though!

math-fehr: Also, just so you know, with this current design, you cannot take the address of a function, which LLVM allows you to do if I understand correctly? Again, we can fix this later but we should probably write a comment here about it.

Veir/Verifier.lean
```diff
@@ -1112,6 +1128,27 @@ private def WfIRContext.graphRegionsHaveAtMostOneBlock (ctx : WfIRContext OpCode
...
+  Check the module-wide invariants needed by LLVM global references: global
+  names are unique and every `llvm.mlir.addressof` names a declared global.
+-/
+private def WfIRContext.verifyLLVMGlobalSymbols (ctx : WfIRContext OpCode) :
+    Except String Unit := do
+  let mut globals : Std.HashMap ByteArray OperationPtr := Std.HashMap.emptyWithCapacity
+  for op in ctx.raw.operations.keys do
+    if op.getOpType! ctx.raw = .llvm .mlir__global then
+      let props := op.getProperties! ctx.raw (.llvm .mlir__global)
+      let symbolName := "@".toUTF8 ++ props.sym_name.value
```
math-fehr, line 1186: This is a bit tricky, because if `sym_name` is `g.x`, we want to check for `@"g.x"` rather than `@g.x`. I think this is failing now?

math-fehr: We can probably also take care of this in another PR, since the lookup system is going to change at that point.

Veir/Verifier.lean
```diff
@@ -663,6 +658,27 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
...
+    if op.getNumRegions ctx.raw opIn ≠ 1 then
+      throw "Expected 1 region"
+    if op.getNumSuccessors ctx.raw opIn ≠ 0 then
+      throw "Expected 0 successors"
+    let properties := op.getProperties! ctx.raw (.llvm .mlir__global)
+    if properties.alignment.type.bitwidth ≠ 64 then
+      throw "'alignment' must be a 64-bit signless integer attribute"
+    if properties.addr_space.type.bitwidth ≠ 32 then
+      throw "'addr_space' must be a 32-bit signless integer attribute"
+    pure ()
```
math-fehr, line 702: I think you need to check that if the region is empty, then `value` is populated? And vice versa if value is `none`, then the region is nonempty.

Veir/Verifier.lean
```diff
@@ -663,6 +658,27 @@ def OperationPtr.verifyLocalInvariants (op : OperationPtr) (ctx : WfIRContext Op
...
+    if op.getNumOperands ctx.raw opIn ≠ 0 then
+      throw "Expected 0 operands"
+    if op.getNumResults ctx.raw opIn ≠ 0 then
+      throw "Expected 0 results"
+    if op.getNumRegions ctx.raw opIn ≠ 1 then
+      throw "Expected 1 region"
+    if op.getNumSuccessors ctx.raw opIn ≠ 0 then
+      throw "Expected 0 successors"
+    let properties := op.getProperties! ctx.raw (.llvm .mlir__global)
+    if properties.alignment.type.bitwidth ≠ 64 then
```
math-fehr: alignment should also be a power of two

math-fehr (approved):
Nice! I think I only have one two comments that should be addressed, the rest is comments where we should probably create an issue about it and fix it later.

Veir/Verifier.lean
regehr: it seems the 3 of the 4 possible combinations are legal in upstream MLIR!
I'll make us match.

Veir/Verifier.lean
regehr: funnily enough, the behavior of the existing code matches MLIR for `g.x`.
however, `"g.x"` causes serious problems for us. I will not try to address this here.

regehr:
@math-fehr comments addressed and code improved, ready for another quick look, thanks!

math-fehr (approved):

PR #1140, opened 2026-07-27
feat(riscv): absorb offsets into immediate fields of loads and stores, when possible

regehr:
there's a lot more to do here. this copies the isel part of LLVM pretty well, but it has a ton of other machinery for doing a good job here. but this will catch the easy cases!!

### DIFF
--- a/Test/Passes/InstructionSelection/RISCV64/load_store_offset.mlir
+++ b/Test/Passes/InstructionSelection/RISCV64/load_store_offset.mlir
@@ -1,0 +1,115 @@
+// RUN: veir-opt %s -p=isel-riscv64 | filecheck %s
+
+// A `llvm.getelementptr` with a constant index is absorbed into the signed
+// 12-bit offset field of the load or store that uses it, mirroring the
+// `isBaseWithConstantOffset` case of LLVM's `SelectAddrRegImm`. The gep itself
+// becomes dead and is erased, so no address arithmetic remains.
+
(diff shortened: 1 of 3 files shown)


luisacicolini (approved):
lgtm but I am not that familiar with the offsets in the lowering, so don't have a lot of context

regehr:
> lgtm but I am not that familiar with the offsets in the lowering, so don't have a lot of context

the code is a little gross, maybe I can make it better, but the intent is super simple: every RV64 load or store comes with a 12-bit immediate offset field that makes the actual operation land -2048 to 2047 bytes away from where the register argument points

regehr:
using these will make the benchmarks for the paper faster!!! not fastntt but at least some of the others

math-fehr (commented):
I'll let @luisacicolini or @tobiasgrosser review it for the instruction selection part.
Otherwise, I feel there is maybe *too much* documentation here, to the point that I don't really see how people are expected to maintain that level of documentation/comments. Maybe this should be reduced quite a lot to just a few sentences / bullet points? Also, I don't think we should refer ever to line numbers in LLVM, as they change quite a lot. Maybe a permalink if these are that important?

regehr:
my fault, I asked for the line numbers. I'll reduce the amount of comments here.

regehr:
ok this is better now!

tobiasgrosser (approved):

PR #1141, opened 2026-07-28
chore: split function boundary coercion into its own pass

AlexanderViand:
Splits the function boundary type coercion functionality introduced into `reconcile-cast`  into its own pass, `coerce-function-boundaries-to-riscv-reg`, so that `reconcile-casts` can remain a generic pass for non-riscv-usecases.


regehr:
I can't review this until later today but I agree with the approach!

regehr (approved):
this LGTM but one more change is needed: add the new pass to the `riscv` pass grouping that `veir-opt` already knows about

PR #1142, opened 2026-07-28
feat: Add `HasDialect` infrastructure

math-fehr:
`HasDialect` allows a global opcode type to specify that it contains the opcodes of a dialect. This allows for instance to write passes that only requires that the `Arith` dialect is registered, plus arbitrary other dialects.

We can move from dialect opcodes to global opcodes with `toDialect?` and `ofDialect`, and similarly with `ofDialectProperties` and `toDialectProperties` for properties.

### DIFF
--- a/UnitTest.lean
+++ b/UnitTest.lean
@@ -4,6 +4,7 @@ import UnitTest.Parser
 import UnitTest.AttrParser
 import UnitTest.MlirParser
 import UnitTest.IR.Operation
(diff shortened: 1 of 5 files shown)


tobiasgrosser (approved):
Very beautiful.

ineol (approved):
Cool!

nchappe (approved):

PR #1147, opened 2026-07-28
feat: constant evaluation part of constant folding

regehr:
this PR contributes the part of constant folding that invokes the interpreter to actually fold constants!

ok, we're going to need to redo our notions of purity, side-effect-freedom, etc.
I will take care of that soon, none of it belongs in this PR. what I'm saying is that you don't have to love the treatment of that stuff in this PR since it's all going to change soon!


math-fehr (approved):
Nice, thanks!

PR #1148, opened 2026-07-28
fix(riscv): store operand ordering fix

regehr:
ok, argh-- hopefully this is the last of these

a riscv store looks like this: `sd value, address`

this is how it looks in textual RISC-V asm and this is how they look in our own interpreter:
```
  | .sd => do
    let [.reg { val }, .reg addr] := operands.toList | none
    let eaddr := riscvEffectiveAddr addr.val properties.value.value
    let mem := mem.ensureSize (eaddr.toNat + 8)
    let mem ← mem.store eaddr.toNat.toUInt64 (UInt64.ofBitVec val).toByteArrayLE
    return (#[], mem, none)
```

but we were still getting this wrong in a few places, here I'm trying to fix this


tobiasgrosser (approved):
Interesting. @luisacicolini, can you double-check.

luisacicolini (approved):

PR #1149, opened 2026-07-28
feat(mod_arith): add interpreter semantics for `mod_arith` operations

AlexanderViand:
Adds interpreter semantics for `mod_arith` operations.

This also introduces a little `LLVM.Int.toNat?` helper to make the semantic definitions a bit prettier, fixes a brittle grind in one of the interpreter lemmas by  removing @[grind <=] from `RuntimeValue.Conforms`, and swaps the `testEvaluateUninterpreted` to a different dialect now that `mod_arith` is no longer uninterpreted.

### DIFF
--- a/Test/Interpreter/ModArith/constant.mlir
+++ b/Test/Interpreter/ModArith/constant.mlir
@@ -1,0 +1,10 @@
+// RUN: veir-interpret %s | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{sym_name = "main", function_type = () -> !mod_arith.int<17 : i32>}> ({
+    %c3 = "mod_arith.constant"() <{ "value" = 3 : i32 }> : () -> !mod_arith.int<17 : i32>
+    "func.return"(%c3) : (!mod_arith.int<17 : i32>) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Program output: #[0x00000003#32]
--- a/UnitTest/Evaluate.lean
+++ b/UnitTest/Evaluate.lean
@@ -44,12 +44,11 @@ info: "ok"
 #eval! testEvaluateUB
 
 /-- An operation the interpreter does not implement evaluates to `none`. The
-    `mod_arith` dialect is lowered to `arith` before interpretation, so it has
-    no runtime semantics of its own to evaluate against. -/
+    `datapath` dialect models hardware structures with no runtime semantics of
+    their own to evaluate against. -/
 private def testEvaluateUninterpreted : String := Id.run do
-  let m17 : TypeAttr := ModArithType.mk (IntegerAttr.mk 17 (IntegerType.mk 32))
   let operands : Array RuntimeValue := #[.int 32 (.val 13), .int 32 (.val 7)]
-  let none := (foldEvaluate (.mod_arith .add) () #[m17] operands : Option (UBOr (Array RuntimeValue)))
+  let none := (foldEvaluate (.datapath .compress) () #[i32, i32] operands : Option (UBOr (Array RuntimeValue)))
     | return "an uninterpreted operation was evaluated"
   return "ok"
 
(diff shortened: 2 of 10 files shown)


Test/Interpreter/ModArith/constant.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: veir-interpret %s | filecheck %s
+
+"builtin.module"() ({
+  "func.func"() <{sym_name = "main", function_type = () -> !mod_arith.int<17 : i32>}> ({
+    %c3 = "mod_arith.constant"() <{ "value" = 3 : i32 }> : () -> !mod_arith.int<17 : i32>
+    "func.return"(%c3) : (!mod_arith.int<17 : i32>) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Program output: #[0x00000003#32]
```
regehr: missing newline after last line

AlexanderViand: Fixed!

regehr (approved):
this LGTM!

optional:
- if you want to make the interpreter cases more concise, they're mostly the same code and could be factored out into a common helper
- I feel like there are some subtleties in the Int/Nat mixed expressions here? everything looks probably right but I'm not good at Lean's type coercion rules and I'm not sure what might be done to make it clearer.

but if you're happy with this, I'm happy!

AlexanderViand:
> * I feel like there are some subtleties in the Int/Nat mixed expressions here? everything looks probably right but I'm not good at Lean's type coercion rules and I'm not sure what might be done to make it clearer.

Yep, the subtraction case worked more by accident than by intentional design, I've made the coercion explicit now and added a test for the "bad" case + for negative constants.

AlexanderViand:
> * if you want to make the interpreter cases more concise, they're mostly the same code and could be factored out into a common helper

If no one minds, I'd keep it as-is for now and do more aggressive factoring once the complexity increases when we introduce the idea of multiple possible representations.

regehr:
I don't mind-- please merge

Seonyoung-Cheon (commented):
I mainly checked the examples and the interpreter, and the behavior of the operations and the mathematical logic look correct to me.

Just one minor question: Just one minor question: is the mod = 0 case checked somewhere?

Other than that, looks good to me!

Veir/Interpreter/Lemmas.lean
```diff
@@ -201,7 +201,7 @@ theorem VariableState.getVar?_setArgumentValues?_loop {block : BlockPtr}
   next => grind
   next =>
     simp only [Option.bind_eq_bind, Nat.succ_eq_add_one, Option.bind]
-    grind [cases BlockArgumentPtr, OperationPtr.getResult_def, cases ValuePtr]
+    grind [-RuntimeValue.Conforms, cases BlockArgumentPtr, OperationPtr.getResult_def, cases ValuePtr]
```
tobiasgrosser: This is ugly. Why is this needed. @math-fehr should have a look.

tobiasgrosser (commented):
This looks good, with the exception of the one grind lemma. @math-fehr likely has opinions.

AlexanderViand:
> Just one minor question: is the mod = 0 case checked somewhere?

It's disallowed in the verifiers for the mod_arith operations, and I think (but am not sure) that the convention for interpreter semantics is to not re-check verifier invariants/etc in interpretOp'? Pinging @math-fehr  for that, too :)

Veir/Interpreter/Lemmas.lean
AlexanderViand: Nothing currently seems to depend on RuntimeValue.Conforms being tagged for [grind], so I've removed that globally rather than manually exclude it in this proof, but not sure if there's an obvious future need for it to be [grind] tagged that I'm not aware of.

UnitTest/Evaluate.lean
```diff
@@ -44,12 +44,11 @@ info: "ok"
...
-    no runtime semantics of its own to evaluate against. -/
+    `datapath` dialect models hardware structures with no runtime semantics of
+    their own to evaluate against. -/
 private def testEvaluateUninterpreted : String := Id.run do
-  let m17 : TypeAttr := ModArithType.mk (IntegerAttr.mk 17 (IntegerType.mk 32))
   let operands : Array RuntimeValue := #[.int 32 (.val 13), .int 32 (.val 7)]
-  let none := (foldEvaluate (.mod_arith .add) () #[m17] operands : Option (UBOr (Array RuntimeValue)))
+  let none := (foldEvaluate (.datapath .compress) () #[i32, i32] operands : Option (UBOr (Array RuntimeValue)))
     | return "an uninterpreted operation was evaluated"
   return "ok"
```
AlexanderViand, line 53: @regehr I had to change your recently-added `testEvaluateUninterpreted` since mod_arith is no longer uninterpreted, let me know if this still captures the intent of your test

regehr: yes that's great!

Veir/Interpreter/Lemmas.lean
AlexanderViand: @ineol and @math-fehr are saying they're happy with the new approach and there's no need for RuntimeValue.Conforms to still be tagged  @[grind]

PR #1150, opened 2026-07-29
chore: use the preset risc-v optimization pipeline instead of specifying them individually

regehr:
this changes nothing except instead of specifying a bespoke list of RISC-V backend passes, it uses the newish `-p=riscv` option. I think this is useful because we want to keep that list up to date and also avoid random lists of passes being specified lots of different places

### DIFF
--- a/Test/Passes/InstructionSelection/RISCV64/ashr_exec.mlir
+++ b/Test/Passes/InstructionSelection/RISCV64/ashr_exec.mlir
@@ -1,5 +1,5 @@
 // RUN: veir-interpret %s | filecheck %s --check-prefix=SRC
-// RUN: veir-opt %s -p=canonicalize,instcombine,canonicalize,cse,dce,isel-br-riscv64,isel-sdag-riscv64,isel-riscv64,canonicalize,riscv-combine,coerce-function-boundaries-to-riscv-reg,reconcile-cast,dce > %t && veir-interpret %t | filecheck %s
+// RUN: veir-opt %s -p=riscv > %t && veir-interpret %t | filecheck %s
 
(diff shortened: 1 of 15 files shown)


tobiasgrosser (approved):

AlexanderViand (changes requested):
LGTM in what it does, but I think there's several more tests that need the same simplification? 

- add_ext_exec.mlir
- bit_intrinsics_exec.mlir
- fshl_exec.mlir
- fshl_general_exec.mlir
- fshl_reg_exec.mlir
- fshr_exec.mlir
- fshr_general_exec.mlir
- fshr_reg_exec.mlir
- sdiv_pow2_exec.mlir
- sdiv_pow2_exact_exec.mlir
- sdiv_pow2_i32_exec.mlir
- udiv_pow2_exec.mlir
- udiv_pow2_i32_exec.mlir

Also, -p=riscv isn't quite the same as the pass list it replaces (e.g., no canonicalize) but I assume that's intentional?

regehr:
> LGTM in what it does, but I think there's several more tests that need the same simplification?

yeah, I wasn't intending to be complete here, but rather just making sure `-p=riscv` has clients in our test suite

> Also, -p=riscv isn't quite the same as the pass list it replaces (e.g., no canonicalize) but I assume that's intentional?

yes, I went with the phase ordering that we used in the POPL submission, it's found in Luisa's repo for the paper evaluation, I figure that's the right one to be using for now.
but also, I want us to keep updating that pass list as we learn what's best, so if more canonicalize invocations are needed we should add them to `veir-opt -p=riscv`

AlexanderViand (approved):

PR #1156, opened 2026-07-29
feat: Move `toAttrDict` and `fromAttrDict` to `HasDialectOpInfo`

math-fehr:
`HasOpInfo` should be merged with `HasDialectOpInfo` in the future, so its fields are moved to `HasDialectOpinfo`. Each dialect file now implements the two functions, and they are merged in the end `OpCode`.


regehr (approved):
well, it's too big to be certain about all of the details, but for sure the shape of this looks good. 

I asked codex to check that nothing got dropped and it seems happy.

so LGTM!

tobiasgrosser (approved):

PR #1157, opened 2026-07-29
feat(riscv): alloca improvements

regehr:
at the RISC-V level allocas should just be bags of bytes at some alignment-- no types are wanted, this PR moves in that direction

after this we can lower LLVM allocas to RISC-V allocas, and then also start to turn RISC-V allocas into stack pointer math


tobiasgrosser (approved):
That makes sense, what a nice incremental step.

math-fehr (approved):

PR #1158, opened 2026-07-29
feat(mod_arith): extend boundary coercion to `mod_arith`

AlexanderViand:
Ports the boundary coercion part of Seonyoung's `seonyoung/reconcile-cast-arith` branch to the newly split-out boundary coercion and makes that more modular, now offering both a `riscv` and a `mod_arith` flavoured pass.

### DIFF
--- a/VeirOpt.lean
+++ b/VeirOpt.lean
@@ -43,7 +43,6 @@ def availablePasses : Std.HashMap String (Pass OpCode) :=
     |>.insert ModArithToArithPass.name ModArithToArithPass
     |>.insert RemuiToBarrettReductionPass.name RemuiToBarrettReductionPass
     |>.insert ArithToLLVMPass.name ArithToLLVMPass
-    |>.insert RemuiToBarrettReductionPass.name RemuiToBarrettReductionPass
     |>.insert CanonicalizePass.name CanonicalizePass
 
 /--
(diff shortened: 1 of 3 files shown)


VeirOpt.lean
```diff
@@ -38,10 +38,12 @@ def availablePasses : Std.HashMap String (Pass OpCode) :=
     |>.insert DCEPass.name DCEPass
     |>.insert CastReconcilePass.name CastReconcilePass
     |>.insert CoerceFunctionBoundariesToRiscvRegPass.name CoerceFunctionBoundariesToRiscvRegPass
+    |>.insert CoerceModArithFunctionBoundariesPass.name CoerceModArithFunctionBoundariesPass
     |>.insert RISCV.Combine.name RISCV.Combine
     |>.insert ModArithToArithPass.name ModArithToArithPass
     |>.insert RemuiToBarrettReductionPass.name RemuiToBarrettReductionPass
     |>.insert ArithToLLVMPass.name ArithToLLVMPass
+    |>.insert RemuiToBarrettReductionPass.name RemuiToBarrettReductionPass
```
AlexanderViand: Thanks, good catch!

Seonyoung-Cheon (approved):
It looks like everything has been integrated well. Looks great!

regehr (approved):
this is cool! 
I like making this code parametric, since it'll need to be extended in the (hopefully near!) future to support rewriting other stuff like calls to other functions and function declaration. this way we'll support all dialects at once when we do that work

PR #1159, opened 2026-07-29
fix(mod_arith): fix shift factor and intermediate width in barret reduction

AlexanderViand:
While building some end-to-end test, I stumbled across a miscompilation it the modulus is large relative to the storage type.

### DIFF
--- a/Veir/Passes/ModArithToArith.lean
+++ b/Veir/Passes/ModArithToArith.lean
@@ -130,7 +130,7 @@ def emitBarrettReduction (rewriter : PatternRewriter OpCode) (r : ValuePtr) (mod
   let (rewriter, sh) ← emitArithConstant rewriter shift bw ip
 
 
-  -- reduced := (r * m) >> shift // (r * m) should be at most 2^(2*width) - 1, so we can safely truncate to width bits after the shift
+  -- reduced := (r * m) >> shift // (r * m) should be at most 2^(2*bw) - 1, so we can safely truncate to width bits after the shift
   let (rewriter, product) ← rewriter.createOp! (.arith .muli)
     #[ty] #[r, m] #[] #[] { attr := { nsw := false, nuw := true } } (some ip)
   let (rewriter, reduced) ← rewriter.createOp! (.arith .shrui)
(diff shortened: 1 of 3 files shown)


Veir/Passes/ModArithToArith.lean
```diff
@@ -108,16 +107,27 @@ def emitArithBinOp (rewriter : PatternRewriter OpCode) (arithOp : Arith)
...
+      pure (rewriter, r)
 
-  let (rewriter, m) ← emitArithConstant rewriter mu width ip
-  let (rewriter, sh) ← emitArithConstant rewriter shift width ip
+  let (rewriter, q) ← emitArithConstant rewriter modulus bw ip
+  let (rewriter, m) ← emitArithConstant rewriter mu bw ip
+  let (rewriter, sh) ← emitArithConstant rewriter shift bw ip
 
 
   -- reduced := (r * m) >> shift // (r * m) should be at most 2^(2*width) - 1, so we can safely truncate to width bits after the shift
```
Seonyoung-Cheon:
```suggestion
  -- reduced := (r * m) >> shift // (r * m) should be at most 2^(2*bw) - 1, so we can safely truncate to width bits after the shift
```
Minor comment modification

Veir/Passes/ModArithToArith.lean
```diff
@@ -108,16 +107,27 @@ def emitArithBinOp (rewriter : PatternRewriter OpCode) (arithOp : Arith)
...
-def emitBarrettReduction (rewriter : PatternRewriter OpCode) (r q : ValuePtr) (modulus : Int)
+def emitBarrettReduction (rewriter : PatternRewriter OpCode) (r : ValuePtr) (modulus : Int)
     (width : Nat) (ip : InsertPoint) : Option (PatternRewriter OpCode × ValuePtr) := do
-  let ty : TypeAttr := IntegerType.mk width
-  let ty_i1 : TypeAttr := IntegerType.mk 1
   let k : Nat := (modulus - 1).toNat.log2 + 1     -- ceil(log2 modulus)
-  let shift : Nat := 4 * k                        -- 4k (2k if you want 2 * k)
+  let shift : Nat := 2 * k
   let mu : Int := (2 ^ shift) / modulus           -- floor(2^shift / modulus)
+  let bw := max width (3 * k)                     -- width required for r *
```
Seonyoung-Cheon, line 115: Just curious, why was 3 * k chosen here ?

AlexanderViand: Because the 🤖 suggested it was sufficient ;) but it does seem to check out:
Assuming r < q² (which Barrett needs anyway), r·μ ≤ (q²−1)·⌊2^2k/q⌋ < q·2^2k ≤ 2^3k (since q ≤ 2^k)

Seonyoung-Cheon (approved):
I checked the fix! Adding the widened Barrett reduction case makes sense!

PR #1160, opened 2026-07-29
feat(mod_arith): add `mod-arith` pass group

AlexanderViand:
Adds a pass group that runs the mod-arith-to-arith and remui-to-barret-reduction passes + the various cleanup passes around it (cast reconciliation, function boundary coercion, canonicalisation, constant folding, etc)

Tagging @regehr for review since he's the author of the only other pass group / how I learnt about pass groups existing :)


regehr (approved):
this looks great!

something I think we should do (I can help) is to manually + using an LLM look at a number of before/after code examples to check for missed transformation opportunities or other problems

Seonyoung-Cheon (approved):
Everything seems fine :)

PR #1163, opened 2026-07-29
test(mod_arith): test that compiles to clang and runs

AlexanderViand:
After all the recent changes, we're now able to use veir to actually lower `mod_arith` to LLVM, compile it with clang and run it! 

Or, well, nearly: 
* we still need `mlir-opt --convert-func-to-llvm` since we don't have that pass (yet) and rewriting an operation with regions seems not-quite-fully-supported yet
* we also use `mlir-translate --mlir-to-llvmir`... we might eventually want that inside veir, but it seems less important than the prior one

The reference computes everything "naively" (explicit `%` everywhere), while the current lowering, though still far from optimal, at least uses Barrett reductions (though I haven't timed it to see if it's actually any faster).

Thanks to @Seonyoung-Cheon  and @regehr  for all their work on getting to this point!

### DIFF
--- a/Test/Passes/mod_arith_runner.mlir
+++ b/Test/Passes/mod_arith_runner.mlir
@@ -1,0 +1,58 @@
+// REQUIRES: clang, mlir-translate, mlir-opt
+// RUN: split-file %s %t
+// RUN: veir-opt %t/input.mlir -p=mod-arith,arith-to-llvm | mlir-opt --convert-func-to-llvm | mlir-translate --mlir-to-llvmir -o %t/kernel.ll
+// RUN: clang -Wno-override-module %t/driver.c %t/kernel.ll -o %t/test
+// RUN: %t/test | filecheck %s
+
+// CHECK: OK
+
+//--- input.mlir
+"builtin.module"() ({
+  "func.func"() <{function_type = (!mod_arith.int<12289 : i32>, !mod_arith.int<12289 : i32>) -> !mod_arith.int<12289 : i32>, sym_name = "mod_arith_chain"}> ({
+  ^bb0(%a : !mod_arith.int<12289 : i32>, %b : !mod_arith.int<12289 : i32>):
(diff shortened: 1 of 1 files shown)


regehr (approved):
super cool! LGTM!

Test/Passes/mod_arith_runner.mlir
```diff
@@ -0,0 +1,58 @@
...
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+uint32_t mod_arith_chain(uint32_t, uint32_t);
+
+static uint32_t reference(uint32_t a, uint32_t b) {
+  const uint32_t q = 12289;
+  uint64_t sum = (a + b) % q;
+  uint64_t difference = (a + q - b) % q;
```
Seonyoung-Cheon:
```suggestion
  const uint32_t q = 12289;
  uint64_t sum = ((uint64_t)a + (uint64_t)b) % q;
  uint64_t difference = ((uint64_t)a + q - (uint64_t)b) % q;
```

Since a and b are always between 0 and q, this should be fine, but it might still be worth casting them to uint64_t before the addition. Not required, though.

regehr:
I tried to compile this all the way to RISC-V using Veir but we can't do isel for `i33` or `i42` yet :(

AlexanderViand:
> I tried to compile this all the way to RISC-V using Veir but we can't do isel for `i33` or `i42` yet :(

Thanks for trying! We generate these weird sizes just in case something down the line will be able to take advantage of it, but in practice I expect them to be realised as the next power-of-two size anyway. Since, by design, mod-arith-to-arith should never generate code that'll over/underflow the IntegerTypes, it should be safe to add a variant of the pass that bumps all sizes to the next power of two. I'll get working on that later today :)

EDIT: I'm really starting to feel the lack of pass/pipeline options :(

PR #1164, opened 2026-07-29
feat(mod_arith): teach veir about mod_arith constants and also canonlizalize them to the right

regehr:
@AlexanderViand if you don't mind a fairly quick review on this change (which I hope is harmless and correct) I can make use of this in my ongoing work on constant folding


AlexanderViand (approved):
This looks great, thanks!

math-fehr (approved):

PR #1165, opened 2026-07-30
feat: Move `fromName` and `toName` to `HasDialectOpInfo`

math-fehr:
These two functions should be part of `HasOpInfo` at least. Since we can already define them on `HasDialectOpInfo`, I feel it should be moved there instead.


regehr (approved):
this LGTM!

tobiasgrosser (approved):

ineol (approved):

PR #1166, opened 2026-07-30
feat: Move the parser to `OpInfo` instead of `OpCode`

math-fehr:
This means the parser can now be used with any collection of dialects.

### DIFF
--- a/Veir/Parser/MlirParser.lean
+++ b/Veir/Parser/MlirParser.lean
@@ -1,14 +1,14 @@
 module
 
 public import Veir.Parser.Parser
-public import Veir.GlobalOpInfo
+public import Veir.Dialects.Builtin.OpInfo
 public import Veir.IR.WellFormed
 public import Veir.Rewriter.InsertPoint
 
-import Veir.Parser.AttrParser
-import Veir.Parser.DecidableInBounds
-import Veir.Rewriter.WellFormed
-import Veir.Rewriter.WfRewriter
+public import Veir.Parser.AttrParser
+public import Veir.Parser.DecidableInBounds
+public import Veir.Rewriter.WellFormed
+public import Veir.Rewriter.WfRewriter
 
 public section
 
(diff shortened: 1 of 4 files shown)


Veir/Parser/MlirParser.lean
```diff
@@ -558,28 +572,44 @@ def parseTypedValue : MlirParserM (ByteArray × TypeAttr × Location) := do
...
+-/
+private def optionallySetUnregisteredOpName (opCode : OpInfo)
+    (properties : HasOpInfo.propertiesOf opCode) (opName : ByteArray) :
+    HasOpInfo.propertiesOf opCode :=
+  if h : some .unregistered = toDialect? Builtin opCode then
+    have h' : ofDialect OpInfo Builtin.unregistered = opCode := by grind
+    let properties : UnregisteredProperties :=
+      HasDialect.toDialectProperties Builtin.unregistered (h' ▸ properties)
+    let properties := { properties with opName := opName }
+    h' ▸ HasDialect.ofDialectProperties OpInfo Builtin.unregistered properties
```
GZGavinZhao, line 603: Is this idiomatic Lean code? Maybe it is, I'm just not used to seeing the little triangle.

GZGavinZhao (approved):
Parser changes look good to me outside of a small, non-blocking style comment.

tobiasgrosser (approved):
Very cool. Its amazing that this breaks the dependency on `GlobalOpInfo`. This will make rebuilds a lot faster and even allow external users. Nice!

Veir/Parser/MlirParser.lean
ineol: I think it's fine in leaf functions. It's common when you have dependently typed data structures. For example `lean4/src/Std/Data` uses it 253 times.

ineol (approved):
LGTM!

math-fehr:
I merged a stacked commit my bad, will fix it when I get time today

PR #1169, opened 2026-07-30
fix(mod_arith): reduce constant during lowering

AlexanderViand:
The `mod-arith-to-arith` lowering still assumed that all `mod_arith.constants` would have canonicalised values, this removes that assumption, instead reducing the value during the lowering (happens at compile time, so basically free) 

In addition, this fixes a few naming inconsistencies, typos, etc.


Seonyoung-Cheon (approved):
Looks good to me! I missed quite a bit there 🥲

AlexanderViand:
> Looks good to me! I missed quite a bit there 🥲

I'm pretty sure this is a bug I introduced 😅 The rest is just AI-suggested minor cleanup

PR #1170, opened 2026-07-30
feat(mod_arith): add variant of lowering that uses only power-of-two integer types.

AlexanderViand:
Introduces the concept of `legalizeWidth: Nat -> Nat` to the mod-arith-to-arith lowering. The existing pass uses `id`, the new pass `Nat.nextPowerOfTwo`.

The same idea is also introduced to mod-arith function boundary coercion, and the new pass variants also get their own pass group.

### DIFF
--- a/Veir/Passes/ModArithToArith.lean
+++ b/Veir/Passes/ModArithToArith.lean
@@ -14,17 +14,17 @@ namespace Veir
   The current lowering is trivial, eagerly reducing at all times.
 
   Since Veir has no Dialect Conversion framework, this pass eagerly inserts `unrealized_conversion_casts`
-  to handle the type conversions between `!mod_arith.int<q : iN>` and `iN` that are needed.
+  to handle the type conversions between `!mod_arith.int<q : iN>` and `iM` that are needed.
 -/
 
 /-! ## Unrealized Conversion Casts -/
 
-/-- Emit `unrealized_conversion_cast v : !mod_arith.int<q:iN> → iN`. -/
-def castToStorage (rewriter : PatternRewriter OpCode) (v : ValuePtr) (ip : InsertPoint) :
-    Option (PatternRewriter OpCode × ValuePtr) := do
+/-- Emit `unrealized_conversion_cast v : !mod_arith.int<q:iN> → i(legalizeWidth N)`. -/
+def castToStorage (legalizeWidth : Nat → Nat) (rewriter : PatternRewriter OpCode)
+    (v : ValuePtr) (ip : InsertPoint) : Option (PatternRewriter OpCode × ValuePtr) := do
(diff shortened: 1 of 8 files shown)


math-fehr (approved):
That looks good to me!

Seonyoung-Cheon (commented):
I have a question about what`legalizeWidth` means exactly. I understand that it converts the width to a power of two, but I’m not sure what counts as “legal” here. I’m not too familiar with the syntax yet, so I may have missed the definition.

For example:
* If the input width is i17, would legalizing it mean casting it to i32?
* Would casting i17 to i64 also be considered legal?

It looks like the argument specifies the legal widths, but how are those widths chosen?

Veir/Passes/ModArithToArith.lean
```diff
@@ -286,12 +289,17 @@ def ModArithToArithPass.impl (ctx : WfIRContext OpCode) (op : OperationPtr)
 public def ModArithToArithPass : Pass OpCode :=
   { name := "mod-arith-to-arith"
     description := "Lower mod_arith operations to the arith dialect."
-    run := ModArithToArithPass.impl }
+    run := ModArithToArithPass.impl id }
+
+public def ModArithToArithPassPow2Width : Pass OpCode :=
+  { name := "mod-arith-to-arith-pow2-width"
+    description := "Lower mod_arith operations to the arith dialect, using power-of-two bitwidths."
```
AlexanderViand, line 296: @Seonyoung-Cheon In the rewrites, `legalizeWidth` is just a function the caller passes in that can map a width to any other width (there's nothing in the rewrites that checks its still valid, but if you mapped every width to i1, you'd probably get some errors somewhere).

The pass definitions down here are the ones that actually supply a definition for `legalizeWidth` - in the case of the original pass, it's `id`, so the widths are used as-is. In thecae if the `-pow2-width` pass, we use `Nat.nextPowerOfTwo` so `i17` would always be mapped to `i32` and never to `i64`.

Veir/Passes/ModArithToArith.lean
```diff
@@ -286,12 +289,17 @@ def ModArithToArithPass.impl (ctx : WfIRContext OpCode) (op : OperationPtr)
...
+    run := ModArithToArithPass.impl Nat.nextPowerOfTwo }
```
AlexanderViand, line 297: Previous comment didn't show the line(s) I wanted

Seonyoung-Cheon: Ah, I see—you’re using `Nat.nextPowerOfTwo`. That makes sense now. Everything looks good to me!

regehr (approved):
this is very cool!

LLVM can take advantage of limited-width values to aid its value range analysis, but Veir is far from being able to do this. 
and we should be aiming for mod_arith-specific range analyses anyhow.

PR #1173, opened 2026-07-30
chore: Split generic verifier utils to `Verifier/Basic|Lemmas`

math-fehr:
This change is the first step towards removing the dependency between `Verifier.lean` and `OpCode`.

This PR only moves files around, and makes the `Verifier/Basic.lean` files use an arbitrary `OpInfo` rather than an `OpCode` one. It also shows in `Cf/OpInfo.lean` how a verifier would be defined for a dialect.

The goal is to move all dialect-specific verifiers to the dialect files. For this, we first need to move all verifier utils to their own file which does not depend on any dialect. `Verifier/Basic.lean` is this utils file, and will in the medium term replace `Verifier.lean`.

### DIFF
--- a/Veir/Dialects/Cf/OpInfo.lean
+++ b/Veir/Dialects/Cf/OpInfo.lean
@@ -2,6 +2,7 @@ module
 
 public import Veir.IR.Simp
 public import Veir.IR.OpInfo
+public import Veir.Verifier.Basic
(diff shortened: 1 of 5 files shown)


tobiasgrosser (approved):

regehr (approved):
LGTM!

PR #1175, opened 2026-07-30
feat: make folding decisions

regehr:
ok-- this is a large patch, but most of it is just tables of folds, and none of them is very complicated or interesting! also they come with proofs.

the new thing here is the non-mutating interface which asks whether an operation + operand list would fold. we do this by:
1. consulting the operation tables, which handle the cases where a fold does not require all operands to be constants
2. if all operands are constant, running the interpreter

the folder is poison-preserving, and immediate UB is refined to poison


regehr:
adding @AlexanderViand to make sure the mod_arith bit of this patch (just a few lines) makes sense

math-fehr:
Nice! Do you think you could split this between the infrastructure for folding, and the actual fold implementation? That would make it quite easier for me to review I would say!

Also, talking about infrastructure, do you think you could integrate this somehow as a field in the `HasDialectOpInfo`, the same way we do for the other fields? That way each folder can be defined in a `Dialects/MyDialect/Fold/Basic.lean` and `Dialects/MyDialect/Fold/Lemmas.lean`? This would also help reviewing it!

regehr:
> Nice! Do you think you could split this between the infrastructure for folding, and the actual fold implementation? That would make it quite easier for me to review I would say!
> 
> Also, talking about infrastructure, do you think you could integrate this somehow as a field in the `HasDialectOpInfo`, the same way we do for the other fields? That way each folder can be defined in a `Dialects/MyDialect/Fold/Basic.lean` and `Dialects/MyDialect/Fold/Lemmas.lean`? This would also help reviewing it!

I'll split it up across dialects, but otherwise this patch is really not very substantative, it just looks big because of the operation tables, which I've been pretty careful about and anyway are cross-checked with proofs, do I really need to split it out even more?

regehr:
well, probably I can just LLM-out the folding tables...

math-fehr:
I think actually if the dialect-specific stuff is moved away from the infrastructure that should be fine to keep it in one go!

math-fehr:
And yeah, LLMs are getting pretty good at things like splitting a file into multiple files. Just make sure to run another LLM after that and ask it to check carefully if this is split correctly.

PR #1176, opened 2026-07-31
feat: interpreter-based folding

regehr:
ok, it wasn't really possible to keep splitting the big folding patch into smaller and smaller parts while also fanning it out across dialects while also keeping changes very small to make them easy to review

so let's try something different-- this is just the interpreter-based folding decision, nothing dialect-specific at all


Veir/Fold.lean
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Veir
+
+inductive FoldDecision where
+  /-- Use operand `j` in place of the result. -/
+  | useOperand (j : Nat)
+  /-- Use the runtime constant `rv` in place of the result. -/
+  | useConstant (rv : RuntimeValue)
+  /-- The operation does not fold with the supplied operand information. -/
+  | noFold
```
math-fehr: Instead of having `noFold`, can you return `Option FoldDecision` from your functions instead? This would be more idiomatic Lean.

Veir/Fold.lean
```diff
@@ -0,0 +1,68 @@
...
+    FoldDecision :=
+  if !constOperands.all (·.isSome) then .noFold else
+  let values := constOperands.map (·.get!)
+  match resultTypes.toList with
+  | [resultType] =>
+    match (foldEvaluate opCode properties #[resultType] values : Option (UBOr _)) with
+    | none => .noFold
+    | some (.ok results) =>
+      match results[0]? with
+      | some result => if result.Conforms resultType then .useConstant result else .noFold
```
math-fehr: This `if` should always evaluate to true, and we should at some point have a proof about it

Veir/Fold.lean
```diff
@@ -0,0 +1,68 @@
...
+  Query whether an operation folds, given its result types and the values of
+  its constant-defined operands (`constOperands[i] = some rv` iff operand `i`
+  is defined by a constant-like operation with value `rv`).
+-/
+def OpCode.foldsTo (opCode : OpCode) (properties : HasOpInfo.propertiesOf opCode)
+    (resultTypes : Array TypeAttr) (constOperands : Array (Option RuntimeValue)) :
+    FoldDecision :=
+  if !constOperands.all (·.isSome) then .noFold else
+  let values := constOperands.map (·.get!)
+  match resultTypes.toList with
```
math-fehr: If you return an Option from this function, it will look much nicer for example

math-fehr (commented):
Thanks!
Is there a reason why you have both Veir/Fold and Interface/Fold?
I think merging both of these would be better, so then everything is in one file (and thus you can mark most of the functions private)?

Also, do you think `OperationPtr.foldTo` could be a better naming for the public operations? So then we can write `op.foldTo`?

regehr:
> Is there a reason why you have both Veir/Fold and Interface/Fold?
> I think merging both of these would be better, so then everything is in one file (and thus you can mark most of the functions private)?

just trying to keep interface files small-- but I agree it's simpler your way

Veir/Fold.lean
regehr: it does!

regehr:
ready for another quick look @math-fehr

Veir/Interfaces/FoldInterfaces.lean
```diff
@@ -0,0 +1,54 @@
...
+  | .ok results =>
+    let result ← results[0]?
+    guard (result.Conforms resultType)
+    return .useConstant result
+  | .ub => return .useConstant (← RuntimeValue.getPoisonForType resultType)
+
+/--
+  Convenience wrapper around `foldsTo`.
+-/
+def opFoldsTo (op : OperationPtr)
```
math-fehr:
```suggestion
def OperationPtr.foldsTo (op : OperationPtr)
```
So this can look like `op.foldsTo ...`?

Veir/Interfaces/FoldInterfaces.lean
```diff
@@ -0,0 +1,54 @@
...
+  | useOperand (j : Nat)
+  /-- Use the runtime constant `rv` in place of the result. -/
+  | useConstant (rv : RuntimeValue)
+
+/--
+  Decide whether an operation folds, given its opcode, properties, result
+  types, and the values of its constant-defined operands (`constOperands[i] =
+  some rv` iff operand `i` is known to hold the constant `rv`).
+-/
+def foldsTo (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
```
math-fehr:
```suggestion
def OpCode.foldsTo (opType : OpCode) (properties : HasOpInfo.propertiesOf opType)
```
So then we can use things like `(.arith .addi).foldsTo ...`?

math-fehr (approved):
Nice! I just have some request on the naming/scoping, but otherwise good to me to merge!

PR #1177, opened 2026-07-31
chore(mod_arith): combine barett reduction pass into mod-arith-to-arith pass

AlexanderViand:
Integrate the Barret Reduction lowering into the mod-arith-to-arith pass, both in preparation for future augmentations to the pass (ability to do lazy reductions) and to make the pass-group more robust (barett reduction is only a suitable substitute for arith.remui under certain conditions which hold for mod-frith but the previous implementation would replace _all_ arith.remui, including any potentially introduced by another pass, which would not be safe to replace). 

This also updates the tests in order to better cover the now four different variants of the pass.


AlexanderViand:
@Seonyoung-Cheon It's still WIP (need to fix tests/etc), but here's an early iteration of combining the mod-arith-to-arith and barrett reduction passes 

Currently, which one gets used is still controlled by "flags", switching to per-op attributes will be in a follow-up PR

AlexanderViand:
@Seonyoung-Cheon this is ready for review now :)

Seonyoung-Cheon (approved):
Everything looks good! Once this and (https://github.com/opencompl/veir/pull/1182) are merged, I’ll open a PR for the range analysis this week :)

PR #1178, opened 2026-07-31
feat(PDL): add PDL dialect with attribute op

tobiasgrosser:


jumerckx (approved):
VEIR moves fast 😅

I'm not in a place to review the dialect structuring in Lean, but the filecheck tests seem to cover everything I'd expect. I'll see if I can add `pdl.type` and fix the unit test.

Veir/Dialects/PDL/OpInfo.lean
```diff
@@ -0,0 +1,98 @@
...
+  readsMemory := PDL.readsMemory
+  isConstantLike := PDL.isConstantLike
+  hasSSADominance := PDL.hasSSADominance
+
+/--
+Verify the local invariants of a `pdl` operation in any operation-info type
+containing the `pdl` dialect.
+
+TODO: The optional `valueType` operand is a `!pdl.type` handle produced by
+`pdl.type`, which is not modelled yet, so the operand type is left unchecked.
```
jumerckx, line 69: Does VEIR have something like a test op that can return an arbitrary type? The operand mustn't be a `pdl.type`-op result, it should have type `!pdl.type`, this can also be produced by external constraints for example.

tobiasgrosser: I fixed the test case.

math-fehr (approved):
Nice!

PR #1182, opened 2026-07-31
feat(mod_arith): attr-based reduction selection in lowering

AlexanderViand:
This makes the `mod-arith-to-arith` pass a two-step procedure: first, we add a (temporary/discardable) attribute to each `mod_arith` operation that describes what kind of reduction should be done afterwards, then the second step performs the lowering, using the annotated reduction strategy. 

Currently this does not change the behavior of the mod-arith-to-arith lowering, since we simply always assign the reduction type selected in the "pseudo pass-argument", but this gives a clear way forward for integrating range anaylsis and more optimised reduction selection strategies in the future.

### DIFF
--- a/Veir/Passes/ModArithToArith.lean
+++ b/Veir/Passes/ModArithToArith.lean
@@ -16,6 +16,11 @@ namespace Veir
   to handle the type conversions between `!mod_arith.int<q : iN>` and `iM` that are needed.
 -/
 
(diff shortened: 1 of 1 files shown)


Seonyoung-Cheon (approved):
LGTM 👍 I think this would be useful for selectively choosing the reduction

PR #1184, opened 2026-07-31
feat: simple pass option system

regehr:
this is syntactically similar to how upstream MLIR does it, but simpler! every option is a Boolean the defaults to false.


Test/Passes/pass_options.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: not veir-opt %s -p='mod-arith-to-arith{pow2-widht}' 2>&1 | filecheck %s --check-prefix=MISSPELLED
```
math-fehr, line 1:
```suggestion
// RUN: not veir-opt %s -p='mod-arith-to-arith{unknown-option}' 2>&1 | filecheck %s --check-prefix=MISSPELLED
```
When I read that I thought it was a typo at first, so maybe it's better to make it an obvious name?

math-fehr (approved):
Very nice! I'll wait for @AlexanderViand for merging it, but it's really nice!

AlexanderViand (approved):
LGTM, but why the restriction to boolean options? In #1182 we add 
```lean4
inductive ReductionKind
  | none
  | barrett
  | full
```
And while we technically don't need the "none" option right now, it'd be nice to have enum-style pass options?

regehr:
> LGTM, but why the restriction to boolean options? 

ah, I just wanted to add a minimal feature. I'll merge this one and then non-boolean options are easy to add later (or sooner)

PR #1187, opened 2026-08-01
feat: mirror more of MLIR's side effect system

regehr:
this is a shallow reimplementation of MLIR's side effect tracking:
https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/

these have to be property-sensitive, otherwise we can't tell the difference between volatile and non-volatile memory operations

I'm not really sure about PDL, I've marked them as not reading or writing memory, which is true but it would be weird if DCE or CSE messed with PDL code.

this PR does not eliminate our existing, parallel infrastructure nor does it start to actually use the new infrastructure, that's for future PRs

this PR also leaves out a lot of machinery such as memory resources, recursive region checking, etc.

### DIFF
--- a/Veir/Dialects/Builtin/OpInfo.lean
+++ b/Veir/Dialects/Builtin/OpInfo.lean
@@ -41,8 +41,17 @@ def Builtin.hasSideEffects (op : Builtin) (_props : Builtin.propertiesOf op) : B
   | .unrealized_conversion_cast => false
   | _ => true
 
-def Builtin.readsMemory (_op : Builtin) : Bool :=
-  false
+def Builtin.readsMemory
+    (op : Builtin) (_props : Builtin.propertiesOf op) : Bool :=
+  match op with
+  | .unrealized_conversion_cast => false
+  | _ => true
+
+def Builtin.writesMemory
+    (op : Builtin) (_props : Builtin.propertiesOf op) : Bool :=
+  match op with
+  | .unrealized_conversion_cast => false
+  | _ => true
 
 def Builtin.isConstantLike (_op : Builtin) : Bool :=
   false
@@ -69,6 +78,7 @@ instance : HasDialectOpInfo Builtin where
   toAttrDict := Builtin.toAttrDict
   hasSideEffects := Builtin.hasSideEffects
   readsMemory := Builtin.readsMemory
+  writesMemory := Builtin.writesMemory
   isConstantLike := Builtin.isConstantLike
   hasSSADominance := Builtin.hasSSADominance
   hasNoTerminator := Builtin.hasNoTerminator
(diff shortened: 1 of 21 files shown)


Veir/Dialects/Builtin/OpInfo.lean
```diff
@@ -41,8 +41,17 @@ def Builtin.hasSideEffects (op : Builtin) (_props : Builtin.propertiesOf op) : B
   | .unrealized_conversion_cast => false
   | _ => true
 
-def Builtin.readsMemory (_op : Builtin) : Bool :=
-  false
+def Builtin.readsMemory
+    (op : Builtin) (_props : Builtin.propertiesOf op) : Bool :=
+  match op with
+  | .unrealized_conversion_cast => false
```
math-fehr, line 47: This might be incorrect in the future, I don't really know.
For instance, what happens when we have an unrealized conversion cast between a `memref` and a `tensor`?
But we can keep it that way for now.

Veir/Dialects/Func/OpInfo.lean
```diff
@@ -51,8 +51,15 @@ def Func.toAttrDict
 def Func.hasSideEffects (_op : Func) (_props : Func.propertiesOf _op) : Bool :=
   true
 
-def Func.readsMemory (_op : Func) : Bool :=
-  false
+def Func.readsMemory (op : Func) (_props : Func.propertiesOf op) : Bool :=
+  match op with
+  | .call => true
+  | _ => false
```
math-fehr: Do you think we could try not having these default be `false`?
I'm fine with defaults being `true` because that's always safe, but here I think I would prefer if the operations are enumerated, just to be safe!

Veir/Interfaces/SideEffectInterfaces.lean
```diff
@@ -31,6 +31,28 @@ def OperationPtr.hasSideEffects {OpInfo : Type} [HasOpInfo OpInfo]
   let opType := op.getOpType! ctx
   HasDialectOpInfo.hasSideEffects opType (op.getProperties! ctx opType)
 
+/--
+  May this operation read memory?
+
+  TODO: recursively walk regions to get a less conservative answer
```
math-fehr, line 37: Just as a note, the way MLIR handles this is having an interface to state that the memory effects of an operation inherits the ones from its regions.

math-fehr (approved):
Very cool, thanks! I just have 1 nit suggestion, but otherwise feel free to merge!

Veir/Interfaces/SideEffectInterfaces.lean
regehr: yes! we can do that later, I think it's a nice mechanism

tobiasgrosser (approved):

PR #1188, opened 2026-08-01
fix: a few things should have been marked ConstantLike or not-side-effecting

regehr:
the introduction of ConstantLike missed a couple of things:
- riscv.lui (trivially constant-like)
- llvm.mlir.addressof (not as obvious, but it's constant-like in upstream MLIR so we can do it too)

here we also make llvm.mlir.addressof not side-effecting (I think trivially true)


math-fehr (approved):
Cool!

PR #1189, opened 2026-08-02
ci: add speed-optimized Namespace CI workflow

tobiasgrosser:
This second workflow runs the same build, unit tests and filecheck tests as `Lean Action CI` using a Namespace runner with 32 cores and high-performance caching. As a result, the build time for warm runs reduces from 160s to 16s (GitHub vs. Namespace).

The Namespace CI and the GitHub runners have the same test coverage, with the exception of the
ExArray steps, which we leave out as they rarely change. However, the Namespace CI workflow uses a custom docker base image with `mlir-opt`, `uv`, `lit`, `filecheck` and `elan` included. Namespace caches these in base docker image, such that we spend no time installing these tools at each CI run,. The  namespace runner also uses a cache volume to cache the lean compiler, temporary build artifacts (e.g., the .lake folder), and offers a local git mirror that caches the git checkout.
 
Namespace builds the image for its runner from a profile in the Namespace dashboard, so the Dockerfile we add is a copy to edit and paste back. A profile build has no build context, so`lit` and `filecheck` are pinned by hand rather than resolved from `uv.lock`.

### DIFF
--- a/.github/docker/namespace-base-image.Dockerfile
+++ b/.github/docker/namespace-base-image.Dockerfile
@@ -34,9 +34,6 @@ RUN apt-get update \
       mlir-22-tools \
       # leanc links through the system toolchain.
       build-essential \
-      # ExArray's C code resolves GMP through pkg-config.
-      libgmp-dev \
-      pkg-config \
  && ln -s /usr/bin/mlir-opt-22 /usr/bin/mlir-opt \
  && rm -rf /var/lib/apt/lists/*
 
--- a/.github/docker/namespace-base-image.Dockerfile
+++ b/.github/docker/namespace-base-image.Dockerfile
@@ -42,7 +42,7 @@ RUN apt-get update \
 
 # `uv` provides both the Python interpreter and the test dependencies, so no
 # separate Python install is needed.
-COPY --from=ghcr.io/astral-sh/uv:0.9.7 /uv /usr/local/bin/uv
+COPY --from=ghcr.io/astral-sh/uv:0.12.2 /uv /usr/local/bin/uv
 
 USER runner
 
--- a/.github/docker/namespace-base-image.Dockerfile
+++ b/.github/docker/namespace-base-image.Dockerfile
@@ -1,0 +1,79 @@
+# Custom base image for the Namespace runner profile used by
+# `lean_action_ci_namespace.yml`.
+#
+# This file is not built by CI. Namespace builds it from the Dockerfile field of
+# the runner profile in the dashboard, so the copy here is the source of truth
+# to edit and paste back. A profile build has no build context: nothing can be
+# `COPY`ed from the repository, which is why the Python tool versions below are
+# spelled out rather than resolved from `uv.lock`.
+#
+# The Namespace runner image ships neither `mlir-opt` nor a Python tool cache,
+# so the workflow installed both on every run. Baking them in removes the
+# install, which caching alone cannot do: a cache saves the download, not the
+# unpacking onto a fresh machine.
+
+ARG NAMESPACE_BASE_IMAGE_REF=""
+
+# Your image must build FROM NAMESPACE_BASE_IMAGE_REF
+FROM ${NAMESPACE_BASE_IMAGE_REF} AS base
+
+# Installing needs root; the image switches back to `runner` below, which the
+# GitHub runner software requires.
+USER root
+
+# `noble` matches the Ubuntu 24.04 base image. On a different base image, change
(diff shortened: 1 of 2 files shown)


.github/docker/namespace-base-image.Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+      build-essential \
+      # ExArray's C code resolves GMP through pkg-config.
+      libgmp-dev \
+      pkg-config \
+ && ln -s /usr/bin/mlir-opt-22 /usr/bin/mlir-opt \
+ && rm -rf /var/lib/apt/lists/*
+
+# `uv` provides both the Python interpreter and the test dependencies, so no
+# separate Python install is needed.
+COPY --from=ghcr.io/astral-sh/uv:0.9.7 /uv /usr/local/bin/uv
```
ineol: Maybe use a more recent version of `uv`? from what I understand the CI action we use on the github runner installs the most recent version unless instructed otherwise?

.github/workflows/lean_action_ci_namespace.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Create the build directory
+        run: mkdir -p ${{ github.workspace }}/.lake
+
+      # `~/.elan` holds the Lean toolchain and `.lake` the build tree; together
+      # they dominate the run.
+      - name: Set up caches
+        uses: namespacelabs/nscloud-cache-action@v1
+        with:
+          path: |
+            ~/.elan
```
ineol, line 36: I don't think it's a big issue, but will the cache accumulate unused toolchain versions over time?

ineol (approved):
I'm no namespace specialist but LGTM

.github/workflows/lean_action_ci_namespace.yml
tobiasgrosser: That should be fine. They will eventually be deleted when the cache overflows.

.github/docker/namespace-base-image.Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+#
+# `HOME` is set explicitly because `USER` alone does not reliably set it during
+# a build, and every default path below is derived from it.
+ENV HOME=/home/runner
+ENV PATH=/home/runner/.local/bin:/home/runner/.elan/bin:$PATH
+
+# Pinned to the versions `uv.lock` resolves. Keep them in step by hand: the
+# lockfile cannot be read here, and `filecheck` versions differ enough to change
+# which tests pass.
+RUN uv python install 3.13 \
```
Jonrempel, line 59: these are pinned by hand and the comment says filecheck versions change which tests pass, maybe worth a small check in the github workflow that diffs them against uv.lock + fails if they drift? this way drift won't silently pass

.github/docker/namespace-base-image.Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+      -o /etc/apt/trusted.gpg.d/llvm-snapshot.asc \
+ && echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-22 main" \
+      > /etc/apt/sources.list.d/llvm.list \
+ && apt-get update \
+ && apt-get install -y --no-install-recommends \
+      mlir-22-tools \
+      # leanc links through the system toolchain.
+      build-essential \
+      # ExArray's C code resolves GMP through pkg-config.
+      libgmp-dev \
```
Jonrempel: from what I understand the workflow leaves out the ExArray steps. so are libgmp-dev and pkg-config still needed here? the comment says they're for ExArray's C code, so maybe they can drop out together?

.github/docker/namespace-base-image.Dockerfile
tobiasgrosser: We now use 0.12.2.

.github/docker/namespace-base-image.Dockerfile
tobiasgrosser: Good point. I dropped these for now.

.github/docker/namespace-base-image.Dockerfile
tobiasgrosser: This is a great idea. I tried this and this works, but it adds 1s to the CI time. Given that we try hard to avoid any extra CI time, and I feel the filecheck drifting is less serious than the comment suggests, I decided to not keep it.

Jonrempel: makes sense, thanks for checking.

PR #1190, opened 2026-08-03
fix(passes): support mixed reduction attrs in mod_arith lowering

Seonyoung-Cheon:
Fixed a bug where an existing Barrett reduction attribute was being overwritten with `full`, and added an example file (`Test/Passes/ModArithToArith/reduction.mlir`) to test lowering with different reduction attrs.


tobiasgrosser:
@Seonyoung-Cheon, feel invited to already address the copilot issues until you get feedback from @AlexanderViand.

AlexanderViand (approved):
Nice catch! LGTM :)

PR #1191, opened 2026-08-03
feat: Prove a legalization pattern based on anyext

nchappe:
widening of 32-bit adds to 64-bit adds

### DIFF
--- a/Veir/Passes/Legalization/Proofs.lean
+++ b/Veir/Passes/Legalization/Proofs.lean
@@ -1,0 +1,30 @@
+module
+
+public import Veir.Data.LLVM.Int.Basic
+public import Veir.Data.Casting
+public import Veir.Data.Refinement
+
+meta import Std.Tactic.BVDecide
+meta import Std.Tactic.BVDecide.Reflect
+
+import Veir.ForLean
+
+public section
+
+namespace Veir.Data.LLVM
+
+/--
+Prove the correctness of `llvm.add` widening with anyext.
+-/
+theorem add_widening (i i' : Veir.Data.LLVM.Int 32) (ext ext' : BitVec 32) (nuw nsw : Bool) :
+    LLVM.Int.add i i' nuw nsw ⊒
+      Veir.Data.LLVM.Int.trunc (Veir.Data.LLVM.Int.add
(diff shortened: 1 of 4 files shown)


Veir/Passes/Legalization/Proofs.lean
```diff
@@ -0,0 +1,30 @@
...
+public import Veir.Data.Refinement
+
+meta import Std.Tactic.BVDecide
+meta import Std.Tactic.BVDecide.Reflect
+
+import Veir.ForLean
+
+public section
+
+namespace Veir.Data.LLVM
```
naveen-seth, line 13: Can we apply this to LLVM.Int / Veir.Data.LLVM.Int in add_widening?

naveen-seth (commented):
Minor nit: I'd keep the naming aligned (`ext` -> `anyext`).
Hopefully we can soon use https://github.com/opencompl/pbvnto1 to make add_widening generic in a follow-up and avoid having `add_widening16_64`, `add_widening8_64`, etc.

Otherwise, LGTM!
(Could we get another review as well? I'm still pretty new to the project.)

nchappe:
Thanks for the review! I applied your suggestions except for the `anyext` renaming.
The reason for naming it `ext` and not `anyext` is that it extends an int with a specific bitvector. The "any" part of anyext then comes from the forall in the theorem statement. I clarified the documentation comment.

Veir/Passes/Legalization/Proofs.lean
```diff
@@ -0,0 +1,29 @@
...
+
+/--
+Prove the correctness of `llvm.add` widening with anyext.
+-/
+theorem add_widening_32_64 (i i' : LLVM.Int 32) (ext ext' : BitVec 32) (nuw nsw : Bool) :
+    LLVM.Int.add i i' nuw nsw ⊒
+      Veir.Data.LLVM.Int.trunc (LLVM.Int.add
```
naveen-seth:
```suggestion
      LLVM.Int.trunc (LLVM.Int.add
```

naveen-seth (approved):
I see, that makes sense to me now! 
LGTM

tobiasgrosser (approved):

PR #1193, opened 2026-08-04
ci: drop the no-op `source .envrc` step

tobiasgrosser:
Each `run:` block executes in its own shell, so the `LEAN_CC` and `LEAN_AR` exports never reached any later step. The build has always run without them.

Introduced in 188fc1754 ("Add ExArray to CI", #620), alongside the ExArray build and test steps it was presumably meant to configure.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -13,9 +13,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Source environment variables
-        run: source .envrc
-
       - name: Add LLVM apt repository
         uses: gerlero/add-apt-repository@v1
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
@ineol, what was the intention behind adding this?

ineol:
I think it's only necessary if another package uses the `ExArray` package, which is not happening so I didn't realise it's a noop. I think we could also just remove the `ExArray` package all together until we need it.

tobiasgrosser:
Will this reduce test coverage?

ineol:
No I don't think it will.

tobiasgrosser:
Let's leave it as it is for now. We should eventually establish test coverage for ExArray.

PR #1194, opened 2026-08-04
chore: git ignore Vim swap files

tobiasgrosser:
The repository already ignores Emacs backup and auto-save files, but not the Vim equivalents, so an open buffer leaves a `.<name>.swp` in the working tree that `git add -A` happily stages.

`*.sw[a-p]` covers the whole rotation Vim uses: `.swp` first, then `.swo`, `.swn` and downwards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ lean_packages/
 *~
 \#*
 .#*
+*.sw[a-p]
 build
 GPATH
 GRTAGS


luigirinaldi (approved):
Looks good!

tobiasgrosser:
@ineol, can you check the history of this change?

ineol:
What do you mean by history?

PR #1195, opened 2026-08-04
test(PDL): cross-check the PDL round-trips against MLIR

tobiasgrosser:
`MLIR_ROUNDTRIP` pipes the test through `mlir-opt --mlir-print-op-generic` before and after `veir-opt`, so the generic form VeIR accepts and prints is validated against real MLIR rather than only against VeIR itself.

Only these three tests qualify. `pdl.operand` carries `HasParent<pdl::PatternOp>` in MLIR, so `mlir-opt` rejects it outside a `pdl.pattern`:

    error: 'pdl.operand' op expects parent op 'pdl.pattern'

That rules out operand.mlir, and operation.mlir with it, since it sources its value handles from `pdl.operand`. Both can be enabled once `pdl.pattern` is modelled and the tests can be wrapped in one.

### DIFF
--- a/Test/PDL/attribute.mlir
+++ b/Test/PDL/attribute.mlir
@@ -1,4 +1,5 @@
 // RUN: VEIR_ROUNDTRIP
+// RUN: MLIR_ROUNDTRIP
 
 "builtin.module"() ({
   // Define an attribute:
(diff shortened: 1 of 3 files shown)


tobiasgrosser:
Merged as obvious.

PR #1197, opened 2026-08-04
feat(PDL): add pdl.pattern and pdl.rewrite operations 

tobiasgrosser:
`pdl.pattern` and `pdl.rewrite` arrive as a pair because `pdl.rewrite` is the
terminator of a `pdl.pattern` body and neither is well-formed without the
other. Together they are the first PDL operations to carry a region, and the
first container the entity operations can actually live in.

`pdl.pattern` holds a non-negative 16-bit `benefit` and an optional `sym_name`,
and its single-block body must end in a `pdl.rewrite`; a `pdl.rewrite` is
either external, naming a native function and leaving its region empty, or
inline, supplying a body and taking no external arguments. Its body sets the
`hasNoTerminator` hook from #1198, so it stays an ordinary SSACFG region as
MLIR models it, rather than being encoded as a graph region.

The whole-pattern invariants MLIR checks in `PatternOp::verifyRegions` — `pdl`
operations only, at least one `pdl.operation`, and a connected component — are
not modelled yet, and neither is `SingleBlock` on the rewrite body.

### DIFF
--- a/Veir/Dialects/PDL/OpInfo.lean
+++ b/Veir/Dialects/PDL/OpInfo.lean
@@ -15,7 +15,9 @@ inductive PDL where
 | attribute
 | operand
 | operation
+| pattern
 | result
+| rewrite
 | type
 deriving Inhabited, Repr, Hashable, DecidableEq
 
@@ -25,7 +27,9 @@ match op with
 | .attribute => PDLAttributeProperties
 | .operand => Unit
 | .operation => PDLOperationProperties
+| .pattern => PDLPatternProperties
 | .result => PDLResultProperties
+| .rewrite => PDLRewriteProperties
 | .type => PDLTypeProperties
 
 def PDL.fromAttrDict
@@ -40,7 +44,9 @@ def PDL.fromAttrDict
     else
       .ok ()
   | .operation => PDLOperationProperties.fromAttrDict attrDict
+  | .pattern => PDLPatternProperties.fromAttrDict attrDict
   | .result => PDLResultProperties.fromAttrDict attrDict
+  | .rewrite => PDLRewriteProperties.fromAttrDict attrDict
   | .type => PDLTypeProperties.fromAttrDict attrDict
 
 def PDL.toAttrDict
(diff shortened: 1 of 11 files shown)


tobiasgrosser:
This is conceptually `stacked` on top of #1197, but we do not use stacking as GitHubs stack-support currently does not support merge queues.

jumerckx:
> This is conceptually stacked on top of https://github.com/opencompl/veir/pull/1197, but we do not use stacking as GitHubs stack-support currently does not support merge queues.

Could this then still be opened as a PR into `tobias/pdl-pattern-rewrite`, perhaps as draft. That way the diff is easier to parse?

Veir/Dialects/PDL/OpInfo.lean
```diff
@@ -193,6 +237,42 @@ def PDL.verifyLocalInvariants {OpInfo : Type} [HasOpInfo OpInfo] [HasDialect OpI
...
+        (op.getProperties! ctx.raw (ofDialect OpInfo PDL.rewrite))
+    /- The operands are the optional `root` followed by the `externalArgs`. -/
+    let segments ← op.verifyOperandSegmentSizes ctx opIn props.operandSegmentSizes 2
+    let rootCount := segments[0]!
+    if rootCount > 1 then
+      throw "Expected at most 1 `root` operand"
+    if rootCount = 1 then
+      if ((op.getOperand! ctx.raw 0).getType! ctx.raw).val
+          ≠ (PDL.OperationType.mk : TypeAttr).val then
+        throw "Expected the `root` operand to be of type '!pdl.operation'"
```
luigirinaldi, line 272: Can the `rootCount` be zero?

luigirinaldi (commented):
Syntactically everything looks good, just a quick curiosity about root count?

Veir/Dialects/PDL/OpInfo.lean
jumerckx: Yes, it's optional: https://mlir.llvm.org/docs/Dialects/PDLOps/#pdlrewrite-pdlrewriteop
```
// Specify a rewrite inline using PDL, automatically selecting root:
pdl.rewrite {
  %op1 = pdl.operation "foo.op"(%arg0, %arg1)
  %op2 = pdl.operation "bar.op"(%arg0, %arg1)
  pdl.replace %root1 with %op1
  pdl.replace %root2 with %op2
}

```

Veir/GlobalOpInfo.lean
```diff
@@ -112,6 +113,10 @@ def OpCode.getRegionKind (opCode : OpCode) (_index : Nat) : RegionKind :=
   | .builtin .module
   | .builtin .unregistered
   | .test .test => .Graph
+  /- MLIR marks the body of a `pdl.rewrite` `NoTerminator`. VeIR ties the
+     "needs a terminator" requirement to the region kind, so the body is
+     modelled as a graph region, which is also single-block as MLIR requires. -/
+  | .pdl .rewrite => .Graph
```
jumerckx: Why is this, is this a temporary VEIR thing? I don't see why a graph region should not be able to contain a terminator.

jumerckx (approved):

Veir/GlobalOpInfo.lean
tobiasgrosser: See https://github.com/opencompl/veir/pull/1198/changes. This should allow us to clean it up.

tobiasgrosser: Fixed.

PR #1198, opened 2026-08-04
feat(verifier): add a `NoTerminator` hook to `HasDialectOpInfo`

tobiasgrosser:
VeIR tied the "every block ends in a terminator" requirement to the region
kind, which conflates two independent properties: a graph region implies no
terminator, but the converse does not hold. MLIR gives `pdl.rewrite` a body
that is an ordinary SSACFG region and yet carries `NoTerminator`, so modelling
it as a graph region relaxes the terminator rule only by also dropping SSA
dominance.

`HasDialectOpInfo` gains a `hasNoTerminator` hook, and the verifier's block
check now depends on it alone rather than on the region kind. This is not a
pure deletion: `builtin.module`, `builtin.unregistered` and `test.test` were
relying on their graph region kind to skip the check and now say so directly.
The hook defaults to `false`, so no other dialect changes behaviour or needs
updating, and `getRegionKind` is left governing only SSA dominance and the
at-most-one-block rule.

The hook has no user on `main` yet — the first is `pdl.rewrite` in #1197, which
currently works around this with a graph region and can then keep an SSACFG
body.

### DIFF
--- a/Veir/Dialects/Builtin/OpInfo.lean
+++ b/Veir/Dialects/Builtin/OpInfo.lean
@@ -52,6 +52,13 @@ def Builtin.hasSSADominance (op : Builtin) (_index : Nat) : Bool :=
   | .module | .unregistered => false
   | _ => true
 
+/-- A `builtin.module` body holds no terminator, and an unregistered operation
+    makes no promise about its regions. -/
+def Builtin.hasNoTerminator (op : Builtin) (_index : Nat) : Bool :=
+  match op with
+  | .module | .unregistered => true
+  | _ => false
+
 #generate_dialect Builtin
 
 instance : HasDialectOpInfo Builtin where
@@ -64,6 +71,7 @@ instance : HasDialectOpInfo Builtin where
   readsMemory := Builtin.readsMemory
   isConstantLike := Builtin.isConstantLike
   hasSSADominance := Builtin.hasSSADominance
+  hasNoTerminator := Builtin.hasNoTerminator
 
 end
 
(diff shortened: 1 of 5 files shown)


jumerckx:
> a graph region implies no terminator, but the converse does not hold.

I believe they are fully orthogonal. e.g., `emitc.expression` has a single graph region with a `emitc.yield` terminator.

math-fehr:
Yep, as @jumerckx says, technically graph regions are expected to have terminators by default

tobiasgrosser:
> > a graph region implies no terminator, but the converse does not hold.
> 
> I believe they are fully orthogonal. e.g., `emitc.expression` has a single graph region with a `emitc.yield` terminator.

OK, I now made it fully orthogonal. Thank you for the comment.

Veir/Dialects/Builtin/OpInfo.lean
```diff
@@ -52,6 +52,13 @@ def Builtin.hasSSADominance (op : Builtin) (_index : Nat) : Bool :=
   | .module | .unregistered => false
   | _ => true
 
+/-- A `builtin.module` body holds no terminator, and an unregistered operation
+    makes no promise about its regions. -/
+def Builtin.hasNoTerminator (op : Builtin) (_index : Nat) : Bool :=
+  match op with
+  | .module | .unregistered => true
```
math-fehr, line 59: Just to mention this (no changes needed), this is the kind of things that make `unregistered` a pain.
Here, if we have an unregistered operation with a terminator, then the verifier will fail. If we instead made this `false`, then we would fail if it has no terminator.

math-fehr (approved):
All good, I would personally also make the PR description less verbose, for instance this sentence is something LLMs love using but make no sense in our context:
```
Behaviour is unchanged: lake build is warning-clean, lake test passes, and
lit reports 801 passed as on main, with the same three pre-existing
environmental failures. The hook has no user on main yet — the first is
pdl.rewrite in https://github.com/opencompl/veir/pull/1197, which currently works around this with a graph region
and can then keep an SSACFG body.
```

PR #1200, opened 2026-08-04
feat(PDL): enforce `SingleBlock` on a `pdl.rewrite` body

tobiasgrosser:
MLIR gives `pdl.rewrite` the `SingleBlock` trait, and rejects a body with more than one block. VeIR accepted it: the check had been an incidental consequence of modelling the body as a graph region, and was lost when the body became an ordinary SSACFG region carrying `hasNoTerminator` instead.


tobiasgrosser:
Merge as obvious.

PR #1211, opened 2026-08-05
DRAFT: docs(PDL): plan for a PDL interpreter

tobiasgrosser:
Records what a PDL interpreter has to do and, more usefully, what it does not:
the driver, the name lookup and the property dictionaries already exist, so the
work reduces to turning a `pdl.pattern` into a `RewritePattern`.

Documentation only — adds `docs/pdl-interpreter-plan.md`, creating `docs/`.
First of a stack; the implementation follows in later PRs.

### Reviewer note

Implementing this plan showed two of its claims to be wrong. Both are corrected
by later commits in the stack, but they are worth knowing about while reading
this one:

- **"Interpretation is a single forward pass"** and the accompanying claim that
  MLIR's root-first traversal is an artifact of compiling to PDLInterp. Both are
  false for a matcher of more than one operation. In
  `%inner = pdl.operation "a"; %res = pdl.result 0 of %inner; %outer = pdl.operation "b"(%res)`
  the root is `%outer`, the *last* operation in the body, so program order gets
  the dependency backwards. Root-first follows from the root being the only
  operation known before matching starts.
- **"Every PDL operation is side-effect free"**, used to argue that patterns must
  be consumed. MLIR marks only `pdl.range`, `pdl.result` and `pdl.results`
  `Pure`; the fix is the next commit in the stack.

Merging this as-is is fine — the corrections arrive with the commits that found
them, which is where they are easiest to justify. Flagging it so the design
section is not read as settled.

### DIFF
--- a/docs/pdl-interpreter-plan.md
+++ b/docs/pdl-interpreter-plan.md
@@ -1,0 +1,131 @@
+# PDL Interpreter Implementation Plan
+
+## Goal
+
+Apply the rewrites described by `pdl.pattern` operations to a module, so that a
+pattern written in the PDL dialect has the same effect as a pattern written by
+hand against `PatternRewriter`.
+
+## What already exists
+
+Most of the machinery is in place, which is what makes this tractable:
+
+- `Veir/PatternRewriter/Basic.lean` provides the driver — a worklist,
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
Closing — superseded; the plan will land differently.

PR #1213, opened 2026-08-06
feat: Add `simp_getset` to all `Rewriter` operations

math-fehr:
This will allow us to use the `simp_getset` tactic for proofs about `Rewriter` operations. It is a slightly better simp set as it allows to also `simp` things that require one hypothesis.

`simp_getset` is a tactic to automatically apply these lemmas using `simp`. It is much faster than `grind`, and can be use non-terminally.

This change also modifies the `simp_getset` (`SimpGetSet.lean`) tactic, so that we can annotate both unconditional and conditional rewrites.

### DIFF
--- a/Veir/Rewriter/GetSet/BlockArguments.lean
+++ b/Veir/Rewriter/GetSet/BlockArguments.lean
@@ -3,6 +3,7 @@ module
 public import Veir.Rewriter.Basic
 
 import all Veir.Rewriter.Basic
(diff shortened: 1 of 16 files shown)


tobiasgrosser (approved):
Nice.

I would appreciate if the PR message could (a) explain when we should use this simp set over grind and (b) list the files that have non-trivial changes -- otherwise the actual change is hard to find.

PR #1217, opened 2026-08-06
feat(match): add the `!match.optional<...>` type

tobiasgrosser:
First piece of the `match` dialect, an unmerged upstream proposal that sits between `pdl` and `pdl_interp`: `!match.optional<T>` wraps a handle whose value may be null at match time, and `match.is_not_null` is the only way back to the bare handle. The wrapped type is an arbitrary `Attribute` rather than a fixed enumeration, matching MLIR, where the restriction to PDL handles lives in the type verifier rather than the type — so `Match.OptionalType` is mutually recursive with `Attribute` and joins the existing `mutual` blocks for the inductive, `decEq` and `toString`, with its parser joining the one holding `parseType`, all following `LLVM.ArrayType`. There is no verifier yet, so `!match.optional<i32>` and nested optionals parse and print.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -403,6 +403,19 @@ structure LLVM.ArrayType where
 deriving Repr, Hashable
 
 /--
+  The `!match.optional<...>` type, wrapping a PDL handle type whose value may
+  be null at match time.
+
+  Navigation in the `match` dialect that can fail returns one of these, and
+  `match.is_not_null` is the only way back to the bare handle. The wrapped type
+  is an arbitrary attribute rather than a fixed enumeration because MLIR places
+  no restriction on it beyond it being a type; the verifier narrows it.
+-/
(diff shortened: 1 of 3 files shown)


Veir/IR/Attribute.lean
```diff
@@ -1035,7 +1068,13 @@ def Attribute.toString (attr : Attribute) : String :=
   | .pdlOperationType type => ToString.toString type
   | .pdlValueType type => ToString.toString type
   | .pdlTypeType type => ToString.toString type
+  | .matchOptionalType type => s!"!match.optional<{Attribute.toString type.innerType}>"
```
luigirinaldi: why not define an `ToString` instance for this too?

jumerckx (approved):

Veir/IR/Attribute.lean
tobiasgrosser: Fixed.

PR #1218, opened 2026-08-06
feat(match): add the value-producing `match` operations

tobiasgrosser:
Registers the `match` dialect with its constant and navigation operations — `constant_attribute`, `constant_type`, `constant_types`, the six nullable `get_*` navigations, `get_value_type`, `get_attribute_type`, `get_users`, `extract` and `get_each` — all `Pure`, as MLIR marks them. The two groups left out are left out for concrete reasons: the test operations (`has_name`, `equal`, `check_operand_count`, …) carry an implicit control effect that transfers to the enclosing failure scope on failure, which has no counterpart in VeIR's operation model, and the structural ones (`matcher`, `try`, the switches, `success`) need variadic regions, `NoTerminator` regions with block arguments, and nested symbol references that `FlatSymbolRefAttr` cannot hold. The verifier performs the narrowing MLIR puts in its type verifier and `!match.optional` deliberately does not — a nullable navigation must return `!match.optional<T>` for the right `T`, `extract` and `get_each` must yield the range's element type, and `get_value_type` maps a range to a range. Covered by a round-trip test over all fourteen operations plus five invalid tests; `MLIR_ROUNDTRIP` is skipped because no released `mlir-opt` knows these operations yet.

### DIFF
--- a/Test/Match/constant_types_element_invalid.mlir
+++ b/Test/Match/constant_types_element_invalid.mlir
@@ -1,0 +1,8 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+// Every element of `match.constant_types` is a type.
+"builtin.module"() ({
+  %a = "match.constant_types"() <{value = [i32, 4 : i32]}> : () -> !pdl.range<type>
+}) : () -> ()
+
+// CHECK: match.constant_types: expected 'value' to hold types, but got 4
--- a/Test/Match/extract_element_invalid.mlir
+++ b/Test/Match/extract_element_invalid.mlir
@@ -1,0 +1,10 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+// `match.extract` yields one element of the range, so the result type is the
+// range's element type.
+"builtin.module"() ({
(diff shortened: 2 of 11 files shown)


math-fehr:
Can we rename the dialect to something else than `match` here? This is very ugly when we have to add quotes to it.
Also, would it be possible to not have an LLM-generated descritpion on PRs? Rust adopted yesterday a policy against having LLM-generated PR descriptions, and I think we should adopt it as well (https://blog.rust-lang.org/inside-rust/2026/08/05/rust-langrust-is-adopting-an-llm-policy/)

jumerckx:
> Can we rename the dialect to something else than match here? 

Are you proposing just a different name in the Lean implementation or for the dialect altogether? I think `match` is a descriptive name but I'd be open to alternatives.

math-fehr:
I like `match` as a dialect name, but in the codebase we have to write `<<match>>` because match is a keyword in Lean. Maybe if we name this `match_`/`Match_` in Lean? And keep `match` in `MLIR` syntax?

tobiasgrosser:
Let's close the match dialect for now. I get back to it when there is a bit more time.

PR #1222, opened 2026-08-06
feat: Merge `HasDialectOpInfo` and `HasOpInfo`

math-fehr:
These two classes had no reason anymore to be split anymore, as they are isomorphic. With this, dialect opcodes and the global opcode both define `HasOpInfo`.


tobiasgrosser (approved):
This is perfect. Thank you!

PR #1223, opened 2026-08-06
feat: constant materialization interface

regehr:
ok this is nearly the last piece of constant folding.
what remains are:
- dialect-specific folding tables to capture things like `x * 0 -> 0`
- an actual pass that performs folding -- I'll put this in our canonicalizer if nobody objects
- (hopefully) using a folding interface as part of routine operation creation, like LLVM's `IRBuilder` does

### DIFF
--- a/Veir/Dialects/Arith/OpInfo.lean
+++ b/Veir/Dialects/Arith/OpInfo.lean
@@ -4,6 +4,8 @@ public import Veir.IR.Simp
 public import Veir.IR.OpInfo
 public import Veir.Dialects.Arith.Properties
 public import Veir.Dialects.LLVM.Properties
+public import Veir.Dialects.LLVM.OpInfo
+public import Veir.ConstantMaterialization
 meta import Veir.Meta.OpCode
 
 namespace Veir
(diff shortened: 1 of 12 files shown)


Veir/Dialects/Arith/OpInfo.lean
```diff
@@ -142,6 +144,22 @@ instance : HasDialectOpInfo Arith where
   isConstantLike := Arith.isConstantLike
   hasSSADominance := Arith.hasSSADominance
 
+/--
+Materialize integer results of folded arithmetic operations as `arith.constant`.
+Poison is materialized as `llvm.mlir.poison`.
```
math-fehr, line 147: As a note for another time, MLIR has `ub.poison` for this purpose.

Veir/Dialects/ModArith/OpInfo.lean
```diff
@@ -72,3 +73,17 @@ instance : HasDialectOpInfo Mod_Arith where
...
   hasSSADominance := Mod_Arith.hasSSADominance
+
+/--
+Materialize concrete modular-integer fold results.
+-/
+def Mod_Arith.materializeConstant {OpInfo : Type} [HasOpInfo OpInfo] [HasDialect OpInfo Mod_Arith]
+    (_op : Mod_Arith) (value : RuntimeValue) (type : TypeAttr) : Option (Materialized OpInfo) :=
+  match value, type.val with
+  | .int bw (.val value), .modArithType modType =>
+    if bw = modType.modulus.type.bitwidth then
```
math-fehr, line 80: @AlexanderViand, is this good enough for materialization of modarith?
Or is requiring the same storage type too restrictive?

Veir/Interfaces/FoldInterfaces.lean
```diff
@@ -51,4 +52,15 @@ def OperationPtr.foldsTo (op : OperationPtr)
...
+/--
+Materialize `value` using the materialization hook of `foldingOpType`'s dialect.
+The hook may select a constant-like operation from another dialect.
+-/
+def PatternRewriter.materializeConstant! (rewriter : PatternRewriter OpCode)
+    (foldingOpType : OpCode) (value : RuntimeValue) (resultType : TypeAttr)
+    (insertionPoint : InsertPoint) : Option (PatternRewriter OpCode × OperationPtr) := do
+  let ⟨materializedOpType, properties⟩ ← foldingOpType.materializeConstant value resultType
+  rewriter.createOp! materializedOpType #[resultType] #[] #[] #[] properties
+    (some insertionPoint)
```
math-fehr, line 72: Here, can you return an `Option (PatternRewriter OpCode × Option OperationPtr)` instead?
That would allow us to graciously fail when we do not know how to synthesize a constant, instead of having to fail the entire pass/pipeline.

math-fehr (approved):
Nice! I just have a nit comment on a function, but otherwise all good (maybe just also checking with @AlexanderViand if the modarith folder is good enough or not)

Veir/Dialects/ModArith/OpInfo.lean
AlexanderViand: I think it's fine - we only get the whole i33/i42/etc mess when we start lowering and seeing intermediate types, at the mod_arith.int level, the storage type should usually be the same across a reasonable IR

regehr:
comments addressed!

math-fehr:
Sorry I forgot about this PR!

PR #1224, opened 2026-08-06
feat(interpreter): allow global variables at the top level of a module

regehr:
previously we insisted that main is the only entity at module top level, but we also want to permit globals there


math-fehr (approved):
Nice!
Also, instead of matching `func.func` and `llvm.func`, can we just check if the operation has a function interface?

PR #1227, opened 2026-08-07
feat: add an MLIR-style isFunctionLike interface

regehr:
logically the update to the test case isn't within the scope of this PR but since it's tiny and needs to be done anyway, hopefully that's ok


math-fehr (approved):
Nice! Sorry I missed this PR!

PR #1228, opened 2026-08-07
feat: Add a basic CTree data structure

nchappe:
This is slightly different to the Rocq CTree library as we support only
one kind of nondeterminism, that counts as a tau transition.

This commit only adds the basic data structure and combinators,
with no proofs except for what is needed for the basic definitions
to go through, and no notion of program comparison.

The `Effect` infrastructure is taken as is from `EffectSSA`,
and the rest partly adapts code from `EffectSSA` and `Coinductive`.

Co-authored-by: Alex Keizer [alex@keizer.dev](mailto:alex@keizer.dev)
Co-authored-by: Michael Sammler [michael.sammler@ist.ac.at](mailto:michael.sammler@ist.ac.at)


nchappe:
Fixed minor problems, added some comments, and updated license

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+public import Coinductive
+public import CTree.Effect
+
+public section
+
+namespace CTree
+open Coinductive Lean.Order
+
+
+inductive E1In : Type u where
```
ineol: Add doc?

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+| e1
+
+@[expose]
+def E1 (e : E1In.{u}) : Type u :=
+  match e with
+  | .e1 => PUnit
+
+
+-- Low-level CTree definitions
```
ineol: Use `/-! ... -/` for nice sections?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,378 @@
...
+  | .e1 => PUnit
+
+
+-- Low-level CTree definitions
+
+/--
+The basic coinductive functor for CTree, with constructor resp. for leaves, for n-ary choices, and for n-ary events
+-/
+inductive CTreeF {EIn : Type u} (E : EIn → Type u) {CIn : Type u} (C : CIn → Type u)
+    (R : Type v) (CTree : Type w) : Type (max u v w) where
```
ineol, line 54:
```
inductive CTreeF {EIn : Type u} (E : EIn → Type u) {CIn : Type u} (C : CIn → Type u)
    (R : Type v) (CTree : Type (max u v)) : Type (max u v) where
```
also works (and the whole package compiles) (w := max (u v)) but I don't know which one is better

lean-ctrees/CTree/Effect.lean
```diff
@@ -0,0 +1,131 @@
...
+
+/--
+Sum of two effect families: on the sum of the index types, dispatches to the
+corresponding component family.
+-/
+def SumE {ι₁ ι₂ : Type u} (ε₁ : ι₁ → Type u) (ε₂ : ι₂ → Type u) : (ι₁ ⊕ ι₂) → Type u
+  | .inl i => ε₁ i
+  | .inr i => ε₂ i
+
+@[inherit_doc] infixr:30 " ⊕ₑ " => CTree.SumE
```
ineol:
```suggestion
@[inherit_doc] infixr:30 " ⊕ₑ " => SumE
```

lean-ctrees/CTree/Effect.lean
```diff
@@ -0,0 +1,131 @@
...
+
+class Subeffect {ι₁ ι₂} (ε₁ : ι₁ → Type u) (ε₂ : ι₂ → Type v) where
+  map : (i₁ : ι₁) → ((i₂ : ι₂) × (ε₂ i₂ → ε₁ i₁))
+  map_surj : ∀ i₁, Function.Surjective (map i₁).snd := by
+    grind [Function.Surjective]
+
+infix:20 " -< " => Subeffect
+attribute [grind! .] Subeffect.map_surj
+
+/-! ## Subeffect Definitions -/
```
ineol, line 46: I'd lift that before the definition of subeffect

lean-ctrees/CTree/Effect.lean
```diff
@@ -0,0 +1,131 @@
...
+  map t := let ⟨i, f⟩ := (sub.map t); ⟨.inl i, f⟩
+
+/-- The `ε' -< (ε₁ ⊕ₑ ε₂)` instance derived from `ε' -< ε₁` maps to `Sum.inl`. -/
+@[simp] theorem map_eq_inl [ε' -< ε₁] (e : ι') :
+    map (ε₁ := ε') (ε₂ := ε₁ ⊕ₑ ε₂) e = ⟨.inl (map e).fst, (map e).snd⟩ := rfl
+@[simp] theorem mapEff_eq_inl [ε' -< ε₁] (e : ι') :
+    mapEff ε' (ε₁ ⊕ₑ ε₂) e = .inl (mapEff ε' ε₁ e) := rfl
+
+/-- `ε₂` is a sub-effect of `ε₁ ⊕ₑ ε₂`. -/
+instance (priority:=low) instSubSumR {ι₁ ι₂ ι'}
```
ineol:
```suggestion
instance (priority := low) instSubSumR {ι₁ ι₂ ι'}
```

lean-ctrees/CTree/Effect.lean
```diff
@@ -0,0 +1,131 @@
...
+    (map (ε₁ := ε₁ ⊕ₑ ε₂) (ε₂:=ε') <| .inl e) = map e := rfl
+@[simp] theorem mapEff_inl [ε₁ -< ε'] [ε₂ -< ε'] {e : ι₁} :
+    (mapEff (ε₁ ⊕ₑ ε₂) ε' <| .inl e) = mapEff ε₁ ε' e := rfl
+@[simp] theorem map_inr [ε₁ -< ε'] [ε₂ -< ε'] {e : ι₂} :
+    (map (ε₁ := ε₁ ⊕ₑ ε₂) (ε₂:=ε') <| .inr e) = map e := rfl
+@[simp] theorem mapEff_inr [ε₁ -< ε'] [ε₂ -< ε'] {e : ι₂} :
+    (mapEff (ε₁ ⊕ₑ ε₂) ε' <| .inr e) = mapEff ε₂ ε' e := rfl
+
+/-- `ε₁` is a sub-effect of `ε₁ ⊕ₑ ε₂`. -/
+instance (priority:=mid) instSubSumL [sub : ε₁ -< ε₂] : ε₁ -< (ε₂ ⊕ₑ ε') where
```
ineol:
```suggestion
instance (priority := mid) instSubSumL [sub : ε₁ -< ε₂] : ε₁ -< (ε₂ ⊕ₑ ε') where
```

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,378 @@
...
+    | .obj (.ret r) _ => .ret r
+    | .obj (.tau i) k => .tau i k
+    | .obj (.vis i) k => .vis i k
+  unpack_pack := by rintro _ ⟨⟩ <;> simp
+  pack_unpack := by rintro _ (⟨⟨⟩, _⟩ | ⟨⟨⟩⟩) <;> simp <;> funext x <;> cases x
+
+/--
+The high-level CTree datatype
+-/
+abbrev CTree {ι : Type u} (ε : ι → Type u) {κ : Type u} (σ : κ → Type u) (α : Type u) : Type u := CoInd (CTreeF ε σ α)
```
ineol: Are you sure it should be an abbrev? can't we hide it?

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+Folding the head of a CTree
+-/
+def CTree.fold (t : CTreeF E C R (CTree E C R)) : CTree E C R := CoInd.fold _ t
+
+/--
+Unfolding the head of a CTree
+-/
+def CTree.unfold (t : CTree E C R) : CTreeF E C R (CTree E C R) := CoInd.unfold _ t
+
+-- High-level CTree constructors
```
ineol: `/-! -/` ?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,378 @@
...
+
+/--
+A custom n-ary choice that generates a τ transition
+-/
+def CTree.tau (i : CIn) (k : C i → CTree E C R) : CTree E C R := CTree.fold (.tau (.inr i) k)
+
+/--
+A unary choice that generates a τ transition
+-/
+def CTree.tau1 (t : CTree E C R) : CTree E C R := CTree.tauG (.inl .e1) λ _ => t
```
ineol: Isn't it a bit surprising that `tau` is not a `\tau` transition?

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+-/
+def CTree.trigger (i : EIn) : CTree E C (E i) := CTree.vis i (fun x => CTree.ret x)
+
+/--
+A CTree making an n-ary choice and immediately returning
+-/
+def CTree.choose (i : CIn) : CTree E C (C i) := CTree.tau i (fun x => CTree.ret x)
+
+
+-- Basic simp lemmas, notably for proofs of monotonicity required to use partial_fixpoint
```
ineol: `/-! ... -/`?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,378 @@
...
+
+
+-- Basic simp lemmas, notably for proofs of monotonicity required to use partial_fixpoint
+
+@[simp]
+theorem CTree.unfold_fold (t : CTree E C R) :
+  CTree.fold (CTree.unfold t) = t := by simp [CTree.fold, CTree.unfold]
+
+@[simp]
+theorem ret_approx_1 (r : R) n :
```
ineol: I don't think the name is very idiomatic in Lean? @bollu might have a better idea than me

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+inductive CTreeF {EIn : Type u} (E : EIn → Type u) {CIn : Type u} (C : CIn → Type u)
+    (R : Type v) (CTree : Type w) : Type (max u v w) where
+  | ret (r : R)
+  | tau (i : E1In ⊕ CIn) (k : (E1 ⊕ₑ C) i → CTree)
+  | vis (i : EIn) (k : E i → CTree)
+
+open Subeffect (mapEff mapCont)
+
+/--
+The bottom element is defined as the spinning CTree
```
ineol: Is it that spinny? It does not inhabit the coinductive type.

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+    ext
+    constructor
+    · intro h
+      rw [CoInd.le_unfold] at h
+      rcases h with (rfl|⟨i, _, _, _, _, h1, h2⟩); simp
+      rw [<-Coinductive.unfold_fold _ t1, <-Coinductive.unfold_fold _ t2]
+      rw [<-PF.unpack_pack (CoInd.unfold _ t1), <-PF.unpack_pack (CoInd.unfold _ t2)]
+      simp only [h1, h2]
+      right
+      cases i <;> simp [PF.pack, ret, tauG, vis, fold]
```
ineol: non terminal simp instead of `simp only`

ineol:
```suggestion
      cases i <;> simp [PF.pack, ret, tauG, vis, fold] <;> grind
```
also works

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+        right
+        exists ?_, ?_; rotate_left 1
+        constructor; rfl
+        apply Exists.intro
+        constructor; rfl
+        simp_all
+    · rintro (rfl| ⟨_, rfl, rfl⟩ | ⟨_, _, _, rfl, rfl, _⟩|⟨_, _, _, rfl, rfl, _⟩)
+      · simp [CoInd.le_unfold]
+      · apply PartialOrder.rel_refl
+      · simp [CoInd.le_unfold]
```
ineol: also here

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+        right
+        simp [PF.unpack, CTree.tauG, CTree.fold]
+        constructor <;> try rfl
+        grind
+      · simp [CoInd.le_unfold]
+        right
+        simp [PF.unpack, CTree.vis, CTree.fold]
```
ineol: and here

ineol (commented):
A bunch of non terminal `simp`s in the proofs of monotonicity at the bottom of `Definition`.

Also it may be more idiomatic to call the file `Defs.lean` than `Definition.lean`?

lean-ctrees/CTree/Definition.lean
```diff
@@ -0,0 +1,378 @@
...
+    intro hf hg t1 t2 hle
+    apply CoInd.le_leN
+    intro n
+    dsimp only
+    have hlef : (f t1) ⊑ (f t2) := by apply hf; assumption
+    generalize f t1 = t1, f t2 = t2 at hlef
+    induction n generalizing t1 t2; simp [CoIndN.le]
+    unfold CTree.bind
+    rw [CTree.le_unfold] at hlef
+    rcases hlef with (rfl|⟨_, rfl, rfl⟩|⟨_, _, _, rfl, rfl, _⟩|⟨_, _, _, rfl, rfl, _⟩)
```
ineol:
```suggestion
    rcases hlef with (rfl|⟨x, rfl, rfl⟩|⟨_, _, _, rfl, rfl, _⟩|⟨_, _, _, rfl, rfl, _⟩)
```
to get rid of the `rename_i` in the second case.

lean-ctrees/CTree/Defs.lean
nchappe: But it is, the difference is that tau1 is unary, tau is n-ary, and tauG encapsulates both

lean-ctrees/CTree/Definition.lean
nchappe: The coinductive library defines the bottom element as the cofixpoint of the provided inhabitant of the functor applied to unit.
I added a comment.

lean-ctrees/CTree/Defs.lean
nchappe: I don't know either, I added a comment about this

nchappe:
Addressed most of the comments, thanks!
I also renamed `E1` to `C1` as it makes more sense for a unary choice.

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
+-- SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+-- definitions and monotonicity proofs partly adapted from the ones in Coinductive and EffectSSA
```
math-fehr:
```suggestion
/-!
Definitions and monotonicity proofs partly adapted from the ones in Coinductive and EffectSSA
-/
```
?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+public section
+
+namespace CTree
+open Coinductive Lean.Order
+open Subeffect (mapEff mapCont)
+
+
+/--
+The unary choice has only one case
+-/
```
math-fehr, line 31:
```suggestion
/--
Type of the unary choice. It is used to encode Itree-style tau nodes that have only a single children.
-/
```
Or something similar. Just something that explains why we need that type.

Also, just for my knowledge, why is it suffixed with `In`? Is it for `Inductive`?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+@[expose]
+def C1 (c : C1In.{u}) : Type u :=
+  match c with
+  | .c1 => PUnit
+
+
+/-! Low-level CTree definitions -/
+
+-- FIXME is universe w needed?
```
math-fehr: I would mark this as a `TODO` rather than a `FIXME`, since this works either way

lean-ctrees/CTree/Defs.lean
math-fehr: It would be really nice if we could hide it and just use lemmas to manipulate it

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+    | .vis i k => .obj (.vis i) k
+  pack
+    | .obj (.ret r) _ => .ret r
+    | .obj (.tau i) k => .tau i k
+    | .obj (.vis i) k => .vis i k
+  unpack_pack := by rintro _ ⟨⟩ <;> simp
+  pack_unpack := by rintro _ (⟨⟨⟩, _⟩ | ⟨⟨⟩⟩) <;> simp <;> funext x <;> cases x
+
+/--
+The high-level CTree datatype
```
math-fehr, line 94: It would help a bit to describe what epsilon, kappa, sigma, and alpha are here. From the type, I would have no idea what they are

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+-/
+abbrev CTree {ι : Type u} (ε : ι → Type u) {κ : Type u} (σ : κ → Type u) (α : Type u) : Type u := CoInd (CTreeF ε σ α)
+
+/--
+A CTree limited to depth n
```
math-fehr, line 99: Same here, it would be nice to have a description of the arguments.

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+-/
+abbrev CTreeN {ι : Type u} (ε : ι → Type u) {κ : Type u} (σ : κ → Type u) (α : Type u) (n : Nat) : Type u := CoIndN (CTreeF ε σ α) n
+
+
+variable {EIn FIn CIn : Type u} {E : EIn → Type u} {F : FIn → Type u} {C : CIn → Type u} {R R' : Type u} (RR : R → R' → Prop)
+
+/--
+Folding the head of a CTree
```
math-fehr:
```suggestion
Folds the head of a CTree
```

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+-/
+def CTree.fold (t : CTreeF E C R (CTree E C R)) : CTree E C R := CoInd.fold _ t
+
+/--
+Unfolding the head of a CTree
```
math-fehr:
```suggestion
Unfolds the head of a CTree
```

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+-/
+def CTree.unfold (t : CTree E C R) : CTreeF E C R (CTree E C R) := CoInd.unfold _ t
+
+/-! High-level CTree constructors -/
```
math-fehr:
```suggestion
/-! ## High-level CTree constructors -/
```

lean-ctrees/CTree/Defs.lean
alexkeizer: What is the universe of the CTree definition if you do this? I think that's the only place where it really matters

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+/--
+A custom n-ary choice that generates τ transitions
+-/
+def CTree.tau (i : CIn) (k : C i → CTree E C R) : CTree E C R := CTree.fold (.tau (.inr i) k)
+
+/--
+A unary choice that generates a τ transition
+-/
+def CTree.tau1 (t : CTree E C R) : CTree E C R := CTree.tauG (.inl .c1) λ _ => t
+
```
math-fehr, line 138: I feel this naming is quite confusing, and probably hard for people not used to CTrees to understand ?
Would it make sense to rename `tau1` to `silentStep`, `tau` to `choice`, and `tauG` to `choiceG`? I feel it makes more sense to the user than these namings that might onyl be understood by people that understand CTrees already.

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+/-! Low-level CTree definitions -/
+
+-- FIXME is universe w needed?
+/--
+The basic coinductive functor for CTree, with constructor resp. for leaves, for n-ary choices, and for n-ary events.
+-/
+inductive CTreeF {EIn : Type u} (E : EIn → Type u) {CIn : Type u} (C : CIn → Type u)
+    (R : Type v) (CTree : Type w) : Type (max u v w) where
+  | ret (r : R)
+  | tau (i : C1In ⊕ CIn) (k : (C1 ⊕ₑ C) i → CTree)
```
math-fehr, line 56: Should `tau` be renamed to something like `nondet` or something else? People might think that `tau` is the silent step here, while it is also a silent choice here?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+-/
+abbrev CTree {ι : Type u} (ε : ι → Type u) {κ : Type u} (σ : κ → Type u) (α : Type u) : Type u := CoInd (CTreeF ε σ α)
+
+/--
+A CTree limited to depth n
+-/
+abbrev CTreeN {ι : Type u} (ε : ι → Type u) {κ : Type u} (σ : κ → Type u) (α : Type u) (n : Nat) : Type u := CoIndN (CTreeF ε σ α) n
+
+
+variable {EIn FIn CIn : Type u} {E : EIn → Type u} {F : FIn → Type u} {C : CIn → Type u} {R R' : Type u} (RR : R → R' → Prop)
```
math-fehr, line 104: Would it maybe make more sense to use `E₁In` and so on? That would make it more obvious that they are all effects?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+
+/-! High-level CTree constructors -/
+
+/--
+A leaf returning a value
+-/
+def CTree.ret (r : R) : CTree E C R := CTree.fold (.ret r)
+
+/--
+An n-ary choice that generates τ transitions (general case)
```
math-fehr: Can you explain why this is a general case in the documentation?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+-/
+def CTree.trigger (i : EIn) : CTree E C (E i) := CTree.vis i (fun x => CTree.ret x)
+
+/--
+A CTree making an n-ary choice and immediately returning
+-/
+def CTree.choose (i : CIn) : CTree E C (C i) := CTree.tau i (fun x => CTree.ret x)
+
+
+/-! Basic simp lemmas, notably for proofs of monotonicity required to use partial_fixpoint -/
```
math-fehr:
```suggestion
/-!
## Basic simp lemmas

Lemmas used for proofs of monotonicity required to use partial_fixpoint
-/
```

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+
+@[simp]
+theorem CTree.unfold_fold (t : CTree E C R) :
```
math-fehr: This should be renamed `CTree.fold_unfold`. Lean naming is usually from outer to inner

lean-ctrees/CTree/Defs.lean
math-fehr: I think this should be named `CTree.approx_ret_succ` ?

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+  CTree.fold (CTree.unfold t) = t := by simp [CTree.fold, CTree.unfold]
+
+@[simp]
+theorem ret_approx_1 (r : R) n :
+  (CTree.ret (E := E) (C := C) r).approx (n + 1) = CTreeF.ret r := by
+    simp [CTree.ret, CTree.fold, CoInd.fold, PF.map, PF.pack, PF.unpack]
+
```
math-fehr, line 169: In general, for the following lemmas, you want to have names in preorder traversal. So the theorem `A (B (C D)) (E F)` will be named `A_B_C_E`. Also, `+ 1` is `succ` usually.

That's a Lean convention, I don't really remember the reason why.

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,374 @@
...
+@[simp]
+theorem fold_ret_approx_1 (r : R) n :
+  (CTree.fold (CTreeF.ret (E := E) (C := C) r)).approx (n + 1) = CTreeF.ret r :=
+    ret_approx_1 r n
+
+@[simp]
+theorem tau_approx_1 i (k : (C1 ⊕ₑ C) i → CTree E C R) n :
+  (CTree.tauG i k).approx (n + 1) = CTreeF.tau i (fun c => (k c).approx n) := by
+    simp [CTree.tauG, CTree.fold, CoInd.fold, PF.map, PF.pack, PF.unpack]
```
math-fehr: You have quite a few non-terminating `simp`. The issue is that we don't want to have proofs breaking because something new was added to `simp`.
So this (and the following) should be `simp only ...` instead of `simp ...`.

lean-ctrees/CTree/Effect.lean
```diff
@@ -0,0 +1,131 @@
...
+
+infix:20 " -< " => Subeffect
+attribute [grind! .] Subeffect.map_surj
+
+/-! ## Subeffect Definitions -/
+namespace Subeffect
+
+/-- `mapEff` is an abbreviation of the first component of `map`. -/
+@[simp, grind]
+abbrev mapEff {ι₁ ι₂} (ε₁ : ι₁ → Type u) (ε₂ : ι₂ → Type u)
```
math-fehr:
```suggestion
abbrev mapEff {ι₁ ι₂} (ε₁ : ι₁ → Type u) (ε₂ : ι₂ → Type v)
```
Does this work?

lean-ctrees/CTree/Effect.lean
```diff
@@ -0,0 +1,131 @@
...
+    [s : ε₁ -< ε₂] (i₁ : ι₁) : ι₂ :=
+  (s.map i₁).1
+
+/-- `mapCont` is an abbreviation of the second component of `map`. -/
+@[simp, grind]
+abbrev mapCont {ι₁ ι₂} (ε₁ : ι₁ → Type u) (ε₂ : ι₂ → Type u) [s : ε₁ -< ε₂]
```
math-fehr:
```suggestion
abbrev mapCont {ι₁ ι₂} (ε₁ : ι₁ → Type u) (ε₂ : ι₂ → Type v) [s : ε₁ -< ε₂]
```
Same here?

lean-ctrees/CTree/Interp.lean
```diff
@@ -0,0 +1,40 @@
...
+public import CTree.Iter
+
+public section
+
+namespace CTree
+open Coinductive Lean.Order
+
+variable {EIn FIn CIn C'In DIn X Y : Type u} {E : EIn → Type u} {F : FIn → Type u} {C : CIn → Type u} {C' : C'In → Type u} {D : DIn → Type u} {R R' : Type u} (RR : R → R' → Prop)
+
+-- FIXME this should be generalized to a larger result choice family D
```
math-fehr:
```suggestion
-- TODO this should be generalized to a larger result choice family D
```
Since this is not really a bug right?

lean-ctrees/CTree/Interp.lean
```diff
@@ -0,0 +1,40 @@
...
+
```
math-fehr, line 18: Can you add a TODO here to not introduce additional unary `tau` when interpreting a `.tau`?

lean-ctrees/CTree/Interp.lean
```diff
@@ -0,0 +1,40 @@
...
+/--
+Substitute events with subtrees according to the provided handler `h`
+-/
+def interp (h : (i: EIn) → CTree F C (E i)) : CTree E C X → CTree F C X :=
```
math-fehr:
```suggestion
def interp (h : (i : EIn) → CTree F C (E i)) : CTree E C X → CTree F C X :=
```

lean-ctrees/CTree/Interp.lean
```diff
@@ -0,0 +1,40 @@
...
+    | .ret r => return (.inr r)
+    | .tau i k => .tauG i (fun x => return (.inl (k x)))
+    | .vis i k => do
+        let o ← h i
+        return (.inl (k o))
+
+/--
+Substitute choice nodes with subtrees according to the provided handler `h`
+-/
+def refine (h : (i: CIn) → CTree E D (C i)) : CTree E C X → CTree E D X :=
```
math-fehr:
```suggestion
def refine (h : (i : CIn) → CTree E D (C i)) : CTree E C X → CTree E D X :=
```

lean-ctrees/CTree/Iter.lean
```diff
@@ -0,0 +1,23 @@
...
+namespace CTree
+
+/--
+The `iter` combinator loops over `body` while it returns a value in `I`,
+and finally returns when it returns a value in `X`.
+-/
+def iter {I X} (body : I -> CTree E C (I ⊕ X)) (i : I) : CTree E C X := do
+  let r ← body i
+  match r with
+  | .inl i => .tau1 (iter body i)
```
math-fehr, line 18: Is there a reason why we insert a `tau1` here? It doesn't seem like it is needed for the function to be defined?

math-fehr (commented):
Nice! I added a few comments, but most of them are style/documentation ones!
I think it would be nice in `Defs` to have a short description of what `CTrees` are in general as well!

lean-ctrees/CTree/Defs.lean
alexkeizer: You absolutely should be able to hide it, I guess it'd just require some boilerplate. One thing: even though you can hide the definition of CTree, it will still be kinda exposed by the unfold/destructure function (I forget what sammlers library calls it, specifically) which returns a CTreeF, and which needs to return a CTreeF for pattern matching to work.

If you wanted to hide CTreeF completely, you'd likely have to forego `match` notation (or build a custom `match` macro, but that's entirely out of scope)

lean-ctrees/CTree/Defs.lean
alexkeizer: Sammler's library has a bunch of non-idiomatic names; I forget how much of those were in `coinductive`, but I'm sure I saw at least a few in the itree library also!

lean-ctrees/CTree/Iter.lean
alexkeizer: It's not strictly speaking required in Lean, with partial fixpoint, but if you don't have it certain coinductive proofs over `iter` become much more painful (if not impossible).

MackieLoeffel:
Just to confirm, I am fine with the license change for this piece of code, as discussed in private communication.

lean-ctrees/CTree/Defs.lean
nchappe: It's for "Input" as it's the input for the effect. Effects are encoded differently in Rocq so I came up with this naming convention for this development (feel free to suggest another convention).

lean-ctrees/CTree/Defs.lean
nchappe: ITreeExtra heavily uses greek letters, as per Lean conventions.
I chose not to follow that convention in most of the development, because I would have had 6 or so greek letters sometimes, and it felt too much (plus I don't like having too much unicode), but I had left out a few occurences, this is now fixed.

lean-ctrees/CTree/Defs.lean
nchappe: Tau is a silent step. If it's unary (tau1) it's exactly like the "tau" in itrees, and if it's a n-ary silent step it furthermore branches and can thus also be considered a choice.

Naming it nondet may give the impression that it's "just" a choice when it is also a step (unlike the choices in rocq-ctrees). I am also a bit reluctant to name it "choice" for the same reason. Naming should ideally make it clear that it's not exactly the same as in the rocq library.

In the rocq version we would call this operation a "stepping branch" (as opposed to regular non-stepping branches).

Tau is not a perfect name either so I'll think a bit more about it.

lean-ctrees/CTree/Effect.lean
nchappe: Yes to both. Do you want me to change it?

lean-ctrees/CTree/Defs.lean
nchappe: Sometimes I use E/F and sometimes E/E'.
With E1 then there would be a naming conflict between C1 (following the same convention as E1In/E2In/E1/E2/etc.) and C1 (the choice with only one option).
I tend to think it's clear enough as is with the context but I need to think more about it.

nchappe:
I addressed most of @math-fehr's comments, thanks!

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,382 @@
...
+
+variable {EIn FIn CIn : Type u} {E : EIn → Type u} {F : FIn → Type u} {C : CIn → Type u} {R R' : Type u} (RR : R → R' → Prop)
+
+/--
+Folds the head of a CTree
+-/
+def CTree.fold (t : CTreeF E C R (CTree E C R)) : CTree E C R := CoInd.fold _ t
+
+/--
+Unfols the head of a CTree
```
tobiasgrosser:
```suggestion
Unfolds the head of a CTree
```

lean-ctrees/CTree/Defs.lean
```diff
@@ -0,0 +1,382 @@
+-- SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+module
+
+public import Coinductive
+public import CTree.Effect
+
+public section
+
+/-! Definitions and monotonicity proofs are partly adapted from the ones in Coinductive and EffectSSA. -/
```
tobiasgrosser, line 21: I would start with 2-6 sentences of what C-trees are. Maybe a citation of the original C-trees paper and how this implementation differs.

alexkeizer:
> Just to confirm, I am fine with the license change for this piece of code, as discussed in private communication.

The same holds for my contributions included here: I'm ok with them being relicensed

nchappe:
Addressed Tobias' comments! Does anybody have further comments?

nchappe:
It seems like our CI build is failing in coinductive, presumably because of an incompatibility with our lean version, even after updating the library to the latest commit.

tobiasgrosser:
I proposed a patch to update this library to 4.33: https://github.com/ISTA-PLV/coinductive/pull/8. Can you review it?

MackieLoeffel:
Thanks for the PR for updating the coinductive library! I've looked into this briefly before, but got stuck and then had to work on other things. The PR is merged and coinductive is now compatible with 4.33.0.

nchappe:
Thanks for the updates to the coinductive library.
There were some more subtleties with the Lean version bump that @ineol helped me debug, and CI passes now

lean-ctrees/CTree/Defs.lean
tobiasgrosser: Do we want to apply this comment, @nchappe.

nchappe: (A variant of) it was already applied

ineol:
LGTM!

PR #1229, opened 2026-08-07
chore: move isTerminator into HasOpInfo

regehr:
isTerminator was lonely all by itself, and now joins the rest of the family


math-fehr (approved):
Nice!

PR #1230, opened 2026-08-08
feat(test): support MLIR_VALID lit predicate

regehr:
we already have lit support for `MLIR_INVALID` to assert that upstream MLIR does not accept a particular piece of code. this PR adds its friend, asserting that upstream MLIR does accept it. I plan to use this in upcoming PRs.

this PR also uses `MLIR_VALID` once, to ensure that it works

### DIFF
--- a/Test/Verifier/arith_select_float_ok.mlir
+++ b/Test/Verifier/arith_select_float_ok.mlir
@@ -1,4 +1,5 @@
 // RUN: veir-opt %s | filecheck %s
+// RUN: MLIR_VALID
 
 // arith.select accepts values of any type, including floats.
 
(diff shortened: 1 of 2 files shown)


math-fehr (commented):
Hmm, is there any case where we want `MLIR_VALID` but not `MLIR_ROUNDTRIP`?

regehr:
> Hmm, is there any case where we want `MLIR_VALID` but not `MLIR_ROUNDTRIP`?

well, I don't feel that strongly about it, but it feels to me like these express different intent.

MLIR_VALID means "MLIR accepts this thing (that Veir also accepts)"

MLIR_ROUNDTRIP means "do Vier + MLIR act as a fixpoint for this piece of text"

tobiasgrosser:
I am also curious when we would not want to use MLIR_ROUNDTRIP. The reason we have it is to ensure that the syntax veir emits matches the syntax of MLIR. Would we want this not to always be true for output that MLIR accepts?

regehr:
well, should I close this? I don't have a justification beyond what I explained above

PR #1231, opened 2026-08-08
feat(riscv): basic datalayout support

regehr:
this only does the RISC-V side, but we also want datalayout in the LLVM interpreter since it interacts with memory objects as well

### DIFF
--- a/Veir/DataLayout/RISCV64.lean
+++ b/Veir/DataLayout/RISCV64.lean
@@ -1,0 +1,73 @@
+module
+
+public import Veir.Interfaces.DataLayoutInterfaces
+
+/-!
+# RV64 Data Layout
+
+The fixed data layout used by the RV64 backend.
+-/
+
+namespace Veir.DataLayout
+
+/-- Round `size` up to a positive byte alignment. -/
+private def alignTo (size alignment : Nat) : Nat :=
+  if alignment = 0 then size
+  else ((size + alignment - 1) / alignment) * alignment
+
+/-- The smallest power of two greater than or equal to `n` (and `1` for `0`). -/
+private def powerOfTwoCeil (n : Nat) : Nat :=
+  if n ≤ 1 then 1 else 2 ^ (Nat.log2 (n - 1) + 1)
+
+/--
+  RV64 integer ABI alignment, derived from LLVM's
+  `i1:8-i8:8-i16:16-i32:32-i64:64-i128:128` layout entries. As in LLVM and
+  MLIR, an unlisted width uses the next larger entry, or the largest entry when
+  no larger one exists.
+-/
+private def rv64IntegerAlignment (bitwidth : Nat) : Nat :=
+  if bitwidth ≤ 8 then 1
+  else if bitwidth ≤ 16 then 2
+  else if bitwidth ≤ 32 then 4
+  else if bitwidth ≤ 64 then 8
+  else 16
+
+private def scalarInfo (size alignment : Nat) : DataLayoutTypeInfo :=
+  { size
+    abiAlignment := alignment
+    preferredAlignment := alignment }
+
+/-- Layout facts for the LLVM-compatible fixed-size types supported by VeIR. -/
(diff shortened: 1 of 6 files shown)


Veir/DataLayout/RISCV64.lean
```diff
@@ -0,0 +1,73 @@
...
+
+The fixed data layout used by the RV64 backend.
+-/
+
+namespace Veir.DataLayout
+
+/-- Round `size` up to a positive byte alignment. -/
+private def alignTo (size alignment : Nat) : Nat :=
+  if alignment = 0 then size
+  else ((size + alignment - 1) / alignment) * alignment
```
naveen-seth: I think this is an exact duplicate of `Veir/Interfaces/DataLayoutInterfaces.lean:39`. Can we make that one public and reuse it here?

naveen-seth (commented):
Otherwise, LGTM! I’m still new to the project/this area, so I’ll leave the approval to someone else.

Veir/DataLayout/RISCV64.lean
regehr: ack-- good catch, thanks.
there's some duplication that I can't avoid since some other changes need to happen first, but this one is inadvertent and I'll fix it

regehr:
> Otherwise, LGTM! I’m still new to the project/this area, so I’ll leave the approval to someone else.

fine to approve (or not) as you choose! we typically wait for all reviewers to approve, or at least most, depending on the complexity of the PR

Veir/Interfaces/DataLayoutInterfaces.lean
```diff
@@ -0,0 +1,66 @@
...
+
+public import Veir.IR.Attribute
+
+/-!
+# DataLayoutInterface
+
+Target data layouts answer physical representation queries for IR types.  The
+interface deliberately distinguishes the byte size of a type, its ABI and
+preferred alignments, and its allocation size (the stride between consecutive
+objects).  In particular, an odd-width integer can have a three-byte type size
```
math-fehr:
```suggestion
objects). In particular, an odd-width integer can have a three-byte type size
```

Veir/Interfaces/DataLayoutInterfaces.lean
```diff
@@ -0,0 +1,66 @@
...
+  than on how layout information is obtained (currently fixed RV64 values,
+  eventually perhaps parsed DLTI entries).
+-/
+structure DataLayout where
+  query : Attribute → Option DataLayoutTypeInfo
+
+namespace DataLayout
+
+/-- Round `size` up to a positive byte alignment. -/
+private def alignTo (size alignment : Nat) : Nat :=
```
math-fehr: `alignTo` or `alignsTo`?

math-fehr (approved):
I think that's good to me, I don't know how it's done in MLIR/LLVM right now, but you know better!

regehr:
> I think that's good to me, I don't know how it's done in MLIR/LLVM right now, but you know better!

you can find MLIR's version here:

https://mlir.llvm.org/docs/DataLayout/

the goal in both cases is to provide an interface that answers questions about how objects want to be represented in RAM. there isn't any question that we need to be able to answer these questions, but of course we might disagree on how to best answer them.

the answer here is "these are hard-coded, but then exposed through a generic API." this is far less ambitious than what MLIR provides, which includes things like parsing and printing LLVM-style datalayout strings.

anyway I'm happy to talk about this, discuss alternatives, etc. but for now, this solves practical problems that we've already encountered -- and solved in ad hoc ways.

luisacicolini (approved):
lgtm, thank you :)

naveen-seth (approved):

Veir/Interfaces/DataLayoutInterfaces.lean
regehr: unless you feel strongly I'm going to leave this unchanged-- `alignTo` is what MLIR inherits from LLVM

PR #1232, opened 2026-08-11
feat(mod-arith): add range analysis and unit tests

Seonyoung-Cheon:
- Added ModArithRangeAnalysis.lean unit tests for default and none reductions. Currently supports add, mul, and constants.
- Added ModArith/Analysis directory since the analysis is currently specific to the ModArith dialect rather than generally reusable.


Veir/Dialects/ModArith/Analysis/RangeAnalysis.lean
```diff
@@ -0,0 +1,117 @@
...
+  | .interval lhs, .interval rhs =>
+      if lhs.bitwidth = rhs.bitwidth then
+        let lower := max lhs.lower rhs.lower
+        let upper := min lhs.upper rhs.upper
+        if lower ≤ upper then
+          .interval { bitwidth := lhs.bitwidth, lower, upper }
+        else
+          .bottom
+      else
+        .top
```
AlexanderViand: Why does this (mismatched bitwidth) branch return .top?

UnitTest/ModArithRangeAnalysis.lean
```diff
@@ -0,0 +1,184 @@
...
+
+/-- Expected range for one named SSA value. -/
+private structure ExpectedRange where
+  name : String
+  range : IntegerRangeLattice
+
+private def rangeToString : IntegerRangeLattice → String
+  | .bottom => "bottom"
+  | .top => "top"
+  | .interval r => s!"[{r.lower}, {r.upper}] : i{r.bitwidth}"
```
AlexanderViand: This probably isn't the right point in the code to comment on it (since it's just the printer), but I'm not sure why you're tracking bit width? The range already tells us what the bitwidth is, and because of how mod_arith is defined, (intermediate) ranges might actually be larger than what the storage type could hold?

Veir/Dialects/ModArith/Analysis/RangeAnalysis.lean
```diff
@@ -0,0 +1,117 @@
...
+instance : Top IntegerRangeLattice where
+  top := .top
+
+/-- Construct a singleton range for a known integer literal. -/
+def singleton  (bitwidth : Nat) (value : Int) : IntegerRangeLattice :=
+  .interval { bitwidth, lower := value, upper := value }
+
+/-- Construct an exact range from an integer attribute. -/
+def ofIntegerAttr (attr : IntegerAttr) : IntegerRangeLattice :=
+  singleton attr.type.bitwidth attr.value
```
AlexanderViand: The LLM says this function is unused/dead code?

AlexanderViand (approved):
Very cool! I couldn't do a super deep review on this, but I noticed a few small things. 

It'd be great to get a review from someone with prior experience with the DataFlowFramework (@regehr? @axelcool1234?), too.

regehr:
so just one super shallow observation is that the changes to `UnitTest/ModArithRangeAnalysis.lean` seem to be adding actual functionality inside of a unit test, as opposed to in Veir proper

axelcool1234:
> Very cool! I couldn't do a super deep review on this, but I noticed a few small things. 
> 
> It'd be great to get a review from someone with prior experience with the DataFlowFramework (@regehr? @axelcool1234?), too.

This looks great. I'd love to connect this to the dataflow framework properly once https://github.com/opencompl/veir/pull/854 is merged.

axelcool1234:
Note, you can have this in the Analysis folder, even if it's just for arith. It's probably fine. It's what MLIR does for its integer range analysis (and that's for arith)

https://github.com/llvm/llvm-project/blob/main/mlir%2Finclude%2Fmlir%2FAnalysis%2FDataFlow%2FIntegerRangeAnalysis.h#L69

regehr:
> This looks great. I'd love to connect this to the dataflow framework properly once #854 is merged.

Axel what's blocking #854?

@Seonyoung-Cheon and @AlexanderViand something that I think Axel and I could use is a bit of a worked example-- perhaps this is just some additional explanation of what's going on, in a comment attached to your test cases.

axelcool1234:
> Axel what's blocking #854?

Just some code review from you and Mathieu. I was busy wrapping up my internship. I will work on that now that I'm done.

UnitTest/ModArithRangeAnalysis.lean
Seonyoung-Cheon: I added the bitwidth to handle operands with different widths, but I overlooked that intermediate values can exceed the storage bitwidth. That’s a very valid point. This also makes the `top` case above a bit awkward. I’ll remove the bitwidth and update the commit!

Seonyoung-Cheon:
Thanks @regehr for review!

> so just one super shallow observation is that the changes to `UnitTest/ModArithRangeAnalysis.lean` seem to be adding actual functionality inside of a unit test, as opposed to in Veir proper

So you mean it’d be better to move range inference functions like `inferModArithRange` and `canonicalModArithRange?` out of `UnitTest/ModArithRangeAnalysis.lean` and into the `Analysis` code?




> @Seonyoung-Cheon and @AlexanderViand something that I think Axel and I could use is a bit of a worked example-- perhaps this is just some additional explanation of what's going on, in a comment attached to your test cases.

Sure! I’ll add some comments to make the examples a bit easier to follow :)

Seonyoung-Cheon:
Thanks @axelcool1234 for the review !

> Note, you can have this in the Analysis folder, even if it's just for arith. It's probably fine. It's what MLIR does for its integer range analysis (and that's for arith)
> 
> https://github.com/llvm/llvm-project/blob/main/mlir%2Finclude%2Fmlir%2FAnalysis%2FDataFlow%2FIntegerRangeAnalysis.h#L69

I’m also wondering about the directory structure. I first had it as `Analysis/Dataflow/Domains/RangeDomain.lean`, then moved it under `ModArith/Analysis` since it felt ModArith-specific. If I move it back under Analysis, would `Analysis/Dataflow/ModArithRangeAnalysis.lean` make more sense, like MLIR does, or would `Analysis/Dataflow/Domains/RangeDomain.lean` be better?

regehr:
> I’m also wondering about the directory structure. I first had it as `Analysis/Dataflow/Domains/RangeDomain.lean`, then moved it under `ModArith/Analysis` since it felt ModArith-specific. If I move it back under Analysis, would `Analysis/Dataflow/ModArithRangeAnalysis.lean` make more sense, like MLIR does, or would `Analysis/Dataflow/Domains/RangeDomain.lean` be better?

I'm in favor of not making directory hierarchies too deep, so I'd probably vote for `Analysis/Dataflow/ModArithRangeAnalysis.lean`!

axelcool1234:
> Thanks @axelcool1234 for the review !
> 
> > Note, you can have this in the Analysis folder, even if it's just for arith. It's probably fine. It's what MLIR does for its integer range analysis (and that's for arith)
> > 
> > https://github.com/llvm/llvm-project/blob/main/mlir%2Finclude%2Fmlir%2FAnalysis%2FDataFlow%2FIntegerRangeAnalysis.h#L69
> 
> I’m also wondering about the directory structure. I first had it as `Analysis/Dataflow/Domains/RangeDomain.lean`, then moved it under `ModArith/Analysis` since it felt ModArith-specific. If I move it back under Analysis, would `Analysis/Dataflow/ModArithRangeAnalysis.lean` make more sense, like MLIR does, or would `Analysis/Dataflow/Domains/RangeDomain.lean` be better?
> 

So your code seems to be the start of an instantiation of an AbstractDomain, so feel free to have it under Domains/. Once I add support for plugging in sparse forward dataflow analyses into the dataflow framework, we can use this domain to create an analysis that works with the framework

axelcool1234:
I will say `mulRange` and `addRange` will likely belong in an analysis file once it can be plugged in but we can just keep it in Domains/ for now

Seonyoung-Cheon:
Got it! For now, I’ll keep it as `Analysis/Dataflow/RangeAnalysis.lean`, and we can split out the `Domain` when it gets integrated with the dataflow framework. I think keeping the structure simple in this PR will also make it easier to find later. Thanks!

axelcool1234:
> Got it! For now, I’ll keep it as `Analysis/Dataflow/RangeAnalysis.lean`, and we can split out the `Domain` when it gets integrated with the dataflow framework. I think keeping the structure simple in this PR will also make it easier to find later. Thanks!
> 

Yeah that sounds like a good idea. Thanks for this contribution!

regehr (commented):
so just one small thing:
does this abstract domain have the invariant that upper >= lower? or upper > lower? or neither? 
whatever the invariant is, we should be able to directly add it as a proof obligation in lean something like:
```
structure IntegerRange where
  lower : Int
  upper : Int
  lower_le_upper : lower ≤ upper
deriving DecidableEq, Repr
```

Seonyoung-Cheon:
Yeah, `upper ≥ lower` should be the invariant to include constants. Sounds good, I’ll add it!

Seonyoung-Cheon:
> so just one small thing: does this abstract domain have the invariant that upper >= lower? or upper > lower? or neither? whatever the invariant is, we should be able to directly add it as a proof obligation in lean something like:
> 
> ```
> structure IntegerRange where
>   lower : Int
>   upper : Int
>   lower_le_upper : lower ≤ upper
> deriving DecidableEq, Repr
> ```

I committed (https://github.com/opencompl/veir/pull/1232/changes/b6769d9e21e8c5f7386a71f4f414dcf8326dd0a6) based on your suggestion. Could you take a look and see if this looks right? I got some help from AI, so I’m not totally sure about it yet 😅

regehr:
> I committed ([b6769d9](https://github.com/opencompl/veir/commit/b6769d9e21e8c5f7386a71f4f414dcf8326dd0a6)) based on your suggestion. Could you take a look and see if this looks right? I got some help from AI, so I’m not totally sure about it yet 😅

it looks good! was there any trouble convincing Lean that this invariant is maintained? this is not a part of lean that I really know much about, but it seems like such a great use case for this kind of feature!

Seonyoung-Cheon:
Not too much trouble so far, since there are only a few ops. Lean seems useful that we can add the proof for each op as we go. Thanks for the suggestion!

PR #1233, opened 2026-08-11
chore: getEffects refactor

regehr:
another incremental step towards matching the MLIR effects system. 

this is a refactor, NFC 

it moves our previous Boolean readsMemory and writesMemory into a unified getEffects call

MLIR returns a SmallVector of effects, whereas here we use a structure of bools. I just did not like the way that MLIR's return value is order-sensitive. we may want to revisit this decision later if we want our effect system to be extensible. for now I figured that an easy, readily decidable version of this was what we want.

after this we'll add the allocate and free effects (also as Boolean flags).

still todo is removing our old sideEffects interface, but this requires at least the allocate effect (I think)


math-fehr (approved):
Very nice!

PR #1235, opened 2026-08-13
feat(dataflow): use WfIRContext

axelcool1234:
Switched the dataflow framework over from `IRContext` to `WfIRContext`. Eventually I need to use `WfIRContext.Verified` somehow.


math-fehr (approved):
Nice!

PR #1236, opened 2026-08-14
feat(printer): registerable custom printing hooks

GZGavinZhao:
## Idea

This PR adds registerable custom printers for VeIR operations.

Custom printers use a monadic `OpPrinter` API. The API carries the current IR context through a reader and carries indentation through a state monad.

```lean
abbrev OpPrinter (OpCode : Type) [IsOpCode OpCode] :=
  ReaderT (OpPrinterContext OpCode) (StateT Nat IO)

abbrev CustomPrinter (GlobalOpCode : Type) [IsOpCode GlobalOpCode] :=
  OperationPtr → Printer.OpPrinter GlobalOpCode Unit
```

Custom printers use printer operations such as:

```lean
printString
printSymbolName
printType
printAttribute
printOperand
printRegionArgument
printSuccessor
printOptionalAttrDict
printOptionalAttrDictWithKeyword
printRegion
printFunctionalType
```

Custom printers do not receive `PrintEnv`, indentation values, or recursive printer functions as arguments. The main printer supplies recursive region printing through `OpPrinterContext`.

`HasCustomPrinting` records both the dialect and the opcode aggregate used for printing:

```lean
class HasCustomPrinting
    (Dialect GlobalOpCode : Type)
    [IsOpCode Dialect] [IsOpCode GlobalOpCode]
    [HasDialect GlobalOpCode Dialect] where
  /-- The custom printer for an operation of this type, if it has one. -/
  customPrinter? :
    Dialect → Option (Printer.CustomPrinter GlobalOpCode) :=
      fun _ => none
```

A dialect defines one generic registration that works for every aggregate that contains the dialect:

```lean
instance {GlobalOpCode : Type}
    [IsOpCode GlobalOpCode]
    [HasDialect GlobalOpCode Func] :
    HasCustomPrinting Func GlobalOpCode where
  customPrinter?
    | .func => some (Func.printFuncFunc (GlobalOpCode := GlobalOpCode))
    | _ => none
```

The global aggregate dispatches to dialect-local registrations:

```lean
instance : HasCustomPrinting OpCode OpCode where
  customPrinter?
    | .func f =>
        HasCustomPrinting.customPrinter?
          (Dialect := Func)
          (GlobalOpCode := OpCode)
          f
    | _ => none
```

This design avoids a separate per-dialect helper and avoids transitive `HasDialect` composition in custom-printing registration.

## Result

The printer uses custom operation syntax by default. Use `--print-op-generic` to force generic syntax.

For example:

```text
lake exe veir-opt Test/Func/custom_printing.mlir
```

prints `func.func` operations in syntax such as:

```mlir
func.func @identity(%arg6_0: i32) -> i32 {
  "func.return"(%arg6_0) : (i32) -> ()
}

func.func @external(i32, f32) -> i64
func.func @fn_res() -> (() -> i32)
func.func @op_attr() attributes {"extra_op" = 42 : i32} {
  "func.return"() : () -> ()
}
```

The custom `func.func` printer follows MLIR function-printing behavior. It handles:

- defined and external functions
- zero, one, and multiple results
- function-typed results
- variadic signatures
- private, nested, and public visibility
- operation attributes and extra function properties
- escaped symbol names
- malformed visibility attributes
- entry-block argument printing

`printFunctionalType` mirrors MLIR's `OpAsmPrinter::printFunctionalType`. It prints operand and result types in functional form and parenthesizes multiple results and single function-typed results.

This PR adds printer support. Parser support for the custom `func.func` syntax remains separate. Generic syntax remains available for parser round-trip tests and tools that need stable generic output.

## Organization

`Veir/Printer/CustomPrinting.lean` defines custom-printer types and the user-facing printer operations.

`Veir/Printer.lean` owns recursive operation, block, region, and generic printing. Internal traversal helpers remain private to that module.

`Veir/GlobalOpInfo.lean` registers the global custom-printer dispatcher.

## Validation

- Lean Beam checks pass for affected modules and downstream modules.
- `Test/Func/custom_printing.mlir` passes lit.
- Generic output remains available through `--print-op-generic`.


GZGavinZhao:
Why is CI failing... I'll look at it tomorrow and do some code cleanup before requesting for review.

tobiasgrosser:
@GZGavinZhao, should this really be a stacked PR. Or can you unstack it?

GZGavinZhao:
@tobiasgrosser Yep un-stacked as discussed on Zulip. Was just playing around with the new feature XD

GZGavinZhao:
@math-fehr Could you review the design and see whether this is a direction you like? Check PR description for a short explanation of the design.

math-fehr:
This makes a lot of sense to me!
Also, this is how I intend to implement interfaces in VeIR in general (the typeclass thing), so hopefully later on `HasCustomPrinting` will be autogenerated with some macros!

GZGavinZhao:
I'm happy with the current stack, will do a last check and cleanup today and request a review.

tobiasgrosser:
> I'm happy with the current stack, will do a last check and cleanup today and request a review.

Cool. Let us know when it's ready to review.

GZGavinZhao:
@tobiasgrosser @math-fehr Ready for review. Thanks!

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+namespace Veir
+
+public section
+
+namespace Printer
+
+/--
+Recursive printing capabilities available to a custom (non-generic) printer.
+
+A custom printer is defined outside of the recursive printer family (e.g. in a
```
tobiasgrosser:
```suggestion
A custom printer is defined outside of the recursive printer family (e.g., in a
```

Veir/OpPrinters.lean
```diff
@@ -0,0 +1,30 @@
+module
+
+/-
+# Global custom-printer registry
+
+The `HasCustomPrinting` instance for the global `OpCode`, aggregating each
+dialect's custom printers.
+
+This module imports every dialect's printing module (e.g.
```
tobiasgrosser:
```suggestion
This module imports every dialect's printing module (e.g.,
```

Veir/Printer.lean
```diff
@@ -126,42 +130,63 @@ partial def printBlock (ctx: IRContext OpCode) (block: BlockPtr) (indent: Nat :=
...
 
-partial def printOperation (ctx: IRContext OpCode) (op: OperationPtr) (indent: Nat := 0) : IO Unit := do
+partial def printOperation (ctx: IRContext OpCode) (op: OperationPtr) (pretty : Bool := false) (indent: Nat := 0) : IO Unit := do
   let opStruct := op.get! ctx
+  let opType := opStruct.opType
+  /- Custom (non-generic) syntax dispatch. When `pretty` is set, operations
+     whose dialect registered a custom printer (via `HasCustomPrinting`) are
+     printed in their custom form; everything else falls through to the
+     generic form. -/
+  if pretty then
```
tobiasgrosser: I would prefer to name it `generic` and flip the option. Can you check how the MLIR parser does this? In a way, `pretty` should be the default. It is only not the default, because starting with the generic parser was easier.

tobiasgrosser (approved):

Veir/Printer.lean
GZGavinZhao: > I would prefer to name it `generic` and flip the option. Can you check how the MLIR parser does this?

`veir-opt`, as you described, prints custom syntax by default and has a `--mlir-print-op-generic` flag to skip custom syntax and print in generic form. (`veir-opt` parser always accepts both)

I'm okay with changing `--pretty` to `--mlir-print-op-generic` and adjusting the behavior accordingly. One thing to confirm that you're ok: I need to make `--mlir-print-op-generic` on by default, which is different from `veir-opt`'s off by default, because if we print custom syntax by default then all the test files needs to be updated and I don't think it's a good idea to do that in this PR.

GZGavinZhao: Ah I just read your comments on #946, will create a Zulip thread and discuss there.

math-fehr: I agree with Tobias on this, we should have an option similarly named to MLIR's. In MLIR this is `printGenericOpFormFlag`.
Otherwise, instead of passing `pretty` everywhere, can you encapsulate it in its own structure, like `PrinterOptions`? That would make things much clearer for most functions, instead of just passing a `pretty` boolean.

VeirOpt.lean
```diff
@@ -148,20 +150,23 @@ def parseArgs (args : List String) : Except String VeirOptArgs := do
   -- Consume `--disable-verifiers` if present.
   let disableVerifiers := flags.contains "--disable-verifiers"
   let flags := flags.filter (· != "--disable-verifiers")
+  -- Consume `--pretty` if present.
+  let pretty := flags.contains "--pretty"
+  let flags := flags.filter (· != "--pretty")
```
math-fehr: It would be nice if the option was `print-op-generic` as well

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
+module
```
math-fehr: Can you rename this file something like `CustomPrinting` or something? I think it would fit more than `Basic`.

Veir/OpPrinters.lean
```diff
@@ -0,0 +1,30 @@
+module
```
math-fehr: Instead of having this file here, can you move this code to `GlobalOpinfo.lean`?
The idea is that `GlobalOpInfo.lean` is the only file that contains the "merging of all dialects".

Veir/OpPrinters.lean
```diff
@@ -0,0 +1,30 @@
...
+-/
+
+public import Veir.GlobalOpInfo
+public import Veir.Dialects.Func.Printing
+
+namespace Veir
+
+public section
+
+instance : HasCustomPrinting OpCode where
```
math-fehr: Can you also add `HasCustomPrinting Func` to `Func/Printing.lean`, and then use that one to build `HasCustomPrinting OpCode`? Similar to how it is done in `HasOpInfo`

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+/--
+Recursive printing capabilities available to a custom (non-generic) printer.
+
+A custom printer is defined outside of the recursive printer family (e.g. in a
+dialect's `Printing.lean`), so it cannot call `printRegion`/`printOpList`/...
+directly; instead it receives an `env` record with those entry points, mirroring
+MLIR's `Printer` methods.
+-/
+structure PrintEnv (OpCode : Type) [IsOpCode OpCode] where
+  /-- Print an SSA value (`%x` / `%argN_M`). -/
```
math-fehr:
```suggestion
  /-- Print an SSA value. -/
```
Just to try to not add confusion

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+public import Veir.IR.Basic
+
+namespace Veir
+
+public section
+
+namespace Printer
+
+/--
+Recursive printing capabilities available to a custom (non-generic) printer.
```
math-fehr:
```suggestion
Printing capabilities available to a custom (non-generic) operation printer.
```
This is what most users will read, so this should confuse them less about why we actually need this in the first place.

Also, I added `operation` to make it clearer that this is only used by operations.

Veir/Printer/Basic.lean
math-fehr: I would remove the `Printing.lean`, part, because we will likely change it soon and we will forget to remove it.

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+
+A custom printer is defined outside of the recursive printer family (e.g. in a
+dialect's `Printing.lean`), so it cannot call `printRegion`/`printOpList`/...
+directly; instead it receives an `env` record with those entry points, mirroring
+MLIR's `Printer` methods.
+-/
+structure PrintEnv (OpCode : Type) [IsOpCode OpCode] where
+  /-- Print an SSA value (`%x` / `%argN_M`). -/
+  printValue     : IRContext OpCode → ValuePtr → IO Unit
```
math-fehr: Can you split this into `printRegionArgument` and `printOperand`? Some of the logic will change in the medium-term, and these will have to be separated.
The difference will be that `printRegionArgument` will "allocate" a new name for the operand, while `printOperand` should use an existing one.

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+  /-- Print an operation followed by its siblings. -/
+  printOpList    : IRContext OpCode → OperationPtr → Nat → IO Unit
+  /-- Print a block followed by its siblings. -/
+  printBlockList : IRContext OpCode → BlockPtr → Nat → IO Unit
```
math-fehr: Can you try to follow roughly the API that exist in MLIR for these (https://mlir.llvm.org/doxygen/classmlir_1_1OpAsmPrinter.html)?
So here instead of `printOpList` and `printBlockList`, we would have `printRegion` with an option `printEntryBlockArgs` that would take care of all the logic. No need for the other options for now, we can add them later.

Veir/Printer/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+
+/--
+Recursive printing capabilities available to a custom (non-generic) printer.
+
+A custom printer is defined outside of the recursive printer family (e.g. in a
+dialect's `Printing.lean`), so it cannot call `printRegion`/`printOpList`/...
+directly; instead it receives an `env` record with those entry points, mirroring
+MLIR's `Printer` methods.
+-/
+structure PrintEnv (OpCode : Type) [IsOpCode OpCode] where
```
math-fehr: I think it would be nice if we could use a monadic structure for printing. That way, the user would not have to provide any indentation, which they should not control anyway. I think it would make sense to just add the current identation in the state monad, as well as the `PrintEnv`. And then calling that an `OpPrinter` or something.

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,87 @@
...
+    (entryPtr : BlockPtr) (indent : Nat) : IO Unit := do
+  IO.println "{"
+  match (entryPtr.get! ctx).firstOp with
+  | some firstOp => env.printOpList ctx firstOp (indent + 1)
+  | none => pure ()
+  match (entryPtr.get! ctx).next with
+  | some nextBlock => env.printBlockList ctx nextBlock (indent + 1)
+  | none => pure ()
+  env.printIndent indent
+  IO.print "}"
```
math-fehr: As I mentionned in the other file, this logic should be defined in the `printRegion` given to the user.

Veir/Printer/Basic.lean
math-fehr: It would also make sense to not give the `IO` monad to the user, and instead expose more functions (like `OpAsmPrinter` in MLIR). That way we are always sure we use the right interface and "print" to the right place (terminal vs file).

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,87 @@
...
+
+/--
+  Custom (pretty) printer for `func.func`, mirroring MLIR's
+  `hasCustomAssemblyFormat`. The dispatcher prints the `func.func` op name; this
+  printer handles everything after it: ` @name(%args) -> results { body }`.
+-/
+def Func.printFuncFunc : Printer.CustomPrinter GlobalOpCode := fun env ctx op indent => do
```
math-fehr: You need to print the function attributes as well

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,87 @@
...
+      for i in 1...outs.size do
+        IO.print s!", {outs[i]!}"
+      IO.print ")"
+    let extraEntries := props.extra.entries
+    if extraEntries.size > 0 then
+      IO.print " attributes "
+      IO.print (DictionaryAttr.fromArray extraEntries)
+    IO.print " "
+    Func.printFuncBodyRegion env ctx entryPtr indent
+  | none =>
```
math-fehr: External functions are not printed properly here, because you do not print argument types, or whether it is external or not, etc..

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,87 @@
...
+/--
+  Custom (pretty) printer for `func.func`, mirroring MLIR's
+  `hasCustomAssemblyFormat`. The dispatcher prints the `func.func` op name; this
+  printer handles everything after it: ` @name(%args) -> results { body }`.
+-/
+def Func.printFuncFunc : Printer.CustomPrinter GlobalOpCode := fun env ctx op indent => do
+  let props : FuncFuncProperties := op.getProperties! ctx Func.func
+  let symName := String.fromUTF8! props.sym_name.value
+  let funcType : FunctionType := props.function_type
+  IO.print s!" @{symName}("
```
math-fehr: Can you also add the printing `f @{symName}` as a function? Just to make sure we escape it at some point

math-fehr (commented):
Nice! I added some comments on the infrastructure, I just want to make sure we have a good design so we don't have to change it too much!
Otherwise that's nice, thanks a lot for adding this feature!

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,103 @@
...
+  -- If s matches, print `@s`, else print `@"escaped"` .
+  let isBare : Bool :=
+    !s.isEmpty &&
+    (s.front.isAlpha || s.front == '_') &&
+    s.all (fun c => c.isAlphanum || c == '_' || c == '.' || c == '$')
+  if isBare then
+    Printer.OpPrinter.printString s!"@{s}"
+  else
+    let escaped := s.replace "\\" "\\\\" |>.replace "\"" "\\\"" |>.replace "\n" "\\n"
+    Printer.OpPrinter.printString s!"@\"{escaped}\""
```
math-fehr: Can you move this to `Printing.lean` (if it doesn't exist there yet)? This will be used by more than just the `func` dialect

math-fehr:
So what I meant with `OpPrinter` is that this would be on top of `Printer.lean`. So it would contain the `Env` rather than passing it as arguments, and then you would have `OpPrinter` functions that would call the elements in `Env`, if that makes sense?

tobiasgrosser:
I just looked over this, and this is going in a great direction. @GZGavinZhao, please ping us as when @math-fehr's comments are addressed and you may need another round of reviews.

GZGavinZhao:
@math-fehr can you give a second round of review when you get a chance? Thanks!

Veir/GlobalOpInfo.lean
```diff
@@ -454,3 +456,16 @@ def OpCode.isCommutative (opCode : OpCode) : Bool :=
...
   | .cir .add | .cir .mul | .cir .and | .cir .or | .cir .xor | .cir .min | .cir .max => true
   | _ => false
+
+instance : HasCustomPrinting OpCode where
+  customPrinter? := fun {GlobalOpCode} _ _ op =>
+    match op with
+    | .func f =>
+      -- TODO(gzgz): automatically generate this
+      let : HasDialect GlobalOpCode Func :=
+        HasDialect.comp
```
GZGavinZhao: Is there a better way of achieving the same behavior without using this `HasDialect` transitivity property? I really want to just use the `HasCustomPrinting Func` instance directly.

Veir/IR/OpCode.lean
```diff
@@ -136,6 +136,38 @@ theorem ofDialect_injective {op₁ op₂ : Dialect} :
   intro h
   grind [congrArg (toDialect? Dialect) h]
 
+/-- Compose two dialect embeddings. -/
+@[instance_reducible]
+def comp
```
GZGavinZhao: My gut feeling tells me `HasDialect.comp` shouldn't be necessary here, but I could be wrong. See the comment in `GlobalOpInfo.lean` for the full context.

VeirOpt.lean
```diff
@@ -157,13 +159,16 @@ def parseArgs (args : List String) : Except String VeirOptArgs := do
   -- Consume `--disable-verifiers` if present.
   let disableVerifiers := flags.contains "--disable-verifiers"
   let flags := flags.filter (· != "--disable-verifiers")
+  -- Consume `--mlir-print-op-generic` if present.
+  let printGenericOpForm := flags.contains "--mlir-print-op-generic"
+  let flags := flags.filter (· != "--mlir-print-op-generic")
```
math-fehr: Can we rename this to `--print-op-generic`? Since we are not `mlir` (and I don't want to have `--veir` as a prefix either).

Veir/GlobalOpInfo.lean
math-fehr: What you can do is replace `HasCustomPrinting` with

```lean4
class HasCustomPrinting (Dialect : Type) (GlobalOpCode : Type)
    [IsOpCode Dialect] [IsOpCode GlobalOpCode] [HasDialect GlobalOpCode Dialect] where
  /-- The custom printer for an operation of this type, if it has one. -/
  customPrinter? : Dialect → Option (Printer.CustomPrinter GlobalOpCode) :=
    fun _ => none
```
That way, you never have to do this `comp` thing, and you still only have to define this only once.

Veir/Printer/CustomPrinting.lean
```diff
@@ -0,0 +1,256 @@
...
+
+namespace Veir
+
+public section
+
+/-- Options controlling printing. Mirrors MLIR's `AsmPrinter` options. -/
+structure PrinterOptions where
+  /-- If true, print every operation in generic form, ignoring custom printers. -/
+  printGenericOpForm : Bool := false
+deriving Inhabited, DecidableEq, Repr
```
math-fehr: This should probably be moved to `Printer.lean`, right? It isn't used anywhere in this file.

Veir/Printer/CustomPrinting.lean
```diff
@@ -0,0 +1,256 @@
...
+
+namespace Printer
+
+variable {OpCode : Type} [IsOpCode OpCode]
+
+/-- Context and recursive services for an `OpPrinter` action. -/
+structure OpPrinterContext (OpCode : Type) [IsOpCode OpCode] where
+  irContext : IRContext OpCode
+  private printRegion : Region → Bool → StateT Nat IO Unit
```
math-fehr: Can you comment here what the `Bool` is?
I'm fine if you just do this with `(name : Bool)`

Veir/Printer.lean
```diff
@@ -2,6 +2,9 @@ module
 
 public import Veir.IR.Basic
 public import Veir.Dialects.Builtin.OpInfo
+public import Veir.GlobalOpInfo
```
math-fehr: Can you remove the `GlobalOpInfo` import from here?

Veir/Printer/CustomPrinting.lean
```diff
@@ -0,0 +1,256 @@
...
+/-- Print `indent` levels of indentation (`  ` per level). -/
+def printIndent : OpPrinter OpCode Unit := do
+  let n ← (get : OpPrinter OpCode Nat)
+  for _ in List.range n do
+    printString "  "
+
+/-- Print a newline and then the current indentation. -/
+def printNewlineAndIndent : OpPrinter OpCode Unit := do
+  printNewline
+  printIndent
```
math-fehr: These should not be available to users in general, can you mark these ones private?

Veir/Printer/CustomPrinting.lean
```diff
@@ -0,0 +1,256 @@
...
+/-
+# Custom printing interface
+
+Types for custom (non-generic) operation printing, plus the `HasCustomPrinting`
+interface through which dialects expose their operations' custom printers.
+
+This module sits below the printer machinery (`Veir/Printer.lean`): it only
+contains the types, so that dialect printing modules can define custom
+printers without depending on the printer itself, and so that the printer can
+depend on the dialect modules without an import cycle.
```
math-fehr: This documentation needs to change a bit, because now that most things have been moved, this isn't really correct anymore.

In general, it would be nice if we can get to this state in the future, where `CustomPrinting` almost only defines the types, not really with the implementation. Right now `CustomPrinting` is mostly a file with a lot of implementation, not only for custom parsing.

We can do this in a separate PR btw, at least now we have a better user interface, we can clean up this file later.

Veir/Printer.lean
```diff
@@ -10,177 +13,155 @@ public section
...
-
-partial def printOperation (ctx: IRContext OpCode) (op: OperationPtr) (indent: Nat := 0) : IO Unit := do
+    printRegionImpl region true options
+    OpPrinter.printString ", "
+  let lastRegion := (op.getRegion! ctx (op.getNumRegions! ctx - 1)).get! ctx
+  printRegionImpl lastRegion true options
+  OpPrinter.printString ")"
+
+/-- Print a single operation, dispatching to a custom printer when available and not in generic form. -/
+partial def printOperation (op : OperationPtr) (options : PrinterOptions) : OpPrinter OpCode Unit := do
```
math-fehr: One reason I think we should have a `CustomOpPrinter` monad at some point is that we want to avoid having users calling these operations in the custom operation printer.

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,96 @@
...
+public section
+
+variable {GlobalOpCode : Type} [IsOpCode GlobalOpCode] [HasDialect GlobalOpCode Func]
+
+/--
+Custom printer for `func.func`, mirroring MLIR's `hasCustomAssemblyFormat`.
+The dispatcher prints the `func.func` op name; this printer handles everything
+after it: ` @name(%args) -> results attributes { body }`.
+-/
+def Func.printFuncFunc : Printer.CustomPrinter GlobalOpCode := fun op => do
```
math-fehr, line 27: I think this function do not print any extra operation attributes?

Veir/Dialects/Func/Printing.lean
```diff
@@ -0,0 +1,96 @@
...
+  let ctx ← getContext
+  let props : FuncFuncProperties := op.getProperties! ctx Func.func
+  let funcType : FunctionType := props.function_type
+  -- ` @name` with optional visibility, like MLIR's `printFunctionOp` does
+  -- `visibility` is stored in `extra` if present (e.g. `"private"`).
+  let visibility? := props.extra.entries.find? fun (k, _) => k == "sym_visibility".toUTF8
+  let extraWithoutVisibility := DictionaryAttr.fromArray (props.extra.entries.filter fun (k, value) =>
+    k != "sym_visibility".toUTF8 || match value with
+      | .stringAttr _ => false
+      | _ => true)
```
math-fehr: now you should be able to write `value.isa StringAttr` instead of the match and I believe it should work?

math-fehr (commented):
Wow, that's much nicer, nice!
I added some comments, but I think overall it's getting there.

I think later on we should fix the file organization of it, but we can do this in later PR to not complexify too much that one!

Veir/Dialects/Func/Printing.lean
GZGavinZhao: This custom printing does print extra operation attributes, cf `Test/Func/custom_printing.mlir`:
```mlir
func.func @op_attr() attributes {"extra_op" = 42 : i32} {}
```

math-fehr: my bad!

math-fehr (approved):
I think we are there now! Thanks a lot for all these iterations!

GZGavinZhao:
Hmmm merge queue failed. Likely needs a rebase. Will do that tonight.

PR #1237, opened 2026-08-14
feat: add control flow branch interface

axelcool1234:
Implemented a very simple interface for branch operations. I skipped a lot of extra stuff (such as supporting branch ops that produce operands internally by the branch op - something VeIR has none of right now) that's in here: https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Interfaces/ControlFlowInterfaces.h but it's a start. This will help https://github.com/opencompl/veir/pull/854 which would use this interface. The reason `SuccessorOperands` is a one field structure is to later have additional fields like the number of produced operands like I mentioned. Right now it'd be unused so I left it as is.

### DIFF
--- a/Veir/Interfaces/ControlFlowInterfaces.lean
+++ b/Veir/Interfaces/ControlFlowInterfaces.lean
@@ -41,13 +41,9 @@ def getSuccessorOperands?
   let opType := branchOp.getOpType! raw
   match opType with
   | .cf .br | .llvm .br | .riscv_cf .branch =>
-    -- Unconditional branches have one successor and forward every operation operand.
-    if branchOp.getNumSuccessors! raw ≠ 1 || successorIndex ≠ 0 then
-      none
-    else
-      some {
-        forwardedOperands := branchOp.getOperands! raw
-      }
+    some {
+      forwardedOperands := branchOp.getOperands! raw
+    }
   | _ => do
     -- Determine whether this is a supported conditional branch and how many
     -- fixed operands appear before its successor operand segments.
@@ -66,26 +62,11 @@ def getSuccessorOperands?
     -- Read the operand segment metadata from the operation's typed properties.
     let attrs := Properties.toAttrDict opType (branchOp.getProperties! raw opType)
     let some (.denseArrayAttr sizes) := attrs["operandSegmentSizes".toUTF8]? | none
-    let successorCount := branchOp.getNumSuccessors! raw
     let segmentSizes := sizes.values
-    -- Validate the requested successor and the basic shape of the segment metadata.
-    if successorIndex ≥ successorCount then
-      none
-    if segmentSizes.size ≠ fixedOperandCount + successorCount then
-      none
-    if segmentSizes.extract 0 fixedOperandCount |>.any (· ≠ 1) then
-      none
-    if segmentSizes.any (· < 0) then
-      none
     -- Select the segment corresponding to the requested successor.
     let segmentIndex := fixedOperandCount + successorIndex
     let forwardedCountRaw ← segmentSizes[segmentIndex]?
     let forwardedCount := forwardedCountRaw.toNat
-    let operandCount := branchOp.getNumOperands! raw
-    -- Ensure the segment metadata accounts for every operation operand.
-    let segmentSum := segmentSizes.foldl (init := 0) fun acc value => acc + value.toNat
-    if segmentSum ≠ operandCount then
-      none
     -- Compute the operation operand index where this successor's forwarded values begin.
     let forwardedStart := fixedOperandCount +
       (segmentSizes.extract fixedOperandCount segmentIndex).foldl
--- a/Veir/Interfaces/ControlFlowInterfaces.lean
+++ b/Veir/Interfaces/ControlFlowInterfaces.lean
@@ -19,16 +19,9 @@ structure SuccessorOperands where
   forwardedOperands : Array ValuePtr
 deriving Inhabited, Repr, DecidableEq
 
-namespace SuccessorOperands
-
-/--
-  Return the SSA value forwarded to a successor block argument.
--/
-def getForwardedOperand?
-    (operands : SuccessorOperands) (blockArgumentIndex : Nat) : Option ValuePtr :=
-  operands.forwardedOperands[blockArgumentIndex]?
-
-end SuccessorOperands
+instance : GetElem SuccessorOperands Nat ValuePtr
+    (fun operands blockArgumentIndex => blockArgumentIndex < operands.forwardedOperands.size) where
+  getElem := fun operands blockArgumentIndex h => operands.forwardedOperands[blockArgumentIndex]'h
 
 namespace BranchOpInterface
 
@@ -83,7 +76,7 @@ def getSuccessorOperand?
     (branchOp : OperationPtr) (successorIndex blockArgumentIndex : Nat)
     (raw : IRContext OpCode) : Option ValuePtr :=
   getSuccessorOperands? branchOp successorIndex raw >>= fun operands =>
-    operands.getForwardedOperand? blockArgumentIndex
+    operands[blockArgumentIndex]?
 
 end BranchOpInterface
 
--- a/Veir/Interfaces/ControlFlowInterfaces.lean
+++ b/Veir/Interfaces/ControlFlowInterfaces.lean
@@ -59,6 +59,7 @@ def getSuccessorOperands?
       | .riscv_cf .bltu
       | .riscv_cf .bgeu => some 2
       | _ => none
+    -- TODO: Move the operandSegmentSizes logic to the respective dialects
     -- Read the operand segment metadata from the operation's typed properties.
     let attrs := Properties.toAttrDict opType (branchOp.getProperties! raw opType)
     let some (.denseArrayAttr sizes) := attrs["operandSegmentSizes".toUTF8]? | none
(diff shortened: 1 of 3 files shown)


Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,110 @@
...
+
+namespace SuccessorOperands
+
+/--
+  Return the SSA value forwarded to a successor block argument.
+-/
+def getForwardedOperand?
+    (operands : SuccessorOperands) (blockArgumentIndex : Nat) : Option ValuePtr :=
+  operands.forwardedOperands[blockArgumentIndex]?
+
```
math-fehr: Is there a reason to create a structure here instead of just passing an `Array ValuePtr`?
I feel that would make things simpler here, and expose all functions on `Array` directly instead of having to rewrite them.
Unless you expect new functions to appear?

Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,110 @@
...
+end SuccessorOperands
+
+namespace BranchOpInterface
+
+/--
+  Return the operands passed to `successorIndex` of a branch operation.
+-/
+def getSuccessorOperands?
```
math-fehr, line 35: Can you move this function to `HasOpInfo`?
Look at how it is done for things like `getEffects` for instance.

math-fehr (commented):
Nice! Can you move `getSuccessorOperands` to `HasOpInfo` though? And just return an `Array ValuePtr`?

Veir/Interfaces/ControlFlowInterfaces.lean
regehr: (getEffects just landed now)

Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,110 @@
...
+    (branchOp : OperationPtr) (successorIndex : Nat) (raw : IRContext OpCode) :
+    Option SuccessorOperands :=
+  let opType := branchOp.getOpType! raw
+  match opType with
+  | .cf .br | .llvm .br | .riscv_cf .branch =>
+    -- Unconditional branches have one successor and forward every operation operand.
+    if branchOp.getNumSuccessors! raw ≠ 1 || successorIndex ≠ 0 then
```
regehr: do you need the != 1 check? isn't this guaranteed by the module verifier?

regehr: (and same with successor index)

regehr:
looks nice overall but I definitely agree that the dialect-specific parts of this need to be pushed into the dialect files

regehr:
also use a WfIRContext instead of a raw context?

Veir/Interfaces/ControlFlowInterfaces.lean
axelcool1234: This is using raw `IRContext` rather than `WfIRContext`, to match the other interfaces also handling raw `IRContext`. So I do all the checks myself. Thoughts on this @math-fehr ?

axelcool1234:
> also use a WfIRContext instead of a raw context?

None of the other interfaces I've looked at does this

> looks nice overall but I definitely agree that the dialect-specific parts of this need to be pushed into the dialect files

Will do, I'll move things around

Veir/Interfaces/ControlFlowInterfaces.lean
axelcool1234: > Unless you expect new functions to appear?

I mentioned it in my PR description - this interface is a heavily slimmed down version of the more generic version in MLIR, which can also support "produced" operands generated by the branch op itself (this'd be in the structure). VeIR lacks an op like this so I didn't include it.

Veir/Interfaces/ControlFlowInterfaces.lean
axelcool1234: Problem: `HasOpInfo` is defined before `ValuePtr`. What should I do?

Veir/Interfaces/ControlFlowInterfaces.lean
math-fehr: My bad, I tunneled vision and didn't read the PR description, that's on me sorry

Veir/Interfaces/ControlFlowInterfaces.lean
math-fehr: Okay let's not do that then, I'll be fixing this later then once I figure out where to move interfaces (since `HasOpInfo` seems to be the wrong place now).

Veir/Interfaces/ControlFlowInterfaces.lean
math-fehr: Even `WfIRContext` wouldn't allow this, you would need an `WfIRContext.Verified`.
I think for now that should be fine since we don't have a good way to handle verified operations, so let's keep this?

Veir/Interfaces/ControlFlowInterfaces.lean
axelcool1234: It's okay I forget to read PR descriptions a lot haha

Veir/Interfaces/ControlFlowInterfaces.lean
math-fehr: Actually, we should just remove the check, and return `some` all the time here for the time being?

axelcool1234: > Actually, we should just remove the check, and return `some` all the time here for the time being?

Should I remove all checks? There's probably 7 or 8 of them in here.

math-fehr: Yes, I would say the same for the checks in the second branch, you have a lot of checks about `operandSegmentSizes` that we could remove

Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,110 @@
...
+      none
+    if segmentSizes.extract 0 fixedOperandCount |>.any (· ≠ 1) then
+      none
+    if segmentSizes.any (· < 0) then
+      none
+    -- Select the segment corresponding to the requested successor.
+    let segmentIndex := fixedOperandCount + successorIndex
+    let forwardedCountRaw ← segmentSizes[segmentIndex]?
+    let forwardedCount := forwardedCountRaw.toNat
+    let operandCount := branchOp.getNumOperands! raw
```
math-fehr: Realistically, all this logic (accessing the first or second operand segment) should be moved to somewhere else.
Since we are going to move this code soon-ish somewhere else, I guess it's fine if we keep this for now to move forward.
Can you add a comment like "TODO: Move the operandSegmentSizes logic to the respective dialects"?

Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,110 @@
...
+  /-- The SSA values forwarded to the successor. -/
+  forwardedOperands : Array ValuePtr
+deriving Inhabited, Repr, DecidableEq
+
+namespace SuccessorOperands
+
+/--
+  Return the SSA value forwarded to a successor block argument.
+-/
+def getForwardedOperand?
```
math-fehr: Do you think having a `GetElem` notation would make sense here? By that I mean being able to write `[]` to index an element in the array, instead of calling `getForwardedOperand?`. This would make the syntax much nicer IMO.

Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,110 @@
...
+      forwardedOperands := (branchOp.getOperands! raw).extract
+        forwardedStart (forwardedStart + forwardedCount)
+    }
+
+/-- Return the SSA value forwarded to a successor block argument. -/
+def getSuccessorOperand?
+    (branchOp : OperationPtr) (successorIndex blockArgumentIndex : Nat)
+    (raw : IRContext OpCode) : Option ValuePtr :=
+  getSuccessorOperands? branchOp successorIndex raw >>= fun operands =>
+    operands.getForwardedOperand? blockArgumentIndex
```
math-fehr: This shouldn't be needed if you add a `GetElem` notation

math-fehr (commented):
Okay nice, I just have some comments on the notation then (since you are removing the checks that are not necessary). And then that should be good!

Veir/Interfaces/ControlFlowInterfaces.lean
```diff
@@ -0,0 +1,85 @@
...
+
+/-- The SSA values forwarded from a branch operation to one of its successors. -/
+structure SuccessorOperands where
+  /-- The SSA values forwarded to the successor. -/
+  forwardedOperands : Array ValuePtr
+deriving Inhabited, Repr, DecidableEq
+
+instance : GetElem SuccessorOperands Nat ValuePtr
+    (fun operands blockArgumentIndex => blockArgumentIndex < operands.forwardedOperands.size) where
+  getElem := fun operands blockArgumentIndex h => operands.forwardedOperands[blockArgumentIndex]'h
```
math-fehr, line 24: Can you add a `GetElem?` implementation as well? (Sorry I should have mentionned)

math-fehr (commented):
Nice! Sorry I just have one last request, but I'll merge it right after!

math-fehr:
Thanks!

PR #1238, opened 2026-08-14
chore: Move arith verifier to `Arith/` folder

math-fehr:
This is part of the plan of making all verifiers depend on an arbitrary `OpInfo` instead of `OpCode`.

### DIFF
--- a/Veir/Verifier.lean
+++ b/Veir/Verifier.lean
@@ -126,30 +126,6 @@ def OperationPtr.verifyRISCVuimm6 (op : OperationPtr) (ctx : WfIRContext OpCode)
   else
     pure ()
 
-/--
-  Verify an `arith` extended operation with two same-typed integer operands and
-  two results. The low result always matches the operand type; the high result
-  is either an `i1` overflow flag (`addui_extended` / `subui_extended`) or
-  another value of the operand type (`mulsi_extended` / `mului_extended`).
--/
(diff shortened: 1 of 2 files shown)


Veir/Verifier.lean
```diff
@@ -1086,7 +1017,8 @@ theorem OperationPtr.Verified.arith_constant {op : OperationPtr} {opInBounds}
     op.getNumRegions! ctx.raw = 0 ∧
     ((op.getResult 0).get! ctx.raw).type =
       ⟨(op.getProperties! ctx.raw Arith.constant).value.type, (by grind)⟩ := by
-  simp only [Verified, verifyLocalInvariants, ← getOpType!_eq_getOpType, opType, ne_eq,
+  simp only [Verified, verifyLocalInvariants, Arith.verifyLocalInvariants,
+    ← getOpType!_eq_getOpType, opType, ne_eq,
```
tobiasgrosser, line 1021: I am curious, should these theorems not also be moved to the `Arith/` folder?

math-fehr: That's my plan, but I can do this only after moving the verifier to `OpInfo`.
The issue is that there is no `verifyLocalInvariants` in the `HasOpInfo`. To add this function, we need to first move the verifiers to their dialect folder.
So the order is going to be :
1) Moving the dialect verifiers to `OpInfo`
2) Adding `verifyLocalInvariants` to `HasOpInfo`
3) Moving these theorems to their respective dialects

tobiasgrosser (approved):

PR #1247, opened 2026-08-15
feat: allocate effect

regehr:
this adds an allocate effect to the values returned by `getEffect()`.
as far as I know, this will be sufficient for our current requirements in DCE, CSE, and related passes. as in, once this lands, we can remove our other side effect machinery and use this alone to reason about effects.


tobiasgrosser (approved):
This LGTM. Thanks. 🙏

math-fehr (approved):
Nice!

PR #1248, opened 2026-08-16
feat(verifier): Align llvm.ashr with the other llvm shift operations

naveen-seth:
The verifier previously required the shift amount operand of `llvm.ashr` to have the same type as the result. 
This aligns `llvm.ashr` with the other shift operations in the LLVM dialect, allowing `ashr` to be used on a `llvm.byte` type.

This also aligns the `ashr` operation in the interpreter.

### DIFF
--- a/Test/Interpreter/LLVM/byte.mlir
+++ b/Test/Interpreter/LLVM/byte.mlir
@@ -1,7 +1,7 @@
 // RUN: veir-interpret %s | filecheck %s
 
 "builtin.module"() ({
-  "func.func"() <{sym_name = "main", function_type = () -> (!llvm.byte<64>, i32, i32, !llvm.byte<64>, !llvm.byte<64>)}> ({
+  "func.func"() <{sym_name = "main", function_type = () -> (!llvm.byte<64>, i32, i32, !llvm.byte<64>, !llvm.byte<64>, !llvm.byte<64>)}> ({
     %1 = "llvm.mlir.constant"() <{ "value" = 1 : i64 }> : () -> i64
     %2 = "llvm.mlir.constant"() <{ "value" = 17 : i64 }> : () -> i64
     %3 = "llvm.mlir.poison"() : () -> i8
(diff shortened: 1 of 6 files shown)


regehr:
Naveen, if you want to assert positive compatibility with upstream MLIR, the `MLIR_ROUNDTRIP` lit predicate can be useful!

naveen-seth:
Ah thanks, I didn't know about that! But I don't think it would be possible to use it right now.
`mlir-opt` does not seem to support shift operations on `llvm.byte` yet.

For example, the current `Test/Interpreter/LLVM/byte.mlir` would also fail the `MLIR_ROUNDTRIP` with:
```
veir/Test/Interpreter/LLVM/byte.mlir:16:10: error: 'llvm.lshr' op operand #0 must be signless integer or LLVM dialect-compatible vector of signless integer, but got '!llvm.byte<64>'
    %9 = "llvm.lshr"(%7, %8) : (!llvm.byte<64>, i64) -> !llvm.byte<64>
         ^
```
(Tested with a trunk build of `mlir-opt`.)

Upstream MLIR should also support the shift operations on `llvm.byte` soon, but it looks like we are a bit ahead at the moment:
https://discourse.llvm.org/t/rfc-add-a-new-byte-type-to-llvm-ir/89522#p-357055-h-3-extend-existing-instructions-11

regehr:
> Ah thanks, I didn't know about that! But I don't think it would be possible to use it right now. `mlir-opt` does not seem to support shift operations on `llvm.byte` yet.

do we know LLVM/MLIR's plan for supporting arithmetic on bytes? we might not want to get very far ahead of them here, I don't know-- @math-fehr probably has opinions

math-fehr (commented):
Code-wise everything is good. My only question is why should `llvm.ashr` be extended for the byte type? Is there a document somewhere that describes which operations should be handled? I understand why `llvm.lhsr` makes sense, but I'm not sure about `llvm.ashr`?

naveen-seth:
I didn’t put much thought into why `llvm.ashr` should be extended for the byte type. 
I stumbled over this while working on legalization and mostly assumed it was a bug because it was the odd one out, so maybe it makes sense to close this PR.

PR #1251, opened 2026-08-16
feat: Change definition of `getDefiningOp` and rename to  `definingOp?`

math-fehr:
`getDefiningOp` was defined in two places: In `IR/Basic.lean`, and in `Matching/Basic.lean`. This removes the `Matching/Basic.lean` definition and moves its implementation to `IR/Basic.lean`. As the implementation only requires a `ValuePtr` and no context, it was renamed from `ValuePtr.getDefiningOp` to `ValuePtr.definingOp?`. The function should also now run faster than before, as the previous implementation had 2 unnecessary indirections.

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -155,10 +155,9 @@ private def compareNamedDominators
 /-- Resolve a named SSA value to the operation that defines it. -/
 private def getNamedOperation?
     (recovered : RecoveredNames)
-    (name : String)
-    (irCtx : WfIRContext OpCode) : Option OperationPtr := do
+    (name : String) : Option OperationPtr := do
   let value ← recovered.values[name]?
-  value.getDefiningOp! irCtx.raw
+  value.definingOp?
 
 /--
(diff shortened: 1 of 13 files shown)


tobiasgrosser (approved):
Nice. You may want to add 1 sentence to the PR that briefly explains why/how the definition became simpler.

regehr (approved):
super cool!

I'm seeing this but otherwise LGTM!
```
~/veir$ lake test
⚠ [148/166] Replayed UnitTest.DataFlowFramework.Dominance
warning: UnitTest/DataFlowFramework/Dominance.lean:159:5: Variable name `irCtx` is not explicitly referenced.

The binding can be removed (if unused) or named `_` (if used implicitly).

Note: This linter can be disabled with `set_option linter.unusedVariables false`
~/veir$ 
```

math-fehr:
Oh, thanks a lot, I completely missed it!
I think the CI should definitely start to catch that!

tobiasgrosser:
> Oh, thanks a lot, I completely missed it! I think the CI should definitely start to catch that!

I am surprised it did not catch it. We are building with `--iofail`, so any IO should fail the CI.

PR #1255, opened 2026-08-17
chore: update lean to 4.33.0

tobiasgrosser:
After factoring out bigger changes, this actual upgrade only resolves an ambiguity on the `Int` type, changes some`simp` and `grind` arguments to make sure they continue to work, and one adjustment to changes in the List API.

Finally, we work around a grind bug that will likley be resolved in the next Lean release.

### DIFF
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -520,6 +520,7 @@ theorem isPoison_select {w : Nat} (x y : Int w) (c : Int 1) :
 @[veir_bv_normalize, grind =]
 theorem getValue_select {w : Nat} (x y : Int w) (c : Int 1) (h : (select c x y).isPoison = false) :
     (select c x y).getValue h = if _ : c.getValue = 1#1 then x.getValue else y.getValue := by
+  /- TODO: use `grind` after https://github.com/leanprover/lean4/pull/14829 is resolved. -/
   cases c with
   | poison => simp [select, Id.run, isPoison] at h
   | val c' => by_cases hc : c' = 1#1 <;> simp [select, Id.run, hc]
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -135,8 +135,7 @@ theorem isPoison_add {w : Nat} (x y : Int w) {nsw nuw : Bool} :
       else
         (nsw ∧ BitVec.saddOverflow x.getValue y.getValue) ∨
         (nuw ∧ BitVec.uaddOverflow x.getValue y.getValue) := by
-  simp only [isPoison, add, Id.run, getValue, Bool.decide_or, Bool.decide_and,
-    Bool.decide_eq_true]
+  simp only [isPoison, add, Id.run, getValue]
   simp only [pure]
   grind
 
@@ -153,9 +152,7 @@ theorem isPoison_sub {w : Nat} (x y : Int w) {nsw nuw : Bool} :
       else
         (nsw ∧ BitVec.ssubOverflow x.getValue y.getValue) ∨
         (nuw ∧ BitVec.usubOverflow x.getValue y.getValue) := by
-  simp only [isPoison, sub, Id.run, getValue, Bool.decide_or, Bool.decide_and,
-    Bool.decide_eq_true]
-  simp only [pure]
+  simp only [isPoison, sub, Id.run, getValue, pure]
   grind
 
 @[veir_bv_normalize, grind =]
@@ -523,8 +520,7 @@ theorem isPoison_select {w : Nat} (x y : Int w) (c : Int 1) :
 @[veir_bv_normalize, grind =]
 theorem getValue_select {w : Nat} (x y : Int w) (c : Int 1) (h : (select c x y).isPoison = false) :
     (select c x y).getValue h = if _ : c.getValue = 1#1 then x.getValue else y.getValue := by
-  simp [select, Id.run]
-  grind
+  sorry
 
 @[veir_bv_normalize, grind =]
 theorem isPoison_smax {w : Nat} (x y : Int w) :
@@ -820,7 +816,7 @@ theorem icmp_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w) (p : IntPred)
 theorem select_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w) (c₁ c₂ : Int 1)
     (h₁ : x₁ ⊒ y₁) (h₂ : x₂ ⊒ y₂) (h₃ : c₁ ⊒ c₂) :
     select c₁ x₁ x₂ ⊒ select c₂ y₁ y₂ := by
-  grind
+  sorry
 
 @[veir_bv_normalize, grind =]
 theorem isPoison_freeze {w : Nat} (x : Int w) :
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -819,6 +819,7 @@ theorem icmp_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w) (p : IntPred)
 theorem select_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w) (c₁ c₂ : Int 1)
     (h₁ : x₁ ⊒ y₁) (h₂ : x₂ ⊒ y₂) (h₃ : c₁ ⊒ c₂) :
     select c₁ x₁ x₂ ⊒ select c₂ y₁ y₂ := by
+   /- TODO: use `grind` after https://github.com/leanprover/lean4/pull/14829 is resolved. -/
   cases c₁ with
   | poison => simp [select, Id.run, isRefinedBy_iff]
   | val v =>
--- a/Veir/Data/LLVM/Int/Bitblast.lean
+++ b/Veir/Data/LLVM/Int/Bitblast.lean
@@ -520,7 +520,9 @@ theorem isPoison_select {w : Nat} (x y : Int w) (c : Int 1) :
 @[veir_bv_normalize, grind =]
 theorem getValue_select {w : Nat} (x y : Int w) (c : Int 1) (h : (select c x y).isPoison = false) :
     (select c x y).getValue h = if _ : c.getValue = 1#1 then x.getValue else y.getValue := by
-  sorry
+  cases c with
+  | poison => simp [select, Id.run, isPoison] at h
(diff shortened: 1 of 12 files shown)


Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -523,8 +520,9 @@ theorem isPoison_select {w : Nat} (x y : Int w) (c : Int 1) :
 @[veir_bv_normalize, grind =]
 theorem getValue_select {w : Nat} (x y : Int w) (c : Int 1) (h : (select c x y).isPoison = false) :
     (select c x y).getValue h = if _ : c.getValue = 1#1 then x.getValue else y.getValue := by
-  simp [select, Id.run]
-  grind
+  cases c with
+  | poison => simp [select, Id.run, isPoison] at h
+  | val c' => by_cases hc : c' = 1#1 <;> simp [select, Id.run, hc]
```
math-fehr, line 526: Do you know why this `grind` fails now? It seems like your `by_cases` is necessary from what you wrote, but since you already have `cases c`, why doesn't grind succeed after that? Is there a missing grind rule about `(.val c').getValue`?

Veir/Data/LLVM/Int/Bitblast.lean
```diff
@@ -820,7 +818,15 @@ theorem icmp_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w) (p : IntPred)
 theorem select_mono {w : Nat} (x₁ x₂ y₁ y₂ : Int w) (c₁ c₂ : Int 1)
     (h₁ : x₁ ⊒ y₁) (h₂ : x₂ ⊒ y₂) (h₃ : c₁ ⊒ c₂) :
     select c₁ x₁ x₂ ⊒ select c₂ y₁ y₂ := by
-  grind
```
math-fehr, line 823: Same here, it is a bit sad that `grind` doesn't work anymore here

Veir/Interpreter/Refinement/Monotonicity.lean
```diff
@@ -171,7 +171,9 @@ theorem interpretOpList_mono
         r₁.1.isRefinedBy r₂.1 mapping ∧ ControlFlowAction.optionIsRefinedBy r₁.2 r₂.2)
       (interpretOpList ops state) (interpretOpList ops state') := by
   induction ops generalizing state state' with
-  | nil => simpa [Interp]
+  | nil =>
+    simp only [interpretOpList_nil]
+    exact ⟨hState, trivial⟩
```
math-fehr: `grind` doesn't work here?

Veir/Interpreter/Refinement/Monotonicity.lean
```diff
@@ -200,7 +202,7 @@ theorem interpretOpList_mono
         /- A control-flow action: the list stops here for both the source and the target. -/
         have ⟨cf', hact', hcfRef⟩ : ∃ cf', act' = some cf' ∧ cf.isRefinedBy cf' := by grind
         subst hact'
-        simp [Interp, hsRef, ControlFlowAction.optionIsRefinedBy, hcfRef]
+        exact ⟨hsRef, hcfRef⟩
```
math-fehr: Same here for `grind`

Veir/Interpreter/EquationLemma.lean
```diff
@@ -287,15 +287,15 @@ theorem InterpreterState.DefinesDominating.interpretOpList_ne_none
       state.DefinesDominating (.before head) (by grind [List.mem_of_head? hhead])) :
     interpretOpList ops state ≠ none := by
   induction ops generalizing state with
-  | nil => simp
+  | nil => simp only [interpretOpList_nil]; exact Option.some_ne_none _
```
math-fehr: Can this be replaced by `simp [Option.some_ne_none]`?

Veir/Interpreter/Lemmas.lean
```diff
@@ -529,7 +529,8 @@ variable {state : InterpreterState ctx}
 @[simp, grind =]
 theorem interpretOpList_nil :
     interpretOpList [] state inBounds = some (.ok (state, none)) := by
-  simp [interpretOpList, pure]
+  simp only [interpretOpList, pure]
```
math-fehr: I'm curious, do you know what happened here? And in the following theorems?

Veir/Passes/CastsReconciliation/Reconciliation.lean
```diff
@@ -74,7 +74,7 @@ def reconcileRegIntCastLocal (ctx : WfIRContext OpCode) (op : OperationPtr) :
   let inputType := parentInput.getType! ctx.raw
   if resultType ≠ inputType then return (ctx, none)
   /- And the reconciliation involves the right types -/
-  if inputType ≠ RegisterType.mk then return (ctx, none)
+  if inputType.val ≠ (RegisterType.mk : TypeAttr).val then return (ctx, none)
```
math-fehr:
```suggestion
  if inputType.val ≠ (RegisterType.mk : Attribute) then return (ctx, none)
```
?

Veir/Verifier/Lemmas.lean
```diff
@@ -32,11 +32,13 @@ theorem OperationPtr.verifyIntegerBinop_eq_ok
...
-    Except.bind, throw, throwThe, MonadExceptOf.throw, pure, Except.pure] at h ⊢
-  simp only [TypeAttr.inj]
-  grind
+  simp only [verifyIntegerBinop, verifyPlainOpCounts, verifyOperandTypesMatch,
+    verifyResultTypeMatches, TypeAttr.verifyIntegerType, ne_eq, bind, Except.bind, throw, throwThe,
+    MonadExceptOf.throw, pure, Except.pure] at h
+  obtain ⟨integerType, _⟩ :
+      ∃ it, ((op.getOperand! ctx.raw 0).getType! ctx.raw).val = .integerType it := by grind
+  refine ⟨by grind, by grind, by grind, by grind, integerType, ?_, ?_, ?_⟩ <;>
+    exact TypeAttr.inj.mpr (by grind)
```
math-fehr:
```suggestion
  refine ⟨by grind, by grind, by grind, by grind, integerType, ?_, ?_, ?_⟩ <;>
    simp [TypeAttr.inj]
```
Does this work?

math-fehr: Same for the few lemmas below

Veir/Verifier.lean
```diff
@@ -284,8 +284,7 @@ theorem OperationPtr.Verified.arith_constant {op : OperationPtr} {opInBounds}
     ← getOpType!_eq_getOpType, opType, ne_eq,
     bind, Except.bind, throw, throwThe, MonadExceptOf.throw, pure, Except.pure, dite_not,
     ite_not] at opVerify
-  simp only [TypeAttr.inj]
-  grind
+  refine ⟨by grind, by grind, by grind, by grind, TypeAttr.inj.mpr (by grind)⟩
```
math-fehr: Do you have an idea why this fails? Why can't we `simp only [TypeAttr.inj]` anymore?

regehr (approved):
LGTM but it's more like "I can see why these would make sense for a Lean update" rather than I understand the specific reasoning

math-fehr (commented):
Thanks! I just have a few questions in the proofs, most importantly I feel there are two regressions compared to Lean 4.32, and I would like to understand why. 1) grind doesn't seem to case on some values, and 2) `TypeAttr` seems to be more annoying to deal with than before. What would be nice is figuring out a good solution for both.

luisacicolini:
I share @math-fehr's concerns wrt. the `grind` failures and would be happy to look into it if it's not too urgent (I have time today evening or tomorrow morning)

tobiasgrosser:
> I share @math-fehr's concerns wrt. the `grind` failures and would be happy to look into it if it's not too urgent (I have time today evening or tomorrow morning)

It can wait 1-2 days. Happy for you (@luisacicolini) to debug this. Thank you.

Veir/Data/LLVM/Int/Bitblast.lean
tobiasgrosser: This is a grind bug: https://github.com/leanprover/lean4/pull/14829. I added a comment, such that we can revert back to just `grind` from the next Lean release. Thank you @luisacicolini for investigating.

Veir/Data/LLVM/Int/Bitblast.lean
tobiasgrosser: I add a comment ` /- TODO: use `grind` after https://github.com/leanprover/lean4/pull/14829 is resolved. -/`.

Veir/Interpreter/Refinement/Monotonicity.lean
tobiasgrosser: It does. Fixed.

Veir/Interpreter/Refinement/Monotonicity.lean
tobiasgrosser: Yes. it does.

Veir/Interpreter/Lemmas.lean
tobiasgrosser: > Interp is a plain def, i.e. semireducible. simp works at reducible/instances transparency, and at that level the goal isn't even type-correct, so its eq_self closing step never fires. rfl checks defeq at default transparency, unfolds Interp, and closes it.

It seems this can actually be closed just by `rfl`.

Veir/Passes/CastsReconciliation/Reconciliation.lean
tobiasgrosser: This requires https://github.com/opencompl/veir/pull/1277. I pulled this out into a separate PR.

Veir/Interpreter/Refinement/Monotonicity.lean
tobiasgrosser: Resolved.

Veir/Interpreter/EquationLemma.lean
tobiasgrosser: Not needed anymore.

Veir/Verifier/Lemmas.lean
tobiasgrosser: Resolved via earlier PRs.

Veir/Verifier.lean
tobiasgrosser: Resolved via earlier PRs.

tobiasgrosser:
@math-fehr @luisacicolini, I spent a couple of hours pulling proper fixes out of this PR. The upgrade requires now only minor changes.

luisacicolini (approved):
lgtm!

math-fehr (approved):
That looks much nicer, thanks!

PR #1256, opened 2026-08-17
chore: remove old side effect interface

regehr:
finally `getEffects()` starts to pay off-- we can remove our old side effect infrastructure!

(there's more of this kind of code to remove but I'll work on it in other PRs)

an MLIR-style `ConditionallySpeculatable` is orthogonal to all of this and remains to be implemented, but nothing in Veir needs it yet, as far as I know


math-fehr (approved):
Nice!

PR #1257, opened 2026-08-17
chore: executable memory independence and CSE cleanup

regehr:
This PR reduces (but does not eliminate) dialect-specific code in our CSE pass. It does this by removing tables of pure opcodes and instead relying on the externally-provided lack of memory effects. It remains to connect memory independence with our operation semantics.


math-fehr (approved):
Very cool!

PR #1260, opened 2026-08-17
chore: Split `IsOpCode` to its own file

math-fehr:
This will allow `HasOpInfo` to import `IR/Basic.lean`, which is necessary whenever we want to include interfaces that mention `OperationPtr` for instance.


tobiasgrosser (approved):
Nice. I love these cleanups.

PR #1261, opened 2026-08-18
chore: Remove dependency between `Printer` and `OpCode`

math-fehr:
The printer can now take any `OpInfo` as long as it has `Builtin` in it.


regehr (approved):
nice!
LGTM

PR #1262, opened 2026-08-18
feat: explicit values for boolean pass arguments

regehr:
just a simple change: instead of a pass argument always implicitly setting a boolean to true, they can now be set to either true or false.

the immediate use case I have for this is canonicalization. I want to make its components turned on by default, but they can be disabled by setting them to false. thus, when canonicalization does constant folding (PR for this is ready to go except for this PR), we can avoid breaking a lot of test cases by turning off the folder.

### DIFF
--- a/Test/Passes/Canonicalize/options.mlir
+++ b/Test/Passes/Canonicalize/options.mlir
@@ -1,0 +1,32 @@
+// RUN: veir-opt %s -p=canonicalize | filecheck %s --check-prefix=DEFAULT
+// RUN: veir-opt %s -p='canonicalize{mod-arith-constant=false}' | filecheck %s --check-prefix=NO-MOD
+// RUN: veir-opt %s -p='canonicalize{commutative-constant-rhs=false}' | filecheck %s --check-prefix=NO-COMMUTE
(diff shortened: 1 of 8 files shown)


Test/Passes/Canonicalize/options.mlir
```diff
@@ -0,0 +1,32 @@
+// RUN: veir-opt %s -p=canonicalize | filecheck %s --check-prefix=DEFAULT
+// RUN: veir-opt %s -p='canonicalize{mod-arith-constant=false}' | filecheck %s --check-prefix=NO-MOD
+// RUN: veir-opt %s -p='canonicalize{commutative-constant-rhs=false}' | filecheck %s --check-prefix=NO-COMMUTE
+// RUN: veir-opt %s -p='canonicalize{mod-arith-constant=false commutative-constant-rhs=false}' | filecheck %s --check-prefix=NEITHER
```
math-fehr: To me, that seems weird that we do not separate the two options with a comma? Is that how it's done in MLIR?

tobiasgrosser (approved):
Nice!

PR #1264, opened 2026-08-18
feat: constant folding finally happens

regehr:
this isn't the last PR for constant folding, but it does finally close the loop and make folding happen for real. I've wired it into the canonicalization pass, where it breaks a bunch of test cases as expected, so I've just turned it off for those.


math-fehr (approved):
Very cool!

PR #1265, opened 2026-08-18
feat: implement MLIR's isolatedFromAbove trait

regehr:
we currently accept illegal MLIR that violates isolatedFromAbove, this fixes that


Veir/Verifier.lean
```diff
@@ -67,6 +67,55 @@ def OperationPtr.verifyTerminatorPosition (op : OperationPtr) (ctx : WfIRContext
   if operation.opType.isTerminator && operation.next.isSome then
     throw "Expected a terminator to be the last operation of its block"
 
+/-- Return the region containing a value's definition, if it is linked into one. -/
+private def ValuePtr.getParentRegion?
+    (value : ValuePtr) (ctx : WfIRContext OpCode) : Option RegionPtr :=
+  match value with
+  | .opResult result => result.op.getParentRegion! ctx.raw
+  | .blockArgument argument => (argument.block.get! ctx.raw).parent
```
math-fehr: Can you move this to `IR/Basic.lean` and rename it to `ValuePtr.getParentRegion!`?
I'm using `!` because it is using the `get!` functions.

Veir/Verifier.lean
```diff
@@ -67,6 +67,55 @@ def OperationPtr.verifyTerminatorPosition (op : OperationPtr) (ctx : WfIRContext
...
+/--
+Whether `ancestor` is `descendant` or one of its enclosing regions. This is the
+executable counterpart of MLIR's `Region::isAncestor` query.
+-/
+private partial def RegionPtr.isAncestorOf
+    (ancestor descendant : RegionPtr) (ctx : WfIRContext OpCode) : Bool :=
+  ancestor = descendant ||
+    match (descendant.get! ctx.raw).parent.bind (·.getParentRegion! ctx.raw) with
+    | none => false
+    | some parentRegion => ancestor.isAncestorOf parentRegion ctx
```
math-fehr: Can you move this to `IRNesting.lean`?

Veir/Verifier.lean
```diff
@@ -67,6 +67,55 @@ def OperationPtr.verifyTerminatorPosition (op : OperationPtr) (ctx : WfIRContext
...
+-/
+def OperationPtr.verifyOperandIsolation
+    (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) : Except String PUnit := do
+  let some useRegion := op.getParentRegion! ctx.raw | return
+  let some isolatedScope := useRegion.nearestIsolatedScope? ctx | return
+  for value in op.getOperands ctx.raw opIn do
+    let some defRegion := value.getParentRegion? ctx
+      | throw "operand is unlinked from any region"
+    if !isolatedScope.isAncestorOf defRegion ctx then
```
math-fehr, line 59: Can we have a fast path for this check? By that I mean first checking if `defRegion` is equal to `useRegion`, in which case this should always be correct!
That way, we don't have to always traverse up to the nearest isolatedFromAbove all the time, which should be faster.

math-fehr (commented):
Thanks a lot! I just had a few code comments, but otherwise very cool!

Veir/Verifier.lean
regehr: ah, nice.
I had thought about a region -> nearest isolated scope hashMap but I didn't like how the code was looking, and I think your idea is better anyhow

regehr: we can also bail out early if there are no operands...

regehr:
thanks for feedback-- plz take another quick look @math-fehr

math-fehr (approved):
Nice, thanks!

PR #1266, opened 2026-08-18
fix: Add missing import for `PatternRewriter/Semantics.lean`

math-fehr:
The file was never imported from anywhere, so it did not actually compile.


regehr (approved):
yikes!
LGTM.

tobiasgrosser (approved):

PR #1267, opened 2026-08-18
feat: Querying for branch ops in HasOpInfo

axelcool1234:
Now you can check if an op is a branch op :sunglasses:

### DIFF
--- a/Veir/Dialects/Cf/OpInfo.lean
+++ b/Veir/Dialects/Cf/OpInfo.lean
@@ -50,6 +50,10 @@ def Cf.isConstantLike (_op : Cf) : Bool :=
 def Cf.hasSSADominance (_op : Cf) (_index : Nat) : Bool :=
   true
 
+/-- Every `cf` operation is a branch. -/
+def Cf.isBranchOp (_op : Cf) : Bool :=
+  true
+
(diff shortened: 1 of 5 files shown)


regehr:
this is pretty close to the existing isTerminator interface, I wonder if instead of this we might want to be able to query for conditional and unconditional branches?

axelcool1234:
> this is pretty close to the existing isTerminator interface, I wonder if instead of this we might want to be able to query for conditional and unconditional branches?

That's a good idea - thoughts @math-fehr?

math-fehr:
Why do you need this interface? I think you should look at `BranchOpInterface`, this is what is used in MLIR

axelcool1234:
> Why do you need this interface? I think you should look at `BranchOpInterface`, this is what is used in MLIR

Hmm you're right. Like here:
https://github.com/llvm/llvm-project/blob/b88c1dafe890f17cfa473d3ad2d8d1350c2fc9c5/mlir/lib/Analysis/DataFlow/SparseAnalysis.cpp#L208-L210

I think this is actually querying for if its a terminator, and if it is, it grabs the successor operands. I could use `isTerminator` and then use the Branch Op interface. Does that make sense?

PR #1268, opened 2026-08-18
feat: Move `verifyLocalInvariants` to `HasOpInfo`

math-fehr:
This continues the work of making everything use a generic `OpCode` type, rather than the specialized one that contains all dialects.

This also exposes the reason why we split `IsOpCode` and `HasOpInfo`, as here the verifier requires `IsOpCode`, and the verifier needs to be put in an `HasOpInfo.


regehr (approved):
very cool!
this looks mechanical and good. obviously I did not read every line.

tobiasgrosser (approved):
Nice.

PR #1270, opened 2026-08-19
fix: require function names and types in `llvm.func` and `func.func`

math-fehr:
`llvm.func` and `func.func` properties were not requiring their function type or `sym_name` to be present.  This PR makes them mandatory in the propreties, and also fixes their type to be function types rather than arbitrary types.


regehr:
let's be a little careful here! I believe that LLVM supports nameless functions

regehr:
hmm, it looks like `mlir-translate` doesn't enjoy nameless LLVM functions, so probably adding this requirement is safe

regehr (approved):

tobiasgrosser (approved):
Nice. That cleans up quite some code.

PR #1271, opened 2026-08-19
feat: creates `FunctionOpInterface` in `HasOpInfo`

math-fehr:
`FunctionOpInterface` was only partially implemented in `HasOpInfo`, and only contained `isFunctionLike`. This PR bundles `getSymName` and `getFunctionType` into the same structure, that can be queried from an `HasOpInfo`.


tobiasgrosser (approved):
Nice.

regehr (approved):
LGTM!

PR #1272, opened 2026-08-19
chore: fail CI when there exist lean files disconnected from the build

regehr:
also there's an allowlist with the currently disconnected files

shall we try this out and see how it works??

### DIFF
--- a/Tools/check-orphan-modules
+++ b/Tools/check-orphan-modules
@@ -41,6 +41,31 @@ def path_of(module, source_dir=pathlib.Path(".")):
     return source_dir.joinpath(*module.split(".")).with_suffix(".lean")
 
 
+def without_comments(text):
+    """Remove Lean line comments and nested block comments."""
+    result = []
+    depth = 0
+    index = 0
+    while index < len(text):
+        if text.startswith("/-", index):
+            depth += 1
+            index += 2
+        elif depth and text.startswith("-/", index):
+            depth -= 1
+            index += 2
+        elif not depth and text.startswith("--", index):
+            index = text.find("\n", index)
+            if index == -1:
+                break
+        else:
+            if not depth or text[index] == "\n":
+                result.append(text[index])
+            index += 1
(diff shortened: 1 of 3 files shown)


tobiasgrosser (approved):
I think that's a neat idea. @math-fehr may have an opinion, though.

Tools/check-orphan-modules
```diff
@@ -0,0 +1,207 @@
...
+        )
+    if new or stale:
+        print(f"\n`{SELF} --update-allowlist` accepts the current state.")
+        return 1
+
+    print(
+        f"{len(all_modules)} Lean modules reachable from a build target, "
+        f"{len(allowed)} allowlisted orphan(s)"
+    )
+    return 0
```
math-fehr: I think we should probably remove this logic, and just return `1` and printing in `stderr` the list of orphans?
I'm not sure we would ever want an allowlist, we should first fix the current broken files, and then never have this issue?

Tools/check-orphan-modules
```diff
@@ -0,0 +1,207 @@
...
+    for a real one. Block comments have to be skipped rather than treated as
+    the end of the header, because a couple of files put their module docstring
+    between `module` and their first import.
+    """
+    result = []
+    in_comment = False
+    for line in path.read_text(encoding="utf-8").splitlines():
+        line = line.strip()
+        if in_comment:
+            in_comment = "-/" not in line
```
math-fehr: Maybe add a comment here stating that this will not work with lines like `/- -/ /-`, but 1) that is not really realistic, and 2) this will only return false negatives in the worse case.

math-fehr (commented):
Nice!
I just think we should remove some complexity of this code by just remove the allowlist, but that means fixing it first. If you want, I can just set the imports in another PR so you can depend on it, otherwise if you want to do it feel free!

math-fehr (approved):
Nice, thanks!

PR #1273, opened 2026-08-19
feat: fold operations that return multiple values

regehr:
not incredibly pleased with this code but I'm not sure what would be better.
we definitely do not want to refuse to fold ops that return multiple values.


math-fehr (approved):
Looks great, thanks!

PR #1275, opened 2026-08-19
chore: Remove generic functions about region kind out of `Verifier.lean`

math-fehr:
One of these function was moved to `OpInfo.lean`, since it can be useful somewhere else. The other one was removed since an equivalent function already existed in a different file.


regehr (approved):
nice!
LGTM

tobiasgrosser (approved):

PR #1277, opened 2026-08-20
chore: build TypeAttr coercions via Attribute.asType

tobiasgrosser:
This makes us less reliant on semi-reducible definitions, which `simp` and `grind` become more strict about in Lean 4.33.

The `Coe` and `CoeDep` instances targeting `TypeAttr` built their results with an anonymous constructor, so the elaborator inlined the coe field into a bare `Subtype.mk` and left goals such as `@Ne TypeAttr inputType (Subtype.mk ..)` that only typecheck once the semireducible `TypeAttr` def is unfolded. Lean 4.33 flips the default of `backward.isDefEq.respectTransparency.types` from false to true, so `isDefEq` no longer bumps to `.default` transparency when comparing types, and instance synthesis — which runs at reducible transparency — stops finding `Decidable` for comparisons like `inputType ≠ RegisterType.mk`. Routing every coercion through `Attribute.asType`, whose result type is literally `TypeAttr`, keeps the coerced term type-correct at reducible transparency, so `reconcileRegIntCastLocal` can keep comparing `TypeAttr` values directly after the upgrade.

This prepares the upgrade to Lean 4.33 https://github.com/opencompl/veir/pull/1255.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -1383,58 +1383,58 @@ def Attribute.asType (attr : Attribute) (isType : attr.isType := by grind) : Typ
 -/
 
 instance : Coe IntegerType TypeAttr where
-  coe type := ⟨.integerType type, by rfl⟩
+  coe type := Attribute.asType (.integerType type) (by rfl)
 
 instance : Coe FloatType TypeAttr where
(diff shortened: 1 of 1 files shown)


math-fehr (approved):
Nice, though again I would avoid AI-generated PR descriptions like this one

PR #1278, opened 2026-08-20
chore: make `Interp` an inductive with `.ok`/`.ub`/`.fail`

tobiasgrosser:
Lean 4.33 has become stricter about the use of semireducible definitions, which makes `simp` and `grind` more fragile. This PR reduces our reliance on semireducible definitions by lifting `UBOr` into `Interp`.

`Interp α` was a semireducible `def` for `Option (UBOr α)`, so statements mixing an `Interp`-typed position with an `Option` constructor (`interpretOpList ops state ≠ none`, `… = some (.ok r)`) were only type-correct once `Interp` unfolded, which `simp`, `grind` and instance synthesis will not do at their reducible transparency. This replaces the `def` with a three-constructor inductive — `.ok` / `.ub` / `.fail`, with `UBOr` folded into it — so those goals are genuinely typed at `Interp` and the constructor lemmas apply directly.

This prepares for the Lean 4.33 upgrade in #1255.

### DIFF
--- a/UnitTest/Evaluate.lean
+++ b/UnitTest/Evaluate.lean
@@ -10,8 +10,8 @@ private def i32 : TypeAttr := IntegerType.mk 32
 
 private def testEvaluateAddi : String := Id.run do
   let operands : Array RuntimeValue := #[.int 32 (.val 7), .int 32 (.val 8)]
-  let some (.ok results) :=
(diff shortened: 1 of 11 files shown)


math-fehr (approved):
Thanks!

PR #1279, opened 2026-08-20
feat: build the `IsVerified*` type specs with `Attribute.asType`

tobiasgrosser:
Lean 4.33 does not reduce subtypes anymore in `simp` and `grind`.  This PR avoids breakage in the Lean 4.33 update in #1255.


math-fehr (approved):
This looks good to me, but can you please change the PR descriptions? AI-generated PR descriptions are both too verbose but also not very good at explaining the problem clearly (for instance `This builds the fourteen specs with Attribute.asType,` is quite cryptic). I think just mentionning that `Subtype` is not going to be reduced anymore in the next Lean version should be good enough.

Also, claude code just added a "concise" setting, maybe that would be best for these kind of things?

PR #1282, opened 2026-08-20
fix(dataflow): Dominance facts now attach directly to BlockPtrs instead of InsertPoints

axelcool1234:
Also added a test to show it fixing a bug (when the first operation of a block is erased, it'd invalidate dominance facts. This is now no longer the case).

### DIFF
--- a/UnitTest/DataFlowFramework/Dominance.lean
+++ b/UnitTest/DataFlowFramework/Dominance.lean
@@ -2,6 +2,7 @@ import UnitTest.DataFlowFramework.Helpers
 
 import Veir.Analysis.DataFlow.DominanceAnalysis
 import Veir.IR.Dominance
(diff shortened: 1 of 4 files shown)


regehr (approved):
this LGTM, thanks!

math-fehr (approved):
Nice! Just for next time, can you split your PRs so that it is obvious what is changing? Here you are having both the BlockPtr change, and also a function signature change (removing IRContext argument).

axelcool1234:
> Nice! Just for next time, can you split your PRs so that it is obvious what is changing? Here you are having both the BlockPtr change, and also a function signature change (removing IRContext argument).

These are directly related in this case as switching to `BlockPtr` mitigates the need for `IRContext` since only `InsertPoint` required it

PR #1283, opened 2026-08-20
chore: more closely approximate MLIR's terminator verification rules

regehr:


Veir/Verifier.lean
```diff
@@ -60,11 +68,39 @@ def OperationPtr.verifyOperandIsolation
       throw "operand uses a value defined outside the isolated region that encloses its use"
 
 /--
-  Check that a block is non-empty and its last operation is a
-  terminator.
+  Whether this operation might be a terminator, mirroring MLIR's
+  `mightHaveTrait<OpTrait::IsTerminator>`.
+-/
+def OpCode.mightBeTerminator (opCode : OpCode) : Bool :=
```
math-fehr: Do you think we can remove both special cases here?

My issue is that I have soon to start reasonning about terminators in programs, and here I really don't want to reason about `.builtin .unregistered` and `.test .test`. I think it's fine at this point to still support `.builtin .unregistered` in the parser, but here this change would require me to reason about it when reasonning about terminators, which will become complex for not that much value anymore :/

math-fehr (approved):
Very cool!
I'm just wondering if we can remove the `.builtin .unregistered` logic here, as it is really annoying to reason about it in the proof perspective. Otherwise, thanks a lot for this!

regehr:
@math-fehr changes done!

PR #1285, opened 2026-08-21
chore: Remove `dominates` and `properlyDominates` axioms

math-fehr:
These were already implemented in a different file, so we use these definitions instead. The actual definitions were named `Dominates` and `ProperlyDominates`, so we also do a renaming.


regehr (approved):
LGTM -- thank you!

PR #1286, opened 2026-08-21
feat: verify that the entry block has no predecessors

regehr:


Veir/Verifier.lean
```diff
@@ -29,6 +29,20 @@ def OperationPtr.verifyTerminatorPosition (op : OperationPtr) (ctx : WfIRContext
...
+  entry block's arguments and dominance treat it as unconditionally
+  reaching every other block in the region. A branch back to it would
+  contradict that.
+-/
+def OperationPtr.verifyDoesNotBranchToEntryBlock (op : OperationPtr) (ctx : WfIRContext OpCode)
+    (opIn : op.InBounds ctx.raw) : Except String PUnit := do
+  for succ in op.getSuccessors ctx.raw opIn do
+    let some region := (succ.get! ctx.raw).parent | pure ()
+    if (region.get! ctx.raw).firstBlock = some succ then
+      throw "entry block of region may not have predecessors"
```
math-fehr: Is this how MLIR implements it?
There is actually a faster way to implement this.
For all regions, if you look at the first block, you can check if `(block.get! ctx).firstUse = none`, and that should be exactly what you want here!

regehr:
thanks-- I ended up with this, hopefully ok @math-fehr ?
```
def BlockPtr.verifyNoEntryBlockPredecessors (block : BlockPtr) (ctx : WfIRContext OpCode)
    (blockIn : block.InBounds ctx.raw) : Except String PUnit := do
  let b := block.get ctx.raw blockIn
  if (b.parent.get!.get! ctx.raw).firstBlock ≠ some block then return
  if b.firstUse.isSome then
    throw "entry block of region may not have predecessors"
```

math-fehr:
Yes that makes sense! Can you just add a comment in the first `if` that states that if the block has no parent, then the default is `.firstBlock = none`, and thus the algorithm is correct?
Because here the first `get!` will create a default in that case.

math-fehr (approved):
otherwise, feel free to merge after adding the comment!

regehr:
@math-fehr instead of the comment I refactored the code a tiny bit to make the thing you said apparent
```
def BlockPtr.verifyNoEntryBlockPredecessors (block : BlockPtr) (ctx : WfIRContext OpCode)
    (blockIn : block.InBounds ctx.raw) : Except String PUnit := do
  let b := block.get ctx.raw blockIn
  let some parent := b.parent | return
  if (parent.get! ctx.raw).firstBlock ≠ some block then return
  if b.firstUse.isSome then
    throw "entry block of region may not have predecessors"
```

PR #1288, opened 2026-08-21
chore(PBV): replace omega with lia

tobiasgrosser:
`lia` is the modern cutast based version of omega. We want to use `lia` consistently in the PBV development. This change moves the last old use of `omega` to `lia`, such that `PBV` is `lia`-only.


luigirinaldi (approved):
LGTM!

PR #1289, opened 2026-08-21
chore: create a separate `LLVMFunctionType`

math-fehr:
This creates `LLVMFunctionType`, a wrapper around `FunctionType`, and makes `Attribute.llvmFunctionType` use `LLVMFunctionType`. I also added a coercion between the two, to avoid disrupting other parts of the code.

I am adding in a follow-up PR a coercion between concrete attributes to `Attribute`. This coercion makes only sense if an attribute can only be projected to a single constructor. Here, both `llvmFunctionType` and `functionType` had the same concrete attribute `FunctionType`.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -373,6 +373,16 @@ structure FunctionType where
 deriving Inhabited, Repr, Hashable
 
 /--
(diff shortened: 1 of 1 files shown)


regehr (approved):
LGTM

tobiasgrosser (approved):

PR #1290, opened 2026-08-21
feat: add `IsAttr` class

math-fehr:
This allows to write functions that take a generic attribute type as input, instead of taking an `Attribute`.


regehr (approved):
LGTM!

tobiasgrosser (approved):

PR #1291, opened 2026-08-21
feat: add `IsTypeAttr` class

math-fehr:
This allows to write functions that take a generic attribute type as input, instead of taking a `TypeAttr`.


regehr (approved):
makes sense!
I am just learning about coercions and typeclasses in Lean

tobiasgrosser (approved):
Nice!

PR #1292, opened 2026-08-21
feat(puddle): introduce `Puddle` pattern DSL 

math-fehr:
This introduces the `Puddle` DSL, which is meant to mimic MLIR's PDL dialect as an embedded DSL in Lean. Its objective is to allow simple reasoning over (local) rewrite patterns. By using an embedded DSL, we can write decision procedures to automatically discharge multiple fields from `LocalRewritePattern.Valid`, as well as reasoning more easily about `LocalRewritePattern.PreservesSemantics`. In particular, we can reason about semantics without having to reason about the equation lemma, or dominance, which makes things much easier.

This PR introduces the matcher DSL, as well as a do-notation-based builder, to make the creation of DAG matcher more user-friendly. The matcher DSL is a list of declarative matching instructions, with variables called `Handle`. Each handle is a reference to some part of the IR. Each instruction requires some handle to exist, constrain it, and then potentially bind a new handle. The execution of a matcher will be implemented in a following PR. 

`UnitTest/Puddle.lean` contains an example on how to build a matcher.

### DIFF
--- a/Veir/PatternRewriter/Puddle/Builders.lean
+++ b/Veir/PatternRewriter/Puddle/Builders.lean
@@ -1,0 +1,152 @@
+module
+
+public import Veir.PatternRewriter.Puddle.Definitions
+
+/-!
+# Puddle Builders
+
+This file contains author-facing builders for constructing declarative Puddle rewrite patterns
+without having to manually manage handles for matched IR elements. The builders are based on a state
+monad that automatically allocates handles, allowing the use of do-notations for authoring
+match programs. In particular, the programs are built in the reverse order of the final match
+program, so the root operation is declared last, and the leaves are declared first.
+-/
+
+namespace Veir.Puddle
+
+public section
+
+variable {OpInfo : Type} [HasOpInfo OpInfo]
+
+/-!
+## Matcher builder
+
+`MatchProg.Builder` provides a monadic interface for authoring the matching phase of a Puddle
+pattern. Each builder operation adds one constraint and returns fresh handles for later constraints
+or for export to the following phases. Although the source reads from leaves to root, declarations
+are prepended so the finished program can be interpreted from root to leaves.
+-/
+
+/-- Internal accumulator used by `MatchProg.Builder`. -/
+structure MatchProg.BuilderState where
+  /-- The next free identifier for a handle. -/
+  nextId : Nat := 0
+  /-- Declarations in interpreter order; after a well-formed build, the root is first. -/
+  decls : List (MatchDecl OpCode) := []
+
+/--
+A stateful builder for a match program that exports a value of type `α`.
+The exported value should consist of handles needed by the creation or replacement phase.
+-/
+structure MatchProg.Builder (α : Type) where
+  /--
+  Run the builder from an explicit internal state. Should not be used directly, use
+  `MatchProg.build` instead.
+  -/
+  run : MatchProg.BuilderState → α × MatchProg.BuilderState
+
+/--
+The handles allocated for a matched operation. It contains the operation handle, an array of
+handles for its SSA results, and a handle for its concrete properties.
+-/
+structure OpHandle (opCode : OpCode) where
+  /-- The operation handle. -/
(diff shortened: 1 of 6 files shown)


Veir/PatternRewriter/Puddle/Definitions.lean
```diff
@@ -0,0 +1,144 @@
...
+  top-down from that root. Every matched operation must be reachable from the root and fully
+  constrained: its operands, result types, and properties must be specified.
+2. A `CreateProg` derives metadata and creates the operations that will replace the matched root.
+  Operations are created in order and inserted before the root. They may consume non-root values
+  matched in the first phase or values produced by earlier creation declarations.
+3. A `Replacement` selects, in order, the values that replace the root's results before the root is
+  erased.
+
+Puddle patterns are not meant to be constructed directly. Instead, they are built with
+`Pattern.Builder`, defined in `Veir.PatternRewriter.Puddle.Builders`.
```
math-fehr, line 31: This describes the full implementation that will come in later PRs, I just wanted to avoid the work of updating the documentation bit by bit.

regehr (approved):
I can't yet see how this will all play out-- but this looks great and I trust you @math-fehr!

Veir/PatternRewriter/Puddle/Builders.lean
```diff
@@ -0,0 +1,152 @@
+module
+
+public import Veir.PatternRewriter.Puddle.Definitions
+
+/-!
+# Puddle Builders
+
+This file contains author-facing builders for constructing declarative Puddle rewrite patterns
+without having to manually manage handles for matched IR elements. The builders are based on a state
+monad that automatically allocates handles, allowing the use of a do-notation style for authoring
```
tobiasgrosser:
```suggestion
monad that automatically allocates handles, allowing the use of do-notation for authoring
```

This somehow flows better. Is this still correct?

Veir/PatternRewriter/Puddle/Builders.lean
```diff
@@ -0,0 +1,152 @@
...
+  ⟨fun state =>
+    let result := Handle.mk (OpInfo := OpCode) state.nextId
+    let matcher := fun (attr : TypeAttr) => ((attr.cast? Attr).map matcher).getD false
+    (result, {
+      nextId := state.nextId + 1
+      decls := .type matcher result :: state.decls
+    })⟩
+
+/--
+Match a value with a given type. This should be used to match values where their definitions
```
tobiasgrosser:
```suggestion
Match a value with a given type. This should be used to match values where their kind
```

`definitions` does not sound right.

Veir/PatternRewriter/Puddle/Definitions.lean
```diff
@@ -0,0 +1,144 @@
...
+
+/-- A predicate on an IR type. It accepts a type when it returns `true`. -/
+abbrev TypeMatcher := TypeAttr → Bool
+
+/--
+A declarative matcher instruction for a Puddle pattern graph.
+
+This is an internal representation; use `MatchProg.Builder` rather than constructing declarations
+directly.
+-/
```
tobiasgrosser, line 100: If this is internal, should this be private? You currently make everything public but placing everything in a `public` section.

tobiasgrosser (approved):
Wow, I love it. Now, I am not sure it would call it a DSL, as its just an API for creating matchers. When you talked about DSL earlier, I expected that you would use Lean syntax extensions to create something that ends us in meta-land. This is meta-code free. How amazing. I love it! 🥳

Veir/PatternRewriter/Puddle/Definitions.lean
math-fehr: I removed the sentence, it's internal in the sense that users should not manually build/introspect it, but the type needs to be exposed

math-fehr:
Well, I agree that the definition of DSL is more loose here. I would still call it an embedded DSL because it uses the Lean syntax, and it needs to be written in a certain way to be "effective".

PR #1293, opened 2026-08-22
feat(verifier): dynamically check that operands are dominated by their defs

regehr:
this required passing the top-level operation into the verifier, no big deal

no attempt is made to link executable dominance with our dominance properties, but we'll need to do that!


Test/Parsing/forward-reference-nested.mlir
```diff
@@ -7,7 +8,9 @@
 // the nested region; the later definition in the function body is an independent value.
 //
 // Whether this is legal under dominance / IsolatedFromAbove is a verifier concern that MLIR
-// checks after parsing. Implementation in VeIR TBD. For details see `ForwardValue` in 
+// checks after parsing, so the parse is checked with the verifier disabled. It is not legal:
+// the outer use reads a value defined in a child region, which does not dominate it, and the
+// verifier rejects it below exactly as MLIR does. For details see `ForwardValue` in
```
math-fehr: Can you improve a bit the first sentences in this comment? The first sentence seems to say that we don't know if it's legal or not, but the second one says it is not.

Veir/Verifier.lean
```diff
@@ -166,12 +167,41 @@ private def WfIRContext.verifyPDLPatternBodies (ctx : WfIRContext OpCode) :
...
+-/
+private def WfIRContext.verifyDominance
+    (ctx : WfIRContext OpCode) (root : OperationPtr) : Except String Unit := do
+  let some dfCtx := Veir.fixpointSolve root #[DominanceAnalysis] ctx
+    | throw "dominance analysis did not reach a fixpoint"
+  ctx.raw.forOpsDepM fun op opIn => do
+    let some block := (op.get ctx.raw opIn).parent | return
+    let some region := (block.get! ctx.raw).parent | return
+    if !region.hasSSADominance ctx then return
+    let some metadata := region.getRegionMetadataFact? dfCtx ctx | return
```
math-fehr: Is there a better API to query this information now? I feel that this function should never be called by a user.

Veir/Verifier.lean
```diff
@@ -166,12 +167,41 @@ private def WfIRContext.verifyPDLPatternBodies (ctx : WfIRContext OpCode) :
...
+    if !metadata.postOrderIndex.contains block then return
+    for (value, index) in (op.getOperands ctx.raw opIn).zipIdx do
+      let dominated :=
+        match value with
```
math-fehr: Is there no `properlyDominates` function on `ValuePtr` directly? If not we should add one I think.

Veir/Verifier.lean
```diff
@@ -192,22 +222,24 @@ def WfIRContext.verify (ctx : WfIRContext OpCode) : Except String Unit := do
...
   OperationPtr.verifyLocalInvariants
   Llvm.verifyLocalInvariants Arith.verifyLocalInvariants Mod_Arith.verifyLocalInvariants
 
 /--
-Assert that the IR context satisfies its structural and local invariants.
+Assert that the IR context verifies successfully for some root operation.
 -/
 def WfIRContext.Verified (ctx : WfIRContext OpCode) : Prop :=
-  ctx.verify = .ok ()
+  ∃ root, ctx.verify root = .ok ()
```
math-fehr: Can you pass the `root` as argument here?

math-fehr (commented):
Nice! I have a few small comments, but otherwise all good!

Test/Parsing/forward-reference-nested.mlir
regehr: sorry, bad update to old comment

regehr:
@math-fehr thanks for feedback! comments addressed. please take another look when you have a minute.

Veir/IR/Dominance.lean
```diff
@@ -207,4 +207,22 @@ partial def opsInDominanceOrder
 
 end OperationPtr
 
+namespace ValuePtr
+
+/--
+Does the definition of `value` dominate the use of it by `op`?
+-/
+def dominatesUse
```
math-fehr: Can you rename this to `properlyDominatesUse`?
Otherwise, an operation might have an operand which is one of its result, because an operation dominates itself but not properly.

math-fehr (approved):
Very cool! I just had one comment about a name, but feel free to merge this PR once you addressed it!

PR #1294, opened 2026-08-23
chore: perf fix

regehr:
ok I am *NOT* good at reasoning about performance of Lean code. this change avoids a copy and turns quadratic time replacement into linear time.


math-fehr (approved):
Ohh, sorry I missed this one when reviewing.
If you think in terms of ownership like in Rust, what happened here is that `rewriter` was "dropped" and given ownership to `newRewriter`. So here, because you returned `rewriter`, which was dropped, Lean made sure it was still alive by copying it.
moving your line made this work because now `rewriter` is just `newRewriter` at that point.

PR #1295, opened 2026-08-23
feat(canonicalizer): run to global fixpoint

regehr:
my view is that if we want to canonicalize, then we should canonicalize for real, and not mess around with not reaching a fixpoint


math-fehr:
@regehr, do you have an example where the code didn't reach a fixpoint? AFAIK the GreedyPatternRewriter should already reaches a fixpoint (unless I forgot to implement it)?
In any case, if it doesn't reach a fixpoint, then we should fix the GreedyPatternRewriter instead

regehr:
> @regehr, do you have an example where the code didn't reach a fixpoint? AFAIK the GreedyPatternRewriter should already reaches a fixpoint (unless I forgot to implement it)? In any case, if it doesn't reach a fixpoint, then we should fix the GreedyPatternRewriter instead

discussion of this moved to zulip

PR #1296, opened 2026-08-23
feat: createOrFold 

regehr:
MLIR has -- and heavily uses -- a createOrFold function. this one is better than MLIR's. theirs creates the operation and then tries to fold it. ours doesn't create it unless it won't fold.

this PR also uses createOrFold from ArithToLLVM. we probably want to use it broadly since this will tend to reduce the number is circumstances where we need to explicitly invoke folding.

you might think that fully folded Arith would not need createOrFold, but this isn't the case. the new test `Test/Passes/ArithToLLVM/fold_after_canonicalize.mlir` shows why the new way of doing things is better


Veir/Interfaces/FoldInterfaces.lean
```diff
@@ -131,4 +131,38 @@ def foldOperation (rewriter : PatternRewriter OpCode) (op : OperationPtr)
...
 
+/--
+Create an operation, but only if it can't fold first.
+-/
+def PatternRewriter.createOrFold! (rewriter : PatternRewriter OpCode) (opType : OpCode)
+    (resultTypes : Array TypeAttr) (operands : Array ValuePtr)
+    (blockOperands : Array BlockPtr) (regions : Array RegionPtr)
+    (properties : propertiesOf opType) (insertionPoint : InsertPoint) :
+    Option (PatternRewriter OpCode × Array ValuePtr) :=
+  foldedResults <|> created
```
math-fehr: There is an issue here is `materializeConstant` triggers an actual error.
The issue is that we should be returning `none` from that entire function, but here you will return the result of `created`. Also, this will trigger an entire copy of `IRContext`.

regehr:
@math-fehr this needed refactoring anyway! I think it's better now, please take a quick look when you have time

Veir/Interfaces/FoldInterfaces.lean
```diff
@@ -110,25 +123,31 @@ dialect declines to represent, leave the IR alone.
...
+    (blockOperands : Array BlockPtr) (regions : Array RegionPtr)
+    (properties : propertiesOf opType) (insertionPoint : InsertPoint) :
+    Option (PatternRewriter OpCode × Array ValuePtr) := do
+  let (rewriter, results) ←
+    rewriter.tryFold! opType properties resultTypes operands insertionPoint
+  let some results := results | do
+    let (rewriter, op) ← rewriter.createOp! opType resultTypes operands blockOperands regions
+      properties (some insertionPoint)
+    return (rewriter, op.getResults! rewriter.ctx.raw)
+  return (rewriter, results)
```
math-fehr: Can you use an if let or a match instead of a `|` syntax here?
Having a do notation after a `|` is a bit confusing

math-fehr (approved):
Thanks! Just a style comment, but feel free to merge after a fix!

PR #1297, opened 2026-08-23
tests: mixed width constants

regehr:
a bunch of these are XFAIL, since the issues remain unfixed

all tests are for the LLVM dialect, we should fix that first and then look at the rest of them


math-fehr (approved):
Nice! ~~Can you also add one/multiple issues on github about it, so we track them?~~
(I just realized you already did ;) )

PR #1299, opened 2026-08-24
feat(puddle): Add creation and replacement to patterns

math-fehr:
This PR adds the second part of the Puddle embedded DSL.

It contains the definitions for the rewrite part of the pattern. It is split in two parts:
* Directly after the matching phase, there is the creation phase, which creates a sequence of new operations.
* Then, after the creation phase, a single instruction specifies which values should replace the original root operaiton matched in the matching phase.

Similar to the match phase, we define a monadic interface to define the creation phase. We also define a monadic interface for the entire pattern, which composes the 3 different phases.


regehr (approved):
this looks cool! 
I asked claude to take a look and it had some concerns but I won't repeat them since I assume you're doing the same thing :)

PR #1302, opened 2026-08-25
feat(puddle): Add execution of puddle patterns

math-fehr:
This PR adds `Puddle.Pattern.compile`, which defines an interpreter for a `Puddle.Pattern`, to convert it into a `LocalRewritePattern`. With Lean's compiler inlining, we have a partial evaluation for a given pattern, which is exactly a futamura projection.

The interpretion is quite simple, we traverse the puddle declarations one by one while carrying an assignment structure (which should be partially evaluated).

I haven't measured the performance of this yet (this is not the end goal of this DSL anyway), but I think I removed most of the interpreter code from the compiled code with inlining.


tobiasgrosser (approved):
Very cool.

PR #1304, opened 2026-08-26
feat(RISCV): use Puddle for isel patterns

luisacicolini:
Port lowering pattern for unary riscv operations to puddle. If we are happy with this, I will port the remaining rewrites in the next PR.

Co-authored by Claude.


math-fehr (approved):
Nice, thanks!

PR #1305, opened 2026-08-26
feat(veir-interpret,veir2mir): read the input program from stdin

GZGavinZhao:
`veir-opt` has read from stdin since #605 (when the CLI receives no positional argument or a `-` filename), but `veir-interpret` and `veir2mir` have always required a file name. This forces pipelines like `veir-opt ... > %t && veir-interpret %t` to materialize a temporary file.

This PR ports the stdin handling to `veir-interpret` and `veir2mir`, and lift the duplicated read-and-parse logic into a shared `Veir.Input` module:
- `getFileContent`: read a file, or stdin when the filename is `none`
- `parseContent`: parse program bytes with caret-style diagnostics
- `parseOperation`: read and parse, naming the source `<stdin>`
- `inputSourceOfArgs`: map positional args (`[]` -> stdin)

Also update the lit tests that used the `veir-opt > %t && veir-interpret %t` pattern to the more elegant `veir-opt | veir-interpret`. Some tests use `veir-opt | tee %t | veir-interpret` because the intermediate MLIR output also needs to be checked.

This also removes `-` from the list of valid file names.

### DIFF
--- a/Veir/Input.lean
+++ b/Veir/Input.lean
@@ -1,0 +1,62 @@
+import Veir.Parser.MlirParser
+import Veir.Parser.ParserError
+import Veir.GlobalOpInfo
+
+/-!
+  Helpers shared by the VeIR command-line tools (`veir-opt`, `veir-interpret`,
+  `veir2mir`) for reading an input program from a file or from standard input
+  and parsing it into a well-formed IR context.
+
+  When no file argument is given, the program is read from standard input.
+ -/
+
+open Veir.Parser
+open Veir
+
+namespace Veir.Input
+
+/-- Read the input program from `filename`, or from standard input when it is
+    `none`. -/
+def getFileContent (filename : Option String) : ExceptT String IO ByteArray := do
+  if let some f := filename then
+    try
+      return ← IO.FS.readBinFile f
+    catch e =>
+      throw s!"Error reading file '{f}': {e}"
+  return ← IO.FS.Stream.readBinToEnd (←IO.getStdin)
+
+/-- Parse program bytes into a well-formed IR context and its top-level
+    operation, formatting parse errors caret-style against `sourceName`. -/
+def parseContent (content : ByteArray) (sourceName : String)
+    (allowUnregisteredDialect : Bool) :
+    ExceptT String IO (WfIRContext OpCode × OperationPtr) := do
+  let some (ctx, _) := WfIRContext.create OpCode
+    | throw "Failed to create IR context"
+  match ParserState.fromInput content with
+  | .ok parser =>
+    let state := MlirParserState.fromContext ctx allowUnregisteredDialect
+    match parseTopLevelOp.run state parser with
+    | .ok (op, state, _) =>
+      return (state.ctx, op)
+    | .error err =>
+      throw (err.format sourceName content)
+  | .error err =>
+    throw (err.format sourceName content)
+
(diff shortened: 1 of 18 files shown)


Test/Passes/ArithToLLVM/addui_extended.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: veir-opt %s -p=arith-to-llvm > %t && veir-interpret %t | filecheck %s --check-prefix=EXEC
+// RUN: veir-opt %s -p=arith-to-llvm | tee %t | veir-interpret | filecheck %s --check-prefix=EXEC
 // RUN: filecheck %s --check-prefix=LOWERED --input-file=%t
```
math-fehr, line 2: I don't really understand the point in changing these ones? This is just adding noise I feel, because otherwise the file is still created anyway?

Test/Passes/InstructionSelection/RISCV64/add_ext_exec.mlir
```diff
@@ -1,6 +1,5 @@
 // RUN: veir-interpret %s | filecheck %s --check-prefix=SRC
-// RUN: veir-opt %s -p='canonicalize{fold=false},instcombine,canonicalize{fold=false},cse,dce,isel-br-riscv64,isel-sdag-riscv64,isel-riscv64,canonicalize{fold=false},riscv-combine,coerce-function-boundaries-to-riscv-reg,reconcile-cast,dce' > %t
```
math-fehr, line 2: Same for these ones to be honest

Veir/Input.lean
```diff
@@ -0,0 +1,64 @@
...
+/-- Map positional CLI arguments to an input source: `[]` and `["-"]` mean
+    standard input; a single argument names the input file. -/
+def inputSourceOfArgs (positional : List String) : Except String (Option String) :=
+  match positional with
+  | [] => .ok none
+  | ["-"] => .ok none
+  | [filename] => .ok (some filename)
+  | _ => .error "Expected at most one positional argument for the input filename."
+
+end Veir.Input
```
math-fehr: Missing newline at the end

math-fehr (commented):
Nice! I think we should actually rollback some of the changes in `.mlir` files (the ones where we don't actually remove temporary files, since it's just adding noise).
Otherwise, should we really still support `-` files as stdin? Because right now, `parseArgs` will consider `-` to be a flag, not a positional argument, so it will anyway fail.
I'm in the opinion that we do not care about `-`, and this is just an artifact of LLVM/MLIR code that no one really cares about, so I would just remove it (since this PR anyway breaks this)

Test/Passes/ArithToLLVM/addui_extended.mlir
GZGavinZhao: Reverted, I got too carried away with my fondness of pipes.

GZGavinZhao:
> I'm in the opinion that we do not care about `-`, and this is just an artifact of LLVM/MLIR code that no one really cares about, so I would just remove it (since this PR anyway breaks this)

Agreed, I forgot to test the `-` case :sweat_smile: I've dropped support for `-`. In the future, when/if we move our hand-rolled CLI to a relatively battle-tested CLI library like `lean4-cli`, we can consider adding support back just for consistency since `lean4-cli` already provides proper `-` parsing.

math-fehr (approved):
All good now, thanks!

PR #1308, opened 2026-08-27
feat(GMIR): Add initial gMIR dialect

naveen-seth:
This introduces a basic gMIR (Generic Machine IR) dialect with a minimal set of
operations and a Puddle pass for lowering LLVM IR to gMIR.

This is required for to implement legalization similar to how it's done in upstream's GlobalISel framework (see: https://github.com/opencompl/paper-legalization/pull/1).

`GMIR.genericOpInfo` will be used by #1234 to describe groups of operands/results targeted by a single legalization rule.

### DIFF
--- a/Veir/Dialects/GMIR/OpInfo.lean
+++ b/Veir/Dialects/GMIR/OpInfo.lean
@@ -1,0 +1,125 @@
+module
+
+public import Veir.IR.OpInfo
+public import Veir.Verifier.Basic
+public import Veir.Dialects.LLVM.Properties
+meta import Veir.Meta.OpCode
+
+namespace Veir
+
+public section
+
+@[opcodes]
+inductive GMIR where
+| g_add
+| g_sub
+| g_icmp
+deriving Inhabited, Repr, Hashable, DecidableEq
+
+@[expose, properties_of]
+def GMIR.propertiesOf : GMIR → Type
+  | .g_add | .g_sub => NswNuwProperties
+  | .g_icmp => IcmpProperties
+
+def GMIR.fromAttrDict
+    (op : GMIR) (attrDict : Std.HashMap ByteArray Attribute) :
+    Except String (GMIR.propertiesOf op) := by
+  cases op
+  case g_add | g_sub => exact NswNuwProperties.fromAttrDict attrDict
+  case g_icmp => exact IcmpProperties.fromAttrDict attrDict
+
+def GMIR.toAttrDict
+    (op : GMIR) (props : GMIR.propertiesOf op) :
+    Std.HashMap ByteArray Attribute :=
+  match op with
+  | .g_add | .g_sub => Id.run do
+    let mut dict := Std.HashMap.emptyWithCapacity 1
+    let val := (if props.nsw then 1 else 0) + (if props.nuw then 2 else 0)
+    if val > 0 then
+      dict := dict.insert "overflowFlags".toUTF8
+        (.integerAttr (IntegerAttr.mk (Int.ofNat val) (IntegerType.mk 32)))
+    dict
+  | .g_icmp =>
+    let value := IntegerAttr.mk (Int.ofNat props.predicate.toNat) (IntegerType.mk 64)
+    (Std.HashMap.emptyWithCapacity 1).insert
+      "predicate".toUTF8 (Attribute.integerAttr value)
+
+#generate_dialect GMIR
+
+instance : IsOpCode GMIR where
+  fromName := GMIR.fromName
+  name := GMIR.name
+  propertiesOf := GMIR.propertiesOf
+  fromAttrDict := GMIR.fromAttrDict
+  toAttrDict := GMIR.toAttrDict
+
+/--
+This replicates the type-group information in LLVM's `GenericOpcodes.td`
+(llvm/include/llvm/Target/GenericOpcodes.td).
+
+Immediate operands and operands with LLVM's `unknown` (special) type are omitted and represented
+as operation properties instead.
+-/
+inductive TypeGroup where
+| type (id : Nat)
+| ptype (id : Nat)
+deriving Inhabited, Repr, BEq, DecidableEq, Hashable
+
+structure GenericOpInfo where
+  outOperandList : Array TypeGroup
+  inOperandList : Array TypeGroup
+deriving Inhabited, Repr
+
+def GMIR.genericOpInfo : GMIR → GenericOpInfo
+  | .g_add | .g_sub =>
+    { outOperandList := #[.type 0]
+      inOperandList := #[.type 0, .type 0] }
+  | .g_icmp =>
+    { outOperandList := #[.type 0]
+      inOperandList := #[.type 1, .type 1] }
+
+-- Mirrors upstreams `MachineVerifier::verifyPreISelGenericInstruction`.
+def GMIR.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
+    [HasDialect OpInfo GMIR] (op : GMIR) (opPtr : OperationPtr)
+    (ctx : WfIRContext OpInfo) (opIn : opPtr.InBounds ctx.raw) :
+    Except String PUnit := do
+  let info := op.genericOpInfo
+  -- Verify operand and result counts.
+  opPtr.verifyPlainOpCounts ctx opIn info.inOperandList.size info.outOperandList.size
+  -- Verify that every member of a type group has the same concrete type.
+  let typedGroups := info.outOperandList.zip (opPtr.getResultTypes! ctx.raw) ++
+                     info.inOperandList.zip (opPtr.getOperandTypes! ctx.raw)
+  let canon : Std.HashMap TypeGroup TypeAttr :=
+    typedGroups.foldl (init := {}) fun canon (slot, ty) => canon.insertIfNew slot ty
+  for (group, type) in typedGroups do
+    let expected := canon[group]!
+    if expected != type then
+      let name := String.fromUTF8! (IsOpCode.name (opPtr.getOpType ctx.raw opIn))
+      throw s!"{name}: type mismatch: expected {expected}, got {type}"
+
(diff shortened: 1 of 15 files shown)


Veir/Data/LLVM/Int/Lemmas.lean
```diff
@@ -32,6 +33,43 @@ theorem add_comm {w : Nat} (x y : Int w) : add x y = add y x := by
   simp only [add]
   cases x <;> cases y <;> simp [BitVec.add_comm]
 
+/-- Adding zero never wraps, so the result is unchanged even when overflow flags are set. -/
+theorem add_zero {w : Nat} (x : Int w) (nsw nuw : Bool) :
```
luisacicolini, line 37: are these lemmas necessary? In my intuition we should not need them

Veir/Dialects/GMIR/OpInfo.lean
```diff
@@ -0,0 +1,123 @@
...
+public import Veir.Verifier.Basic
+public import Veir.Dialects.LLVM.Properties
+meta import Veir.Meta.OpCode
+
+namespace Veir
+
+public section
+
+@[opcodes]
+inductive Gmir where
```
luisacicolini:
```suggestion
inductive GMIR where
```
for consistency with llvm I would suggest using capslock

naveen-seth: Thanks! Added.

Veir/Dialects/GMIR/OpInfo.lean
```diff
@@ -0,0 +1,125 @@
...
+
+def GMIR.genericOpInfo : GMIR → GenericOpInfo
+  | .g_add | .g_sub =>
+    { outOperandList := #[.type 0]
+      inOperandList := #[.type 0, .type 0] }
+  | .g_icmp =>
+    { outOperandList := #[.type 0]
+      inOperandList := #[.type 1, .type 1] }
+
+-- Mirrors upstreams `MachineVerifier::verifyPreISelGenericInstruction`.
```
luisacicolini:
```suggestion
/-- Mirrors upstreams `MachineVerifier::verifyPreISelGenericInstruction`. -/
```
can you make this into a docstring and maybe add a sentence about what `verifyPreISelGenericInstruction` does in llvm?

Veir/Passes/LLVMToGMIR.lean
```diff
@@ -0,0 +1,70 @@
...
+  # LLVMToGMIR pass
+
+  This file replicates LLVM's GlobalISel IRTranslator pass.
+
+  All properties on LLVM IR operations are carried over to gMIR, similar to
+  upstream's `MachineInstr::copyFlagsFromInstruction`
+
+  This lowering differs from the upstream implementation in the following ways:
+  - Upstream converts all poison values to undef values; we only have poison but no undef.
+  - Some informations represented as operation properties (e.g. the icmp condition code) are
```
luisacicolini, line 20: can we list these precisely?

Veir/Passes/LLVMToGMIR.lean
```diff
@@ -0,0 +1,70 @@
...
+    represented as special operands in upstream gMIR.
+-/
+
+/-! ## Lowering Patterns -/
+
+/--
+  gMIR 1:1 lowerings with Puddle for binary operations.
```
luisacicolini:
```suggestion
  GMIR lowerings with Puddle for binary operations.
```
we always call it GMIR in veir afaiu, it's good to be consistent

luisacicolini (commented):
thank you for the PR! It's pretty good already, I left a couple comments mostly on documentation but the overall design makes sense to me.

regehr:
let's fix the error message for this:
```mlir
  "builtin.module"() ({
    ^bb0(%lhs: i32, %rhs: i32):
      %result = "gmir.g_icmp"(%lhs, %rhs) : (i32, i32) -> i1
  }) : () -> ()
```

right now it says:
```
~/veir$ lake exec veir-opt -p=legalize foo.mlir
foo.mlir:3:43: error: llvm.icmp: missing predicate
      %result = "gmir.g_icmp"(%lhs, %rhs) : (i32, i32) -> i1
```

naveen-seth:
Thanks for the reviews! Fixed it.

regehr:
this code looks good! but please tune up the verifier to reject things like this (icmp should not accept FP operands):

```mlir
"builtin.module"() ({
  ^bb0(%lhs: f32, %rhs: f32):
    %result = "gmir.g_icmp"(%lhs, %rhs) <{predicate = 0 : i64}> : (f32, f32) -> i1
}) : () -> ()
```

naveen-seth:
Apologies for the delay in addressing this!
Added the verifier changes and some more tests covering the diagnostics.

Test/GMIR/icmp_non_i1_return.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: veir-opt %s 2>&1
+
+"builtin.module"() ({
+  %lhs = "llvm.mlir.constant"() <{value = 1 : i64}> : () -> i64
+  %rhs = "llvm.mlir.constant"() <{value = 2 : i64}> : () -> i64
+  %result = "gmir.g_icmp"(%lhs, %rhs) <{predicate = 0 : i64}> : (i64, i64) -> i64
+}) : () -> ()
```
regehr, line 7: do you want to add a CHECK: line in this test case?

regehr: (or maybe VEIR_ROUNDTRIP if you just want to assert that it passes through)

regehr (approved):
nothing references `TypeGroup.ptype` is that what you intend?

anyway-- this LGTM overall, but let's see what others say!

Test/GMIR/icmp_non_i1_return.mlir
naveen-seth: Yes, that’s exactly what I wanted! Thanks, added.

naveen-seth:
> nothing references `TypeGroup.ptype` is that what you intend?
> 
> anyway-- this LGTM overall, but let's see what others say!

Yes, that’s intentional. I didn’t want to use `.ptype` in this PR yet, but we’d need it for GMIR operations with pointer-only operands/results, like `G_LOAD` and `G_STORE`.
(But also wouldn't mind removing it for now!)

Veir/Dialects/GMIR/OpInfo.lean
```diff
@@ -0,0 +1,147 @@
...
+  toAttrDict := GMIR.toAttrDict
+
+/--
+This replicates the type-group information in LLVM's `GenericOpcodes.td`
+(llvm/include/llvm/Target/GenericOpcodes.td).
+
+Immediate operands and operands with LLVM's `unknown` (special) type are omitted and represented
+as operation properties instead.
+-/
+inductive TypeGroup where
```
luigirinaldi, line 63: document what is the `id : Nat` doing here. Also explain the difference between `type` and `ptype` (property type I assume?). Atm it's not used anywhere right, so maybe don't include it?

Veir/Dialects/GMIR/OpInfo.lean
```diff
@@ -0,0 +1,147 @@
...
+
+def GMIR.genericOpInfo : GMIR → GenericOpInfo
+  | .g_add | .g_sub =>
+    { outOperandList := #[.type 0]
+      inOperandList := #[.type 0, .type 0] }
+  | .g_icmp =>
+    { outOperandList := #[.type 0]
+      inOperandList := #[.type 1, .type 1] }
+
+def OperationPtr.verifyGMIRICmp {OpInfo : Type} [IsOpCode OpInfo]
```
luigirinaldi, line 81: this function is not specific to GMIR? perhaps it could be named `verifyIntegerOrPointerICmp`? Or keep the name and check/document that the opcode is must be a gmir one?

Veir/Passes/LLVMToGMIR.lean
```diff
@@ -0,0 +1,71 @@
...
+                    (fun p => some (cast h p))
+      let newOp ← Puddle.CreateProg.operation (.gmir gOp) #[lhs, rhs] #[resultType] props
+      return newOp)
+    (fun newOp => newOp)
+
+def g_add := (lowerBinop .add .g_add).compile
+
+def g_sub := (lowerBinop .sub .g_sub).compile
+
+def g_icmp := (lowerBinop .icmp  .g_icmp).compile
```
luigirinaldi: so how is the `icmp` condition code lowered in this case?

luigirinaldi (commented):
looks good overall, a few minor things, mainly for me to understand what is happening.

Veir/Dialects/GMIR/OpInfo.lean
luigirinaldi: I see now from other comments that it's not used, probably best to remove it then.

Veir/Dialects/GMIR/OpInfo.lean
naveen-seth: It should only be specific to GMIR. I've made it private for now but will follow up on addressing this because I believe that I still have to add some corresponding lemmas in Verifier.lean.

Veir/Dialects/GMIR/OpInfo.lean
naveen-seth: Thanks for reviewing! Removed it.

Veir/Dialects/GMIR/OpInfo.lean
naveen-seth: I think the Verifier.lean lemmas will be a bit larger and I'd like to split them into a new PR. Would it be fine keeping `verifyGMIRICmp` as private for now, and completing the missing parts in a follow-up?

luigirinaldi: yes I think that sound reasonable

luigirinaldi (approved):

naveen-seth:
Thanks for the reviews!

PR #1309, opened 2026-08-27
feat(puddle): Improve runtime of puddle patterns

math-fehr:
`Pattern.compile` was not compiled as efficiently as it could by Lean, and was reinitializing the closure at each invocation. This PR changes `compile` to create a structure that contains this closure, which makes it more efficient. In the worse case I measured, this makes us go from a `2.62x` slowdown to a `1.85x` slowdown.


tobiasgrosser (approved):
Very cool.

PR #1315, opened 2026-08-31
feat(puddle): check structural well-formedness of puddle patterns

math-fehr:
This PR add the field `structurallyWellFormed` to `Pattern.Valid` that checks that:
* the match program begins with an operation declaration for its root;
* matcher bindings that share an identifier also share a runtime kind;
* every creation input is bound by the matcher or by an earlier creation declaration;
* every creation output has an identifier that is globally fresh;
* each created operation has as many result handles as result-type handles;
* every replacement value is bound by the matcher or the creation program; and
* the matched root and its results are not available to the creation or replacement phases.

This proposition is decidable, and even better we can just check it with `cbv`, which is quite efficient (even if the algorithm is O(n^2), so this might need to change in the future).

### DIFF
--- a/Veir/PatternRewriter/Puddle/Validity.lean
+++ b/Veir/PatternRewriter/Puddle/Validity.lean
@@ -74,6 +74,207 @@ def MatchProg.ConstrainsRoot (prog : MatchProg OpInfo α) : Prop :=
   | .operation _ _ _ _ _ opHandle _ _ :: _ => opHandle = prog.rootHandle
   | _ => False
 
+/--
+Return the hidden SSA-result handles from the program's root constraint.
+
+The root is assumed to be the first declaration in the match program. If it is not, then this
+function returns `none`.
+-/
+@[expose]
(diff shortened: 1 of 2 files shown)


tobiasgrosser:
> This PR add the field `structurallyWellFormed` in `Pattern.Valid` that checks that:

 This PR add the field `structurallyWellFormed` **to** `Pattern.Valid` that checks that:

Veir/PatternRewriter/Puddle/Validity.lean
```diff
@@ -123,9 +327,8 @@ macro "provePuddleValid" : tactic =>
     unfoldPuddleBuilder
     constructor
     · provePuddleSupported
-    · rfl
+    · cbv
+    · cbv
   ))
 
-end
-
```
tobiasgrosser, line 130: Whyi is this `end` moved here? If this change is needed, should the earlier PR in this stack of PRs not just create this `end` at the right place?

tobiasgrosser (approved):

PR #1322, opened 2026-09-02
feat(dataflow): add ModArithRangeAnalysis

axelcool1234:


Veir/Passes/PrintModArithRanges.lean
AlexanderViand, line 1: For a follow-up PR: it might be nicer to have PrintAnalysis.lean that takes an analysis and  a Lattice-to-string function

AlexanderViand (changes requested):
@axelcool1234 Can you please split this into two PRs: the needed changes to the DataFlowFramework and the mod_arith integer range analysis? Thanks!

Veir/Passes/PrintModArithRanges.lean
axelcool1234: > For a follow-up PR: it might be nicer to have PrintAnalysis.lean that takes an analysis and a Lattice-to-string function

Okay thanks

shankarapailoor:
I think one potential issue is that widening should only be run on loop back edges but now it is run on every CFG merge which is pretty imprecise.

shankarapailoor:
Can you explain the logic behind `applyReduction`? It seems to not handle cases like "barrett" or "full" and I'm not sure why?

Test/Analysis/ModArithRangeAnalysis.mlir
```diff
@@ -0,0 +1,158 @@
+// RUN: split-file %s %t
```
AlexanderViand: Split-file isn't available on CI, so this will unfortunately need to be split into actual files for now.

Test/Analysis/ModArithRangeAnalysis.mlir
```diff
@@ -0,0 +1,158 @@
...
+    %add2 = "mod_arith.add"(%add1, %a) {"reduction" = "none"} : (!mod_arith.int<12289 : i32>, !mod_arith.int<12289 : i32>) -> !mod_arith.int<12289 : i32>
+    %out = "mod_arith.mul"(%add2, %small) {"reduction" = "none"} : (!mod_arith.int<12289 : i32>, !mod_arith.int<12289 : i32>) -> !mod_arith.int<12289 : i32>
+    "func.return"(%out) : (!mod_arith.int<12289 : i32>) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+//--- storage-overflow.mlir
+
+// Raw ModArith operations still produce values of their unsigned storage type.
+// When a mathematical interval can overflow i8, use the full i8 range.
```
AlexanderViand: I think we already discussed this, but for what we want to use this analysis for, the container type does not actually upper-bound the ranges.  For example, here we'd need the answer to be either `[0, 500]` for the add and `[0, 62500]` for the mul, or something like `[0, 256]` for both to indicate that they overflow.

I'm guessing the first might make life a bit easier for us but is more painful for the analysis, as you'll get some crazy large ranges, while the second one I'd have to think about a bit more carefully if it's really sufficient.

Test/Analysis/ModArithRangeAnalysis.mlir
```diff
@@ -0,0 +1,158 @@
...
+// NONE-NEXT: // mod_arith.range mod_arith.add result 0 = [46, 24622]
+// NONE-NEXT: // mod_arith.range mod_arith.add result 0 = [46, 36910]
+// NONE-NEXT: // mod_arith.range mod_arith.mul result 0 = [138, 110730]
+
+"builtin.module"() ({
+  "func.func"() <{function_type = (!mod_arith.int<12289 : i32>, !mod_arith.int<12289 : i32>) -> !mod_arith.int<12289 : i32>, sym_name = "no_reduction"}> ({
+  ^entry(%a : !mod_arith.int<12289 : i32>, %b : !mod_arith.int<12289 : i32>):
+    %c = "mod_arith.constant"() <{"value" = 46 : i32}> : () -> !mod_arith.int<12289 : i32>
+    %small = "mod_arith.constant"() <{"value" = 3 : i32}> : () -> !mod_arith.int<12289 : i32>
+    %add0 = "mod_arith.add"(%a, %c) {"reduction" = "none"} : (!mod_arith.int<12289 : i32>, !mod_arith.int<12289 : i32>) -> !mod_arith.int<12289 : i32>
```
AlexanderViand, line 37: I see you're using the `{ "reduction" = "..."}` attributes on the `mod_arith` ops to determine what's going on, nice!

AlexanderViand (changes requested):
Generally LGTM, but in order to make this useful for the intended optimisation, we need to be able to differentiate "just barely fits into container type" from "could overflow container type", see comment below.

AlexanderViand (approved):
LGTM 👍  Thanks for putting this together!

Test/Analysis/ModArithRangeAnalysis.mlir
```diff
@@ -0,0 +1,106 @@
+// RUN: veir-opt %s -p=print-mod-arith-ranges | filecheck %s
+
+// CHECK:      // mod_arith.range block argument 0 = [0, 12288]
+// CHECK-NEXT: // mod_arith.range block argument 1 = [0, 12288]
+// CHECK-NEXT: // mod_arith.range mod_arith.constant result 0 = [46, 46]
+// CHECK-NEXT: // mod_arith.range mod_arith.constant result 0 = [3, 3]
+// CHECK-NEXT: // mod_arith.range mod_arith.add result 0 = [0, 12288]
+// CHECK-NEXT: // mod_arith.range mod_arith.add result 0 = [0, 12288]
+// CHECK-NEXT: // mod_arith.range mod_arith.add result 0 = [0, 12288]
+// CHECK-NEXT: // mod_arith.range mod_arith.mul result 0 = [0, 12288]
```
math-fehr, line 10: Can you move each check under the operation it prints the analysis for?
Otherwise, it is a bit hard to understand this test.

math-fehr (approved):
Looks good to me!
I would just change the tests a bit but otherwise all good!

PR #1326, opened 2026-09-02
fix(LLVM): extend properly on mixed integer width (#1297)

AdUhTkJm:
We need to zero-extend when the width of attribute integer type is 1, and sign-extend otherwise.

### DIFF
--- a/Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
+++ b/Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
@@ -1,6 +1,4 @@
 // RUN: veir-interpret %s | filecheck %s
-// Expected to fail until `llvm.mlir.constant` handles the value attribute's
-// integer width the way MLIR does; drop the XFAIL with the fix.
 
 // MLIR special-cases width-1 (and unsigned) integer attributes: they are
 // *zero*-extended to the result width, not sign-extended.  See
--- a/Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
+++ b/Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
@@ -1,7 +1,6 @@
 // RUN: veir-interpret %s | filecheck %s
 // Expected to fail until `llvm.mlir.constant` handles the value attribute's
 // integer width the way MLIR does; drop the XFAIL with the fix.
-// XFAIL: *
 
 // MLIR special-cases width-1 (and unsigned) integer attributes: they are
 // *zero*-extended to the result width, not sign-extended.  See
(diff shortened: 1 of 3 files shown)


tobiasgrosser (approved):
This LGTM. Let's see what @regehr thinks.

AdUhTkJm:
By the way, do we have a prettier for VeIR code, one like clang-format?

Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
```diff
@@ -1,7 +1,6 @@
 // RUN: veir-interpret %s | filecheck %s
 // Expected to fail until `llvm.mlir.constant` handles the value attribute's
 // integer width the way MLIR does; drop the XFAIL with the fix.
```
tobiasgrosser: The comment here is outdated. We do not have an `XFAIL` anymore.

Test/Interpreter/LLVM/constant_attr_sext_to_wider_result.mlir
```diff
@@ -1,7 +1,6 @@
 // RUN: veir-interpret %s | filecheck %s
 // Expected to fail until `llvm.mlir.constant` handles the value attribute's
 // integer width the way MLIR does; drop the XFAIL with the fix.
```
tobiasgrosser: The comment here is outdated. We do not have an `XFAIL` anymore.

Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
AdUhTkJm: Could you elaborate on this? I believe the upstream [file](https://github.com/opencompl/veir/blob/main/Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir) on current head still contains an XFAIL on line 4.

AdUhTkJm: Oh, did you mean I should remove these comments?

tobiasgrosser: Yes.

tobiasgrosser: Nice!

Test/Interpreter/LLVM/constant_attr_sext_to_wider_result.mlir
tobiasgrosser: Nice!

tobiasgrosser:
@regehr, can you have a look?

regehr (approved):
I suspect this change will cause us to miss some optimizations, for example we currently iirc match `select -1, x, y` but we'll need to change that to `select 1, x, y`.
but we can fix that up later.

regehr:
argh I think the consequences might be worse than I thought, there's some RISC-V fallout. I'll work on it.

PR #1327, opened 2026-09-02
parser: accept trailing type on unregistered attributes

kumarak:
MLIR's generic parser accepts two forms of dialect attribute that VeIR rejected: the body may be omitted (`#foo.bar`), and any dialect attribute may carry a trailing `: type` (`#foo.bar<baz> : i32`, `#foo.int<1> : !foo.int<s, 32>`). Both show up in IR from dialects VeIR does not register, typically as typed constants, and `veir-opt --allow-unregistered-dialect` failed on them.

The changes stores the trailing type structurally on `UnregisteredAttr` and prints it back after the body, mirroring `mlir::OpaqueAttr`. Types never carry one, so the field is always `none` when `isType` is true.

### DIFF
--- a/Test/Parsing/unregistered_typed_attr.mlir
+++ b/Test/Parsing/unregistered_typed_attr.mlir
@@ -1,0 +1,23 @@
+// RUN: VEIR_UNREGISTERED_ROUNDTRIP
+
+// Attributes from unregistered dialects may omit the `<body>` and may carry a
+// trailing `: type`, as MLIR's generic parser allows for any dialect attribute.
+// The type may be builtin or itself unregistered, and the attribute may appear
+// in properties, discardable attributes, and nested containers.
+"builtin.module"() ({
+  "func.func"() <{function_type = () -> (), sym_name = "main"}> ({
+    "unregistered.op_one"() {foo = #foo.bar} : () -> ()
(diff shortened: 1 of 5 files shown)


Test/Parsing/unregistered_typed_attr.mlir
```diff
@@ -0,0 +1,26 @@
+// RUN: VEIR_UNREGISTERED_ROUNDTRIP
+// `MLIR_UNREGISTERED_ROUNDTRIP` only enables unregistered dialects in veir-opt; the ops here
+// are unregistered in MLIR too, so mlir-opt needs the flag as well.
```
tobiasgrosser: Can you create a separate PR for this? Changing the lit.cfg file?

tobiasgrosser (approved):
This is great. Let's see what @math-fehr thinks.

tobiasgrosser:
@kumarak, it seems there is a small merge conflict. Could you fix it?

tobiasgrosser:
And rebase, please.

Test/Parsing/unregistered_typed_attr.mlir
kumarak: Removed custome LIT run since it is added to lit config.

tobiasgrosser:
@kumarak, the tests seem to fail due to a problem with the guard messages.

PR #1329, opened 2026-09-02
feat(llzk): add the Felt dialect

AlexanderViand:
This adds the Felt (Field Element) dialect which is how the [LLZK compiler](https://github.com/project-llzk) models finite field arithmetic.

### DIFF
--- a/Veir/Dialects/LLZK/Felt/OpInfo.lean
+++ b/Veir/Dialects/LLZK/Felt/OpInfo.lean
@@ -1,0 +1,156 @@
+module
+
+public import Veir.IR.Simp
+public import Veir.IR.OpInfo
+public import Veir.Verifier.Basic
+public import Veir.Dialects.LLZK.Felt.Properties
+meta import Veir.Meta.OpCode
+
+namespace Veir
+
+public section
+
+@[opcodes]
+inductive Felt where
+| const
+| add
+| sub
+| mul
+| pow
+| div
+| uintdiv
+| sintdiv
+| umod
+| smod
+| neg
+| inv
+| bit_and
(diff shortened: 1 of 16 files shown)


Veir/Dialects/LLZK/Felt/OpInfo.lean
```diff
@@ -0,0 +1,156 @@
...
+    .ok ()
+
+def Felt.fromAttrDict
+    (op : Felt) (attrDict : Std.HashMap ByteArray Attribute) :
+    Except String (Felt.propertiesOf op) :=
+  match op with
+  | .const => FeltConstProperties.fromAttrDict attrDict
+  | .add | .sub | .mul | .pow | .div | .uintdiv | .sintdiv | .umod | .smod
+  | .neg | .inv | .bit_and | .bit_or | .bit_xor | .bit_not | .shl | .shr =>
+    noProperties attrDict
```
AlexanderViand, line 57:
```suggestion
  | _ => noProperties attrDict
```

shankarapailoor (approved):
LGTM. We have other validation/attributes on the Felt ops that will come in when we import the func dialect.

Veir/Dialects/LLZK/Felt/OpInfo.lean
AlexanderViand: Never mind, wildcard does't work because Lean can't figure out `Felt.propertiesOf op` is `Unit` in the catch-all branch because the return type depends on `op`.

ineol (commented):
I didn't look super carefuly but LGTM

ineol (approved):

PR #1333, opened 2026-09-02
chore: allow unregistered dialects in MLIR_UNREGISTERED_ROUNDTRIP

tobiasgrosser:
We forgot to allow unregistered dialects in mlir-opt, but some test cases benefit of operations that do not need to exist in mlir.


kumarak (approved):
LGTM

PR #1334, opened 2026-09-02
feat(Verif): Add initial support for verif.assume and assert

### LINKED ISSUE #1320: Support CIRCT's `verif.assume` and `verif.assert`
To start supporting CIRCT's constructs for model checking, we need to be able to represent some of the `verif` dialect. This will require creating a new `Verif` directory in `Veir/Dialects` and defining some operations/verifiers (as in other dialects).

A good start would be [`verif.assume`](https://circt.llvm.org/docs/Dialects/Verif/#verifassume-circtverifassumeop) and [`verif.assert`](https://circt.llvm.org/docs/Dialects/Verif/#verifassert-circtverifassertop), that express properties we want to assume/check about a hardware circuit. Although these allow an arbitrary `ltl.property` for the `property` argument, for now we can just implement it in the case where `property` is an `i1`.

This can target parsing the following with `lake exe veir-opt input.mlir`:

```mlir
"builtin.module"() ({
  "hw.module"() <{comment = "", module_type = !hw.modty<input a : i1, input b : i1>, parameters = [], per_port_attrs = [], result_locs = [], sym_name = "top"}> ({
  ^bb0(%arg0: i1, %arg1: i1):
    "verif.assume"(%arg0) : (i1) -> ()
    "verif.assume"(%arg0, %arg1) : (i1, i1) -> ()
    "verif.assert"(%arg0) : (i1) -> ()
    "verif.assert"(%arg0, %arg1) : (i1, i1) -> ()
    "hw.output"() : () -> ()
  }) : () -> ()
}) : () -> ()
```

Scheremo:
This patch fixes #1320 - made in collaboration with @mafeguimaraes and @georgerennie!

### DIFF
--- a/Veir/Dialects/Verif/OpInfo.lean
+++ b/Veir/Dialects/Verif/OpInfo.lean
@@ -2,7 +2,6 @@ module
 
 public import Veir.IR.Simp
 public import Veir.IR.OpInfo
-public import Veir.ConstantMaterialization
 public import Veir.Verifier.Basic
 meta import Veir.Meta.OpCode
 
--- a/Veir/Dialects/Verif/OpInfo.lean
+++ b/Veir/Dialects/Verif/OpInfo.lean
@@ -1,0 +1,87 @@
+module
+
+public import Veir.IR.Simp
+public import Veir.IR.OpInfo
+public import Veir.ConstantMaterialization
+public import Veir.Verifier.Basic
+meta import Veir.Meta.OpCode
+
+namespace Veir
+
+public section
+
+@[opcodes]
+inductive Verif where
+| assume
+| assert
+deriving Inhabited, Repr, Hashable, DecidableEq
+
+@[expose, properties_of]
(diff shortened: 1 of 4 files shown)


Veir/Dialects/Verif/OpInfo.lean
```diff
@@ -0,0 +1,87 @@
+module
+
+public import Veir.IR.Simp
+public import Veir.IR.OpInfo
+public import Veir.ConstantMaterialization
```
georgerennie: I dont think we need this import

georgerennie (commented):
Small unneeded import but I think otherwise this is great! We can add the label attributes later.

Veir/Dialects/Verif/OpInfo.lean
Scheremo: 🫡  Removed it, thanks!

tobiasgrosser (approved):
Nice!

PR #1340, opened 2026-09-02
feat(llzk): add String and Global dialects

AlexanderViand:
Adds LLZK String and Global operation definitions, properties, parsing, verification, and focused identity and invalid-input tests.


AlexanderViand:
@shankarapailoor Can you let us know if this matches what you think these LLZK dialects should be doing?

ineol (approved):
lgtm

PR #1344, opened 2026-09-02
feat(RISCVCombines): add combines to speed-up ChaCha20

tobias-rothmann:
Adds verified RISC-V combines:

- `roriw (xor (zextw x) (zextw y)), imm` → `roriw (xor x y), imm`, removing redundant extensions in the ChaCha20 idiom while preserving shared full-width XOR uses.
- `zextw (zextb x)` → `zextb x`, including fixed-register retargeting.

Includes correctness proofs and regression tests for matching, negative, shared-use, and fixed-register cases.


luisacicolini:
thanks a lot for the PR, the code looks good to me, I left a couple comments! I will ask a second look by @regehr and @naveen-seth - do you guys think these rewrites will interfere with the legalization work?

Veir/Passes/RISCVCombines/Combine.lean
```diff
@@ -1558,6 +1560,202 @@ def zexth_zexth := drop_redundant_ext .zexth
 def sextb_sextb := drop_redundant_ext .sextb
 def sexth_sexth := drop_redundant_ext .sexth
 
+/-- `riscv.zextw (riscv.zextb x) -> riscv.zextb x`. A byte zero-extension
+    already clears every bit which `zextw` would clear.
+
+    LLVM: `CastInst::isEliminableCastPair` folds a `ZExt` followed by a `ZExt`
+    to the first cast.
+    https://github.com/llvm/llvm-project/blob/c536b0aa030474672e293dccdb27b97c36c4e1af/llvm/lib/IR/Instructions.cpp#L2922-L2967 -/
```
luisacicolini: instead of this link to Instructions is there a link to an actual rewrite we can use? (same below)

Veir/Passes/RISCVCombines/Combine.lean
```diff
@@ -1558,6 +1560,202 @@ def zexth_zexth := drop_redundant_ext .zexth
...
+private def zextw_zextb_pattern : Puddle.Pattern OpCode :=
```
luisacicolini: why is this definition private? (same for all the defs in this file)

naveen-seth:
Upstream seems to make these optimizations during the legalization step, but I'd prefer landing this and to then move/revert this later once we have gotten further on legalization.
Thanks for tagging me; I'll keep this in PR in mind!

Veir/Passes/RISCVCombines/Combine.lean
tobias-rothmann: I'm not sure. I don't think there's a coherent pattern, but it seems like you only want to expose the compiled version for some combines. I've removed the "private" tag from all definitions now.

Veir/Passes/RISCVCombines/Combine.lean
tobias-rothmann: Yes, sorry, updated!

Issue #975, opened 2026-07-03
Support --split-input-file


AlexanderViand:
Especially for "negative" tests (invalid IR triggering verifiers, etc), it quickly becomes rather tedious to separate everything into its own *.mlir file. `mlir-opt` handles this with `--split-input-file`, which allows you to have multiple failing test in a single *.mlir file.

alethkit:
Hi - I've been working on a fix for this, but i'm not sure what input test cases to test this with.

Do you have some examples of the desired behaviour?

naveen-seth:
I think this test would be a good example and we could adopt a similar test:
https://github.com/llvm/llvm-project/blob/main/mlir/test/mlir-opt/split-markers.mlir

PR #1348, opened 2026-09-03
ci: install split-file for lit tests

AlexanderViand:
I think for the namespace CI, the docker file in the repo is just documentation, and @tobiasgrosser  or someone else will need to actually update/recreate the runner manually?


tobiasgrosser:
Let's not make every user install LLVM. The right way is to create a simple split-file clone in Python and install it via uv. No need to update the dockerfile.

AlexanderViand:
Closing in favor of 
* #975

PR #1349, opened 2026-09-03
cir: lower the cir integer core to arith, cf and llvm

kumarak:
PR adds `cir-to-std`, which gives the cir dialect its semantics by lowering to `arith`, `cf` and `llvm`. It follows the `mod-arith-to-arith` approach, since VeIR has no dialect-conversion framework: each value operation is rewritten in place with unrealized_conversion_casts at the type boundary, function and block boundaries are retyped with bridging casts, and the cast pairs are reconciled in the same pass so no cir type survives. Signedness selects the arith operation (divsi/divui, shrsi/shrui, signed or unsigned cmpi predicates); saturating arithmetic and cast kinds without a lowering are rejected.

cir.func and cir.return stay in place, because the rewriter cannot change an opcode or move a region between operations. cir.func implements the function interface and cir.return gets the same interpreter arm as func.return, so veir-opt -p=cir | veir-interpret runs lowered programs; the cir pass group expands to cir-to-std,canonicalize,cse,dce. Tests cover signed and unsigned arithmetic, comparisons with branches, a flat loop, casts, and the two rejected cases. 

Not covered yet: pointers and memory operations, cir.call, and structured control flow.

### DIFF
--- a/Test/Interpreter/Cir/add_mul.mlir
+++ b/Test/Interpreter/Cir/add_mul.mlir
@@ -1,0 +1,15 @@
+// RUN: veir-opt %s -p=cir > %t.mlir && veir-interpret %t.mlir | filecheck %s
+
+// (7 + 5) * 3 = 36.
+"builtin.module"() ({
+  "cir.func"() <{function_type = !cir.func<() -> !cir.int<s, 32>>, sym_name = "main"}> ({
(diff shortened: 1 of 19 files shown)


AlexanderViand (commented):
Only did a quick skim, but it looks good to me :)

ineol:
LGTM

PR #1350, opened 2026-09-03
[Draft] feat(Parser): Adds more floating-point types and their parser

AdUhTkJm:
Draft PR.

To unify internal representation of floating-type values of different settings, we chose to store them in bit vectors.


UnitTest/AttrParser.lean
```diff
@@ -68,13 +48,19 @@ def expectSuccessType (s : String) (expected : TypeAttr)
   testType s allowUnregisteredDialect = .ok expected
 
 /--
-  Test that parsing an attribute in the given string succeeds and matches the expected attribute.
+   Test that parsing an attribute in the given string succeeds and matches the expected attribute.
```
bollu, line 71: Unrelated whitespace change?

Test/Interpreter/LLVM/constant_attr_i1_zext_to_wider_result.mlir
```diff
@@ -1,7 +1,6 @@
 // RUN: veir-interpret %s | filecheck %s
 // Expected to fail until `llvm.mlir.constant` handles the value attribute's
 // integer width the way MLIR does; drop the XFAIL with the fix.
-// XFAIL: *
```
bollu, line 4: Unrelated change

Test/Interpreter/LLVM/constant_attr_sext_to_wider_result.mlir
```diff
@@ -1,7 +1,6 @@
 // RUN: veir-interpret %s | filecheck %s
 // Expected to fail until `llvm.mlir.constant` handles the value attribute's
 // integer width the way MLIR does; drop the XFAIL with the fix.
-// XFAIL: *
```
bollu, line 4: Unrelated change

Veir/IR/Attribute.lean
```diff
@@ -150,31 +192,25 @@ deriving Inhabited, Repr, DecidableEq, Hashable
...
-        have hval : a.value = b.value := floatEqOfToBitsEq hv
-        isTrue (by
-          cases a
-          cases b
-          simp_all)
-      else
-        isFalse (by intro h; exact ht (congrArg FloatAttr.type h))
-    else
-      isFalse (by intro h; exact hv (congrArg (Float.toBits ∘ FloatAttr.value) h))
+  value : BitVec type.bitwidth
```
bollu: having `value` be a `BitVec type.bitwidth` is OK for now. However, we will probably eventually want to mirror, or maybe even reuse, core parts of Lean's floating point specification via [`Lean.Init.Data.Float.Model.Float`](https://github.com/leanprover/lean4/blob/master/src/Init/Data/Float/Model/Float.lean).

PR #1351, opened 2026-09-03
feat(HW): Add support for hw.instance

### LINKED ISSUE #1318: Support CIRCT's `hw.instance`
Although we can define `hw.module`, we can't yet instantiate them! This requires supporting [`hw.instance`](https://circt.llvm.org/docs/Dialects/HW/#hwinstance-circthwinstanceop). This will require defining a new op/verifier in `Veir/Dialects/HW`. Wed do not need to support all parameters of instance yet - `inner_sym` and `doNoPrint` can be ignored for now.

The example below can be tested with `lake exe veir-opt input.mlir`

```mlir
"builtin.module"() ({
  "hw.module"() <{comment = "", module_type = !hw.modty<input a : i8, input b : i8, output out : i8>, parameters = [], per_port_attrs = [], result_locs = [loc(unknown)], sym_name = "add2"}> ({
  ^bb0(%arg3: i8, %arg4: i8):
    %2 = "comb.add"(%arg3, %arg4) : (i8, i8) -> i8
    "hw.output"(%2) : (i8) -> ()
  }) {sym_visibility = "private"} : () -> ()
  "hw.module"() <{comment = "", module_type = !hw.modty<input a : i8, input b : i8, input c : i8, output out : i8>, parameters = [], per_port_attrs = [], result_locs = [loc(unknown)], sym_name = "add3"}> ({
  ^bb0(%arg0: i8, %arg1: i8, %arg2: i8):
    %0 = "hw.instance"(%arg0, %arg1) <{argNames = ["a", "b"], instanceName = "a0", moduleName = @add2, parameters = [], resultNames = ["out"]}> {sv.namehint = "s"} : (i8, i8) -> i8
    %1 = "hw.instance"(%arg2, %0) <{argNames = ["a", "b"], instanceName = "a1", moduleName = @add2, parameters = [], resultNames = ["out"]}> : (i8, i8) -> i8
    "hw.output"(%1) : (i8) -> ()
  }) : () -> ()
}) : () -> ()
```

mafeguimaraes:
This patch adds support to `instanceName`, `moduleName`, `argNames`, `resultNames`, and `parameters` attributes from the `hw` dialect. 

Fixes #1318

### DIFF
--- a/Veir/Dialects/HW/OpInfo.lean
+++ b/Veir/Dialects/HW/OpInfo.lean
@@ -16,6 +16,7 @@ inductive HW where
 | constant
 | module
 | output
+| instance
 deriving Inhabited, Repr, Hashable, DecidableEq
 
 @[expose, properties_of]
@@ -23,6 +24,7 @@ def HW.propertiesOf (op : HW) : Type :=
 match op with
 | .constant => HWConstantProperties
 | .module => HWModuleProperties
+| .instance => HWInstanceProperties
 | _ => Unit
 
 def HW.fromAttrDict
@@ -31,6 +33,7 @@ def HW.fromAttrDict
   cases op
   case constant => exact HWConstantProperties.fromAttrDict attrDict
   case module => exact HWModuleProperties.fromAttrDict attrDict
+  case «instance» => exact HWInstanceProperties.fromAttrDict attrDict
   all_goals exact .ok ()
 
 def HW.toAttrDict
@@ -45,6 +48,14 @@ def HW.toAttrDict
     let dict := dict.insert "module_type".toUTF8 (.hwModuleType props.module_type)
     let dict := dict.insert "sym_name".toUTF8 (.stringAttr props.sym_name)
     let dict := dict.insert "per_port_attrs".toUTF8 (.arrayAttr props.per_port_attrs)
+    let dict := dict.insert "parameters".toUTF8 (.arrayAttr props.parameters)
+    dict
+  | .instance => Id.run do
+    let dict := Std.HashMap.emptyWithCapacity 5
+    let dict := dict.insert "instanceName".toUTF8 (.stringAttr props.instanceName)
+    let dict := dict.insert "moduleName".toUTF8 (.flatSymbolRefAttr props.moduleName)
+    let dict := dict.insert "argNames".toUTF8 (.arrayAttr props.argNames)
+    let dict := dict.insert "resultNames".toUTF8 (.arrayAttr props.resultNames)
     let dict := dict.insert "parameters".toUTF8 (.arrayAttr props.parameters)
     dict
   | _ => Std.HashMap.emptyWithCapacity 0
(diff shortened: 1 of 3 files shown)


tobiasgrosser (commented):
LGTM. Let's see what @georgerennie and @Scheremo say.

Veir/Dialects/HW/OpInfo.lean
```diff
@@ -47,6 +50,14 @@ def HW.toAttrDict
     let dict := dict.insert "per_port_attrs".toUTF8 (.arrayAttr props.per_port_attrs)
     let dict := dict.insert "parameters".toUTF8 (.arrayAttr props.parameters)
     dict
+  | .instance => Id.run do
+    let dict := Std.HashMap.emptyWithCapacity 5
+    let dict := dict.insert "instanceName".toUTF8 (.stringAttr props.instanceName)
+    let dict := dict.insert "moduleName".toUTF8 (.flatSymbolRefAttr props.moduleName)
+    let dict := dict.insert "argNames".toUTF8 (.arrayAttr props.argNames)
+    let dict := dict.insert "resultNames".toUTF8 (.arrayAttr props.resultNames)
+    let dict := dict.insert "parameters".toUTF8 (.arrayAttr props.parameters)
```
Scheremo, line 63: According to [HW Dialect spec](https://circt.llvm.org/docs/Dialects/HW/#hwinstance-circthwinstanceop) there are two more attributes: `inner_sym` and `doNotPrint`. Not sure if they should be handled, but it might be nice to add a comment here to make sure it's documented as an explicit choice not to model them.

Scheremo (approved):
One comment about documenting the "missing" (?) attributes, otherwise LGTM!

Veir/Dialects/HW/OpInfo.lean
mafeguimaraes: We chose not to add support for `inner_sym` and `doNotPrint` because we don't have support for `::circt::hw::InnerSymAttr` and `::mlir::UnitAttr` types yet.

I've just added a comment to document this. 

Thanks!

Veir/Dialects/HW/OpInfo.lean
```diff
@@ -31,8 +33,13 @@ def HW.fromAttrDict
   cases op
   case constant => exact HWConstantProperties.fromAttrDict attrDict
   case module => exact HWModuleProperties.fromAttrDict attrDict
+  case «instance» => exact HWInstanceProperties.fromAttrDict attrDict
   all_goals exact .ok ()
 
+/--
```
georgerennie: I think this comment is actually most relevant in `HWInstanceProperties.fromAttrDict`, as these fields doesn't exist in the object passed to `toAttrDict`, but do exist in the hashmap parsed in by `fromAttrDict` - but this is a tiny nit so not a dealbreaker for me

georgerennie (approved):
Thanks! This looks great to me. And yeah I think it is ok for us to just ignore those last two attributes for now, they aren't too important for us at the moment and are optional.

Veir/Dialects/HW/OpInfo.lean
mafeguimaraes: That makes sense, I'll fix that. Thanks!

PR #1353, opened 2026-09-03
feat(pass): add `apply-patterns` pass

math-fehr:
This pass takes as input a set of strings that represent peephole rewrites, and apply the peephole rewrites until a fixpoint in the given IR.

It is currently limited to `Veir.Passes.InstCombine` rules, but the goal is to extend it with all available peephole rewrites in the future.


tobiasgrosser (approved):
Nice. That makes sense to me.

Issue #952, opened 2026-06-30
LLVM: support parsing omitted `!llvm.` prefix when nested


GZGavinZhao:
## Context

LLVM dialect has a custom directive `PrettyLLVMType` that allows one to write types from the LLVM dialect without `!llvm.` prefix when it's already nested inside another LLVM-dialect type. For example, `!llvm.array<2 x struct<...>>` is the same as `!llvm.array<2 x !llvm.struct<...>>`.

This convenience syntax is only available for types wrapped inside the `PrettyLLVMType` custom directive. In practice, this convenience syntax is available wherever a nested type is allowed, so in VeIR, we just always allow it when we know we're parsing inside a `!llvm.` type.

## Status

Since `PrettyLLVMType` applies to all types from the LLVM dialect, `parseLLVMType` needs to handle the shorthand form for all 10 types:
- [x] `void`
- [x] `ptr`
- [x] `array`
- [ ] `func`
- [ ] `struct` #951 
- [ ] `target`
- [ ] `x86_amx`
- [ ] `ppc_fp128`
- [ ] `label`
- [ ] `metadata`

PR #1354, opened 2026-09-03
fix(parser): accept shorthand LLVM function types

sueszli:
Part of #952.

`parseLLVMType` already accepts `void`, `ptr`, `array`, and `struct` without the `!llvm.` prefix. This adds the missing `func` case:

```mlir
func<i32 (ptr)>
```

It now parses to the same type as `!llvm.func<i32 (!llvm.ptr)>`.


regehr (approved):
LGTM but in the future please consider also adding MLIR-level tests, not just unit tests

sueszli:
Thanks for reviewing @regehr

Noted for future PRs

PR #1355, opened 2026-09-03
feat(io): add `io` dialect for I/O operations. 

ArquintL:
This PR adds the dialect `Io`, which consists of the operations `send`, `recv`, and `rand` for sending a buffer to the network, receiving a buffer from a network, and sampling a random buffer, respectively.
The operations are intentionally left high-level and are supposed to be configured via attributes in the future. E.g., `send` and `recv` take a value of type `address` as their first argument without specifying what an address concretely is. More specifically, attributes like `tcp`, `udp`, `tls`, ... would then select a particular lowering, i.e., a concrete address format and implementations for sending and receiving messages.

Based on #1357


AlexanderViand:
> Based on #1357

I think we can do ops + types in one go (so close 1357 and mark this one ready-for-review), I'd just keep the semantics as a separate PR

Veir/Verifier/Basic.lean
```diff
@@ -181,6 +187,16 @@ def TypeAttr.verifyI1
...
 
+def TypeAttr.verifyI64
+    (ty : TypeAttr) (errMsg : String) : Except String PUnit :=
+  match ty.val with
+  | .integerType intType =>
+    if intType.bitwidth ≠ 64 then
+      throw errMsg
+    else
+      pure ()
+  | _ => throw errMsg
```
AlexanderViand, line 198: small nit:
```suggestion
  match ty.val with
  | .integerType { bitwidth := 64 } => pure ()
  | _ => throw errMsg
```

Veir/Dialects/IO/OpInfo.lean
```diff
@@ -0,0 +1,154 @@
...
+than `len`. A negative status is one of the error codes in `Io.Error`.
+-/
+@[opcodes]
+inductive Io where
+/--
+`io.send` writes the `len` bytes starting at `ptr` to the peer `dest`. Its
+operands are `(dest : !io.address, ptr : !llvm.ptr, len : integer)` and its
+result is the `i64` status. It reads the buffer, and its output is observable,
+so it is modelled as both reading and writing memory to keep it from being
+removed or reordered.
```
AlexanderViand, line 28: You might want to add an `io` flavoured test line to the CSE and DCE test files we already have so that something breaks if this is ever (accidentally) relaxed/changed.

Veir/GlobalOpInfo.lean
```diff
@@ -367,12 +341,10 @@ def OpCode.materializeConstant (opCode : OpCode) (value : RuntimeValue)
     | .llvm op => Llvm.materializeConstant op value type
     | .mod_arith op => Mod_Arith.materializeConstant op value type
     | .riscv op => Riscv.materializeConstant op value type
-    | .felt op => Felt.materializeConstant op value type
     -- Listed rather than folded into a catch-all so that adding a dialect
     -- fails to compile until it decides how, or whether, to materialize.
     | .riscv_cf _ | .riscv_stack _ | .rv64 _ | .cf _ | .builtin _
-    | .verif _
-    | .func _ | .datapath _ | .pdl _ | .test _ | .cir _ => none
+    | .func _ | .datapath _ | .pdl _ | .io _ | .test _ => none
```
AlexanderViand, line 347: There's also `testDialectsWithoutMaterializerDecline` which you might want to add `io` too.

AlexanderViand (approved):
LGTM, just a few nits (see above)

PR #1356, opened 2026-09-03
fix(parser): accept shorthand LLVM byte types

sueszli:
Part of #952.

MLIR introduced [`!llvm.byte<N>`](https://github.com/llvm/llvm-project/blob/e559d43e47e1264962831f936d1ea56d0fd3465d/mlir/include/mlir/Dialect/LLVMIR/LLVMTypes.td#L63-L87) in [llvm/llvm-project@e559d43](https://github.com/llvm/llvm-project/commit/e559d43e47e1264962831f936d1ea56d0fd3465d). Its [`PrettyLLVMType` dispatch](https://github.com/llvm/llvm-project/blob/e559d43e47e1264962831f936d1ea56d0fd3465d/mlir/lib/Dialect/LLVMIR/IR/LLVMTypeSyntax.cpp#L261-L278) accepts bare `byte<N>`, and [`!llvm.array`](https://github.com/llvm/llvm-project/blob/e559d43e47e1264962831f936d1ea56d0fd3465d/mlir/include/mlir/Dialect/LLVMIR/LLVMTypes.td#L45-L48) uses that directive for its element type. `byte` is omitted from the issue checklist.

This makes:

```mlir
!llvm.array<2 x byte<8>>
```

equivalent to `!llvm.array<2 x !llvm.byte<8>>`.


math-fehr (approved):
I missed this sorry, but all good, just fix the merge commits and we can merge it!

sueszli:
Awesome, yeah, i just rebased :)
Thanks for reviweing @math-fehr

PR #1359, opened 2026-09-03
feat: parser for floating point that reuses Lean internals

bollu:
This PR changes the data representation of FloatTypes, storing values as bit vectors, and introduces several floating-point types used in MLIR. It also adds a floating-point parser that reuses Lean's internal `UnpackedFloat` implementation to perform rounding: we mimic Lean's unpacked float implementation to create a correctly-rounded result. 

Based on the PR by @aduhtkjm (#1350)

Co-Authored By: Yue Huang <yh548@cam.ac.uk>


UnitTest/AttrParser.lean
```diff
@@ -68,13 +68,19 @@ def expectSuccessType (s : String) (expected : TypeAttr)
   testType s allowUnregisteredDialect = .ok expected
 
 /--
-  Test that parsing an attribute in the given string succeeds and matches the expected attribute.
+   Test that parsing an attribute in the given string succeeds and matches the expected attribute.
```
tobiasgrosser: Unrelated.

Veir/Parser/AttrParser.lean
```diff
@@ -206,19 +204,155 @@ def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
...
+      let pf := UnpackedFloat.packComponents type.toFormat
+        s (BitVec.ofNat _ biasedExponent) (BitVec.ofNat _ m)
+      pf.cast (by simp)
+
+    else
+      -- subnormal
+      let pf := UnpackedFloat.packComponents type.toFormat s 0#_ (BitVec.ofNat _ m)
+      pf.cast (by simp)
+
+
```
tobiasgrosser:
```suggestion

```

tobiasgrosser (approved):
LGTM. Let's get reviews from @math-fehr and @ineol.

Veir/IR/Attribute.lean
```diff
@@ -46,12 +46,74 @@ structure IntegerType where
...
 -/
 structure FloatType where
-  bitwidth : Nat
+  mantissa : Nat
+  exponent : Nat
+  bias: Nat
+  hasInf : Bool := true
+  hasNaN : Bool := true
+  hasNegZero : Bool := true
+  canonicalName : String
```
math-fehr: To me, it feels a lot that this should be moved to a structure to something like `Veir.Data.Float`, and then `FloatType` is just a singleton structure around it?
I feel like we are breaking a layer of abstraction here in general (see below).

Veir/IR/Attribute.lean
```diff
@@ -154,31 +216,25 @@ deriving Inhabited, Repr, DecidableEq, Hashable
 A floating point attribute storing a Lean `Float` value with an associated float type.
 -/
 structure FloatAttr where
-  value : Float
   type : FloatType
-deriving Inhabited, Repr
+  value : BitVec type.bitwidth
```
math-fehr: Same thing here, I feel like this `BitVec` should be hidden, the same way we hide `LLVM.Int` in `Data.Int`.
So we could have something like `structure FloatValue (t : FloatType)` that hides this abstraction and makes it dependent (which we could easily cast to different types as well). Making it dependent would just be so that it is less complex to make mistakes I feel.

Veir/Parser/AttrParser.lean
```diff
@@ -251,19 +249,153 @@ def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
...
+     let format : Model.Format := type.toFormat
+     let uf := UnpackedFloat.ofScientific format f.significand f.exponent
+     let uf := if f.negative then uf.neg else uf
+     packUnpackedFloatToFloatType type uf
+
+/--
+  Returns `true` if the numeric literal is in `0x`-prefixed hexadecimal form.
+-/
+private def isHexValue (value : ByteArray) : Bool :=
+  value.size > 2 && (value[1]! == 'x'.toUInt8 || value[1]! == 'X'.toUInt8)
```
math-fehr: Can you move this to `Parser/Parser.lean`?

Veir/Parser/AttrParser.lean
```diff
@@ -251,19 +249,153 @@ def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
...
+
+/--
+  Convert a numeric literal byte sequence to a `Nat`, accepting decimal and `0x`-prefixed
+  hexadecimal forms.
+-/
+private def numericValueToNat? (value : ByteArray) : Option Nat :=
```
math-fehr: Can you move this to `Parser/Parser.lean`?

Veir/Parser/AttrParser.lean
```diff
@@ -251,19 +249,153 @@ def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
   return some (StringAttr.mk bytes)
 
 /--
-Parse a float attribute, if present.
-Only `1.0 : f64` is supported; the value is stored as Lean's `Float`.
+Use Lean's builtin floating point packing algorithm,
+with a custom bias instead of the default IEEE-based exponent bias
+for floating point packing. This is an adaptation of `UnpackedFloat.pack`
+-/
+def packUnpackedFloatToFloatType
```
math-fehr: Should this be moved to a `Data.Float` as well? This doesn't seem to be related to the parser at all.

Veir/Parser/AttrParser.lean
```diff
@@ -251,19 +249,153 @@ def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
...
+using round-to-nearest, ties-to-even. Overflow saturates to infinity (or the
+largest finite value for types without infinity), and underflow produces a
+subnormal value or zero.
 -/
-def parseOptionalFloatAttr : AttrParserM (Option FloatAttr) := do
-  let some tok ← parseOptionalToken .floatLit | return none
-  let str := String.fromUTF8! (tok.slice.of (← getThe ParserState).input)
-  if str ≠ "1.0" then
-    throwAtCurrentPos s!"unsupported floating-point literal '{str}', only '1.0 : f64' is supported"
+def convertFPToBitvec (f : ParsedFloat) (type: Veir.FloatType) : BitVec (type.bitwidth) :=
```
math-fehr: Same here

Veir/Parser/AttrParser.lean
```diff
@@ -251,19 +249,153 @@ def parseOptionalStringAttr : AttrParserM (Option StringAttr) := do
...
+  For an integer type, `value` must be a (possibly negated) integer literal in decimal or
+  `0x`-prefixed hexadecimal form.
+  For a floating-point type, `value` may be a (possibly negated) decimal floating-point
+  literal, or a `0x`-prefixed hexadecimal integer literal interpreted as the raw IEEE-754
+  bit pattern of the type. A decimal integer literal is rejected for a floating-point type.
+
+  MLIR accepts the following floating point numbers: [-+]?[0-9]+[.][0-9]*([eE][-+]?[0-9]+)?
+  The regex is taken from AsmParser/Lexer.cpp in LLVM repository.
+-/
+def parseOptionalNumericAttr : AttrParserM (Option Attribute) := do
```
math-fehr, line 256: Can you move the parsing logic of this code to `Parser.lean`, which would return something like `Boolean | Integer | Float`?
And then we would call this function here and return `IntegerAttr | FloatAttr`?

math-fehr (commented):
Nice! I just had some design/architecture requests, but otherwise all good!

regehr:
I'll post a few tests that have arrived via LLM. I have no idea if these issues are severe enough that you must fix them before merging, perhaps not!

```mlir
// RUN: VEIR_ROUNDTRIP

// f8E4M3FNUZ has exponent bias 8 rather than the IEEE-standard 7, so its
// subnormal boundary is one binade below that of a standard 4-bit-exponent
// format: the smallest normal is 2^-7 (0x08) and the subnormal step is 2^-10
// (0x01).  Rounding, not just packing, has to use the type's own bias.
//
// Oracle: `mlir-opt --canonicalize` on `arith.bitcast` of the same literal
// (LLVM APFloat), cross-checked against ml_dtypes float8_e4m3fnuz and against
// the encoding by hand (normal: 2^(E-8) * 1.M; subnormal: M * 2^-10).
//
//   0.0078125     = 2^-7     -> 0x08  (smallest normal)
//   -0.0078125               -> 0x88
//   0.013         = 1.664*2^-7 -> 0x0d  (E=1, M=5)
//   0.0048828125  = 5*2^-10  -> 0x05  (subnormal)
//   0.0009765625  = 2^-10    -> 0x01  (smallest subnormal)

"builtin.module"() ({
    "test.test"() { a = 0.0078125 : f8E4M3FNUZ, b = -0.0078125 : f8E4M3FNUZ, c = 0.013 : f8E4M3FNUZ, d = 0.0048828125 : f8E4\
M3FNUZ, e = 0.0009765625 : f8E4M3FNUZ } : () -> ()
    // CHECK:     "test.test"() {"a" = 0x08 : f8E4M3FNUZ, "b" = 0x88 : f8E4M3FNUZ, "c" = 0x0d : f8E4M3FNUZ, "d" = 0x05 : f8E\
4M3FNUZ, "e" = 0x01 : f8E4M3FNUZ} : () -> ()
}) : () -> ()
```

regehr:
```mlir
// RUN: VEIR_ROUNDTRIP

// f8E4M3FN and f8E4M3FNUZ have no infinity, so overflow rounds to their own NaN
// encoding (0x7f for FN, 0x80 for FNUZ), and f8E4M3FNUZ has no negative zero, so
// -0.0 must round to +0.0.
//
// Oracle: `mlir-opt --canonicalize` on `arith.bitcast` of the same literal
// (LLVM APFloat, fltNonfiniteBehavior::NanOnly / fltNanEncoding::NegativeZero),
// cross-checked against ml_dtypes float8_e4m3fn / float8_e4m3fnuz.
//
//   f8E4M3FN:   max normal 448 = 0x7e, tie boundary 464, NaN = 0x7f
//     448.0  -> 0x7e   464.0  -> 0x7e (ties-to-even)   500.0 -> 0x7f   1000.0 -> 0x7f
//   f8E4M3FNUZ: max normal 240 = 0x7f, tie boundary 248, NaN = 0x80
//     240.0  -> 0x7f   244.0  -> 0x7f                  250.0 -> 0x80   1000.0 -> 0x80
//     -0.0   -> 0x00

"builtin.module"() ({
    "test.test"() { a = 448.0 : f8E4M3FN, b = 464.0 : f8E4M3FN, c = 500.0 : f8E4M3FN, d = 1000.0 : f8E4M3FN } : () -> ()
    // CHECK:     "test.test"() {"a" = 0x7e : f8E4M3FN, "b" = 0x7e : f8E4M3FN, "c" = 0x7f : f8E4M3FN, "d" = 0x7f : f8E4M3FN} : () -> ()
    "test.test"() { a = 240.0 : f8E4M3FNUZ, b = 244.0 : f8E4M3FNUZ, c = 250.0 : f8E4M3FNUZ, d = 1000.0 : f8E4M3FNUZ } : () -> ()
    // CHECK:     "test.test"() {"a" = 0x7f : f8E4M3FNUZ, "b" = 0x7f : f8E4M3FNUZ, "c" = 0x80 : f8E4M3FNUZ, "d" = 0x80 : f8E4M3FNUZ} : () -> ()
    "test.test"() { a = -0.0 : f8E4M3FNUZ, b = 0.0 : f8E4M3FNUZ } : () -> ()
    // CHECK:     "test.test"() {"a" = 0x00 : f8E4M3FNUZ, "b" = 0x00 : f8E4M3FNUZ} : () -> ()
}) : () -> ()
```

regehr:
this one I would definitely prefer for you to fix before merging this

```mlir
// RUN: not veir-opt %s 2>&1 | filecheck %s --strict-whitespace

// Verify that a 0x-prefixed bit pattern wider than the float type is rejected
// rather than silently truncated to the low bits of the type.
//
// Oracle: mlir-opt rejects this with
//   "error: hexadecimal float constant out of range for type"
// (AsmParser/AttributeParser.cpp), and there is no correct value to fall back
// on -- 0xdeadbeefdeadbeef is not an f32 bit pattern.  veir-opt today accepts
// it and silently keeps the low 32 bits, yielding 0xdeadbeef : f32.
//
// NOTE: the message text below is proposed, matching MLIR's wording; adjust it
// to whatever the fix emits.  The 17:24 location is the first byte of the
// literal, which is where the neighbouring float-attr error tests point.

"builtin.module"() ({
  "test.test"() {"v" = 0xdeadbeefdeadbeef : f32} : () -> ()
}) : () -> ()

// CHECK:float-attr-hex-out-of-range.mlir:17:24: error: hexadecimal float constant out of range for type
// CHECK-NEXT:  "test.test"() {"v" = 0xdeadbeefdeadbeef : f32} : () -> ()
// CHECK-NEXT:                       ^
```

regehr:
```mlir
// RUN: VEIR_ROUNDTRIP

// A decimal literal that overflows the type must round to a signed infinity, not
// to a NaN.  These values sit in the binade immediately above the largest finite
// value, where the biased exponent lands on the reserved all-ones pattern.
//
// Oracle: `mlir-opt --canonicalize` on `arith.bitcast` of the same literal
// (LLVM APFloat), cross-checked against ml_dtypes/numpy and Python `struct`.
//
//   3.5e38   : f32    -> 0x7f800000          (+inf; max normal is 3.4028235e38)
//   -3.5e38  : f32    -> 0xff800000          (-inf)
//   2.0e308  : f64    -> 0x7ff0000000000000  (+inf; max normal is 1.7976931e308)
//   70000.0  : f16    -> 0x7c00              (+inf; max normal is 65504)
//   4.0e38   : bf16   -> 0x7f80              (+inf)
//   1.0e5    : f8E5M2 -> 0x7c                (+inf; max normal is 57344)

"builtin.module"() ({
    "test.test"() { a = 3.5e38 : f32, b = -3.5e38 : f32, c = 2.0e308 : f64, d = 70000.0 : f16, e = 4.0e38 : bf16, f = 1.0e5 : f8E5M2 } : () -> ()
    // CHECK:     "test.test"() {"a" = 0x7f800000 : f32, "b" = 0xff800000 : f32, "c" = 0x7ff0000000000000 : f64, "d" = 0x7c00 : f16, "e" = 0x7f80 : bf16, "f" = 0x7c : f8E5M2} : () -> ()
}) : () -> ()
```

AdUhTkJm:
We don't have support for FN and UZ types yet. But I'll have a look at normal floating points.

Veir/Parser/AttrParser.lean
AdUhTkJm: Is this possible? We might need to know the exact float type before returning an appropriate floating point value.

bollu: We can't do this, because we need to know the type of a float before we start parsing a float.

bollu:
@regehr we ignore the types with no infinity and no neg-zero for now, since they add quite a few edge cases to the rounding :(

AdUhTkJm:
@regehr The last 2 bugs should have been fixed. The first 2 are not supported yet.

regehr (approved):
this LGTM!

PR #1360, opened 2026-09-03
feat(dataflow): add pessimistic entry state hook to sparse API

axelcool1234:


AlexanderViand (commented):
Not the right person to review this, so just leaving a note that we probably won't need widening for mod_arith ;) 

But entry_state we do need, so that part LGTM

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
```diff
@@ -26,6 +26,21 @@ for that result.
 abbrev VisitOperationFn (Domain : Type) :=
   OperationPtr -> Array Domain -> WfIRContext OpCode -> Array (Option Domain)
 
+/-- Compute the analysis specific pessimistic state when control flow loses precision. -/
+abbrev EntryStateFn (Domain : Type) :=
+  ValuePtr -> WfIRContext OpCode -> Domain
```
math-fehr: You are defining this here but you don't use it in `WideningPolicy`?
Maybe let's remove it?

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
```diff
@@ -26,6 +26,21 @@ for that result.
...
+/--
+Delayed widening policy for framework merge sites.
+
+After `budget` merge site changes, `upperBound` computes an abstract domain
+upper bound admitted by the target's type. Ordinary transfer function joins do not
+consume this budget and are never widened.
+-/
+structure WideningPolicy (Domain : Type) where
+  budget : Nat
+  upperBound : ValuePtr -> WfIRContext OpCode -> Domain
```
math-fehr: I think we should directly implement the "right" widening operator, where we have the Domain before and after, which can provide a better strategy than just "Top".
For instance, if we are in a loop and we increase some value, integer range analysis could be widened to `[0, infinity[` instead of `]-infinity, infinity[`.

regehr:
this seems sensible and good but is it not testable in this form?

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
math-fehr: Otherwise, if you keep this, `upperBound` should just be `Top` I think?

math-fehr (commented):
Just had an issue with how widening is defined, but otherwise looks good!

Veir/Analysis/DataFlow/SparseForwardDataFlowAnalysis.lean
axelcool1234: @math-fehr This allows for the definition of a better widening than just top

axelcool1234: Will remove widening for now and properly learn the theory before attempting again.

AlexanderViand (approved):
LGTM

PR #1361, opened 2026-09-03
SQLite scoreboard CI

zaz:
# SQLite3 scoreboard in CI

@regehr this is the CI version of the per-function sqlite board.

## What it does

Splits the sqlite3 amalgamation into one generic-MLIR chunk per function and per global initializer, runs each through `veir-opt`, and reports how many pass. Because every function is tested on its own, one blocker no longer hides the ones behind it, and the report ranks blockers by how many functions they affect.

Each function gets one of three statuses:

- **supported**: passes `veir-opt` strictly, every op and attribute registered
- **parsed**: passes only with `--allow-unregistered-dialect`, so some ops are carried opaquely
- **fail**

Two corpora are scored: `-O0` + sroa, which is vcc's own pipeline, and `-O3`.

## Where we stand

| corpus | supported | parsed | total |
|---|---:|---:|---:|
| -O0+sroa functions | 834 | 2096 | 2603 |
| -O0+sroa globals | 115 | 154 | 215 |
| -O3 functions | 34 | 893 | 1573 |
| -O3 globals | 93 | 131 | 185 |

The census (which distinct ops, types and attribute forms the corpus needs, each tested in isolation) currently says 38/58 ops, 564/565 types and 27/37 attrs are supported at `-O0`.

The single biggest blocker is one verifier bug: `llvm.mlir.addressof` cannot resolve quoted symbol names like `@".str.1218"`. It accounts for 394 of the 507 failing functions and 59 of the 61 failing globals. Registering `llvm.mlir.zero` is the equivalent for the supported tier; it appears in 1368 functions.

## What CI does

- Runs on PRs and on pushes to main, about 2.5 minutes with warm caches.
- Says nothing when nothing changed.
- When a PR changes the numbers, the workflow commits the updated baselines to the PR branch as `github-actions[bot]`, with the counts in the subject, and posts one sticky comment with the report. Fork PRs get the commands to run instead.
- Improvements only: green. Any regression: red, but the regressed baselines are still committed so the diff shows exactly which functions changed.
- A corpus that no longer matches the baselines (toolchain drift) is reported as its own outcome, never as a regression.

## Running it locally

```
Tools/sqlite-scoreboard            # full run; downloads the amalgamation, caches chunks and results
Tools/sqlite-scoreboard --check-baseline
Tools/mlir-feature-census --check-baseline
```

Needs clang 19 and MLIR 22 (`mlir-translate` from LLVM 21 or later emits the integer `overflowFlags` encoding VeIR expects). Every failing chunk is kept on disk as a one-file reproducer, and `--only NAME` reruns a single function.

## Caveats for maintainers

- Merging a red PR bakes the regression into main's baseline. That is deliberate: the diff, not the check, is the record.
- The baselines are Linux x86_64 specific; sqlite compiles different code on macOS.

Rehearsed end to end on a fork: https://github.com/zaz/veir/pull/1


tobiasgrosser:
@zaz, to be able to move fast I moved most of this PR to https://github.com/opencompl/veir-sqlite.

tobiasgrosser:
It does not have the cool CI integration, but there is already a leadership board https://github.com/opencompl/veir-sqlite/blob/main/LEADERSHIP.md, which allows us to triage the open issues. I gave you admin access to the repo, such that you can advance this independently. Let's close this and move to the veir-sqlite repo! Happy to work with you on adding better CI or web integration. 🙏

zaz:
@tobiasgrosser I have that page set up to pull from veir and re-grade every 10 minutes. Does that work for you?

PR #1362, opened 2026-09-03
feat: support dialect mnemonics that differ from lean declaration name

AlexanderViand:
Allow opcode dialect declarations to override their MLIR mnemonic with `@[dialect_name "..."]`, while retaining the lowercased Lean declaration name as the default.


ineol:
LGTM :)

ineol (approved):

PR #1363, opened 2026-09-03
feat: support namespace-qualified dialect declarations

AlexanderViand:
Preserve namespace-qualified Lean names when generating dialect helpers and global opcodes. This lets dialects such as Veir.LLZK.Function use natural Lean names while keeping their MLIR mnemonics independent.


AlexanderViand (commented):
Reviewed by @ineol on my laptop xD

PR #1365, opened 2026-09-03
perf: speed up critical path

hargoniX:
This PR speeds up two files on the critical build path, saving about 10s of rebuild time:
1. Make an explicit case split to avoid grind thinking too much about both cases
2. Move up several definitions into one contiguous block (from interspersing with theorems) to enable better in file parallelism.


math-fehr (approved):
❤️

PR #1366, opened 2026-09-03
feat: add `VectorType` type

math-fehr:
This PR adds the `VectorType` type, which contains a single element type and an array of positive dimensions. All element types are currently allowed, though an interface will be required in the future, like in MLIR. Similarly, it is not yet enforced that dimensions are non-null.

This change also add support for printing and parsing.


luisacicolini (approved):
lgtm :)

PR #1371, opened 2026-09-04
feat(LLVM): add constant range metadata

tobiasgrosser:
`#llvm.constant_range<iN, lo, hi>` is the half-open range a value is known to lie in. It reaches VeIR as `llvm.range` inside the `arg_attrs` and `res_attrs` dictionaries of `llvm.func` and `llvm.call`. Like the other LLVM metadata VeIR registers but store its body as an uninterpreted string.

Collapse the ten attributes that share this parse into a table. This cuts the build time from 45s to 6s, and makes the next such attribute a one-line entry.

### DIFF
--- a/Test/LLVM/constant_range.mlir
+++ b/Test/LLVM/constant_range.mlir
@@ -1,0 +1,24 @@
+// RUN: VEIR_ROUNDTRIP
+// RUN: MLIR_ROUNDTRIP
+//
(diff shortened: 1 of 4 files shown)


mafeguimaraes (approved):
LGTM

luisacicolini (approved):

PR #1374, opened 2026-09-04
feat(dataflow): known bits analysis

axelcool1234:


regehr:
one thing you should add here is an invariant on the known bits struct that no bit is known both zero and one.

another thing you should do is to think carefully about top. here you have an explicit representation for it, but there's another representation, which is that no bits are known. it's a mistake to be able to represent the same value two different ways. you should either remove the explicit top or else use another invariant on the struct to enforce that at least one bit is always known.

PR #1375, opened 2026-09-04
feat(Verif): Add support for label attr

mafeguimaraes:
This patch adds support for the `label` attribute from `verif.assume` and `verif.assert`.

### DIFF
--- a/Veir/Dialects/Verif/OpInfo.lean
+++ b/Veir/Dialects/Verif/OpInfo.lean
@@ -18,9 +18,9 @@ deriving Inhabited, Repr, Hashable, DecidableEq
 
 @[expose, properties_of]
 def Verif.propertiesOf (op : Verif) : Type :=
-match op with
-| .assume => VerifAssumeAssertProperties
-| .assert => VerifAssumeAssertProperties
+  match op with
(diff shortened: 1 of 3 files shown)


georgerennie:
Looks good, thanks!

Veir/Dialects/Verif/OpInfo.lean
```diff
@@ -16,19 +17,27 @@ inductive Verif where
 deriving Inhabited, Repr, Hashable, DecidableEq
 
 @[expose, properties_of]
-def Verif.propertiesOf (_op : Verif) : Type :=
-  Unit
+def Verif.propertiesOf (op : Verif) : Type :=
+match op with
+| .assume => VerifAssumeAssertProperties
+| .assert => VerifAssumeAssertProperties
```
georgerennie: Small nit actually

```suggestion
def Verif.propertiesOf (op : Verif) : Type :=
  match op with
  | .assume => VerifAssumeAssertProperties
  | .assert => VerifAssumeAssertProperties
```

mafeguimaraes: Fixed it, thanks!

PR #1378, opened 2026-09-04
feat(io): Tweaks I/O Dialect

ArquintL:
Adapts `recv` to return the sender's address and adds `self` to refer to one's address

### DIFF
--- a/Test/IO/recv_non_address_result.mlir
+++ b/Test/IO/recv_non_address_result.mlir
@@ -3,8 +3,9 @@
 "builtin.module"() ({
   "func.func"() <{sym_name = "main", function_type = () -> ()}> ({
     %len = "llvm.mlir.constant"() <{value = 4 : i64}> : () -> i64
-    %n, %from = "io.recv"(%len, %len) : (i64, i64) -> (i64, !io.address)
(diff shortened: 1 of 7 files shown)


Test/IO/recv_non_pointer_buffer.mlir
```diff
@@ -2,10 +2,9 @@
 
 "builtin.module"() ({
   "func.func"() <{sym_name = "main", function_type = () -> ()}> ({
-    %peer = "test.test"() : () -> !io.address
     %len = "llvm.mlir.constant"() <{value = 4 : i64}> : () -> i64
-    %n = "io.recv"(%peer, %len, %len) : (!io.address, i64, i64) -> i64
-    // CHECK: io.recv: Expected operand 1 to have !llvm.ptr type
+    %n, %from = "io.recv"(%len, %len) : (i64, i64) -> (i64, !io.address)
```
AlexanderViand, line 6: I probably should have asked this on the original PR, but why does this pass `%len` twice?

ArquintL: To test that this is correctly rejected as `i64` is not an `!llvm.ptr` type

AlexanderViand (approved):
👍

PR #1379, opened 2026-09-04
feat(io): Interpreter for the I/O Dialect

ArquintL:
Extends the Interpreter's state by a source of entropy and a network abstraction.
The source of entropy is supposed to be initialized before interpretation starts and allows operations like random number generation to consume random bytes from this source. As soon as no entropy is left, the behavior is UB.
The network abstraction holds the in-flight messages as a FIFO queue and `recv` consumes (& drops) the first message in the queue for the given receiver.

PR #1384, opened 2026-09-04
feat(io): CTree Interpreter for I/O

ArquintL:
Adds an interpreter for the I/O dialect using CTrees to capture (network) send & receive effects and random number generation as observable effects.
This new, CTree-based interpreter (`Io.interpretOpCTree`) operates of course on `MemoryState`. However, it does not touch the `EntropyState` and `NetworkState` parts of `MemoryState` (except for getting the interpreter's own address; cf. #1379). Thus, there are two ways forward: (1) remove these unused state components as soon as we delete the existing non-CTree-based interpreters or (2) use different state representations for CTree-based and non-CTree-based interpreters.


Veir/Dialects/IO/Interpreter.lean
```diff
@@ -0,0 +1,82 @@
...
+  match opType with
+  | .self => do
+    return (#[.ioAddr mem.selfAddress], mem, none)
+  | .send => do
+    let [.ioAddr dest, .addr ptr, .int _ len] := operands.toList | ub
+    let .val len := len | ub
+    let lenNat := len.toNat
+    if lenNat ≥ UInt64.size then ub
+    let len : UInt64 := UInt64.ofNat lenNat
+    if ← mem.hasPoisonCTree ptr len then ub
```
nchappe, line 54: One of my recently-merged CTree PRs includes a MonadLift instance from the Interp monad to CTree. You should be able to write `monadLift $ mem.hasPoison ptr len` here instead of duplicating the hasPoison definition, and same for memory load/store/etc. You can look at the memory operations in the CTree-based LLVM dialect interpreter for more specific examples.

Veir/Dialects/IO/Interpreter.lean
```diff
@@ -0,0 +1,82 @@
...
+    let buf ← mem.loadCTree ptr len
```
nchappe, line 55: A note about CTrees and memory accessses: in the medium term, we will want to have memory access (load/store) effects for CTrees, so you will be able to introduce such a load node here instead of accessing the actual memory state directly. But we don't have this level of modularity for now.

Veir/Interpreter/Memory.lean
```diff
@@ -1,10 +1,12 @@
 module
 
 public import Veir.RuntimeValue
+public import Veir.Interpreter.CTree
```
nchappe, line 4: All the changes to this file should become unnecessary if you use monadLift.

nchappe (commented):
The network state component you added in the other PR will be useful if you want to interpret the CTree IoE effects later.
I would move these components from MemoryState to InterpreterState (as I suggested in the other PR). InterpreterState (and the whole old interpreter) will become obsolete once we have a complete CTree-based stack, because we will probably have one separate (interpreter+state) pair for each kind of effect. So the problem of unused components will solve by itself.

PR #1389, opened 2026-09-04
feat(llzk): more local invariants checks (array), refactor common checks

Maschmalow:
- added common LLZK type checks in `Verifier.Basic` and refactor existing usage
- added array operations local invariants checks

### DIFF
--- a/Veir/Dialects/LLZK/Array/OpInfo.lean
+++ b/Veir/Dialects/LLZK/Array/OpInfo.lean
@@ -91,12 +91,62 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
   if op.getNumSuccessors ctx.raw opIn ≠ 0 then
     throw s!"{instrName}: Expected 0 successors"
   match opType with
-  | .new => requireResults 1
-  | .read | .extract => do requireAtLeastOperands 1; requireResults 1
-  | .write | .insert => do requireAtLeastOperands 2; requireResults 0
+  | .new => do
+    requireResults 1
+    if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayType then
+      throw s!"{instrName}: Expected valid LLZK array.type return type"
+    -- TODO: check operands types/count
+  | .read => do
+    requireAtLeastOperands 1
+    requireResults 1
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to have a valid LLZK array type"
+      else if !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+    if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayElemType then
+      throw s!"{instrName}: Expected return to have a valid LLZK array element type"
+  | .extract => do
+    requireAtLeastOperands 1
+    requireResults 1
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to have a valid LLZK array type"
+      else if !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+    if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayType then
+      throw s!"{instrName}: Expected return to have a valid LLZK array type"
+  | .write => do
+    requireAtLeastOperands 2
+    requireResults 0
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to be a valid LLZK array"
+      else if idx ≠ (op.getNumOperands! ctx.raw)-1 && !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+      else if !operType.val.isLLZKArrayElemType then
+        throw s!"{instrName}: Expected operand {idx} to have a valid LLZK array element type"
+  | .insert => do
+    requireAtLeastOperands 2
+    requireResults 0
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to be a valid LLZK array"
+      else if idx ≠ (op.getNumOperands! ctx.raw)-1 && !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+      else if !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand {idx} to have a valid LLZK array type"
   | .len => do
+    requireResults 1
     if op.getNumOperands ctx.raw opIn ≠ 2 then
       throw s!"{instrName}: Expected 2 operands (the array and the dimension)"
+    if !(op.getOperandTypes! ctx.raw)[0]!.val.isLLZKArrayType then
+      throw s!"{instrName}: Expected operand 0 to be a valid LLZK array"
+    if !((op.getOperandTypes! ctx.raw)[1]!.val matches .indexType _) then
+      throw s!"{instrName}: Expected operand 1 to have an index type"
+    if !((op.getResultTypes! ctx.raw)[0]!.val matches .indexType _) then
+      throw s!"{instrName}: Expected return to have index type"
+
     requireResults 1
 
 instance : HasOpInfo LLZK.Array where
--- a/Veir/Dialects/LLZK/Array/OpInfo.lean
+++ b/Veir/Dialects/LLZK/Array/OpInfo.lean
@@ -102,7 +102,7 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
     for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
       if idx = 0 && !operType.val.isLLZKArrayType then
         throw s!"{instrName}: Expected operand 0 to have a valid LLZK array type"
-      else if !(operType.val matches .indexType _) then
+      if idx ≠ 0 && !(operType.val matches .indexType _) then
         throw s!"{instrName}: Expected operand {idx} to have an index type"
     if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayElemType then
       throw s!"{instrName}: Expected return to have a valid LLZK array element type"
@@ -112,7 +112,7 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
     for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
       if idx = 0 && !operType.val.isLLZKArrayType then
         throw s!"{instrName}: Expected operand 0 to have a valid LLZK array type"
-      else if !(operType.val matches .indexType _) then
+      if idx ≠ 0 && !(operType.val matches .indexType _) then
         throw s!"{instrName}: Expected operand {idx} to have an index type"
     if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayType then
       throw s!"{instrName}: Expected return to have a valid LLZK array type"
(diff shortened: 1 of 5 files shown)


tobiasgrosser (approved):
This LGTM. @AlexanderViand, what do you think?

Veir/Verifier/Basic.lean
```diff
@@ -38,6 +38,25 @@ def Attribute.isKnownNonZero (attr : Attribute) : Bool :=
   | .floatAttr fltAttr => fltAttr.value != 0.0
   | _ => false
 
+/-- Verify that a type belongs to the subset of LLZK types currently represented by VeIR.
+  TypeVarType, PodType, and NoneType are currently missing.
+  https://github.com/project-llzk/llzk-lib/blob/265d68f678ab15018e3f6253b85557fbaeac9c0d/lib/Util/TypeHelper.cpp#L482-L511 -/
+def Attribute.isLLZKType (attr : Attribute)  : Bool :=
+  match attr with
+  | .integerType intType => decide (intType.bitwidth = 1)
+  | .indexType _ | .feltType _ | .structType _ | .stringType _ | .arrayType _ => true
```
AlexanderViand: `isLLZKType` returns true for every `.arrayType`, but llzk's `isValidTypeImpl` recurses in the array case (`caseArray` -> `isValidArrayTypeImpl` -> `isValidArrayElemTypeImpl`), so a nested array is not a valid LLZK type there. That makes `!array.type<2 x !array.type<3 x !felt.type>>` legal as a `struct.member` type / `struct.readm` result / `struct.writem` value here and illegal in llzk (`'array.type' element type cannot be 'array.type'`).

Awkwardly `isLLZKArrayType` ten lines down does check the element, so the two predicates disagree about the same type.

```lean
| .arrayType arrT => !(arrT.elementType matches .arrayType _) && arrT.elementType.isLLZKType
```

You may need a `termination_by`, since the recursion goes through the nested `elementType`.

Veir/Verifier/Basic.lean
```diff
@@ -38,6 +38,25 @@ def Attribute.isKnownNonZero (attr : Attribute) : Bool :=
...
+  | _ => false
+
+def Attribute.isLLZKArrayElemType (attr : Attribute)  : Bool :=
+  match attr with
+  | .arrayType _ => false
+  | _ => attr.isLLZKType
+
+def Attribute.isLLZKArrayType (attr : Attribute) : Bool :=
+  match attr with
+  | .arrayType arrT  => arrT.elementType.isLLZKArrayElemType
```
AlexanderViand: This skips `areValidArrayDimSizes`. llzk requires at least one dimension, and `LLZK.ArrayType`'s `Inhabited` instance is literally `{ dims := #[] }` (`Veir/IR/Attribute.lean:749`).

Symbolic and affine dims aren't representable in Veir, so `arrT.dims.size > 0 &&` is the whole of that check here.

Veir/Verifier/Basic.lean
```diff
@@ -38,6 +38,25 @@ def Attribute.isKnownNonZero (attr : Attribute) : Bool :=
   | .floatAttr fltAttr => fltAttr.value != 0.0
   | _ => false
 
+/-- Verify that a type belongs to the subset of LLZK types currently represented by VeIR.
+  TypeVarType, PodType, and NoneType are currently missing.
+  https://github.com/project-llzk/llzk-lib/blob/265d68f678ab15018e3f6253b85557fbaeac9c0d/lib/Util/TypeHelper.cpp#L482-L511 -/
+def Attribute.isLLZKType (attr : Attribute)  : Bool :=
```
AlexanderViand: Two small things.

`Veir/Verifier/Basic.lean` is the dialect-agnostic verifier toolbox and now holds three LLZK predicates. There's no shared `Veir/Dialects/LLZK` module today; a `Types.lean` there would be a better home, and would give `isLLZKSubArrayOrElementType` somewhere to live too.

`isLLZKArrayElemType` and `isLLZKArrayType` have no docstrings. This one links the llzk source, those two should link `isValidArrayElemType` and `isValidArrayType`. The list of what's missing (TypeVarType, PodType, NoneType) should mention array elements as well, or just fix the recursion and drop the note.

Also a double space in `(attr : Attribute)  : Bool` here and on `isLLZKArrayElemType`, and in `| .arrayType arrT  =>` below.

Veir/Dialects/LLZK/Constrain/OpInfo.lean
```diff
@@ -65,11 +65,11 @@ instance : IsOpCode LLZK.Constrain where
...
-private def Attribute.isSupportedLLZKConstrainEqType (type : Attribute) : Bool :=
-  match type with
-  | .integerType intType => intType.bitwidth = 1
-  | .indexType _ | .feltType _ | .arrayType _ => true
-  | _ => false
+
+private def Attribute.isLLZKEmitEqType (attr : Attribute) : Bool :=
+  match attr with
+  | .structType _ | .stringType _ => false
+  | _ => attr.isLLZKType
```
AlexanderViand: Same shape of problem as `isLLZKType`. llzk builds this as `AllowedTypes().noString().noStruct().isValidTypeImpl(...)` and those flags reach the array element, so `!array.type<3 x !string.type>` is rejected there and accepted here.

```lean
| .arrayType arrT => arrT.elementType.isLLZKEmitEqType
```

The refactor didn't introduce this, the old `isSupportedLLZKConstrainEqType` accepted exactly the same set (I diffed them), but you're rewriting the function anyway.

Veir/Dialects/LLZK/Constrain/OpInfo.lean
```diff
@@ -65,11 +65,11 @@ instance : IsOpCode LLZK.Constrain where
...
 
 /-- Whether `candidate` is an element or trailing-dimensional subarray of `arrayType`. -/
 private def isLLZKSubArrayOrElementType
```
AlexanderViand: This already covers the extract/insert half of the array shape checking that `Array/OpInfo.lean` is missing. Worth moving somewhere shared and dropping the `private` rather than writing it a second time.

Veir/Dialects/LLZK/Array/OpInfo.lean
```diff
@@ -91,12 +91,62 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
+  | .read => do
+    requireAtLeastOperands 1
+    requireResults 1
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to have a valid LLZK array type"
+      if idx ≠ 0 && !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+    if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayElemType then
+      throw s!"{instrName}: Expected return to have a valid LLZK array element type"
```
AlexanderViand: Two checks missing here that llzk does locally.

The result type isn't compared against the array's element type, only against "is some valid element type", and `index` qualifies for a felt array. llzk enforces it with `ArrayTypeElemsUnifyWithResultCustomInfer` on read and `ArrayElemTypeUnifyWith` on write; `array.read %a[%i] : !array.type<2 x !felt.type>, index` gets `result type matches with arr_ref element type`.

Index count isn't compared against rank. llzk's `verifyScalarArrayAccess` wants `numIndices == rank` for read and write; `array.read %a[%i]` on a rank-2 array gets `insufficient indexed dimensions: expected 2 but found 1`. `dims` is `Array Int` here, always concrete, so both are local and cheap.

Same applies to `.write` below.

Veir/Dialects/LLZK/Array/OpInfo.lean
```diff
@@ -91,12 +91,62 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
+  | .insert => do
+    requireAtLeastOperands 2
+    requireResults 0
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to be a valid LLZK array"
+      if idx ≠ 0 && idx ≠ (op.getNumOperands! ctx.raw)-1 && !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+      if idx = (op.getNumOperands! ctx.raw)-1 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand {idx} to have a valid LLZK array type"
```
AlexanderViand: `array.insert` of `<7 x felt>` into `<2,3 x felt>` passes here; llzk gives `cannot unify array dimensions [3] with [7]`. Its verifier wants `numIndices + rhsRank == lhsRank` plus `verifySubArrayType`, and `extract` wants `numIndices < rank` with the result equal to `getSelectionType(numIndices)`.

`isLLZKSubArrayOrElementType` in `Constrain/OpInfo.lean` already does this.

Veir/Dialects/LLZK/Array/OpInfo.lean
```diff
@@ -91,12 +91,62 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
+  | .extract => do
+    requireAtLeastOperands 1
+    requireResults 1
+    for (operType,idx) in (op.getOperandTypes! ctx.raw).zipIdx do
+      if idx = 0 && !operType.val.isLLZKArrayType then
+        throw s!"{instrName}: Expected operand 0 to have a valid LLZK array type"
+      if idx ≠ 0 && !(operType.val matches .indexType _) then
+        throw s!"{instrName}: Expected operand {idx} to have an index type"
+    if !(op.getResultTypes! ctx.raw)[0]!.val.isLLZKArrayType then
+      throw s!"{instrName}: Expected return to have a valid LLZK array type"
```
AlexanderViand: `.read` and `.extract` have identical bodies except for the last line, and `.write`/`.insert` likewise. That's four copies of one `zipIdx` loop; a helper taking the result check as a parameter would collapse them.

Inside the `.write` and `.insert` copies, `op.getNumOperands! ctx.raw` is recomputed twice per iteration when the array is already in hand and has a `.size`.

Veir/Dialects/LLZK/Array/OpInfo.lean
```diff
@@ -91,12 +91,62 @@ def LLZK.Array.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
     if op.getNumOperands ctx.raw opIn ≠ 2 then
       throw s!"{instrName}: Expected 2 operands (the array and the dimension)"
+    if !(op.getOperandTypes! ctx.raw)[0]!.val.isLLZKArrayType then
+      throw s!"{instrName}: Expected operand 0 to be a valid LLZK array"
+    if !((op.getOperandTypes! ctx.raw)[1]!.val matches .indexType _) then
+      throw s!"{instrName}: Expected operand 1 to have an index type"
+    if !((op.getResultTypes! ctx.raw)[0]!.val matches .indexType _) then
+      throw s!"{instrName}: Expected return to have index type"
+
     requireResults 1
```
AlexanderViand: `requireResults 1` again, it's already at the top of the `.len` branch.

Veir/Dialects/LLZK/Struct/Properties.lean
```diff
@@ -48,12 +41,10 @@ def StructMemberProperties.fromAttrDict (opName : String)
     | throw s!"{opName}: expected 'sym_name' to be a string attribute, got {symAttr}"
   let typeAttr ← match attrDict["type".toUTF8]? with
     | some attr =>
-      if _ : attr.isType = true then
-        attr.asType.verifySupportedLLZKType
-          s!"{opName}: expected 'type' to be a supported LLZK type"
+      if _ : attr.isType = true && attr.isLLZKType then
```
AlexanderViand: This parses as `decide (attr.isType = true) && attr.isLLZKType` (checked with `#print`). It compiles, but `asType`'s `by grind` now has to pull its hypothesis out of a conjunction, and the separate "expected 'type' to be a type attribute" message is gone, so a non-type attribute now reports the LLZK-type error.

Nesting the ifs keeps both diagnostics and the direct proof.

AlexanderViand (approved):
I asked Claude to check this against  LLZK (`ba936ff`) and it looks good, `veir-opt` seems to be a bit more permissive than `llzk-opt`, but that seems OK for now.

It would be nice to add some tests, but probably not super essential given that the source of truth is LLZK anyway.

AlexanderViand (approved):
I asked Claude to check this against  LLZK (`ba936ff`) and it looks good, `veir-opt` seems to be a bit more permissive than `llzk-opt`, but that seems OK for now.

It would be nice to add some tests, but probably not super essential given that the source of truth is LLZK anyway.

PR #1391, opened 2026-09-04
feat(dataflow): Improve transfer function signature

axelcool1234:
Bottom is implicit in the product lattice and is not actually set in the lattice. That means the transfer function doesn't need to return `none` for the same purpose, which improves its function signature and makes proving things about them so much easier. It also frees the burden from transfer function developers from having to reason about this.

### DIFF
--- a/UnitTest/DataFlowFramework/EntryState.lean
+++ b/UnitTest/DataFlowFramework/EntryState.lean
@@ -29,7 +29,7 @@ private instance : SparseFactSpec .test TestDomain where
   payloadEq := rfl
 
 /--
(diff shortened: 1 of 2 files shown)


regehr (approved):
LGTM

axelcool1234:
@math-fehr John gave an LGTM, does this look good to you? It's an additional hook a user can provide when creating an analysis. I think later I'll need a better way to provide hooks as there'll need to be a widening one as well. But for now this is probably fine.

math-fehr (approved):
Yes, that makes sense to me!

PR #1392, opened 2026-09-04
feat(LLVM): add mem[set|cpy|move] to parser

tobiasgrosser:
We add store the properties `isVolatile`, `arg_attrs`, `res_attrs`, `access_groups`, `alias_scopes`, `noalias_scops`, `tbaa`, and parse and ignore `op_bundle_sizes` and `op_bundle_tags` to match MLIR.

In contrast to MLIR, we verify the length of `arg_attrs` and `res_attrs`.

In `sqlite`, we only observe `isVolatile` (always false), `arg_attrs`, and `tbaa`.


tobiasgrosser:
I cleaned this up. @math-fehr @regehr, could you have a look?

Veir/Dialects/LLVM/Properties.lean
```diff
@@ -607,6 +607,42 @@ def LLVMSwitchProperties.caseValues? (props : LLVMSwitchProperties) : Option (Ar
     values := values.push value
   return values
 
+/--
+  Properties of the memory intrinsics `memset`, `memcpy`, and `memmove`.
+-/
+structure LLVMMemIntrinsicProperties where
+  isVolatile : Bool
+  arg_attrs : Option ArrayAttr
+  tbaa : Option ArrayAttr
```
math-fehr, line 630: Is there a reason you are adding only these properties, and not all the others? For instance, why `arg_attrs` but not `res_attrs`?

math-fehr (approved):
Looks good to me, there is just some properties that are missing, I'm not sure the reason to only implement these ones?
But I think at some point we should just try to see what properties are missing from the entire dialect, it's likely other operations are missing properties.

Test/LLVM/mem_intrinsics.mlir
```diff
@@ -0,0 +1,22 @@
...
+  ^bb0(%dst: !llvm.ptr, %src: !llvm.ptr, %byte: i8, %len: i64):
+    "llvm.intr.memset"(%dst, %byte, %len) <{arg_attrs = [{llvm.align = 8 : i64, llvm.nonnull, llvm.noundef}, {}, {}], isVolatile = false}> : (!llvm.ptr, i8, i64) -> ()
+    "llvm.intr.memcpy"(%dst, %src, %len) <{isVolatile = true}> : (!llvm.ptr, !llvm.ptr, i64) -> ()
+    "llvm.intr.memmove"(%dst, %src, %len) <{isVolatile = false}> : (!llvm.ptr, !llvm.ptr, i64) -> ()
+    "llvm.intr.memcpy"(%dst, %src, %len) <{isVolatile = false, access_groups = [], alias_scopes = [], noalias_scopes = [], tbaa = [#llvm.tbaa_tag<base_type = <id = "int", members = {<#llvm.tbaa_root<id = "Simple C/C++ TBAA">, 0>}>, access_type = <id = "int", members = {<#llvm.tbaa_root<id = "Simple C/C++ TBAA">, 0>}>, offset = 0>]}> : (!llvm.ptr, !llvm.ptr, i64) -> ()
+    // `res_attrs` is not meaningful given that memset has no results, but MLIR accepts it,
+    // so we round-trip it rather than refuse it.
+    "llvm.intr.memset"(%dst, %byte, %len) <{isVolatile = false, res_attrs = [{llvm.noundef}]}> : (!llvm.ptr, i8, i64) -> ()
```
math-fehr: This seems to me like a bug in the LLVM dialect? I'm not sure if we should rely on it at all, and instead not test that corner case (and report it as an issue in MLIR)

Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -577,6 +597,24 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
+        | throw s!"Expected operand {i} to have !llvm.ptr type"
+    if opType = .intr__memset then
+      let byteType := (op.getOperand! ctx.raw 1).getType! ctx.raw
+      let .integerType byteType := byteType.val
+        | throw "operand #1 must be 8-bit signless integer"
+      if byteType.bitwidth ≠ 8 then
+        throw s!"operand #1 must be 8-bit signless integer, but got i{byteType.bitwidth}"
+    let lengthType := (op.getOperand! ctx.raw 2).getType! ctx.raw
+    let .integerType _ := lengthType.val
+      | throw "Expected operand 2 to have integer type"
```
math-fehr, line 626: I'm surprised we don't verify that `res_attrs` or `arg_attrs` have the same size as the results or operands? If this is ommited by the LLVM dialect, it should be a bug I believe.

math-fehr (approved):
I think we should add a verifier for res_attrs and arg_attrs, but otherwise all good to me!

tobiasgrosser:
> Looks good to me, there is just some properties that are missing, I'm not sure the reason to only implement these ones?

I currently aim to make `sqlite` compile and add the ones needed there. I now checked for others implemented by MLIR and added them with corresponding tests.

Veir/Dialects/LLVM/OpInfo.lean
tobiasgrosser: I added the verification.

Test/LLVM/mem_intrinsics.mlir
tobiasgrosser: We now check that res_attrs lenght matches the number of results and drop this test case.

Issue #947, opened 2026-06-29
sqlite parent issue


sishtiaq:
Goal: compile and run sqlite3. 
All sqlite issues: https://github.com/opencompl/veir/issues?q=is%3Aissue%20state%3Aopen%20label%3Asqlite. Feel free to pick them up! 

sqlite3 is defined to be the 1-file sqlite amalgamation. 
`curl -O https://sqlite.org/2026/sqlite-amalgamation-3530300.zip` to get it. 

The entire "compile and run" will likely be done in stages: 
1. parse; 
2. emit;
3. interpret;
4. lower to risc-v.

Parsing status: You can get the latest parsing error by running `Tools/vcc /path/to/sqlite3.c`.

PR #1395, opened 2026-09-04
feat(LLVM): add llvm.intr.assume

zaz:
Part of #947.

The condition is followed by the operands of the op's operand bundles, so the properties model `op_bundle_sizes` and the optional `op_bundle_tags` explicitly instead of an `extra` dictionary, and the verifier checks what MLIR's does: an i1 condition, non-negative i32 sizes that account for every operand after the condition, and one string tag per bundle.

@tobiasgrosser


tobiasgrosser (approved):
Very cool.

PR #1396, opened 2026-09-04
feat(LLVM): add llvm.mlir.undef syntax

tobiasgrosser:
This adds syntax support for `llvm.mlir.undef`.

When adding interpreter support, we need to be careful to ensure we only add as much undef support as LLVM actually needs. Undef is deprecated with the exception of memory loads.


naveen-seth (approved):
LGTM

PR #1397, opened 2026-09-05
build(nix): update mlir flags to pass PDL tests

GZGavinZhao:
*puts on his nixpkgs maintainer hat* Great to see a nix setup!

The MLIR provided by nixpkgs doesn't build the `test` dialect used by PDL tests, so we have to build MLIR from source with an extra `MLIR_INCLUDE_TESTS=ON` CMake flag. This does mean that people using the Nix flake need to build MLIR from source, but after discussing with @axelcool1234 we think that this is acceptable and aligns with the spirit of Nix.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,6 +12,7 @@
         "x86_64-linux"
       ];
       forAllSystems = nixpkgs.lib.genAttrs systems;
+      llvmPackages = pkgs: pkgs.llvmPackages_23;
 
       developmentPackages = pkgs: with pkgs; [
         bash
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -1,76 +1,73 @@
 {
   description = "Verified Intermediate Representation";
 
-  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-26.05";
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
```
axelcool1234: Having looked at a lot of flakes, I've noticed there's a trend away from flake-utils simply due to how easy it is to do yourself. In my opinion it's an unneeded dependency.

flake.nix
```diff
@@ -1,76 +1,73 @@
...
+        };
 
-      apps = forAllSystems (system:
-        let
-          pkgs = nixpkgs.legacyPackages.${system};
-        in
-        {
-          default = makeApp pkgs null;
-          build = makeApp pkgs "build";
-          tests = makeApp pkgs "tests";
```
axelcool1234, line 73: We should probably still have a way to execute the make commands from nix I think

axelcool1234 (approved):
LGTM, good catch on the MLIR thing.

PR #1398, opened 2026-09-05
build(nix): expose Lean executable apps

GZGavinZhao:
Expose the `[[lean_exe]]` targets in Nix flake as well, such as `.#veir-opt` and `.#veir-interpret`. This way developers can run `nix run .#veir-opt` and `nix run .#veir-interpret` instead of `nix develop --command lake exe veir-opt`.


tobiasgrosser (commented):
Very cool. Let's see what @axelcool1234 says.

axelcool1234 (approved):
LGTM. This flake is going to be useful. So glad VeIR has one now!

PR #1400, opened 2026-09-07
chore(IR): avoid quadratic case splitting in attribute equality

math-fehr:
Attribute.decEq previously split on both attributes, generating a case for every pair of constructors. This change adds a helper function, `IsAttr.decEqAgainst` to avoid this quadratic blowup.

This reduces the compile time of `Attribute.lean` (which is in the critical build path) from 24s to 8s locally.

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -790,349 +790,6 @@ theorem UnregisteredAttr.sizeOf_type {a : UnregisteredAttr} (h : a.type = some t
   grind [cases UnregisteredAttr]
 
 /-!
-  ## DecidableEq instances
--/
-
-mutual
-def VectorType.decEq (type1 type2 : VectorType) : Decidable (type1 = type2) :=
-  if hshape : type1.shape = type2.shape then
-    match Attribute.decEq type1.elementType type2.elementType with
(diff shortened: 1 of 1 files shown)


Veir/IR/Attribute.lean
```diff
@@ -1738,6 +1395,278 @@ instance : IsAttr Attribute where
...
+analysis on the first attribute constructor, and then checking whether or not the second attribute
+has the same constructor (which is done in constant time with).
+-/
+
+/--
+Implement decidable equality between an attribute of a known kind with an arbitrary attribute,
+given a decidable equality function for the specific attribute kind.
+
+This function does not a linear case split on the attribute kind at runtime, and instead only checks
+whether the attribute constructor is the expected one.
```
tobiasgrosser: Grammar.

Veir/IR/Attribute.lean
```diff
@@ -1738,6 +1395,278 @@ instance : IsAttr Attribute where
...
+
+This function is mostly useful to define an `DecidableEq` instance for `Attribute` in linear time.
```
```suggestion
This function is mostly useful to define a `DecidableEq` instance for `Attribute` in linear time.
```

tobiasgrosser (approved):

PR #1401, opened 2026-09-07
chore(IR): generate attribute Repr and Hashable helpers once

math-fehr:
`deriving Repr for A, B, C`, where `A`, `B`, and `C` are in the same mutual group, creates 3 times the 3 `repr` functions, without any reuse. This change add a new macro to derive `Repr` and `Hashable` so that the functions are only created once.

This reduce the compilation of `Attribute.lean` by 1 second, and will prevent the file compilation time to blowup once we add more attributes. Also, almost all files depend on this one, so this is in the critical path.

### DIFF
--- a/UnitTest.lean
+++ b/UnitTest.lean
@@ -1,4 +1,5 @@
 import UnitTest.Lexer
+import UnitTest.Deriving
 import UnitTest.ParserError
 import UnitTest.Parser
 import UnitTest.AttrParser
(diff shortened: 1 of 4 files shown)


tobiasgrosser (approved):

tobiasgrosser:
I am overall OK with this, but am wondering why we add so. much code for a 1 second win. Maybe. explain this in the per message?

PR #1404, opened 2026-09-07
chore(IR): make recursive attribute printers partial

math-fehr:
Attribute printers are never used during proofs, so they can be marked as partial and remove their termination proofs.

This reduces the compilation time of `Attribute.lean` from 4.7s to 4.2s, which is quite important since most files in the codebase depend on `Attribute.lean`.


tobiasgrosser (approved):
This is OK. But can you add a comment why optimizing for `0.5s` is worth it?

PR #1407, opened 2026-09-07
feat(LLVM): parse llvm.call_intrinsic

tobiasgrosser:
Add `llvm.call_intrinsic`, the op MLIR uses for LLVM intrinsics without a dedicated op, with the properties clang emits: `intrin`, `operandSegmentSizes`, `op_bundle_sizes`, `op_bundle_tags`, `fastmathFlags`, `arg_attrs` and `res_attrs`. The verifier mirrors MLIR's: the name must start with `llvm.`, at most one result, no regions, `operandSegmentSizes` must be an i32 array matching the operands, and the bundle sizes and tags must be consistent. The bundle checks are shared with `llvm.intr.assume` through a new `verifyOperandBundles` helper, which also makes that op reject non-i32 sizes and missing or non-string tags the same way. Effects stay `.unknown`, since they depend on the intrinsic name.

### DIFF
--- a/Veir/Dialects/LLVM/OpInfo.lean
+++ b/Veir/Dialects/LLVM/OpInfo.lean
@@ -827,6 +827,7 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
     pure ()
   | .call_intrinsic => do
     op.checkIsNonNullIntegerType ctx opIn
+    op.verifyLLVMCompatibleTypes ctx opIn
     let props := op.getProperties! ctx.raw Llvm.call_intrinsic
     if !(String.fromUTF8? props.intrin.value).any (·.startsWith "llvm.") then
       throw "intrinsic name must start with 'llvm.'"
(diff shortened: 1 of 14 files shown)


mafeguimaraes (approved):
Looks good!

Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -819,6 +825,21 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
     if op.getNumSuccessors ctx.raw opIn ≠ 0 then
       throw "Expected 0 successors"
     pure ()
+  | .call_intrinsic => do
+    op.checkIsNonNullIntegerType ctx opIn
```
math-fehr, line 829: This only checks that it is not a `i0`. So things like `vector`, `index`, `felt`, are still allowed from this.
I'm pretty sure that most other operations are also buggy in that way :/

math-fehr (approved):
There is a bug due to `OperationPtr.checkIsNonNullIntegerType` being wrong, but otherwise I think it's fine so far

PR #1410, opened 2026-09-07
feat(LLVM): parse llvm.fence

tobiasgrosser:
We introduce the property `ordering` using the new `Data.LLVM.AtomicOrdering`, which mirrors `IntPred` and `FloatPred`. Following MLIR, we leave 3 as an unused id. As a fence orders other accesses rather than performing one, its verifier allows only the four orderings that say something. A fence also carries `syncscope`, which is optional.


mafeguimaraes (approved):
LGTM!

tobiasgrosser:
@math-fehr, this one is good to go.

regehr (approved):
LGTM!

PR #1412, opened 2026-09-07
fix(parser): accept shorthand LLVM parameterless types

sueszli:
Part of #952.

Adds shorthand parsing for the remaining LLVM parameterless types: `x86_amx`, `ppc_fp128`, `metadata`, `label`, `token`, and `target<...>`.

```mlir
!llvm.array<2 x x86_amx>
!llvm.array<4 x ppc_fp128>
!llvm.array<1 x target<"spirv.Image", i32, 0, 0, 0, 0, 0, 0>>
!llvm.func<void (metadata, label, token)>
```

The printer emits the full `!llvm.` prefix, so a second parse site in `parseOptionalType` accepts that form for round-trip correctness.

Both spellings are handled by `parseOptionalUnregisteredSingletonType` and `parseOptionalUnregisteredType`, which take the type name as a parameter, and the names themselves live in a single `llvmParameterlessTypes` list used by both parse sites. The shared name matching replaces the hand-rolled prologue in the array, struct and function type parsers as well.

Tested by `Test/LLVM/pretty_llvm_types.mlir`, which runs `VEIR_ROUNDTRIP` and `MLIR_ROUNDTRIP`, and by unit tests in `UnitTest/AttrParser.lean`.

Two pre-existing limitations that this PR extends to the new types rather than fixing:

- A `target<...>` body is stored as text, like `!llvm.struct<...>`. Type aliases inside a body are not resolved, and two spellings of the same type are not equal.
- VeIR accepts `!llvm.array<N x metadata|label|token>` where `mlir-opt` rejects it, so the round-trip test uses parameter position for those three.

### DIFF
--- a/Veir/Parser/AttrParser.lean
+++ b/Veir/Parser/AttrParser.lean
@@ -509,6 +509,54 @@ partial def parseOptionalDialectType : AttrParserM (Option TypeAttr) := do
     return some (⟨UnregisteredAttr.mk ("!" ++ String.fromUTF8! dialectName) true none, by grind⟩)
 
 /--
+  Consume the type name `!name`, or (exclusively) `name` without the dialect prefix when
+  `short`. Consumes nothing if the name is not there.
+-/
+private def parseOptionalTypeName (name : String) (short : Bool) : AttrParserM Bool := do
+  if short then
+    let mnemonic := match name.splitOn "." with
+      | _dialect :: mnemonic@(_ :: _) => ".".intercalate mnemonic
+      | _ => name
+    return ← parseOptionalKeyword mnemonic.toByteArray
+  let token ← peekToken
+  let .exclamationIdent := token.kind | return false
+  let input := (← getThe ParserState).input
+  let typeName := { token.slice with start := token.slice.start + 1 }.of input
+  if typeName ≠ name.toByteArray then return false
+  let _ ← consumeToken
+  return true
+
+/--
+  The LLVM types that take no parameters. `void` and `ptr` have their own parsers.
+-/
+private def llvmParameterlessTypes : List String :=
+  ["llvm.x86_amx", "llvm.ppc_fp128", "llvm.label", "llvm.metadata", "llvm.token"]
+
+/--
+  Parse the parameterless type `!name`, or (exclusively) `name` without the dialect prefix
+  when `short`.
+-/
+def parseOptionalUnregisteredSingletonType (name : String) (short := false) :
+    AttrParserM (Option TypeAttr) := do
+  if !(← parseOptionalTypeName name short) then return none
+  return some ⟨UnregisteredAttr.mk ("!" ++ name) true none, by grind⟩
+
+/--
+  Parse the type `!name<...>`, or (exclusively) `name<...>` without the dialect prefix when
+  `short`. The body is kept as written, so both spellings print as `!name<...>`.
+-/
+def parseOptionalUnregisteredType (name : String) (short := false) :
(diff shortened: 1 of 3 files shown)


Veir/Parser/AttrParser.lean
```diff
@@ -1209,6 +1220,23 @@ partial def parseOptionalType : AttrParserM (Option TypeAttr) := do
...
+        return some ⟨UnregisteredAttr.mk s!"!llvm.{kw}" true none, by grind⟩
+    if typeName == "llvm.target".toByteArray then
+      let _ ← consumeToken
+      let startPos ← getPos
+      parsePunctuation "<"
+      let _ ← parseUnregisteredAttrBody
+      let endPos := (← peekToken).slice.stop
+      parsePunctuation ">"
+      let body := (Slice.mk startPos endPos).of input
+      return some ⟨UnregisteredAttr.mk ("!llvm.target" ++ String.fromUTF8! body) true none, by grind⟩
```
math-fehr: This code is quite complex in this location without being its own function.
Can you create a `parseUnregisteredSingletonType` that takes a string as parameter (like `"llvm.x86_amx"`) and parses it? And `parseUnregisteredType` that just takes `"llvm.target"` and parses it with its body?
Then this code should just be 4 if/elses, and make this entire code nicer?

sueszli: Thank you - I will rethink the architecture.

I was trying to maximize code locality.

math-fehr:
oh also, let me know if I can review this now or if you still want to make changes

sueszli:
@math-fehr Yes, please! May I ask for your review?

Veir/Parser/AttrParser.lean
```diff
@@ -508,6 +508,54 @@ partial def parseOptionalDialectType : AttrParserM (Option TypeAttr) := do
...
+  The LLVM types that take no parameters. `void` and `ptr` have their own parsers.
+-/
+private def llvmParameterlessTypes : List String :=
+  ["llvm.x86_amx", "llvm.ppc_fp128", "llvm.label", "llvm.metadata", "llvm.token"]
+
+/--
+  Parse the parameterless type `!name`, or (exclusively) `name` without the dialect prefix
+  when `short`.
+-/
+def parseOptionalUnregisteredSingletonType (name : String) (short := false) :
```
math-fehr, line 539:
```suggestion
private def parseOptionalUnregisteredSingletonType (name : String) (short := false) :
```

Veir/Parser/AttrParser.lean
```diff
@@ -508,6 +508,54 @@ partial def parseOptionalDialectType : AttrParserM (Option TypeAttr) := do
...
+    AttrParserM (Option TypeAttr) := do
+  if !(← parseOptionalTypeName name short) then return none
+  return some ⟨UnregisteredAttr.mk ("!" ++ name) true none, by grind⟩
+
+/--
+  Parse the type `!name<...>`, or (exclusively) `name<...>` without the dialect prefix when
+  `short`. The body is kept as written, so both spellings print as `!name<...>`.
```
math-fehr:
```suggestion
  Parse the type `!dialect.name<...>`, or (exclusively) `name<...>` without the dialect prefix when
  `short`. The body is kept as written, so both spellings print as `!dialect.name<...>`.
```

math-fehr (approved):
Nice! I just had two nits comments, I'll address them then merge this PR

sueszli:
Thank you so much!

Issue #1298, opened 2026-08-24
bugs in mixed-width constants


regehr:
we're not always handling `llvm.mlir.constant` correctly when the attribute bitwidth differs from the result bitwidth. I've put together some XFAIL tests here that I think we should merge right away, and then of course remove XFAIL as we fix the issues. then we should make sure we don't have analogous bugs in any other dialects.

https://github.com/opencompl/veir/pull/1297

cc @GZGavinZhao

sueszli:
Relevant: @math-fehr's review in PR #1413

PR #1413, opened 2026-09-07
fix: reject out-of-range integer attrs, fix mixed-width constant folding

sueszli:
Part of #1298.

Parser rejects attrs outside `[-2^(N-1), 2^N)`:

```mlir
// now rejected, matching mlir-opt
%0 = "llvm.mlir.constant"() <{value = 256 : i8}> : () -> i8
```

Constant folder normalizes operands to their attribute bitwidth before folding. Without this, `200 : i8` folded as 200 instead of -56:

```mlir
%0 = "llvm.mlir.constant"() <{value = 200 : i8}> : () -> i32
```

### DIFF
--- a/Veir/IR/Attribute.lean
+++ b/Veir/IR/Attribute.lean
@@ -104,6 +104,16 @@ structure IntegerAttr where
   type : IntegerType
 deriving Inhabited, Repr, DecidableEq, Hashable
 
+def IntegerAttr.normalize (value : Int) (bitwidth : Nat) : Int :=
+  if bitwidth == 0 then 0
+  else
+    let modulus : Int := 2 ^ bitwidth
+    let reduced := value % modulus
+    let unsigned := if reduced < 0 then reduced + modulus else reduced
+    if bitwidth == 1 then unsigned
+    else if unsigned ≥ modulus / 2 then unsigned - modulus
+    else unsigned
+
 /--
  Floating point fastmath flags attribute.
 -/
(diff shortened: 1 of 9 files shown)


Veir/IR/Attribute.lean
```diff
@@ -104,6 +104,16 @@ structure IntegerAttr where
   type : IntegerType
 deriving Inhabited, Repr, DecidableEq, Hashable
 
+def IntegerAttr.normalize (value : Int) (bitwidth : Nat) : Int :=
```
math-fehr, line 107: Can you add documentation to this function to explain what it does?

Veir/IR/Attribute.lean
```diff
@@ -104,6 +104,16 @@ structure IntegerAttr where
...
+  if bitwidth == 0 then 0
+  else
+    let modulus : Int := 2 ^ bitwidth
+    let reduced := value % modulus
+    let unsigned := if reduced < 0 then reduced + modulus else reduced
+    if bitwidth == 1 then unsigned
```
math-fehr, line 113: This is incorrect, and `-1 : i1` should not be normalized to `1 : i1`.
The error we had in LLVM was unrelated to this normalization, and was the fact that the property was `i1` but the result was `i32` for instance.

math-fehr (commented):
I'm a bit worried about this change. This transforms `-1 : i1` into `1 : i1`, which is incorrect AFAIK.
Do you think instead of talking about `Int` here, we could instead change `IntegerAttr` to use `BitVec` rather than `Int`? This would likely solve all of our problems there if we can make it work.

sueszli:
Thanks for the careful review. Yes, I agree and you're right.

Your "`BitVec` instead of `int`" suggestion makes the most sense here, but I'd need some time to rethink this problem from first principles once I have the bandwidth. Rather than leave a half-correct patch open, I'm closing this and will redo it from scratch (I hope this is OK!)

I've referenced your review up on #1298 so it's not lost

PR #1403, opened 2026-09-07
chore(IR): improve generated `IsAttr` proofs

math-fehr:
`attribute_instance` generates for each attribute kind a proof that is linear in the number of attribute kinds. This change makes this proof constant in the number of attribute kinds, making the entire process linear instead of quadratic.

This reduces the compilation of `Attribute.lean` from 7s to 4.7s locally.

PR #1415, opened 2026-09-07
chore(IR): improve generated `IsAttr` proofs

math-fehr:
`attribute_instance` generates for each attribute kind a proof that is linear in the number of attribute kinds. This change makes this proof constant in the number of attribte kinds, making the entire process linear instead of quadratic.

This reduces the compilation of `Attribute.lean` from 7s to 4.7s locally.


math-fehr:
Merged as #1403 was approved but deleted by github stack PRs

PR #1416, opened 2026-09-07
fix: make the interpreter use datalayout

regehr:
the purpose of this minimal PR is to make the interpreter correct with respect to RISC-V.

future PRs should make the interpreter parameterized by a datalayout, and add additional parts of the MLIR/LLVM datalayout mechanism, which is quite elaborate. what we have here is the bare minimum.


luisacicolini (approved):

math-fehr (approved):
I don't have a good understanding how all of this work, but the design looks good to me otherwise.

PR #1417, opened 2026-09-07
feat(LLVM): check the types of the floating point arithmetic

tobiasgrosser:
Floating point operations expect scalar floating point types, all of uniform width.


regehr (approved):
LGTM!

PR #1418, opened 2026-09-07
feat: table-based constant folding

regehr:
up to now we fold two ways:
- poison-propagating instructions return poison when any input is poison
- pure operations fold when all inputs are constant

this PR adds dialect-specific tables of folds. the tables themselves are purposefully minimal: I'll flesh them out after this code lands. this PR is about the mechanism. the interesting mechanism here is that we try all 3 folding methods and take the best result where best is the highest priority of:
0: no fold
1: argument
2: regular constant
3:  poison value

this mechanism is needed because none of the 3 folding methods always produces the best result.

finally, remove a peephole that is now subsumed by a fold table


Veir/Interfaces/FoldInterfaces.lean
```diff
@@ -61,21 +74,26 @@ private def foldPoisonedOperand (opType : OpCode)
...
     (resultTypes : Array TypeAttr) (constOperands : Array (Option RuntimeValue))
-    : Option (Array FoldResult) := do
+    : Option (Array FoldDecision) := do
   guard (!opType.isConstantLike)
-  foldByEvaluation opType properties resultTypes constOperands <|>
-    foldPoisonedOperand opType resultTypes constOperands
+  let tableDecision := foldByTable opType properties resultTypes constOperands
+  let evaluationDecision := foldByEvaluation opType properties resultTypes constOperands
+  let poisonDecision := foldPoisonedOperand opType resultTypes constOperands
+  preferredFold (preferredFold tableDecision evaluationDecision) poisonDecision
```
math-fehr: Should we always evaluate `poisonDecision` first, since it will always be a `poison` decision, which is always the best?
Then `tableDecision`, because it is likely the fastest, then `evaluationDecision`, and grab the preferredFold of the last two if `poisonDecision` is none?

Veir/IR/OpInfo.lean
```diff
@@ -64,6 +65,14 @@ class HasOpInfo (opCode: Type)
     (opIn : op.InBounds ctx.raw) → Except String PUnit :=
       fun _ _ _ _ => pure ()
   /--
+  Apply this opcode set's dialect-local fold table. The array contains the
+  known constant value of each operand, or `none` for a nonconstant operand.
+  Implementations are responsible for returning an in-range operand or a
+  constant conforming to the operation's result type.
+  -/
+  fold : (op : opCode) → propertiesOf op → Array TypeAttr →
```
math-fehr: Can we rename this `tryFold` or `fold?`

Veir/FoldDecision.lean
```diff
@@ -0,0 +1,26 @@
...
+  that `HasOpInfo`, and therefore every dialect's fold table, can refer to it
+  without depending on the folding machinery built on top of it.
+-/
+
+namespace Veir
+
+public section
+
+/-- What an operation folds to, when it folds at all. -/
+inductive FoldDecision where
```
math-fehr, line 18: I feel we should also add `foldPreference` here as well?

math-fehr: And `prefferedFold`?

Also probably in the `FoldDecision` namespace.

Veir/IR/OpInfo.lean
```diff
@@ -64,6 +65,14 @@ class HasOpInfo (opCode: Type)
...
+    Array (Option RuntimeValue) → Option FoldDecision := fun _ _ _ _ => none
```
math-fehr: While we are at it, can you return an `Array (Option FoldDecision)`? We will need this later for things like `muli_extended`

math-fehr (approved):
Nice! I just added a few comments here and there, but the structure looks great!

Veir/Interfaces/FoldInterfaces.lean
regehr: I'll do this-- but I think it probably doesn't have much effect on compiler throughput since I think we expect no-fold to be the common case by far, so we end up running all three anyway

Veir/IR/OpInfo.lean
regehr: I've gone with the `Option (Array FoldDecision)` version

PR #1419, opened 2026-09-07
feat(LLVM): add llvm.intr.vector.reduce.or

zaz:
Part of #947.

Parse and verify `llvm.intr.vector.reduce.or` for fixed integer vectors, with a scalar result matching the element type. Includes VeIR/MLIR round-trip and rejection tests. SQLite function coverage increases from 1,435 to 1,437/1,598, accepting `incrVacuumStep` and `sqlite3BtreeOpen`.


math-fehr (approved):
Looks good to me!

PR #1422, opened 2026-09-07
chore: fast path for canonicalizing binops

regehr:
this simply avoids calling the `partition` function on a list of 2 operands, which is better sorted using a quick comparison and swap


math-fehr (approved):
Nice!

PR #1423, opened 2026-09-08
feat(PBV): Handle sign and zero extend operations 

luigirinaldi:
Introduce support for sign and zero extend operations:
- Expand the `Tm` type to express props on width terms to parse width conditions. By prop we mean:
  - Equality and inequalities on width terms (`>` and `≥` are encoded as flipped `<` and `≤` respectively)
  - `and`s on the props
- Translate conditions on the widths into conditions on the corresponding masks.
- Define new theorems to relate widths and masks.
- Add test cases to `UnitTest`.
- Add an example that uses the `and` of two width conditions.
- (minor) Rename some theorems to match the lean convention.

### DIFF
--- a/UnitTest/BoundedBitblasting/BoundedBitblasting.lean
+++ b/UnitTest/BoundedBitblasting/BoundedBitblasting.lean
@@ -34,3 +34,32 @@ example {w t v: Nat} (a b : BitVec w)
   := by
   pbv_decide 16
   · bv_decide
+
+/-- Zero extending a zero extension -/
+example (p q r : Nat) (x : BitVec p)
+  (hr : r ≤ 8)
+  (hqr : q < r)
+  (hpq : p < q) :
+  (x.zeroExtend q).zeroExtend r = x.zeroExtend r
+  := by
+  pbv_decide 8
+  · bv_decide
+
+/-- Sign extending a zero extensions -/
+example (p q r : Nat) (x : BitVec p)
+  (hr : r ≤ 8)
+  (hqr : q < r)
+  (hpq : p < q) :
+  (x.zeroExtend q).signExtend r = x.zeroExtend r
+  := by
+  pbv_decide 8
+  · bv_decide
+
+/-- Double zero extending with composite width. -/
+example (p q : Nat) (x : BitVec p)
+  (hr : q ≤ 8)
+  (hpq : q > p) :
+  (x.zeroExtend q).zeroExtend (q + q) = x.zeroExtend (q + q)
+  := by
+  pbv_decide 8
+  · bv_decide
--- a/UnitTest/BoundedBitblasting/BoundedBitblasting.lean
+++ b/UnitTest/BoundedBitblasting/BoundedBitblasting.lean
@@ -35,17 +35,7 @@ example {w t v: Nat} (a b : BitVec w)
   pbv_decide 16
   · bv_decide
 
-/-- Zero extending a zero extension -/
-example (p q r : Nat) (x : BitVec p)
-  (hr : r ≤ 8)
-  (hqr : q < r)
-  (hpq : p < q) :
-  (x.zeroExtend q).zeroExtend r = x.zeroExtend r
-  := by
-  pbv_decide 8
-  · bv_decide
-
-/-- Sign extending a zero extensions -/
+/-- Sign extending a zero extension -/
 example (p q r : Nat) (x : BitVec p)
   (hr : r ≤ 8)
   (hqr : q < r)
@@ -55,10 +45,10 @@ example (p q r : Nat) (x : BitVec p)
   pbv_decide 8
   · bv_decide
 
-/-- Double zero extending with composite width. -/
+/-- Double zero extending with composite width -/
 example (p q : Nat) (x : BitVec p)
-  (hr : q ≤ 8)
-  (hpq : q > p) :
+  (hq : q ≤ 8)
+  (hqp : q > p) :
   (x.zeroExtend q).zeroExtend (q + q) = x.zeroExtend (q + q)
   := by
   pbv_decide 8
--- a/UnitTest/BoundedBitblasting/BoundedBitblasting.lean
+++ b/UnitTest/BoundedBitblasting/BoundedBitblasting.lean
@@ -35,6 +35,15 @@ example {w t v: Nat} (a b : BitVec w)
   pbv_decide 16
   · bv_decide
 
+/-- Double zero extend with conjunction condition -/
+example (p q r : Nat) (x : BitVec p)
+  (hr : r ≤ 8)
+  (h : q < r ∧ p < q)
+  : (x.zeroExtend q).zeroExtend r = x.zeroExtend r
+  := by
+  pbv_decide 8
+  · bv_decide
+
 /-- Sign extending a zero extension -/
 example (p q r : Nat) (x : BitVec p)
   (hr : r ≤ 8)
--- a/Veir/Data/PBV.lean
+++ b/Veir/Data/PBV.lean
@@ -50,7 +50,9 @@ which we wish to prove it, the tactic performs the following steps:
 * `Veir.Data.PBV.Lemmas` — `Nat` and `BitVec` lemmas used by the translation.
 * `Veir.Data.PBV.Elim` — steps 3 and 4: `width_elim` and `var_elim`.
 * `Veir.Data.PBV.Mask` — step 5: `maskOfWidth`, the mask constraint.
-* `Veir.Data.PBV.Push` — step 6: pushing `setWidth` from the root to its leaves.
+* `Veir.Data.PBV.Push` — step 5: translating width relations and width
+  arithmetic into mask relations; step 6: pushing `setWidth` from the root to
+  its leaves.
 * `Veir.Data.PBV.Examples` — a manual trace of the whole pipeline.
 
 -/
--- a/Veir/Meta/PBVDecide.lean
+++ b/Veir/Meta/PBVDecide.lean
@@ -125,10 +125,13 @@ meta def Tm.toExpr (this : Tm .width) (env : TmWidthEnv) : Expr :=
 /--
 Generate a `Name` from a `Tm`. Uses the index of the atoms as the basic variable name.
 -/
-meta def Tm.toName (tm : Tm .width) : Name :=
+meta def Tm.toName {k : TmKind} (tm : Tm k) : Name :=
   match tm with
-  | .widthAtom e => Name.mkSimple s!"w{e}"
+  | .widthAtom e  => Name.mkSimple s!"w{e}"
   | .widthAdd v w => Name.mkSimple s!"{v.toName}_add_{w.toName}"
+  | .widthLT  v w => Name.mkSimple s!"{v.toName}_lt_{w.toName}"
+  | .widthLE  v w => Name.mkSimple s!"{v.toName}_le_{w.toName}"
+  | .widthEQ  v w => Name.mkSimple s!"{v.toName}_eq_{w.toName}"
 
 /--
 Compute an upper bound on the value of this width term: an atom is bounded by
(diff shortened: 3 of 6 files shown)


Veir/Meta/PBVDecide.lean
```diff
@@ -235,44 +270,39 @@ meta def introMaskWidth (maxBound : Nat) (g : MVarId) (widthTm : Tm .width) (inf
...
-  else
-    introMaskWidth maxBound g term this
-
 /--
 Recurse through a `.width Tm`, converting `Nat` term into a `BitVec` mask, and
 translating relations on the terms into relations on the `BitVec` (eg. add).
 -/
-meta def introMaskRec (maxBound : Nat) (g : MVarId) (widthTm : Tm .width) (infos : WidthInfos)
-  : MetaM (MVarId × WidthInfo × WidthInfos) :=
+meta def introMaskRec (blastWidth : Nat) (g : MVarId) (widthTm : Tm .width) (infos : WidthInfos)
```
bollu: Naming convention: I call such a function `getOrCreateWidthMask`. The `FooRec` or `FooAux` suffix is used for a helper worker function, if it is wrapped by a function called `Foo`. In this case, since `introMaskRec` is the direct user API, let's call it `getOrCreateWidthMask`. So please rename.

Veir/Meta/PBVDecide.lean
```diff
@@ -235,44 +270,39 @@ meta def introMaskWidth (maxBound : Nat) (g : MVarId) (widthTm : Tm .width) (inf
...
+      g.note (Name.mkSimple s!"bv_{widthTm.toName}")
+      <| ← mkAppM ``add_eq_mul_of_maskOfWidth
+      <| #[
+        vInfo.hypWidthLeBoundNote,
+        wInfo.hypWidthLeBoundNote,
+        thisInfo.hypWidthLeBoundNote,
+        vInfo.widthMaskHypFvar,
+        wInfo.widthMaskHypFvar,
+        thisInfo.widthMaskHypFvar
+      ].map mkFVar
```
bollu:
```suggestion
        mkFVar vInfo.hypWidthLeBoundNote,
        mkFVar wInfo.hypWidthLeBoundNote,
        mkFVar thisInfo.hypWidthLeBoundNote,
        mkFVar vInfo.widthMaskHypFvar,
        mkFVar wInfo.widthMaskHypFvar,
        mkFVar thisInfo.widthMaskHypFvar
      ]
```

While I agree that pulling out the `map mkFVar` is shorter, I believe the suggested code much more strongly reflects intent, since we are passing `Expr`s which happen to be `FVar`s. It so happens that in our code, al the `Expr`s are `FVar`s, but this is not a 'deep' reason.

Veir/Meta/PBVDecide.lean
```diff
@@ -235,44 +270,39 @@ meta def introMaskWidth (maxBound : Nat) (g : MVarId) (widthTm : Tm .width) (inf
...
 
     return (g, thisInfo, infos)
 
-
```
bollu, line 275: Stray change in this PR: please push this directly as a whitespace change in a separate PR.

Veir/Meta/PBVDecide.lean
```diff
@@ -365,20 +395,64 @@ meta partial def visitExprRec (g : MVarId)
...
     return (g, widthTms, bvs)
 
+/--
+Extract `Tm .width`s and corresponding theorem from `Tm .prop`.
+-/
+meta def getThmFromProp (term : Tm .prop) : Name × Tm .width × Tm .width :=
+  match term with
+  | .widthLT v w => (``lt_of_lt_of_eq_maskOfWidth, v, w)
+  | .widthLE v w => (``le_of_le_of_eq_maskOfWidth, v, w)
+  | .widthEQ v w => (``eq_of_eq_of_eq_maskOfWidth, v, w)
```
bollu:
```suggestion
meta def Tm.getThmFromProp (term : Tm .prop) : Name × Tm .width × Tm .width :=
  match term with
  | .widthLT v w => (``lt_of_lt_of_eq_maskOfWidth, v, w)
  | .widthLE v w => (``le_of_le_of_eq_maskOfWidth, v, w)
  | .widthEQ v w => (``eq_of_eq_of_eq_maskOfWidth, v, w)
```

Let's namespace it, so we can write `tm.getThmFromProp`.

bollu: Also, isn't it doing 'too much'? Like, why does is every `prop` about widths a binary operator?
In some sense, this code is special cased against the fact that since all our `Prop`s on widths are binary operators, we can unify the code in this way, right? Trying to check that I understand correctly.

bollu (changes requested):
Overall, looks algorithmically correct to me. Please make the cosmetic/naming convention changes I asked for. Thanks for pushing this through!

Veir/Meta/PBVDecide.lean
luigirinaldi: yes exactly, I don't love it either, but it was to avoid duplicating the `g.note` for each theorem.

luigirinaldi: relatedly, it feels like the `translateWidthPrecond` could be folded into the `getOrCreateMaskWidth` except for the fact that it also requires a fact from the local context in order to state the theorem, hence why it's been pulled out

luigirinaldi:
@bollu I've changed the way the condition on masks is obtained from the `Nat` condition and expanded the scope of this PR a bit to include conjunctions of props, as it motivates having a better structure.

Veir/Meta/PBVDecide.lean
```diff
@@ -365,20 +415,72 @@ meta partial def visitExprRec (g : MVarId)
   else
     return (g, widthTms, bvs)
 
+/--
+Given a `Tm .prop` and the `Expr` is was reified from, construct the expr that
+corresponds to the `Prop` expressed in terms of the masks.
+-/
+meta def getExprFromProp (propTm : Tm .prop) (prop : Expr) (g : MVarId)
```
bollu: Instead of having to keep `propTm` and `prop` in sync, you can use the `Tm.toExpr` function, which will construct the `Expr` that the `Tm` represents. That way, you don't ned to keep `prop` around anymore:

```suggestion
meta def getExprFromProp (propTm : Tm .prop) (g : MVarId)
```

Veir/Meta/PBVDecide.lean
```diff
@@ -365,20 +415,72 @@ meta partial def visitExprRec (g : MVarId)
...
+  -- Helper to apply a theorem for a binary operation on widths.
+  let applyPropBinop (thm : Name) (v w : Tm .width) : MetaM (MVarId × WidthInfos × Expr) := do
+    let (g, vInfo, wInfos) ← getOrCreateWidthMask g v wInfos
+    let (g, wInfo, wInfos) ← getOrCreateWidthMask g w wInfos
+    let expr ← g.withContext <| mkAppM thm <| #[
+          .fvar vInfo.hypWidthLeBoundNote,
+          .fvar wInfo.hypWidthLeBoundNote,
+          .fvar vInfo.widthMaskHypFvar,
+          .fvar wInfo.widthMaskHypFvar,
+          prop,
```
bollu:
```suggestion
          propTm.toExpr,
```

Veir/Meta/PBVDecide.lean
bollu: Also, maybe we should call the function `getMaskedExprFromProp`? or some such?

bollu (changes requested):
Thanks! We're nearly there. I have one more implementation suggestion, and then I'm happy to merge!

luigirinaldi:
@bollu I've switched it to a version where the propTm.toExpr is used to generate `freshExprMvar`s which are returned by the function and the caller of the function has to unify the expr to fill the hole. Let me know if you think this is better and if there's a more idiomatic way of doing this?

bollu:
Aha, my bad: I misunderstood what the previous code was doing :D I didn't realize that we were both interrogating the type, while building proofs with the value! I guess I would do the following: 

(a) revert to the old code
(b) rename the prop variable to `prop`, and the value variable to `proof`, since it's the proof of that given proposition. 
(c) I would add checks, or a comment, that we expect that `isDefEq (<- inferType proof) prop`, to make it clear to the reader what's happening.

Sorry for the confusion, this stuff is tricky, and I'm no more infalliable than anyone else!

bollu:
I'd recommend bundling up the results of complex functions as a struct, because it returns quite a few things that are just raw `Expr`s, but some are propositions and some are proofs. For example, sth like this?

luigirinaldi:
> Aha, my bad: I misunderstood what the previous code was doing :D I didn't realize that we were both interrogating the type, while building proofs with the value! I guess I would do the following:
> 
> (a) revert to the old code (b) rename the prop variable to `prop`, and the value variable to `proof`, since it's the proof of that given proposition. (c) I would add checks, or a comment, that we expect that `isDefEq (<- inferType proof) prop`, to make it clear to the reader what's happening.
> 
> Sorry for the confusion, this stuff is tricky, and I'm no more infalliable than anyone else!

Thanks for the review, and sorry for not explaining this better. 

The only reason we might want to keep this current version is that it catches any mistakes exactly where the proof might not match with the prop. But I suppose the same would be achieved by placing an `isDefEq` at the start of the function as you suggested.

bollu (approved):
LGTM!

UnitTest/BoundedBitblasting/BoundedBitblasting.lean
```diff
@@ -34,3 +34,31 @@ example {w t v: Nat} (a b : BitVec w)
   := by
   pbv_decide 16
   · bv_decide
+
+/-- Double zero extend with conjunction condition -/
+example (p q r : Nat) (x : BitVec p)
+  (hr : r ≤ 8)
+  (h : q < r ∧ p < q)
+  : (x.zeroExtend q).zeroExtend r = x.zeroExtend r
```
tobiasgrosser:
```suggestion
  (h : q < r ∧ p < q) :
  (x.zeroExtend q).zeroExtend r = x.zeroExtend r
```

Veir/Data/PBV.lean
```diff
@@ -50,7 +50,9 @@ which we wish to prove it, the tactic performs the following steps:
 * `Veir.Data.PBV.Lemmas` — `Nat` and `BitVec` lemmas used by the translation.
 * `Veir.Data.PBV.Elim` — steps 3 and 4: `width_elim` and `var_elim`.
 * `Veir.Data.PBV.Mask` — step 5: `maskOfWidth`, the mask constraint.
-* `Veir.Data.PBV.Push` — step 6: pushing `setWidth` from the root to its leaves.
+* `Veir.Data.PBV.Push` — step 5: translating width relations and width
+  arithmetic into mask relations; step 6: pushing `setWidth` from the root to
+  its leaves.
```
tobiasgrosser, line 55: Do we know have two step 5 here?

tobiasgrosser (commented):
Thank you, @luigirinaldi.

This PR needs more test coverage. In particular, `(x.signExtend q).signExtend r = x.signExtend r`. Also make sure `le_of_le_of_eq_maskOfWidth` and `eq_of_eq_of_eq_maskOfWidth` are hit by test cases and that we have examples where `msb_eq_and_signBitOfMask_maskOfWidth_ne_zero` is fully exploited.

You move `namespace Veir.Data.PBV` around. This move is unrelated to this feature. Please create a separate PR for this.

Veir/Data/PBV.lean
luigirinaldi: yes but this was a mistake on my end, I had moved some theorems from Mask to Push, but I think they should actually stay in Mask and follow what was originally in the plan. Not sure if this should be a separate PR?

luigirinaldi:
> Thank you, @luigirinaldi.
> 
> This PR needs more test coverage. In particular, `(x.signExtend q).signExtend r = x.signExtend r`. Also make sure `le_of_le_of_eq_maskOfWidth` and `eq_of_eq_of_eq_maskOfWidth` are hit by test cases and that we have examples where `msb_eq_and_signBitOfMask_maskOfWidth_ne_zero` is fully exploited.
> 
> You move `namespace Veir.Data.PBV` around. This move is unrelated to this feature. Please create a separate PR for this.

I've added some new tests that use all of the conditions on widths, as well as the `msb`. The `trace_zero_sign_extend`  in Examples is already using `msb_eq_and_signBitOfMask_maskOfWidth_ne_zero`, not sure how else it could be fully exploited?.

naveen-seth (approved):
Reviewed mostly for style; LGTM there!

tobiasgrosser (approved):

PR #1424, opened 2026-09-08
chore: Split Felt semantics from interpreter

nchappe:


Veir/Data/Felt.lean
```diff
@@ -0,0 +1,67 @@
+module
+
+public import Veir.IR.Attribute
+public import Veir.RuntimeValue
```
tobiasgrosser: Do we need these two includes? Or rather, can we make the `Data` type independent of the IR and Runtime values?

Veir/Data/Felt.lean
```diff
@@ -0,0 +1,67 @@
...
+
+public section
+
+namespace Veir.Data
+
+namespace FeltSemantics
```
tobiasgrosser, line 9:
```suggestion
namespace Felt
```

Would `Felt` be enough?

tobiasgrosser (commented):
Nice. This is clearly going in a good direction.

nchappe:
Addressed the first point (one import was indeed unnecessary). The second one would be a renaming orthogonal to this PR (with non-trivial consequences), while the PR as it is now is just moving code around. I'm not familiar enough with Felt to have any strong opinion on naming.

math-fehr (approved):

PR #1425, opened 2026-09-08
test: cross-check more tests against mlir-opt

tobiasgrosser:
Add MLIR_ROUNDTRIP to 6 tests and MLIR_UNREGISTERED_ROUNDTRIP to 27 that already pass them, and MLIR_INVALID to the 90 negative tests that mlir-opt also rejects. The Parsing tests embed error line numbers in their CHECK lines, which move by one with the extra RUN line.


Test/Verifier/arith_i0_operand.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: not veir-opt %s 2>&1 | filecheck %s
+// RUN: %if mlir-min-23 %{ MLIR_INVALID %}
```
math-fehr, line 2: Interesting, why did this require MLIR 23? Did something change before that?

Test/Verifier/arith_i0_result.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: not veir-opt %s 2>&1 | filecheck %s
+// RUN: %if mlir-min-23 %{ MLIR_INVALID %}
```
math-fehr, line 2: Interesting, why did this require MLIR 23? Did something change before that?

math-fehr (approved):
Nice! I just had a question about two tests (for my own knowledge, nothing blocking). Feel free to merge after that.

Test/Verifier/arith_i0_operand.mlir
tobiasgrosser: MLIR stopped to support i0 integers in arith from 23, but still allowed them beforehand.

PR #1426, opened 2026-09-08
feat: dominance for graph regions

regehr:
this PR moves us towards compatibility with upstream MLIR-- we weren't really checking anything for dominance in graph regions and this does some of that

the basic thing here is making properDominance, not dominance, the main function, that the other one is implemented in terms of, the other way is inconvenient for graph regions, and I believe the new way is in line with upstream MLIR


Veir/Interfaces/RegionKindInterfaces.lean
```diff
@@ -16,15 +16,34 @@ namespace Veir
...
+owner declares: only a CFG gives the blocks an order, and an operation that
+declares a graph region is separately required to keep it to a single block.
 -/
 @[expose]
 def RegionPtr.hasSSADominance (region : RegionPtr)
     (ctx : WfIRContext OpInfo) : Bool :=
-  match (region.get! ctx.raw).parent with
+  let body := region.get! ctx.raw
+  if body.firstBlock ≠ body.lastBlock then true else
+  match body.parent with
```
math-fehr: Why do we need this check?
I think in general, it only makes sense to check for dominance in a verified program, so we should assume that the region has only a single block, and in that case remove this runtime check?

Veir/Interfaces/RegionKindInterfaces.lean
```diff
@@ -16,15 +16,34 @@ namespace Veir
...
+region has no owner and is an SSACFG region.
+-/
+@[expose]
+def RegionPtr.getRegionKind (region : RegionPtr)
+    (ctx : WfIRContext OpInfo) : RegionKind :=
+  match (region.get! ctx.raw).parent with
+  | none => .SSACFG
+  | some parent =>
+      HasOpInfo.getRegionKind (parent.getOpType! ctx.raw)
+        ((parent.get! ctx.raw).regions.idxOf region)
```
math-fehr: I feel like we should not have a function for `getRegionKind` and `hasSSADominance`, can we try to only use one of them?

Veir/Interfaces/RegionKindInterfaces.lean
```diff
@@ -16,15 +16,34 @@ namespace Veir
...
+A region holding more than one block always uses SSA dominance, whatever its
+owner declares: only a CFG gives the blocks an order, and an operation that
+declares a graph region is separately required to keep it to a single block.
 -/
 @[expose]
 def RegionPtr.hasSSADominance (region : RegionPtr)
     (ctx : WfIRContext OpInfo) : Bool :=
-  match (region.get! ctx.raw).parent with
+  let body := region.get! ctx.raw
+  if body.firstBlock ≠ body.lastBlock then true else
```
math-fehr: MLIR additionally checks that the region is not empty. If it is, it returns `true`, while here we return `false`.

Veir/Interfaces/RegionKindInterfaces.lean
```diff
@@ -16,15 +16,34 @@ namespace Veir
 
 variable {OpInfo : Type} [HasOpInfo OpInfo]
 
+/--
+The region kind `region` is declared to have by its owning operation. A root
+region has no owner and is an SSACFG region.
+-/
+@[expose]
```
math-fehr, line 23: Do we ever want to call this function directly? I feel that we should encouraged people to call the `hasSSADominance` instead right?

Veir/IR/Dominance.lean
math-fehr, line 44: Unrelated to your PR, but this `return false` should be `return true` here, otherwise we don't handle unreachable blocks correctly.

math-fehr (approved):
Nice! There is just a difference with MLIR for regions with no blocks, but otherwise all good!

Veir/IR/Dominance.lean
regehr: I can't make an MLIR test case that behaves differently due to this change and claude can't either!

regehr: oh wait maybe claude succeeded now. at least it thinks it did :). if this works out I'll make a PR with your one liner here.

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -521,12 +532,12 @@ def testOpDomNestedRegions : String :=
     %siblingInner = "test.test"() : () -> i32
   }) : () -> i32
 }) : () -> ()"#
-    #[ { dominator := "outer",      dominated := "outer",        dominates := true,  properDom := false }
+    #[ { dominator := "outer",      dominated := "outer",        dominates := true,  properDom := true  }
```
axelcool1234, line 535: Why is outer properly dominating itself?

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -521,12 +532,12 @@ def testOpDomNestedRegions : String :=
...
      , { dominator := "outer",      dominated := "inner",        dominates := true,  properDom := true  }
      , { dominator := "outer",      dominated := "otherOuter",   dominates := true,  properDom := true  }
      , { dominator := "outer",      dominated := "siblingInner", dominates := true,  properDom := true  }
      , { dominator := "inner",      dominated := "siblingInner", dominates := false, properDom := false }
-     , { dominator := "otherOuter", dominated := "inner",        dominates := false, properDom := false }
+     , { dominator := "otherOuter", dominated := "inner",        dominates := true,  properDom := true  }
```
axelcool1234, line 540: How is outerOther dominating inner?

UnitTest/DataFlowFramework/Dominance.lean
```diff
@@ -547,7 +558,7 @@ def testOpDomTwoLevelNested : String :=
     #[ { dominator := "top",    dominated := "middle", dominates := true, properDom := true  }
      , { dominator := "top",    dominated := "leaf",   dominates := true, properDom := true  }
      , { dominator := "middle", dominated := "leaf",   dominates := true, properDom := true  }
-     , { dominator := "leaf",   dominated := "leaf",   dominates := true, properDom := false }
+     , { dominator := "leaf",   dominated := "leaf",   dominates := true, properDom := true  }
```
axelcool1234, line 561: Same here, how is this proper dominance?

UnitTest/DataFlowFramework/Dominance.lean
axelcool1234: is this graph region stuff? I thought this wasn't a graph region but an SSACFG thingy

Veir/IR/Dominance.lean
regehr: nope-- claude was lying

UnitTest/DataFlowFramework/Dominance.lean
regehr: it's in a graph region, right?

UnitTest/DataFlowFramework/Dominance.lean
regehr: same answer-- it's in a graph region

UnitTest/DataFlowFramework/Dominance.lean
regehr: you've probably guess the answer already-- graph region

axelcool1234:
Addressing my previous review comments:
When I wrote those tests, I didn't realize I created graph regions. Is there a specific syntax in MLIR for graph regions vs SSACFG regions? This stuff continues to confuse me.

Veir/Interfaces/RegionKindInterfaces.lean
regehr: I'm not sure this is easy to fix-- how strongly do you feel about this one, Mathieu?

math-fehr: Not strong enough, let's leave it!

regehr:
> Addressing my previous review comments: When I wrote those tests, I didn't realize I created graph regions. Is there a specific syntax in MLIR for graph regions vs SSACFG regions? This stuff continues to confuse me.

well, first, have you read the available documentation on this topic?
but second, it's kind of complicated and no there's no single syntax you can look for to spot graph regions

math-fehr:
> Addressing my previous review comments: When I wrote those tests, I didn't realize I created graph regions. Is there a specific syntax in MLIR for graph regions vs SSACFG regions? This stuff continues to confuse me.

Operations may specify what are the "kind" of its regions. So an operation decides whether or not its region is a graph region or not. There is no parsing difference sadly

math-fehr (approved):
👍

PR #1428, opened 2026-09-08
feat(LLVM): add llvm.extractvalue

tobiasgrosser:
The op shares its position property and aggregate walk with insertvalue: LLVMInsertValueProperties becomes LLVMPositionProperties, and the walk moves into Llvm.verifyAggregatePosition, which returns the element type it reaches so insertvalue checks it against the value and extractvalue against the result. Arrays are checked fully; struct bodies stay opaque.


math-fehr (approved):
All good!

PR #1429, opened 2026-09-09
fix(LLVM,arith): spell the flags isExact and isDisjoint, as MLIR does

tobiasgrosser:
MLIR prints the flags as exact and disjoint in the pretty syntax but as isExact and isDisjoint in the generic syntax that VeIR uses, so VeIR dropped them on input and mlir-opt dropped VeIR's spelling on output, silently losing the flag in both directions. The arith extensions share the property structures and follow the same spelling.


regehr (approved):
yikes!
LGTM

PR #1430, opened 2026-09-09
fix: dominance and dead blocks

regehr:
over at #1426 @math-fehr suggested this as a fix, but neither claude nor I was able to come up with an MLIR-level test case that changed behavior based on this patch. so here it is anyhow


math-fehr:
It won't change an observed behavior currently, but it does change what happens when you query a region kind in unit tests for instance.
Can you add a test with an unreachable block, you should see a difference there!

regehr:
> It won't change an observed behavior currently, but it does change what happens when you query a region kind in unit tests for instance. Can you add a test with an unreachable block, you should see a difference there!

unit test added

math-fehr (approved):
This is the tests that should behave differently:

```
"builtin.module"() ({
  "func.func"() <{sym_name = "main", function_type = () -> ()}> ({
  ^entry:
    %v = "arith.constant"() <{value = 42 : i32}> : () -> i32
    "func.return"() : () -> ()
  ^dead:
    "test.test"() ({
      %x = "arith.addi"(%v, %v) : (i32, i32) -> i32
    }) : () -> ()
    "func.return"() : () -> ()
  }) : () -> ()
}) : () -> ()
```

This will only behave differently after you rebase on top of PR 1426. After PR 1426 this will fail, and after both PR 1426 and this one this should succeed

PR #1434, opened 2026-09-09
feat(ctree): trigger subeffects instead of effects

nchappe:
This makes things more convenient with non-trivial CTrees.


math-fehr (approved):
👍

PR #1436, opened 2026-09-09
feat(LLVM): add llvm.insertelement

zaz:
Register and verify `llvm.insertelement` for LLVM-compatible fixed vector types represented in VeIR, with pure effects and conservative poison handling. This enables strict round trips for 17 additional vectorized SQLite functions and all 14 insertion cases in the [vector tracker](https://github.com/opencompl/veir-sqlite/blob/main/VECTORS.md).

### DIFF
--- a/Test/Verifier/llvm_insertelement_i0_index.mlir
+++ b/Test/Verifier/llvm_insertelement_i0_index.mlir
@@ -1,0 +1,13 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+// VeIR's LLVM dialect rejects i0, although MLIR accepts it.
+
+"builtin.module"() ({
+  "func.func"() <{function_type = (vector<4xi32>, i32, i0) -> (), sym_name = "f"}> ({
+  ^bb0(%arg0: vector<4xi32>, %arg1: i32, %arg2: i0):
+    %r = "llvm.insertelement"(%arg0, %arg1, %arg2) : (vector<4xi32>, i32, i0) -> vector<4xi32>
+    "func.return"() : () -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: llvm.insertelement: operand 2 has forbidden i0 type
--- a/UnitTest/FoldDecision.lean
+++ b/UnitTest/FoldDecision.lean
@@ -4,9 +4,6 @@ import Veir.Interfaces.FoldInterfaces
 
 open Veir
 
--- A poison vector or inserted element does not poison every result lane.
-#guard !OpCode.propagatesPoison (.llvm .insertelement)
-
 /-- Find the first in-bounds operation with the given opcode. -/
 private def findOp (ctx : IRContext OpCode) (opType : OpCode) :
     Option { op : OperationPtr // op.InBounds ctx } := Id.run do
--- a/UnitTest/FoldDecision.lean
+++ b/UnitTest/FoldDecision.lean
@@ -4,6 +4,9 @@ import Veir.Interfaces.FoldInterfaces
 
 open Veir
 
+-- A poison vector or inserted element does not poison every result lane.
+#guard !OpCode.propagatesPoison (.llvm .insertelement)
+
 /-- Find the first in-bounds operation with the given opcode. -/
 private def findOp (ctx : IRContext OpCode) (opType : OpCode) :
     Option { op : OperationPtr // op.InBounds ctx } := Id.run do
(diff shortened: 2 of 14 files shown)


Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -473,6 +474,10 @@ def Llvm.propagatesPoison : Llvm → Bool
   | .intr__lifetime__start | .intr__lifetime__end | .intr__assume
   | .intr__vastart | .intr__vaend | .va_arg
   | .intr__memset | .intr__memcpy | .intr__memmove
+  -- Inserting into a poison vector can define a lane, and a poison element
+  -- only poisons the inserted lane. A poison index does poison the whole
+  -- result, but the interface is per-op.
```
tobiasgrosser: Not sure we need the comment here. Can you drop it?

Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -894,6 +899,26 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
+      | .floatType type => #[FloatType.bf16, FloatType.f16, FloatType.f32, FloatType.f64].contains type
+      | _ => false
+    if !validElementType then
+      throw s!"Expected an LLVM-compatible vector element type, but got {vectorType.elementType}"
+    let valueType := (op.getOperand! ctx.raw 1).getType! ctx.raw
+    if valueType.val != vectorType.elementType then
+      throw s!"Expected operand 1 to have vector element type {vectorType.elementType}, but got {valueType}"
+    ((op.getOperand! ctx.raw 2).getType! ctx.raw).verifyIntegerType
+      "Expected operand 2 to have integer type"
+    op.verifyResultTypeMatches ctx containerType "Expected the result to have the vector type"
```
tobiasgrosser, line 924: I am wondering if we can factor this out into a separate function. I would expect that this checking will be needed across numerous vector operations.

tobiasgrosser (commented):
Very cool. Let's discuss this tomorrow, but this is already going in a great direction.

tobiasgrosser:
> Includes round-trip, rejection and poison regression tests. Validated with lake build --iofail, lake test --iofail and the full lit suite: 1,069 passed, 5 expected failures. Before/after SQLite audits found no regressions.

You can probably drop this from the commit message. This is an AI update that provices info already tracked by our CI, so no need to replicate it in each PR.

zaz:
Factored out vector type verification.

Test/Verifier/llvm_insertelement_i0_index.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: not veir-opt %s 2>&1 | filecheck %s
+
+// VeIR's LLVM dialect rejects i0, although MLIR accepts it.
```
math-fehr, line 3: What is the reason for rejecting `i0`? We should try to copy MLIR as much as possible.

UnitTest/FoldDecision.lean
```diff
@@ -4,6 +4,9 @@ import Veir.Interfaces.FoldInterfaces
 
 open Veir
 
+-- A poison vector or inserted element does not poison every result lane.
+#guard !OpCode.propagatesPoison (.llvm .insertelement)
+
```
math-fehr, line 9: This test should be removed, there are no other tests like this in this file, and I do not think it adds anything to have this.

math-fehr (commented):
I just added two comments, but otherwise all good!

Test/Verifier/llvm_insertelement_i0_index.mlir
zaz: I was following the convention in #1118, `llvm.zext`, `llvm.trunc`, `llvm.getelementptr`, etc. I can remove the check here if you want, but then maybe other LLVM verifiers should be updated for consistency?

tobiasgrosser: There is a gap between the LLVM dialect in MLIR which accepts `i0`, LLVM itself which rejects `i0`, and VeIR which currently rejects `i0` for the ops @zaz listed above. I suggest to follow the established convention in VeIR and then switch it to accepting `i0` in a separate commit if that is desired. Now, arguably MLIR should reject `i0` as the LLVM dialect is an interface dialect that ought to mirror LLVM.

tobiasgrosser (approved):

tobiasgrosser:
@math-fehr, good to go?

PR #1437, opened 2026-09-10
fix(nix): pin clang and LEAN_AR to LLVM 23

GZGavinZhao:
The flake pinned `llvm` and `mlir` to `llvmPackages_23`, but `clang` and the `LEAN_AR` of the `makeLeanApp` wrappers used nixpkgs' default `llvmPackages` (LLVM 21). This PR makes sure `llvmPackages_23` is used for every LLVM component.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,7 @@
 
       developmentPackages = pkgs: with pkgs; [
         bash
-        clang
+        (llvmPackages pkgs).clang
         elan
         gmp
         gnumake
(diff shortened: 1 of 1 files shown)


axelcool1234:
Nice catch. When I first made the flake it didn't occur to me at all to match the LLVM version to the current version Lean nightly uses (https://github.com/leanprover/lean4-nightly/releases/tag/nightly-2026-08-29) :facepalm:

axelcool1234 (approved):
LGTM

PR #1438, opened 2026-09-10
feat(PBV): Automatically solve goals with grind

luigirinaldi:
Run grind on the auto-generated proof obligations. These the conditions enforcing each parametric width variable to be less-than-or-equal to the blast-width.


tobiasgrosser (approved):
That seems like a good idea.

PR #1440, opened 2026-09-10
chore: split some interpreter utils away

nchappe:
With this, there is a clear separation between
the basic interpreter infrastructure in `Util.lean`
and the actual interpreters for the various dialects
in `Basic.lean`.


tobiasgrosser (approved):
LGTM.

PR #1441, opened 2026-09-10
 feat: proofs for constant folding tables

regehr:
@math-fehr does this look like something we can merge?

### DIFF
--- a/Veir/Interfaces/FoldInterfaces/Correctness.lean
+++ b/Veir/Interfaces/FoldInterfaces/Correctness.lean
@@ -1,0 +1,94 @@
+module
+
+public import Veir.Interpreter.Refinement.Basic
+
+/-!
+# Local correctness of dialect fold tables
+
+These contracts concern the decisions returned by `HasOpInfo.tryFold`. They do
+not concern constant materialization, the folding driver, or rewriting the IR.
+Operand and result types describe a single operation signature; entry proofs
+state the signatures they support explicitly.
+-/
+
+public section
+
+namespace Veir
+
+/-- A replacement has the result's exact type, and any operand index is in range. -/
+@[expose]
+def FoldDecision.HasType (operandTypes : Array TypeAttr) (decision : FoldDecision)
+    (resultType : TypeAttr) : Prop :=
+  match decision with
+  | .useOperand j => operandTypes[j]? = some resultType
+  | .useConstant value => value.Conforms resultType
+
+/-- There is one well-typed replacement for each result, in result order. -/
+@[expose]
+def FoldDecision.HasTypes (decisions : Array FoldDecision)
+    (operandTypes resultTypes : Array TypeAttr) : Prop :=
+  decisions.size = resultTypes.size ∧
+    ∀ i (h : i < decisions.size), HasType operandTypes decisions[i] resultTypes[i]!
+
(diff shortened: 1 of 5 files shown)


Veir/Interfaces/FoldInterfaces/Correctness.lean
```diff
@@ -0,0 +1,94 @@
...
+/-- Correctness of every successful table lookup at the given operation signature.
+
+Typing is required independently of execution, including when the source has UB.
+Semantic correctness quantifies over all well-typed completions of the known
+operands, all memories, successor arrays, and data layouts. It uses the operation
+interpreter directly, so it does not rely on the driver's poison handling or
+evaluation fallback. Returning `none` creates no obligation.
+-/
+@[expose]
+def CorrectAt (op : OpCode) (properties : propertiesOf op)
```
math-fehr: Can you make this definition a structure with:
* One field for `HasTypes`. This field ensures that the replacement we return will always have the same type.
* One field for the rest

You can look at `structure LocalRewritePattern.Sound` in `Semantics.lean` for an example.
This way it makes it a bit easier to write tactics for specific parts of this definition.

Veir/Interfaces/FoldInterfaces/Correctness.lean
```diff
@@ -0,0 +1,94 @@
...
+def Agrees (known : Array (Option RuntimeValue)) (operands : Array RuntimeValue) : Prop :=
+  known.size = operands.size ∧
+    ∀ i, i < known.size → ∀ v, known[i]! = some v → operands[i]! = v
+
+/-- A successful interpretation has no memory or control-flow effect, and its
+results are refined by the replacements. UB permits any replacements, but an
+interpreter failure never establishes correctness. -/
+@[expose]
+def Refines (source : Interp (Array RuntimeValue × MemoryState × Option ControlFlowAction))
+    (replacements : Array RuntimeValue) (initialMemory : MemoryState) : Prop :=
```
math-fehr: Can you either swap the arguments (and put `source` at the end), or rename this to something like `IsRefinedBy`?
Since you check whether or not the first argument is refined by the next 2 others.

Veir/Interfaces/FoldInterfaces/Correctness.lean
```diff
@@ -0,0 +1,94 @@
...
+Typing is required independently of execution, including when the source has UB.
+Semantic correctness quantifies over all well-typed completions of the known
+operands, all memories, successor arrays, and data layouts. It uses the operation
+interpreter directly, so it does not rely on the driver's poison handling or
+evaluation fallback. Returning `none` creates no obligation.
+-/
+@[expose]
+def CorrectAt (op : OpCode) (properties : propertiesOf op)
+    (operandTypes resultTypes : Array TypeAttr) : Prop :=
+  ∀ known decisions, InputsConform known operandTypes →
```
math-fehr: You'll also need to add `op.Verifies` as an hypothesis. I think that's the contract for `tryFold` to be correct.
Otherwise in `tryFold` we would have to think whether or not the operation verifies, and do a bunch of checks that shouldn't be needed.

math-fehr:
(I'm still working on it btw, as I mentionned on zulip I want to clean up some proofs first, but the idea is good to me!)

PR #1443, opened 2026-09-11
feat(LLVM): add llvm.comdat and llvm.comdat_selector

tobiasgrosser:
Adds `llvm.comdat` and `llvm.comdat_selector`, which every clang-emitted C++ module with inline or template functions carries. A selector holds its name and the comdat kind as an i64 checked against MLIR's five values; a comdat body is a single block without arguments that may hold only selectors with distinct names. The `comdat` of an `llvm.func` or `llvm.mlir.global` must name a selector, either `@comdat::@selector` or a flat `@selector` outside any comdat, mirroring what MLIR verifies through its symbol tables. Nine negative tests are cross-checked with `mlir-opt`.

### DIFF
--- a/Veir/Dialects/LLVM/OpInfo.lean
+++ b/Veir/Dialects/LLVM/OpInfo.lean
@@ -71,6 +71,8 @@ inductive Llvm where
 | return
 | func
 | module_flags
+| comdat
+| comdat_selector
 | fadd
 | fsub
 | fmul
@@ -133,6 +135,8 @@ match op with
 | .getelementptr => GetelementptrProperties
 | .insertvalue | .extractvalue => LLVMPositionProperties
 | .fence => LLVMFenceProperties
+| .comdat => LLVMComdatProperties
+| .comdat_selector => LLVMComdatSelectorProperties
 | .fadd | .fsub | .fmul | .fdiv | .frem | .fneg | .intr__fmuladd | .intr__fabs =>
   FastMathFlagsProperties
 | .fcmp => FcmpProperties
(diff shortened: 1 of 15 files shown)


math-fehr:
Can you split your PR between the changes in SymbolAttr and the introduction of `comdat_selector`?

tobiasgrosser:
> Can you split your PR between the changes in SymbolAttr and the introduction of `comdat_selector`?

Done. #1445

Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -936,6 +957,17 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
       if elementType ≠ resultType.val then
         throw s!"Type mismatch: extracting from {containerType} should produce {elementType} \
           but this op returns {resultType}"
+  | .comdat => do
+    if op.getNumOperands ctx.raw opIn ≠ 0 then
+      throw "Expected 0 operands"
+    if op.getNumResults ctx.raw opIn ≠ 0 then
+      throw "Expected 0 results"
+    if op.getNumRegions ctx.raw opIn ≠ 1 then
```
math-fehr, line 965: You should also check that the region has a single block, and that this block doesn't have any arguments

math-fehr (approved):
Just one thing missing, but otherwise all good

Veir/Dialects/LLVM/OpInfo.lean
tobiasgrosser: Done.

PR #1444, opened 2026-09-11
chore: refactor how literal constants are accessed

regehr:
this is intended to be purely mechanical and NFC -- `matchConstantIntVal` now returns an integer option rather than an attribute option. this makes the numerous client sites each a little bit prettier.


math-fehr (approved):
Nice!

PR #1446, opened 2026-09-11
feat(LLVM): add llvm.mlir.alias

tobiasgrosser:
An alias names another global through an initializer region that computes its address. All its properties are modelled explicitly: sym_name, alias_type, linkage, the dso_local and thread_local_ unit flags, the unnamed_addr, visibility_ and tls_mode i64 enumerations, and the symbol's sym_visibility (public, private or nested).

We currently support properties for MLIR 23 and newer. For MLIR 23, we allow the property thread_local_ to express thread-locality even though MLIR 24 has dropped it. For MLIR 24, we have tls_mode (the five thread-local access models) and sym_visibility, neither of which is supported by MLIR 23. We currently store all three independently, such that we can reprint exactly what a given MLIR version had printed initially. As tls_mode is a more precise version of thread_local_ we could consider auto-upgrading. For now, we avoid auto-upgrades to keep interactions across both MLIR versions straightforward.

### DIFF
--- a/Veir/Dialects/LLVM/OpInfo.lean
+++ b/Veir/Dialects/LLVM/OpInfo.lean
@@ -19,6 +19,7 @@ inductive Llvm where
 | mlir__undef
 | mlir__zero
 | mlir__global
+| mlir__alias
 | mlir__addressof
 | and
 | or
@@ -109,6 +110,7 @@ def Llvm.propertiesOf (op : Llvm) : Type :=
 match op with
 | .mlir__constant => LLVMConstantProperties
 | .mlir__global => LLVMGlobalProperties
+| .mlir__alias => LLVMAliasProperties
 | .mlir__addressof => LLVMAddressOfProperties
 | .add => NswNuwProperties
 | .sub => NswNuwProperties
@@ -152,6 +154,7 @@ def Llvm.fromAttrDict
   cases op
   case mlir__constant => exact LLVMConstantProperties.fromAttrDict attrDict
   case mlir__global => exact LLVMGlobalProperties.fromAttrDict attrDict
+  case mlir__alias => exact LLVMAliasProperties.fromAttrDict attrDict
   case mlir__addressof => exact LLVMAddressOfProperties.fromAttrDict attrDict
   case add | sub | mul | shl | trunc =>
     exact NswNuwProperties.fromAttrDict attrDict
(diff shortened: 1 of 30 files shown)


Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -714,6 +733,24 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
       if properties.linkage.value == "common" && value.isKnownNonZero then
         throw "expected zero value for 'common' linkage"
     pure ()
+  | .mlir__alias => do
```
math-fehr, line 767: in LLVM, we check that the `alias_type` in the properties is a compatible LLVM type.

Veir/Dialects/LLVM/OpInfo.lean
```diff
@@ -714,6 +733,24 @@ def Llvm.verifyLocalInvariants {OpInfo : Type} [IsOpCode OpInfo]
...
+      throw "Expected 0 operands"
+    if op.getNumResults ctx.raw opIn ≠ 0 then
+      throw "Expected 0 results"
+    if op.getNumRegions ctx.raw opIn ≠ 1 then
+      throw "Expected 1 region"
+    if op.getNumSuccessors ctx.raw opIn ≠ 0 then
+      throw "Expected 0 successors"
+    let properties := op.getProperties! ctx.raw Llvm.mlir__alias
+    let body := (op.getRegion! ctx.raw 0).get! ctx.raw
+    if body.firstBlock.isNone || body.firstBlock ≠ body.lastBlock then
```
math-fehr: You should also check that the terminator is exactly an `llvm.return`

Veir/Dialects/LLVM/Properties.lean
```diff
@@ -286,6 +286,42 @@ def LLVMGlobalProperties.fromAttrDict (attrDict : Std.HashMap ByteArray Attribut
     extra
   }
 
+/--
```
math-fehr, line 290: What is the reasonning in having an `extra` field for the remaining attributes?
We need to actually check that their types is correct (for instance, if they are unit attribute, etc...).
We cannot do this if we just keep them in an `extra` dictionary attribute.

math-fehr (commented):
There are some verifier parts missing in the operation, but otherwise should be okay. I also believe we should not have the `extra` part in the property, what is the reasonning behind defining the properties that way?

I guass my general question is what is the goal, is it to fully support the operation, or to have a partial support of it that we can improve later on.

Veir/Dialects/LLVM/OpInfo.lean
tobiasgrosser: Done.

Veir/Dialects/LLVM/OpInfo.lean
tobiasgrosser: Done.

Veir/Dialects/LLVM/Properties.lean
tobiasgrosser: I made them all explicit.

tobiasgrosser:
Hi @math-fehr, I addressed your comments as follows:

The alias properties are now explicit instead of using an extra dictionary. The verifier now requires an `llvm.return` terminator, an alias_type that is an LLVM type naming storage, and no symbol redefinitions. Each rule has a negative test; 18 of the 22 are cross-checked with mlir-opt 23, and the remaining 4 cannot be, since MLIR 23 does not know tls_mode and drops sym_visibility.

See the updated PR message for further information.

regehr (approved):
this LGTM

PR #1449, opened 2026-09-12
feat(IR): add the f80 and f128 float types

tobiasgrosser:
f128 is IEEE binary128 and fits the existing float model. f80 is x87 extended precision, whose 64-bit significand stores its leading bit explicitly, so FloatFormat gains an explicitLeadingBit flag and decimal literals of such formats are packed accordingly, with x87's infinity and NaN patterns. clang emits both for long double, so they blocked parsing of ordinary C code.


Veir/Data/Float/Basic.lean
```diff
@@ -161,16 +164,54 @@ def ofUnpackedFloat (format : FloatFormat) (uf : UnpackedFloat)
...
+  else if format.explicitLeadingBit then
+    if hm : format.mantissa - 1 = 0 then
+      .ofBits 0#_
+    else
+      let leanFormat : _root_.Float.Model.Format :=
+        { exponentBits := format.exponent, mantissaBitsWithoutImplicit := format.mantissa - 1,
+          hm := by omega, he := by omega }
+      let uf := UnpackedFloat.ofScientific leanFormat significand exponent
+      let uf := if negative then uf.neg else uf
+      .ofUnpackedFloatExplicit format uf
```
math-fehr, line 214: Should maybe `toLeanFormat` contain the option about `format.explicitLeadingBit`? It feels a bit weird to have the `if/else` here rather than in `toLeanFormat`?
Thatm ight make the API harder to misuse also

math-fehr (approved):
That looks good to me, I'm not sure if this new field is the right solution or not though (I'm not familiar enough with the design).
Just had a comment about a function we could maybe improve.

AdUhTkJm:
The algorithm looks correct, but since f80 is using a new path, it can be better to add these test cases:
```c
// Largest finite value of f80
Value: 1.189731495357231765021e+4932
Pattern: 0x7FFEFFFFFFFFFFFFFFFF

// Smallest subnormal value of f80
Value: 3.645199531882474602528e-4951
Pattern: 0x00000000000000000001

// Ordinary subnormal value, without leading 1 in mantissa
Value: 1.282540566677892115121e-4937
Pattern: 0x00000000200000000000
```

Veir/Data/Float/Basic.lean
tobiasgrosser: Done.

PR #1450, opened 2026-09-12
feat(interfaces): getSuccessorForOperands in ControlFlowInterface

axelcool1234:
This will be replace `getSuccessorForOperands?` in DeadCodeAnalysis.lean.


math-fehr:
Can you try to make this work like in `HasOpInfo`? So adding in `HasOpInfo` a field that would represent this. Then in each `Dialect/OpInfo.lean` file you add the implementation. So that the idea is that you can depend on `ControlFlowInterface` without depending on all dialects.

axelcool1234:
> Can you try to make this work like in `HasOpInfo`? So adding in `HasOpInfo` a field that would represent this. Then in each `Dialect/OpInfo.lean` file you add the implementation. So that the idea is that you can depend on `ControlFlowInterface` without depending on all dialects.

I believe I tried this and ran into an import dependency

PR #1452, opened 2026-09-12
style(dataflow): renaming + documentation of a type in framework

axelcool1234:


regehr (approved):
looks good to me but why are you importing 2 more things from Lean Std in a PR that's just renaming things?

math-fehr (approved):
Can you just remove the new imports? Then all good!

axelcool1234:
Total mistake about the imports, I think I had them when I was messing around with different ways to define the analysis map. Woops!

PR #1454, opened 2026-09-12
feat(interpreter): give memory blocks with a physical address space

tobiasgrosser:
Memory was one flat byte array, so a pointer could walk from one allocation into its neighbour and nothing distinguished allocations. Memory is now an array of objects, one per allocation, and a pointer is an object index with a 64-bit offset. Loads and stores are checked against their own object by `MemoryState.checkAccess`, the one place where every later condition on an access is added, so leaving the object is UB.

Every object has a base address assigned by a bump allocator that honours the allocation's declared alignment, leaves a guard byte between objects, and starts past a 64 KiB arena that machine code may address directly. A pointer converts to an integer as base plus offset, and an integer converts back by binary search over the bases, so bitcasts, unrealized casts to registers and pointers stored in memory keep their meaning. RISC-V accesses decode the register value the same way and grow the object they land in up to the next object's base, which keeps machine code that addresses memory freely working.

Memory refinement lifts from the flat array to the objects: the same number of objects, each at the same address and refined bytewise.


Veir/Interpreter/Memory.lean
```diff
@@ -10,122 +10,213 @@ open Veir.Data
 namespace Veir
 
 /--
-  Memory state during interpretation.
-  Set bits in the poison mask represent poison bits.
+  One allocation during interpretation: its bytes, a poison mask in which set
+  bits mark poison bits, and the physical address `base` at which the object
+  starts, so that byte `i` of the object lives at address `base + i`.
```
math-fehr: Instead of adding documentation for the fields here, can you add the documentation per field? That will also improve the IDE experience

Veir/Interpreter/Memory.lean
```diff
@@ -10,122 +10,213 @@ open Veir.Data
...
 @[ext]
-structure MemoryState where
+structure MemoryObject where
   contents : ByteArray
   poisonMask : ByteArray
   consistentSize : contents.size = poisonMask.size
+  base : UInt64
+
+/-- An object of `size` bytes at address `base`, all of them poison. -/
+def MemoryObject.ofSize (base : UInt64) (size : Nat) : MemoryObject :=
```
math-fehr: maybe renaming this `alloc`? `ofSize` isn't descriptive enough since you also have to pass `base`?

PR #1457, opened 2026-09-12
fix: avoid unsoundness due to or-disjoint

regehr:
the provided test case results in [poison, poison] without this patch -- a failure of refinement


math-fehr (approved):
Yep, makes sense!

PR #1458, opened 2026-09-12
feat: propagate constants

regehr:
ok we finally get to connect the docs and propagate actual constants!

this runs SCP once and then runs rewrites to fixpoint. I don't think this strategy is guaranteed to reach a global fixpoint.


Veir/Passes/Canonicalize/ConstantPropagation.lean
```diff
@@ -0,0 +1,67 @@
+module
+
+public import Veir.PatternRewriter.Basic
+public import Veir.Analysis.DataFlow.SparseConstantPropagationAnalysis
+import Veir.Interfaces.FoldInterfaces
+import Veir.Interfaces.ControlFlowInterfaces
+
+namespace Veir.Canonicalize
+
+private def replaceConstant (rewriter : PatternRewriter OpCode)
```
math-fehr: Can you add documentation for this function?

Veir/Passes/Canonicalize/ConstantPropagation.lean
```diff
@@ -0,0 +1,67 @@
...
+  let mut maybeUse := (argument.block.get! ctx).firstUse
+  while let some use := maybeUse do
+    let useStruct := use.get! ctx
+    maybeUse := useStruct.nextUse
+    let some forwarded :=
+        BranchOpInterface.getSuccessorOperand? useStruct.owner use.index argument.index ctx
+      | continue
+    let some definingOp := forwarded.definingOp? | continue
+    return definingOp.getOpType! ctx
+  .arith .constant
```
axelcool1234: Is it safe to default to arith constant if we cant find a type for this block argument?

Veir/Passes/Canonicalize/ConstantPropagation.lean
```diff
@@ -0,0 +1,67 @@
...
+  let opType := op.getOpType rewriter.ctx.raw opInBounds
+  let operands := op.getOperands rewriter.ctx.raw opInBounds
+  if opType.isConstantLike then return rewriter
+  let mut rewriter := rewriter
+  for result in op.getResults! rewriter.ctx.raw do
+    rewriter ← replaceConstant rewriter dfCtx result opType
+  for operand in operands do
+    let .blockArgument argument := operand | continue
+    rewriter ← replaceConstant rewriter dfCtx operand
+      (blockArgumentMaterializer argument rewriter.ctx.raw)
```
axelcool1234: When passed to `tryFoldWithAnalysis`, does this check all arbitrary op's operands to see if any are block arguments? Unsure of the inefficiency of this - is there not a way to have a pattern for block arguments to avoid a lot of redundant checks?

axelcool1234:
> this runs SCP once and then runs rewrites to fixpoint. I don't think this strategy is guaranteed to reach a global fixpoint.

Does that mean SCP is rerun over and over per step? Like does `CanonicalizePass.impl` get run over and over until fixpoint?
This could mean lots and lots of runs of SCP possibly, right?

axelcool1234:
Wait nevermind you literally said run SCP once and then run rewrites over and over to fixpoint, woops.

Veir/Passes/Canonicalize/ConstantPropagation.lean
regehr: I need to work more on this part

PR #1459, opened 2026-09-12
chore: make integer decoding logic reusable

regehr:
the interpreter had some logic for pulling values out of literal constants, but elsewhere in the compiler we were not using that logic, but rather pulling out the raw value, which can lead to miscompiles.

this PR makes `decodeLLVMIntegerConstant` reusable and also reuses it in a few spots.

### DIFF
--- a/Test/Passes/InstructionSelection/RISCV64/const.mlir
+++ b/Test/Passes/InstructionSelection/RISCV64/const.mlir
@@ -7,22 +7,22 @@
         // CHECK: %[[A:.*]] = "riscv.li"() <{"value" = -1 : i64}> : () -> !riscv.reg
         // CHECK-NEXT: {{.*}} = "builtin.unrealized_conversion_cast"(%[[A]]) : (!riscv.reg) -> i64
         %c2 = "llvm.mlir.constant"() <{"value" = -1 : i32 }> : () -> i32
-        // CHECK: %[[A:.*]] = "riscv.li"() <{"value" = -1 : i64}> : () -> !riscv.reg
+        // CHECK: %[[A:.*]] = "riscv.li"() <{"value" = -1 : i32}> : () -> !riscv.reg
         // CHECK-NEXT: {{.*}} = "builtin.unrealized_conversion_cast"(%[[A]]) : (!riscv.reg) -> i32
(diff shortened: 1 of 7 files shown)


tobiasgrosser:
Hi @regehr, would it make sense to just store a `BitVec` in `IntegerAttr`. I mocked this up here: https://github.com/opencompl/veir/pull/1462.

regehr:
> Hi @regehr, would it make sense to just store a `BitVec` in `IntegerAttr`. I mocked this up here: #1462.

I think this change would be fine and probably a good idea, but I'm not sure it removes the need for a decoding function that reconciles the potentially different widths of the constant attribute and the result of the constant operation.

one possible solution would be to eliminate the different widths early, during parsing, but my take has been that we didn't want to do that since then we stop faithfully roundtripping MLIR that has this feature -- which seems to actually be a deliberate feature, not some weird corner case they didn't intend.

Test/Passes/InstructionSelection/RISCV64/const.mlir
```diff
@@ -15,10 +15,21 @@
         %c4 = "llvm.mlir.constant"() <{"value" = -1 : i1 }> : () -> i1
         // CHECK: %[[A:.*]] = "riscv.li"() <{"value" = -1 : i1}> : () -> !riscv.reg
         // CHECK-NEXT: {{.*}} = "builtin.unrealized_conversion_cast"(%[[A]]) : (!riscv.reg) -> i1
+        // An attribute wider than the result carries bits that are not part of the
+        // value: `300 : i32` in an i8 result is 44.
+        %c5 = "llvm.mlir.constant"() <{"value" = 300 : i32 }> : () -> i8
+        // CHECK: %[[A:.*]] = "riscv.li"() <{"value" = 44 : i32}> : () -> !riscv.reg
```
tobiasgrosser, line 21: I wonder, should we not force the attribute to match the register width? Such that a lowering would create a new attribute with the matching bitwidth.

The riscv is our dialect, so we can keep it simple?

regehr: good point-- let's do this for RISC-V, but not for the other dialects.

tobiasgrosser (approved):
LGTM.

PR #1463, opened 2026-09-13
feat(LLVM): add poison tracking to pointers in interpreter

tobiasgrosser:
In LLVM, values of type `ptr` can be poison. This PR lifts our previously poison-free pointer runtime values to support element-level poison. We add the following interpreter semantics:
- `llvm.mlir.poison` yields a poison value
- `llvm.mlir.zero` yields a non-poison zero address
- `llvm.load` and `llvm.store` trigger UB when accessing a poison address
- `llvm.getelementptr` forwards poison on both operands
- `llvm.bitcast` of a poison pointer yields (all) poison for `llvm.int` and `llvm.byte`

These semantics choices are worth taking a closer look:

- `unrealized_conversion_cast` from `llvm.ptr` to `riscv.reg` where `llvm.ptr` is poison is `0` for now. 

    As `riscv.reg` does not carry poison, we ideally want to take a non-deterministic choice. As we do not yet have ctrees, we pick `0` for now.

- `llvm.load` into a `llvm.ptr` type yields poison if any bit loaded was labeled poison

     Currently, LLVM models certain uninitialized memory as `undef`,  but `undef` is due to be replaced. The paper `Towards Removing Undef Values from LLVM IR` proposes to set uninitialized memory to `poison` and use freezing loads to obtain unknown non-poisonous values. Without ctrees, a freeze is currently equal to picking a fixed concrete value, but eventually this will be a non-deterministic choice of any concrete bit pattern. As we neither want to introduce `undef`, we lack support for ctree, and also do not have freezing loads the current semantics means that in certain cases our semantics allow poison to propagate further than LLVM's. We document this.
     
For better readability of the test cases, we improve the `toString` method used for printing pointers.


Test/Interpreter/LLVM/poison_pointer_freeze.mlir
```diff
@@ -0,0 +1,17 @@
...
+    %one = "llvm.mlir.constant"() <{value = 1 : i64}> : () -> i64
+    %slot = "llvm.alloca"(%one) <{elem_type = !llvm.ptr}> : (i64) -> !llvm.ptr
+    %p = "llvm.load"(%slot) : (!llvm.ptr) -> !llvm.ptr
+    %f = "llvm.freeze"(%p) : (!llvm.ptr) -> !llvm.ptr
+    %bits = "llvm.bitcast"(%f) : (!llvm.ptr) -> !llvm.byte<64>
+    "func.return"(%bits) : (!llvm.byte<64>) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: Program output: #[0b0000000000000000000000000000000000000000000000000000000000000000#64]
```
regehr: do we want a test on this? surely this behavior is going to change soon once we have ctrees!

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,62 @@
...
+  A pointer-typed value: an address, or poison. This mirrors `LLVM.Int`,
+  which carries poison the same way, and it is deliberately separate from the
+  address itself: the memory interface is indexed by a concrete address,
+  because using a poison pointer for an access is undefined behaviour and that
+  is decided by the operation, before memory is reached.
+-/
+inductive Ptr where
+  /-- An address. -/
+  | val (p : UInt64)
+  /-- A poison value indicating delayed undefined behavior. -/
```
regehr: very minor issue, but I always say "deferred" undefined behavior, not delayed

regehr (approved):
this moves us in the right direction, I'd be happy to land this and work on the finer details as we move forward

Veir/Data/LLVM/Ptr.lean
tobiasgrosser: Changed.

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+def isRefinedBy : Ptr → Ptr → Prop
+  | .poison, _ => True
+  | .val p, .val p' => p = p'
+  | .val _, .poison => False
+
+@[inherit_doc] infix:50 " ⊒ " => LLVM.Ptr.isRefinedBy
+
+@[simp, grind .]
+theorem isRefinedBy_refl (p : Ptr) : p ⊒ p := by
+  cases p <;> simp [isRefinedBy]
```
luigirinaldi:
```suggestion
  cases p <;> simp
```

`isRefinedBy` is already in the `simp` set

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+
+@[grind .]
+theorem isRefinedBy_trans {p₁ p₂ p₃ : Ptr}
+    (h12 : p₁ ⊒ p₂) (h23 : p₂ ⊒ p₃) : p₁ ⊒ p₃ := by
+  cases p₁ <;> cases p₂ <;> cases p₃ <;> simp_all [isRefinedBy]
```
luigirinaldi:
```suggestion
  cases p₁ <;> cases p₂ <;> cases p₃ <;> simp_all
```

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+
+/-- Only the same pointer refines a pointer that is not poison. -/
+@[grind .]
+theorem eq_of_val_isRefinedBy {p : UInt64} {q : Ptr}
+    (h : Ptr.val p ⊒ q) : q = .val p := by
+  cases q <;> simp_all [isRefinedBy]
```
luigirinaldi:
```suggestion
  cases q <;> simp_all
```

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+
+/-- The pointer whose bits are `b`, poison if any bit is poison. -/
+def ofByte (b : Byte 64) : Ptr :=
+  if b.poison = 0 then .val b.toUInt64 else .poison
+
+/-- The bits of a pointer: all poison for a poison pointer. -/
+def toByte : Ptr → Byte 64
+  | .val p => Byte.fromUInt64 p
+  | .poison => Byte.allPoison
+
```
luigirinaldi: docs?

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+@[simp, grind =]
+theorem ofByte_toByte (p : Ptr) : ofByte p.toByte = p := by
```
luigirinaldi, line 58: docs?

luigirinaldi (approved):
From what I understand, it looks good. Some really minor nits in the new `Ptr` file.

Veir/Data/LLVM/Ptr.lean
tobiasgrosser: What do you mean? `ofByte_toByte` seems to be self-explanatory? Do you want a comment on this theorem or any other documentation? What do you expect?

luigirinaldi: I suppose it is self-explanatory. I thought all theorems should have some doc-string, but in this case it wouldn't be adding any extra information.

Test/Interpreter/LLVM/poison_pointer_freeze.mlir
tobiasgrosser: I find it helpful to know that this will change in the future and have a test notify us if it does. I added a comment explaining that eventually we would like to have non-determistic choice via ctrees.

Test/Interpreter/LLVM/null_pointer_bitcast.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// The null pointer prints as address zero, and bitcasting it gives that address
+// with no poison bits, both as an `llvm.byte` and as an integer.
```
tobiasgrosser:
```suggestion
// The operation `llvm.mlir.zero` yields the zero address (null-pointer). Bitcasting it gives that address
// with no poison bits, both as an `llvm.byte` and as an integer.
```

nchappe (approved):
This looks good, thanks. Nit: the PR body is unclear, only unrealized casts from *poison* pointers to registers are fail.

tobiasgrosser:
> This looks good, thanks. Nit: the PR body is unclear, only unrealized casts from _poison_ pointers to registers are fail.

Nice. I fixed the PR body wording.

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+import all Veir.Data.LLVM.Byte.Basic
+
+namespace Veir.Data.LLVM
+
+public section
+
+/--
+  A pointer-typed value: an address, or poison.
+-/
+inductive Ptr where
```
math-fehr, line 14: Do you think we should at some point (not in this PR) have a `PoisonOr` definition that we could reuse in some places?
I'm personally not sure if this would help us in any way (not sure we can gain much from it), but I just wanted to ask if you nee any reuse we could have at some point.

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+public import Veir.Data.LLVM.Byte.Basic
+
+import all Veir.Data.LLVM.Byte.Basic
+
+namespace Veir.Data.LLVM
+
+public section
+
+/--
+  A pointer-typed value: an address, or poison.
```
math-fehr, line 12: Can you add a note that we are only considering 64-bits architectures here?
Also, in general, do you think we should keep that restriction or should we at some point consider other architectures?

Veir/Data/LLVM/Ptr.lean
```diff
@@ -0,0 +1,73 @@
...
+  cases p₁ <;> cases p₂ <;> cases p₃ <;> simp_all
+
+/-- Only the same pointer refines a pointer that is not poison. -/
+@[grind .]
+theorem eq_of_val_isRefinedBy {p : UInt64} {q : Ptr}
+    (h : Ptr.val p ⊒ q) : q = .val p := by
+  cases q <;> simp_all
+
+/-- The pointer whose bits are `b`, poison if any bit is poison. -/
+def ofByte (b : Byte 64) : Ptr :=
```
math-fehr, line 49: Should we have the same for conversions with `LLVM.Int`, or is the conversion only with `Byte`?

Veir/Interpreter/Basic.lean
```diff
@@ -656,10 +662,16 @@ def Llvm.interpretOp' (opType : Veir.Llvm) (properties : propertiesOf opType)
...
+          if h : bw = 64 then .ok (.addr (LLVM.Ptr.ofByte (val'.cast h))) else .fail
       | .addr val', .llvmPointerType _ => .ok (val)
       | .addr val', .byteType ⟨bw⟩ =>
-          if h : bw = 64 then .ok ((.byte 64 $ LLVM.Byte.fromUInt64 val')) else .fail
+          if bw = 64 then .ok (.byte 64 val'.toByte) else .fail
+      | .addr val', .integerType ⟨bw⟩ =>
+          if bw = 64 then
+            match val' with
+            | .val v => .ok (.int 64 (LLVM.Int.val v.toBitVec))
+            | .poison => .ok (.int 64 .poison)
```
math-fehr, line 673: For instance here, we should probably have a `LLVM.Ptr.toInt`?

Veir/Interpreter/Basic.lean
```diff
@@ -1284,6 +1296,8 @@ def interpretOp' (opType : OpCode) (properties : propertiesOf opType)
     | .registerType _, [.byte _bw val] =>
       return (#[.reg (LLVM.Byte.toReg val)], mem, none)
     | .registerType _, [.addr val] =>
+      /- A register has no poison to carry, so a poison pointer cannot be cast into one. -/
+      let .val val := val | Interp.fail
       return (#[.reg ⟨val.toNat⟩], mem, none)
```
math-fehr, line 1301: This would be incorrect with

```
%p = "llvm.mlir.poison"() : () -> !llvm.ptr
%b = "llvm.bitcast"(%p) : (!llvm.ptr) -> !llvm.byte<64>
```

Here, the current lowering might create this program:
```
%p = "llvm.mlir.poison"() : () -> !llvm.ptr
%r = "builtin.unrealized_conversion_cast"(%p) : (!llvm.ptr) -> !riscv.reg
%b = "builtin.unrealized_conversion_cast"(%r) : (!riscv.reg) -> !llvm.byte<64>
```

So returning `fail` here is making this transformation incorrect. I *think* the answer is to return non-deterministically any possible value of `!riscv.reg`, like we do with poison (and in that case here, only return 0 in the current interpreter).

Veir/Interpreter/Memory.lean
```diff
@@ -80,6 +80,11 @@ def MemoryState.empoison (state : MemoryState) (addr : UInt64) (n : Nat)
   else
     Interp.ub
 
+/-- Store the 64 bits of `v`, poison bits included, at `addr`. -/
```
math-fehr: Can you add that this is in little endian?

Veir/Interpreter/Memory.lean
```diff
@@ -133,9 +138,27 @@ def MemoryState.hasPoison (state : MemoryState) (addr size : UInt64)
...
   Load an LLVM value from the given memory address.
   Yields UB if access is out of bounds or the address is 0.
+
+  An integer or pointer load with any poison bit is poison as a whole, and a
+  `byte` load keeps poison per bit.
+
+  Together with fresh memory being poison, this is the semantics proposed in
+  "Towards Removing Undef Values from LLVM IR" (Lobo et al., PLDI 2026), not
+  LangRef's, where uninitialized memory reads as `undef`.
+  As Clang on still on poison, e.g., for a bitfield or an integer copy of a
```
math-fehr, line 158: Grammar?

math-fehr (commented):
Nice! I just had 1-2 style comments, but nothing major!
I think however we should change the semantics of unrealized_conversion_cast, otherwise currently our lowering would be unsound

Veir/Data/LLVM/Ptr.lean
tobiasgrosser: I am not yet convinced. Establishing a full API is typically a bit more expensive, but then on the end-user side easier to use than having to partially unfold a PoisonOr. Happy to discuss this though, as a potential API improvement.

Veir/Data/LLVM/Ptr.lean
tobiasgrosser: Let's start with this restriction. We can likely generalize this when needed.

Veir/Data/LLVM/Ptr.lean
tobiasgrosser: Added `ofInt`, `toInt`.

Veir/Interpreter/Basic.lean
tobiasgrosser: Use `toInt` here.

Veir/Interpreter/Memory.lean
tobiasgrosser: Fixed.

Veir/Interpreter/Basic.lean
tobiasgrosser: We are for now returning zero.

math-fehr (approved):
Looks good to me, I think you just have to resolve the merge conflicts though!

PR #1465, opened 2026-09-14
feat: CTree-based LLVM interpreter

nchappe:
This is a minimal adaptation of the Interp-based interpreter,
but it gives more reasonable semantics to freeze.

Results from helper functions (e.g. for memory accesses) are lifted from
the Interp monad.

This is not connected to any proof for the time being.


math-fehr (approved):
Code and all looks good to me.
I'm just a bit worried about duplicating code, especially since we are changing the memory model currently. But to be honest, I don't have a good solution for it yet besides trying as fast as we can to move to CTrees everywhere.

tobiasgrosser (approved):
Nice. Let's get this in.

PR #1466, opened 2026-09-14
chore: risc-v dialects now prefer size-64 bitvectors instead of arbitrary integers

regehr:
in upstream dialects like Arith and LLVM, we're forced to deal with literal constants whose attribute width differs from the width of the operation, making things potentially confusing and error-prone.

in our own RISC-V dialects, we're not constrained by upstream, so this PR is a simplification removing that degree of freedom and furthermore just centralizing on `i64` as the internal data representation for our RISC-V dialects. this removes some possibilities for errors around signed/unsigned conversions and I think it's a good change overall.


Test/RISCV/mem-imm-width-not-i64.mlir
```diff
@@ -0,0 +1,11 @@
...
+
+"builtin.module"() ({
+  "func.func"() <{sym_name = "main", function_type = () -> !riscv.reg}> ({
+    %a = "riscv.li"() <{ "value" = 0 : i64 }> : () -> !riscv.reg
+    %y = "riscv.ld"(%a) <{ "value" = 8 : i12 }> : (!riscv.reg) -> !riscv.reg
+    "func.return"(%y) : (!riscv.reg) -> ()
+  }) : () -> ()
+}) : () -> ()
+
+// CHECK: error: RISC-V memory operation: expected 'value' to be a 64-bit signless integer attribute, but got i12
```
tobiasgrosser: > signless integer attribute

Given that we print integers as negative numbers, the term `signless` sounds confusing to me. Maybe just call them `64-bit integer attribute`?

UnitTest/FoldDecision.lean
```diff
@@ -80,7 +80,7 @@ private def testRiscvConstantMaterialization : String := Id.run do
   let value : RuntimeValue := .reg ⟨BitVec.ofInt 64 (-77)⟩
   match (.riscv .add : OpCode).materializeConstant value regType with
   | some ⟨.riscv .li, props⟩ =>
-    if props.value ≠ IntegerAttr.mk (-77) (IntegerType.mk 64) then
+    if props.value ≠ BitVec.ofInt 64 (-77) then
```
tobiasgrosser:
```suggestion
    if props.value ≠ -77#64 then
```

You can use the short notation.

Veir/Dialects/Builtin/Properties.lean
```diff
@@ -24,6 +24,26 @@ def UnregisteredProperties.fromAttrDict (attrDict : Std.HashMap ByteArray Attrib
...
+  value it denotes as a `BitVec 64`.
+-/
+def getI64Attr (what key : String) (attrDict : Std.HashMap ByteArray Attribute) :
+    Except String (BitVec 64) := do
+  let some attr := attrDict[key.toUTF8]?
+    | throw s!"{what}: missing '{key}' property"
+  let .integerAttr intAttr := attr
+    | throw s!"{what}: expected '{key}' to be an integer attribute, but got {attr}"
+  if intAttr.type.bitwidth ≠ 64 then
+    throw s!"{what}: expected '{key}' to be a 64-bit signless integer attribute, but got i{intAttr.type.bitwidth}"
```
tobiasgrosser:
```suggestion
    throw s!"{what}: expected '{key}' to be a 64-bit integer attribute, but got i{intAttr.type.bitwidth}"
```

tobiasgrosser: Given that the parser accepts `-1`, I am not sure signless is the right term.

Veir/Passes/InstructionSelection/RISCV64.lean
```diff
@@ -350,7 +350,7 @@ def bswap_local (ctx : WfIRContext OpCode) (op : OperationPtr) :
   let (ctx, rev8Op) ← WfRewriter.createOp! ctx Riscv.rev8 #[RegisterType.mk] #[castOp.getResult 0]
       #[] #[] () none
   if opType.bitwidth = 32 then
-    let sh32 := RISCVImmediateProperties.mk (IntegerAttr.mk 32 (IntegerType.mk 64))
+    let sh32 := RISCVImmediateProperties.mk (BitVec.ofInt 64 32)
```
tobiasgrosser:
```suggestion
    let sh32 := RISCVImmediateProperties.mk 32#64
```

tobiasgrosser: The same for the constants below.

Veir/Passes/RISCVCombines/Combine.lean
```diff
@@ -39,7 +39,7 @@ def right_identity_zero_add_local (ctx : WfIRContext OpCode) (op : OperationPtr)
   let some (lhs, rhs, _) := matchRVAdd op ctx.raw | return (ctx, none)
   let some liOp := rhs.definingOp? | return (ctx, none)
   let some cst := matchRVLi liOp ctx.raw | return (ctx, none)
-  if cst.value.value ≠ 0 then return (ctx, none)
+  if cst.value.toInt ≠ 0 then return (ctx, none)
```
tobiasgrosser:
```suggestion
  if cst.value ≠ 0#64 then return (ctx, none)
```

tobiasgrosser (approved):
I had initially thought to just add verifier to check that the type is an i64. This is a bit more involved, but seems to make the API a bit simpler. For BitVec constants, there is a nice syntax.

I am happy with this. Let's see what @math-fehr says.

naveen-seth (approved):
I found nothing else; LGTM!

Veir/Dialects/Builtin/Properties.lean
```diff
@@ -24,6 +24,26 @@ def UnregisteredProperties.fromAttrDict (attrDict : Std.HashMap ByteArray Attrib
     Except String UnregisteredProperties :=
   .ok { opName := .empty, properties := DictionaryAttr.fromArray attrDict.toArray }
 
+/--
+  Read a required integer attribute that must be declared `i64`, returning the
+  value it denotes as a `BitVec 64`.
+-/
+def getI64Attr (what key : String) (attrDict : Std.HashMap ByteArray Attribute) :
```
math-fehr: Maybe renaming `what` to a more descriptive name? Like `contextMsg` or `errorCtx`, not sure

math-fehr (approved):
Tobias pointed most of the issues already, besides that this is good to me!

PR #1467, opened 2026-09-15
chore: fixup test cases for MLIR 24

regehr:
sigh... I track latest LLVM / MLIR due to Alive2 and there's been a behavior change that we need to keep up with:
  - thread_local_ is replaced by tls_mode. The old property was a unit attribute; the new one is an i64 enum.
  - Generic printing now includes tls_mode = 0 : i64 for non-thread-local aliases.
  - silently discards the old thread_local_ property


math-fehr (approved):
Thanks for updating the tests!
LLVM is moving very fast :(

PR #1471, opened 2026-09-15
feat(legalization): make llvm.add widening proof width-generic (bounded)

naveen-seth:
This makes the add_widening proof generic over widths up to a bounded limit using `pbv_decide`.
This is currently bounded to 16 bits for performance reasons; large widths (around 64 bits and above) time out.


Veir/Passes/Legalization/Proofs.lean
```diff
@@ -13,16 +14,22 @@ public section
...
+      LLVM.Int.trunc
+        (LLVM.Int.add (LLVM.Int.ext i t ext hIsWiden)
+          (LLVM.Int.ext i' t ext' hIsWiden) false false)
+        w false false hIsWiden := by
+  veir_bv_normalize
+  constructor
+  · simp
+  · intros
+    pbv_decide 16
+    · bv_decide
```
luigirinaldi:
```suggestion
    bv_decide
```

you don't really need the dot here

luigirinaldi (commented):
LGTM! modulo the minor nit. The `constructor ...` is a bit ugly atm could also be replaced by: 

```
refine ⟨by simp, fun _ _ => ?_⟩
intros
pbv_decide 16
bv_decide
``` 

More in general I should fix `pbv_decide` to handle it without any of this trickery.

naveen-seth:
After #1475 has landed, the performance should be good enough to increase this to 64 bit.

nchappe (approved):
Nice, thanks! Note that #1483 has landed today, so you may be able to increase the upper bound now.

PR #1472, opened 2026-09-15
chore(lean) : update lean to version 4.34.0

math-fehr:
Update lean to the latest release candidate version (4.35.0-rc1). 4.34.0 sadly has a bug in bv_decide that prevents us from using it.

The following changes were made:
* FloatFormat exponents now need to specify an exponent greater than 1, as 1-bit exponents are not supported anymore
* if_neg and if_pos were renamed to ite_eq_right and ite_eq_left
* Some grind patterns were changed in lean so we had to change 2 proofs
* Use `native_decide` instead of `cbv` in a Puddle tactic, as there is a bug in Lean: https://live.lean-lang.org/#codez=JYWwDg9gTgLgBAGQKYEMB2AoJAPF4A2ScAXBnOXABTABmcAKlAK5EwAWSacAzhCEQAUAqmmAwAdE1Hwk%2BbkUpoIaIsTgB5MDGDK4w6QEoDZCuQC8PPoJFjJ0khYBGATzgBjRwDcgA


math-fehr:
All good now, I updated it to `3.35.0-rc1` as `3.34.0` has a bug

Veir/Rewriter/GetSet/BlockOperands.lean
```diff
@@ -436,7 +436,7 @@ theorem OperationPtr.getSuccessors!_initBlockOperands {operation : OperationPtr}
     else
       operation.getSuccessors! ctx := by
   simp only [OperationPtr.getSuccessors!_def]
-  grind
+  grind [-Array.range'_append, -Array.range'_append_1]
```
tobiasgrosser, line 441: Having to drop simp lemmas is a bit concerning. Do we understand why this is happening?

tobiasgrosser (approved):
LGTM, assuming we have a good answer to my comment.

Veir/Rewriter/GetSet/BlockOperands.lean
math-fehr: This is a small reproducer of the bug in latest lean version:
https://github.com/leanprover/lean4/issues/15183

tobiasgrosser: OK, then we can move along and fix it when the bug is resolved. You may want to add a comment to the relevant Lean bug in the code to ensure things we can revert this when the bug is resolved.

PR #1480, opened 2026-09-17
feat: a few missing CTree lemmas, especially around bind

nchappe:


lean-ctrees/CTree/Defs.lean
```diff
@@ -392,5 +414,75 @@ instance : MonoBind (CTree E C) where
...
+theorem unfold_pure (r : R) :
+  CTree.unfold (pure r) = CTreeF.ret (E := E) (C := C) r := by
+    simp [pure]
+
+@[simp]
+theorem approx_pure_succ (r : α) n :
+  (pure r : CTree _ _ _).approx (n + 1) = CTreeF.ret (E := E) (C := C) r := by
+    simp [pure]
+
+theorem id_map : ∀ (x : CTree E C R), id <$> x = x := by
```
math-fehr, line 439: This should be renamed `map_id` instead I think

math-fehr: You can also probably mark this `simp` and `grind =` I think?

lean-ctrees/CTree/Defs.lean
```diff
@@ -392,5 +414,75 @@ instance : MonoBind (CTree E C) where
...
+    rw [<-CTree.fold_unfold t]
+    cases t.unfold
+    · apply ret
+    · apply tau
+    · apply vis
+
+@[simp]
+theorem unfold_pure (r : R) :
+  CTree.unfold (pure r) = CTreeF.ret (E := E) (C := C) r := by
+    simp [pure]
```
math-fehr, line 432: I think the best thing here would be to not have any theorems over `pure`, and instead have the lemmas on `ret`.
Then we have a single lemma to "unfold" `pure` to `ret`.
At least that's how it's done in other monads AFAIK, pure should only be used at the toplevel if I understand correctly.

lean-ctrees/CTree/Defs.lean
```diff
@@ -392,5 +414,75 @@ instance : MonoBind (CTree E C) where
...
+
+@[simp]
+theorem approx_pure_succ (r : α) n :
+  (pure r : CTree _ _ _).approx (n + 1) = CTreeF.ret (E := E) (C := C) r := by
```
math-fehr, line 436: Same here for pure

lean-ctrees/CTree/Defs.lean
```diff
@@ -392,5 +414,75 @@ instance : MonoBind (CTree E C) where
...
+    intro _ _ a _ _ _
+    rw [Bind.bind]
+    apply bind_mono (λ x => a) (λ x => x)
+    · grind [monotone]
+    · grind [monotone]
+    · intro _; grind
+
+@[simp, grind =]
+theorem CTree.ret_pure (r : R) : CTree.ret (E := E) (C := C) r = pure r := by
+  rfl
```
math-fehr, line 477: Here we should have `pure r` in the lhs instead, and rename this theorem `CTree.pure_def`.

math-fehr (approved):
Thanks for splitting this! I just had some comments on a few lemmas namings, otherwise all good!
Btw, we should also in general not use `pure` in our theorems, and instead remove `pure` as early as possible and use `CTree.ret` directly.

nchappe:
`id_map` is the name used in [stdlib](https://lean-lang.org/doc/reference/latest/Functors___-Monads-and--do--Notation/Laws/).
I added the simp/grind annotations.

PR #1481, opened 2026-09-17
feat: add a `PureOrErr` predicate on CTrees

nchappe:


Veir/Interpreter/CTree.lean
```diff
@@ -57,4 +57,24 @@ instance [ErrorE -< E] [UBE -< E] : MonadLift Interp (CTree E C) where
...
+/--
+A `PureOrErr` CTree has a finite depth, and no visible effect except for failure or UB.
+-/
+inductive PureOrErr {EIn CIn R} {E : EIn → Type} {C : CIn → Type}
+    [ErrorE -< E] [UBE -< E] :
+  CTree E C R → Prop where
+| ret (r : R) : PureOrErr (CTree.ret r)
+| tau (c : C1In ⊕ CIn) k : (forall x, PureOrErr (k x)) → PureOrErr (CTree.tauG c k)
+| fail {k} : PureOrErr (CTree.vis (Subeffect.mapEff ErrorE E ErrorEIn.mk) k)
+| ub {k} : PureOrErr (CTree.vis (Subeffect.mapEff UBE E UBEIn.mk) k)
```
math-fehr, line 69: We don't have any coercion so that this only looks something like `CTree.vis UBEIn.mk k`?

math-fehr (approved):
Nice! I just had a question about notation, but we can figure this out later.

PR #1485, opened 2026-09-17
chore(interpreter): move `Arith.interpretOp'` to `Arith/OpInfo.lean`

math-fehr:
This starts the move of all `interpretOp'` to their respective dialects. The objective of this is that most of the interpreter infrastructure is independent of dialect definitions, so that we can reason in isolation per dialect.

To move `Arith.interpretOp'`, I had to change a few imports to avoid circular imports. In particular I had to move `check(un)SignedDivision` to `Data.LLVM.Int`.


Veir/Data/LLVM/Int/Basic.lean
```diff
@@ -326,6 +326,26 @@ def smulHigh {w : Nat} (x y : Int w) : Int w := Id.run do
...
+Returns `true` if the signed division of `x` by `y` triggers undefined behavior.
+This returns `true` if either:
+- `y` is zero
+- `y` is poison (as it refines zero)
+- `y` is `-1` and `x` is the minimum signed value on `w` bits, or poison (as it refines the
+  minimum signed value).
+-/
+def isSignedDivisionUB {w : Nat} (x y : LLVM.Int w) : Bool :=
+  y = .poison ∨ y = .val 0 ∨ y = .val (-1) ∧ (x = .poison ∨ x = .val (BitVec.intMin w))
+
```
tobiasgrosser, line 348: Should we not delete these functions from `Veir/Interpreter/Util.lean`?

tobiasgrosser (approved):
LGTM, otherwise.

Veir/Data/LLVM/Int/Basic.lean
math-fehr: These are still used in the LLVM one, I'll remove them once I move the LLVM functions as well if that's okay

PR #1486, opened 2026-09-17
fix: call decodeLLVMIntegerConstant from more places

regehr:
both ArithToLLVM and the RISC-V backend need to call decodeLLVMIntegerConstant to get the actual value in a constant, in order to avoid miscompilation bugs.


math-fehr (approved):
👍

PR #1491, opened 2026-09-18
refactor(dataflow): flattened lattice map

axelcool1234:


Veir/Analysis/DataFlowFramework.lean
```diff
@@ -19,7 +19,7 @@ The solver state containing all dataflow facts and the worklist of program point
 to call transfer functions on.
 -/
 structure DataFlowContext where
-  lattice : HashMap LatticeAnchor (DHashMap FactKind Fact)
+  lattice : DHashMap FactKey (fun (key : FactKey) => Fact key.kind)
```
math-fehr:
```suggestion
  lattice : DHashMap FactKey (Fact ·.kind)
```
Should make this easier to read I think?

math-fehr (approved):
👍  Just one comment on style

Veir/Analysis/DataFlowFramework.lean
axelcool1234: Ah I forgot, woops!

PR #1492, opened 2026-09-18
feat(dataflow): DependencyGraph

axelcool1234:
Instead of every fact holding an array of dependents, we now store dependents, and its reverse, dependencies, in a central object now stored in the `DataFlowContext`. This allows for custom analyses to reason about the dependencies of a fact, whereas before it could only reason about dependents of a fact.


regehr (approved):
@math-fehr might want to look at this but it looks great to me

Issue #1448, opened 2026-09-11
dominance bug


regehr:
upstream MLIR rejects this as a dominance violation. we accept it-- I believe incorrectly

```mlir
 "func.func"() <{sym_name = "f", function_type = (i1, i1, i1) -> ()}> ({
^entry(%a : i1, %b : i1, %c : i1):
  "cf.cond_br"(%a, %a) [^left, ^def] <{operandSegmentSizes = array<i32: 1, 0, 1>}> : (i1, i1) -> ()
^left:
  "cf.br"() [^join] : () -> ()
^def(%v : i1):
  "cf.br"() [^use] : () -> ()
^mid(%unused : i1):
  "cf.br"() [^tail] : () -> ()
^use:
  "cf.br"(%v) [^mid] : (i1) -> ()
^join:
  "cf.cond_br"(%b, %a) [^mid, ^left] <{operandSegmentSizes = array<i32: 1, 1, 0>}> : (i1, i1) -> ()
^tail:
  "cf.cond_br"(%c) [^join, ^use] <{operandSegmentSizes = array<i32: 1, 0, 0>}> : (i1) -> ()
}) : () -> ()
```

axelcool1234:
Will investigate

axelcool1234:
# Dominance fixpoint trace for issue #1448

This is a handwritten trace of the dominance analysis on the reproducer from
[issue #1448](https://github.com/opencompl/veir/issues/1448). It compares the
buggy dependency handling with the implementation from
[PR #1493](https://github.com/opencompl/veir/pull/1493), built on the
`DependencyGraph` introduced in
[PR #1492](https://github.com/opencompl/veir/pull/1492).

The traces were checked against instrumented executions of:

- buggy base commit `20d91b90fea5`;
- fixed commit `6ffce54aa562`.

## Input CFG

```mlir
"func.func"() <{sym_name = "f", function_type = (i1, i1, i1) -> ()}> ({
^entry(%a : i1, %b : i1, %c : i1):
  "cf.cond_br"(%a, %a) [^left, ^def] <{operandSegmentSizes = array<i32: 1, 0, 1>}> : (i1, i1) -> ()
^left:
  "cf.br"() [^join] : () -> ()
^def(%v : i1):
  "cf.br"() [^use] : () -> ()
^mid(%unused : i1):
  "cf.br"() [^tail] : () -> ()
^use:
  "cf.br"(%v) [^mid] : (i1) -> ()
^join:
[...]

PR #1493, opened 2026-09-18
fix(dataflow): resolved unsoundness in dominance analysis due to missed fact enqueues

axelcool1234:
Dominance analysis wasn't enqueuing all facts that relied on a block that is a part of an immediate dominator chain. This PR fixes that, which caused the bug in https://github.com/opencompl/veir/issues/1448


regehr (approved):
LGTM!
I'll wait for Mathieu to look before merging

PR #1496, opened 2026-09-18
chore(interpreter): reorganize some interpreter infrastructure files

math-fehr:
This PR:
* Moves `RuntimeValue/` inside `Interpreter/`
* Moves `ControlFlowAction` structure to `RuntimeValue/Basic.lean`
* Remove two LLVM.Int functions from `Interpreter/Util.lean`
* Renames `Interpreter/Util.lean` to `Interpreter/VariableState.lean`, as this file only contains the variable state

This makes it such that `ControlFlow` action can be imported without importing all dialects, so that we can define interpreter for each dialect in their respective file.


math-fehr:
This should be ready now, sorry I forgot to fix the CI issue

naveen-seth (approved):
LGTM

PR #1497, opened 2026-09-18
chore(interpreter): move `Dialect.intepretOp'` functions to dialect folders

math-fehr:
This moves dialect-specific `interpretOp'` functions to their `Dialect/` folder. This has for objective that in the long-term, every dialect definitions (including semantic definitions) should only be written in their respective dialect folder. Similarly, the `Interpreter/` should be completely dialect-agnostic.

The only remaning thing is that `interpretOp'` is still in `Intepreter/Basic.lean`.


naveen-seth (approved):
Also LGTM!

PR #1503, opened 2026-09-18
test(riscv): allocate the memory the RISC-V interpreter tests use

tobiasgrosser:
The RISC-V interpreter tests stored to and loaded from literal addresses such as 8 without allocating memory, which only works because the flat memory model grows on demand. Each test now allocates the bytes it accesses with `riscv_stack.alloca` and addresses them through the returned register, the way a lifted machine function addresses its stack frame; the offset tests derive their second base with `riscv.addi`. Outputs are unchanged. This prepares for a memory model in which a machine access outside every allocation is undefined behaviour, as in Alive2's assembly mode.


math-fehr (approved):
Nice! Thanks for making this incremental

PR #1504, opened 2026-09-18
feat(riscv): bounds-check RISC-V memory accesses

tobiasgrosser:
RISC-V loads and stores grew memory with `MemoryState.ensureSize` so that machine code could address memory freely, while LLVM accesses were bounds-checked. Now that the RISC-V tests allocate what they access, the growth is gone: a machine access outside the memory that was allocated is undefined behaviour, exactly as an LLVM access is, and as it is in Alive2's assembly mode. Two tests cover a store past the end of an alloca and a load from an address nothing allocated.

The tests were freed from any accesses to unallocated memory in #1503.


math-fehr (approved):
Nice!

PR #1506, opened 2026-09-19
chore: move the tactics out of `Veir/Meta` into `Veir/Tactic`

tobiasgrosser:
`Veir/Meta` held two kinds of code: the elaborators `Deriving`, `Attrs` and `OpCode`, and the tactics `BVDecide`, `BVNormalizeSimp` and `PBVDecide`. The tactics now live in `Veir/Tactic`.


math-fehr (approved):
That looks fine to me, but I'm wondering if we should instead have a `BVDecide/` folder that contains all the useful code? Not sure, at least this PR is an improvement

luigirinaldi (commented):
This looks good, looking at Lean Core it seems like they put the tactics in `Meta/Tactic`. So maybe we could do that instead?

PR #1507, opened 2026-09-19
refactor(interpreter): an explicit `Pointer` type

tobiasgrosser:
A pointer into interpreter memory was a bare `UInt64`. It is now a `Pointer` in `Veir/Data/Pointer`, representing the object it may access and a byte offset into it, with `Pointer.null` and `Pointer.isNull`. We maintain for now our current flat memory model. This means we allocate only a single object, object 0, and memory addresses map directly to offsets within this 0-object. Pointers print as (object, offset) pair. Even though, semantically, this PR does not change anything, the interfaces (and variable names) are now established. As a result, round-trips from a pointer through an integer or bytes and back no longer hold.

This PR makes API explicit that will be helpful when we later introduce a block memory model.


Veir/Data/Pointer.lean
```diff
@@ -0,0 +1,3 @@
+module
```
math-fehr, line 1: Is there a reason to create a `Pointer.lean` and `Pointer/Basic.lean` for such a small file? I would usually do this only when we have multiple files?

Veir/Data/Pointer/Basic.lean
```diff
@@ -0,0 +1,31 @@
...
+  object : Nat
+  offset : UInt64
+deriving Inhabited, Repr, DecidableEq, Hashable
+
+namespace Pointer
+
+/-- The null pointer. -/
+def null : Pointer := ⟨0, 0⟩
+
+def isNull (p : Pointer) : Bool := p == null
```
nunoplopes: perhaps worth leaving a note saying that the right spec is addr(p) == 0

math-fehr:
```suggestion
def isNull (p : Pointer) : Bool := p = null
```

Veir/Interpreter/Memory.lean
```diff
@@ -7,6 +7,8 @@ public import Veir.Interpreter.Interp
 public section
 
 open Veir.Data
+open Veir.Data (Pointer)
```
math-fehr:
```suggestion
```
Since you already have `open Veir.Data`?

Veir/Interpreter/Memory.lean
```diff
@@ -7,6 +7,8 @@ public import Veir.Interpreter.Interp
...
+open Veir.Data.LLVM (Ptr)
```
math-fehr:
```suggestion
```
This one is just not used

math-fehr (approved):

Veir/Interpreter/Memory.lean
tobiasgrosser: Removed.

Veir/Interpreter/Memory.lean
tobiasgrosser: Removed.

Veir/Data/Pointer/Basic.lean
tobiasgrosser: I was a bit hesitant making this a `Prop`, or mixing Boolean and Prop types. I dropped this definition entirely for now, as it is was only used twice. We can add it when we actually need it.

Veir/Data/Pointer.lean
tobiasgrosser: I expect that we will eventually get more theorems and also find it easier to have no random files at the top of the `Data` directory.

Veir/Data/Pointer/Basic.lean
tobiasgrosser: Added.

PR #1508, opened 2026-09-19
refactor(interpreter): allocation may fail

tobiasgrosser:
An `alloca` computed its size as a `Nat` and truncated it to 64 bits, so a request for more than the address space silently allocated less. Its size now converts through `memorySize`, which makes a size that does not fit undefined behaviour. `MemoryState.alloc` returns in `Interp` and fails the run when memory would pass the end of the address space.


Veir/Interpreter/Memory.lean
```diff
@@ -29,11 +29,21 @@ def MemoryState.empty : MemoryState := {
 }
 
 /--
-  Allocate the given number of bytes of memory.
-  Return the updated memory state and the freshly allocated address.
+  The size of an `alloca` in bytes as a 64-bit value. An `alloca` has no way
+  to report failure, so a size that does not fit in the address space is
+  undefined behaviour, as it is in Alive2.
```
tobiasgrosser:
```suggestion
  undefined behaviour.
```

Test/Interpreter/LLVM/alloca_too_large.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// An `alloca` cannot report failure, so reserving more bytes than the address
+// space holds is UB: 2^62 elements of 8 bytes are 2^65 bytes.
```
math-fehr:
```suggestion
// space holds is UB: 2^62 elements of 8 bytes is 2^65 bytes.
```

math-fehr (approved):

PR #1509, opened 2026-09-19
refactor(interpreter): unify bounds check for memory accesses

tobiasgrosser:
`store`, `load`, `loadPoison`, and `empoison` each did their own access validation, which we now factored out into  `MemoryState.checkAccess`.

- The computation works on 64-bit values, and we ensure that it does not wrap
- An access of zero bytes is always allowed

We also delete redundant code in `empoison` by forwarding a call to it to a plain `store`.


Veir/Interpreter/Memory.lean
```diff
@@ -47,41 +47,38 @@ def MemoryState.alloc (mem : MemoryState) (size : UInt64) : Interp (MemoryState
     mem.poisonMask.extend size.toNat 0xff,
     by simp [mem.consistentSize]⟩, ⟨0, mem.contents.size.toUInt64⟩)
 
+/--
+  Check if an access of `size` bytes at `p` is allowed.
+-/
+def MemoryState.checkAccess (mem : MemoryState) (p : Pointer) (size : UInt64) : Interp Unit := do
+  -- An access of zero bytes is allowed anywhere.
+  if size = 0 then return ()
```
math-fehr, line 55: See my message on Zulip, but it seems that LLVM assumes that if a pointer is accessed (even with 0 bytes), it is addresseable.
So here, technically the pointer of address `0` should be rejected (?)

math-fehr (approved):
Just an issue with `0`-byte null pointers, but we can probably figure this out later.

Veir/Interpreter/Memory.lean
tobiasgrosser: Yes, lets keep this in mind for later. If we match alive, this warrants a separate discussion (and change)

PR #1510, opened 2026-09-20
refactor(interpreter): decode machine addresses into pointers

tobiasgrosser:
We make the need to convert physical addresses to pointers exlicit by adding two functions: `MemoryState.decode` and `MemoryState.address`. In the flat model both are trivial: the address is the offset. RISC-V accesses decode the register value instead of building the pointer by hand.

This cleans up our interface and removes plenty of functions that would not work for a non-trivial memory model.


Veir/Interpreter/Memory.lean
```diff
@@ -26,6 +27,12 @@ def MemoryState.empty : MemoryState := {
   consistentSize := (by grind)
 }
 
+/-- The pointer that the physical address `addr` denotes. -/
+def MemoryState.decode (_mem : MemoryState) (addr : UInt64) : Pointer := ⟨0, addr⟩
```
math-fehr, line 31: This should return an `Option Pointer` right? When we move to the object model, this decoding should not always work AFAIK.

math-fehr (approved):
Looks good to me besides one API type

Veir/Interpreter/Memory.lean
tobiasgrosser: Let's discuss this in the PR where the API should ge extended. I added a comment here: https://github.com/opencompl/veir/pull/1512/changes#r4078872079

PR #1511, opened 2026-09-20
refactor(interpreter): route every access through one memory object

tobiasgrosser:
Memory is now an array of objects. The flat model has a single object, 0, whose offsets are the addresses themselves, so nothing observable changes. Every access looks its object up with `getObject?`, `checkAccess` returns the object it checked, and refinement is stated per object.


Veir/Interpreter/Memory.lean
```diff
@@ -49,21 +66,28 @@ def memorySize (n : Nat) : Interp UInt64 :=
...
 /--
   Check if an access of `size` bytes at `p` is allowed.
 -/
-def MemoryState.checkAccess (mem : MemoryState) (p : Pointer) (size : UInt64) : Interp Unit := do
+def MemoryState.checkAccess (mem : MemoryState) (p : Pointer) (size : UInt64) : Interp MemoryObject := do
+  let some obj := mem.getObject? p | Interp.ub
+
   -- An access of zero bytes is allowed anywhere.
-  if size = 0 then return ()
+  if size = 0 then return obj
```
naveen-seth, line 94: Should an access of zero bytes be allowed even when `getObject?` fails because `p` has an invalid index into the MemoryObjects array?

naveen-seth (approved):
Just one small question; I didn't understand this part fully.
Otherwise LGTM!

Veir/Interpreter/Memory.lean
tobiasgrosser: Good catch. This is only true if the object actually exists. I updated the comment accordingly.

I also documented that the current interpreter creates only objects that are known to be in the `MemoryState`.

Veir/Interpreter/Refinement/Basic.lean
```diff
@@ -59,13 +59,20 @@ def RuntimeValue.arrayIsRefinedBy (source target : Array RuntimeValue) : Prop :=
...
-def MemoryState.isRefinedBy (source target : MemoryState) : Prop :=
+def MemoryObject.isRefinedBy (source target : MemoryObject) : Prop :=
   ∀ addr, source.poisonMask.getD addr 0 ||| ((source.contents.getD addr 0 ^^^ ~~~target.contents.getD addr 0) &&& ~~~target.poisonMask.getD addr 0) = 0xff
 
+@[inherit_doc] infix:50 " ⊒ " => MemoryObject.isRefinedBy
+
+/-- Refinement of memory states: the same objects, each refined bytewise. -/
+@[expose]
+def MemoryState.isRefinedBy (source target : MemoryState) : Prop :=
+  source.objects.size = target.objects.size ∧ ∀ i : Nat, source.objects[i]! ⊒ target.objects[i]!
```
math-fehr: I believe this is too strong as a refinement relation.
I think the refinement should be a mapping between objects, rather than requiring one-to-one comparison between objects. This is how Alive2 does it as well.

Veir/Interpreter/Refinement/Lemmas.lean
```diff
@@ -8,6 +8,8 @@ public section
 
 namespace Veir
 open Veir.Data
+open Veir.Data (Pointer)
```
math-fehr: Is this needed if we have `open Veir.Data`? right before?

Veir/Interpreter/Memory.lean
```diff
@@ -12,20 +12,45 @@ open Veir.Data.LLVM (Ptr)
...
-  poisonMask := (ByteArray.emptyWithCapacity 1024).extend 8 0xff,
-  consistentSize := (by grind)
-}
+/-- An object of `size` bytes, all of them poison. -/
+def MemoryObject.ofSize (size : Nat) : MemoryObject :=
+  ⟨ByteArray.replicate size 0, ByteArray.replicate size 0xff, by grind⟩
+
+instance : Inhabited MemoryObject := ⟨MemoryObject.ofSize 0⟩
+
+def MemoryObject.size (obj : MemoryObject) : Nat := obj.contents.size
```
math-fehr: Can you write in the documentation of this function that the size is expressed in bytes?

Veir/Interpreter/Memory.lean
```diff
@@ -12,20 +12,45 @@ open Veir.Data.LLVM (Ptr)
...
+
+/--
+  Memory state during interpretation.
+-/
+@[ext]
+structure MemoryState where
+  objects : Array MemoryObject
```
math-fehr: Should this be a HashMap instead of an Array? It doesn't seem like we can delete a memory object here, and this is definitely something we will want to enable in the future.

Veir/Interpreter/Memory.lean
```diff
@@ -12,20 +12,45 @@ open Veir.Data.LLVM (Ptr)
...
+
+def MemoryState.empty : MemoryState := { objects := #[MemoryObject.ofSize 8] }
```
math-fehr: Why does an empty memory contains 8 bytes of memory allocated?

Veir/Interpreter/Memory.lean
```diff
@@ -12,20 +12,45 @@ open Veir.Data.LLVM (Ptr)
...
+
+/--
+  The object that `p` points into, or `none` if `p` indexes no object.
+
+  No pointer should trigger the `none` case: every pointer the interpreter builds
+  names an object contained in the state (currently only object 0) which `empty`
+  creates and no operation ever removes. However, this is not enforced, so a
+  bug may break this invariant.
```
math-fehr: What happens if we deallocate a memory object, and we try to access it with an old pointer? I think this should trigger a `none` here? So I wouldn't say that no pointer should trigger the `none` case, but rather that this is only happening if the pointer we are using is referring to old memory.

Veir/Interpreter/Memory.lean
```diff
@@ -12,20 +12,45 @@ open Veir.Data.LLVM (Ptr)
...
+-/
+def MemoryState.getObject? (mem : MemoryState) (p : Pointer) : Option MemoryObject :=
+  mem.objects[p.object]?
```
math-fehr: I have the feeling that API-wise, we should only pass the object ID here.
Otherwise, we would have to create a pointer whenever we want to access any memory object, despite not needed the offset.

Veir/Interpreter/Memory.lean
```diff
@@ -49,21 +74,28 @@ def memorySize (n : Nat) : Interp UInt64 :=
   excecute this program.
 -/
 def MemoryState.alloc (mem : MemoryState) (size : UInt64) : Interp (MemoryState × Pointer) :=
-  if mem.contents.size + size.toNat ≥ 2 ^ 64 then Interp.fail else
-  return (⟨mem.contents.extend size.toNat 0,
-    mem.poisonMask.extend size.toNat 0xff,
-    by simp [mem.consistentSize]⟩, ⟨0, mem.contents.size.toUInt64⟩)
+  match mem.getObject? ⟨0, 0⟩ with
```
math-fehr: This is for example why I think we should not pass a pointer to `getObject?` but only the object ID

PR #1512, opened 2026-09-20
feat(interpreter): give memory blocks a physical address space

tobiasgrosser:
Memory was represented as a single flat object, such that a pointer could walk from one allocation into its neighbour, and nothing distinguished allocations. Every allocation is now its own object, and LLVM and RISC-V accesses alike are checked against the object that their pointer belongs to, so running past an allocation or dereferencing null is undefined behaviour.

Every object has a base address, assigned by a bump allocator that aligns it to 16 bytes, leaves a guard byte between objects, and starts past the low 64 KiB so that a small integer never denotes an object. A pointer converts to an integer as base plus offset, and an integer converts back to the last object whose base is at most it, so casts and pointers stored in memory keep their meaning.  In the future, we may replace the bump allocator with a symbolic interface that generalizes over all permitted allocators.

Memory refinement requires the same number of objects, each at the same address and refined bytewise.


Test/Interpreter/RISCV_Stack/store_null.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: veir-interpret %s | filecheck %s
+
+// The null object at address 0 holds no bytes, so a machine store through a
+// zero register is UB, as a null dereference is in LLVM code.
```
tobiasgrosser: A store to `0` is UB.
```suggestion
// A store to 0 is UB.
```

UnitTest/MemoryLayout.lean
```diff
@@ -0,0 +1,17 @@
+import Veir.Interpreter.Memory
+
+/-!
+  Objects are laid out past every other object: the first lands right past
+  the arena, and the next follows it with a guard byte, aligned.
```
nunoplopes, line 5: maybe leave a not this is a simplification. Addresses need to be non-deterministic.

Veir/Interpreter/Memory.lean
```diff
@@ -36,19 +38,53 @@ def MemoryObject.size (obj : MemoryObject) : Nat := obj.contents.size
...
+def MemoryState.objectOfAddress (mem : MemoryState) (addr : UInt64) : Nat :=
+  let baseOf (i : Nat) : UInt64 := (mem.objects[i]?.map (·.base)).getD 0
+  (List.range mem.objects.size).foldl (init := 0) fun best i =>
+    if baseOf i ≤ addr ∧ baseOf best ≤ baseOf i then i else best
+
 /-- The pointer that the physical address `addr` denotes. -/
-def MemoryState.decode (_mem : MemoryState) (addr : UInt64) : Pointer := ⟨0, addr⟩
+def MemoryState.decode (mem : MemoryState) (addr : UInt64) : Pointer :=
+  let i := mem.objectOfAddress addr
+  ⟨i, addr - (mem.objects[i]?.map (·.base)).getD 0⟩
```
tobiasgrosser, line 83: @math-fehr asked:

> This should return an Option Pointer right? When we move to the object model, this decoding should not always work AFAIK.

UnitTest/MemoryLayout.lean
tobiasgrosser: Addressed. I added a note.

Veir/Interpreter/Memory.lean
tobiasgrosser: I added the following comment:

> Every address is mapped to a pointer object, as we just look for the closest object from below that a
  given address may belong to. If no other object matches, addresses are mapped
  to the null object. Yet, addresses outside of an object's boundaries may be
  out-of-bounds and will be rejected in `checkAccess`.

Veir/Interpreter/Memory.lean
```diff
@@ -36,7 +38,32 @@ def MemoryObject.size (obj : MemoryObject) : Nat := obj.contents.size
...
+
+/--
+  The address of the next object: past the end of every object with a guard
+  byte between them and past the arena, rounded up to `objectAlignment`.
+
+  TODO: This is a simplification. Eventually, addresses should be
+  non-deterministic.
+-/
+def MemoryState.nextBase (mem : MemoryState) : UInt64 :=
+  let past := mem.objects.foldl (init := arenaSize) fun past obj =>
```
Chobbes: I am likely misunderstanding something, but is there a need to fold here? Isn't it the case that the final memory object in the array always has the largest base in the current implementation?

Veir/Interpreter/Memory.lean
```diff
@@ -52,11 +79,29 @@ def MemoryState.getObject? (mem : MemoryState) (p : Pointer) : Option MemoryObje
...
 
-/-- The pointer that the physical address `addr` denotes. -/
-def MemoryState.decode (_mem : MemoryState) (addr : UInt64) : Pointer := ⟨0, addr⟩
+/--
+  The index of the last object whose base is at most `addr`. Object 0 starts
+  at address 0, so some object always qualifies.
+-/
+def MemoryState.objectOfAddress (mem : MemoryState) (addr : UInt64) : Nat :=
+  let baseOf (i : Nat) : UInt64 := (mem.objects[i]?.map (·.base)).getD 0
+  (List.range mem.objects.size).foldl (init := 0) fun best i =>
```
Chobbes: I'm not sure what Lean support for maps is like, but I think we should consider using maps instead of arrays for this. I think this code will become quite a bit simpler. The asymptotics should be better too---not sure how Lean implements arrays, though, if they're contiguous and cache friendly it might be worth benchmarking to see if this is faster (although, in that case I might still recommend having a map-style interface backed by arrays?)

Veir/Interpreter/Memory.lean
Chobbes: HMMMM. This is interesting and I'm not sure what all of the ramifications of this are. Is there a reason for this approach, or is it out of convenience to avoid options?

Veir/Interpreter/Memory.lean
```diff
@@ -74,15 +119,10 @@ def memorySize (n : Nat) : Interp UInt64 :=
...
-  | none => Interp.fail
-  | some obj =>
-    if obj.size + size.toNat ≥ 2 ^ 64 then Interp.fail else
-    return (mem.setObject ⟨0, 0⟩ { obj with
-        contents := obj.contents.extend size.toNat 0,
-        poisonMask := obj.poisonMask.extend size.toNat 0xff,
-        consistentSize := by simp [obj.consistentSize] },
-      ⟨0, obj.size.toUInt64⟩)
+  let base := mem.nextBase
+  if base.toNat + size.toNat ≥ 2 ^ 64 then Interp.fail else
```
Chobbes: What's the expected refinement for Interp.fail? I do worry about restricting to finite memory, as well.

UnitTest/MemoryLayout.lean
Chobbes: I'd love to discuss what the plan is for making addresses non-deterministic. CTrees?

Chobbes (commented):
Added some feedback and questions.

Veir/Interpreter/Memory.lean
```diff
@@ -74,15 +119,10 @@ def memorySize (n : Nat) : Interp UInt64 :=
...
-  match mem.getObject? ⟨0, 0⟩ with
-  | none => Interp.fail
-  | some obj =>
-    if obj.size + size.toNat ≥ 2 ^ 64 then Interp.fail else
-    return (mem.setObject ⟨0, 0⟩ { obj with
-        contents := obj.contents.extend size.toNat 0,
-        poisonMask := obj.poisonMask.extend size.toNat 0xff,
-        consistentSize := by simp [obj.consistentSize] },
-      ⟨0, obj.size.toUInt64⟩)
+  let base := mem.nextBase
```
Chobbes: One thing that's sort of bothering me about this representation for memory is that the `nextBase` must take the size / offsets of previously allocated objects into account. The `nextBase` function does do this, so in principle it's fine, but it seems like it would be really easy to make a mistake and produce bad memory states, e.g., by having overlapping objects in memory. I wonder if this representation would be more difficult when dealing with specifications and non-deterministic allocations.

Chobbes (commented):
The current representation involves maintaining some specific constraints carefully, and I wonder if that will cause problems.

PR #1513, opened 2026-09-20
refactor(isel): give the RISC-V branch lowering a pure core

tobiasgrosser:
The pass ran in `IO`, walked use chains with `while` loops, and returned a half-rewritten context when an operation could not be created. Its core is now a function in `Except String` built from folds over lists, which fails instead, and the `IO` pass only unwraps the result. Branches are converted in one fold over the operations and block arguments in a second fold over the blocks, which visits the same branches as following the use chain of every block did. The generated code is unchanged.


naveen-seth (approved):
LGTM, no issues found!

PR #1515, opened 2026-09-21
feat(interpreter): make its partial fixpoint fail on non-termination

tobiasgrosser:
`interpretBlockCFG` uses partial fixpoint. Without an order on `Interp`, Lean uses for it a flat order whose least element is picked by choice. Runs that loop forever have then an where noting nothing can be proven about. Hence, no refinement between two contexts could be shown for a program with a loop. `Interp` now carries the flat order with `fail` at the bottom, so forever-looping runs fail and do not prevent proofs.


naveen-seth (commented):
I think that UB would make sense for this behavior.
Can we add a small unit test?

I tried the patch locally. A MonoBind instance might make sense to make this work under do-notation in the interpreter, but maybe this could also be added when needed.
Besides that, LGTM!

PR #1517, opened 2026-09-21
feat(verifier): the entry block of a verified region is not branched to

tobiasgrosser:
`verify` rejects a region whose entry block has a predecessor, but no lemma gave that fact to a proof that assumes `Verified`. This adds `Verified.entryBlock_firstUse_eq_none`. It rests on a small lemma that a loop over the keys of a hash map in `Except` that succeeds ran its body successfully on every key, lifted to `forOpsDepM` and `forBlocksDepM`. No behaviour changes.


Veir/Verifier.lean
```diff
@@ -383,6 +383,40 @@ theorem WfIRContext.Verified.graph_region_firstBlock_eq_lastBlock
   have hregionCheck := (List.all_eq_true.mp hcheck) region hregionKeys
   grind [WfIRContext.graphRegionsHaveAtMostOneBlock]
 
+/-- A verified context passes the checks that `verify` runs on every block. -/
+private theorem WfIRContext.Verified.blockChecks
+    {ctx : WfIRContext OpCode} {root : OperationPtr} (ctxVerified : ctx.Verified root)
+    {block : BlockPtr} (blockIn : block.InBounds ctx.raw) :
+    (do block.verifyTerminator ctx blockIn
+        block.verifyNoEntryBlockPredecessors ctx blockIn : Except String PUnit.{1}) = .ok ⟨⟩ := by
```
math-fehr, line 391: @ineol knows this more, but I believe we should not have `do` notation inside theorems in general, and use the `bind` directly?

Veir/Verifier.lean
```diff
@@ -383,6 +383,40 @@ theorem WfIRContext.Verified.graph_region_firstBlock_eq_lastBlock
...
+  unfold WfIRContext.Verified WfIRContext.verify at ctxVerified
```
math-fehr, line 392: Should we use `unfold` instead of `simp`? @ineol what's the usual style for this (I recalled using `simp` was preferred instead of `unfold`, but might be misremembering)

Veir/Verifier.lean
```diff
@@ -383,6 +383,40 @@ theorem WfIRContext.Verified.graph_region_firstBlock_eq_lastBlock
...
+/-- The entry block of a region in a verified context is not branched to. -/
+theorem WfIRContext.Verified.entryBlock_firstUse_eq_none
+    {ctx : WfIRContext OpCode} {root : OperationPtr} (ctxVerified : ctx.Verified root)
+    {block : BlockPtr} (blockIn : block.InBounds ctx.raw) {region : RegionPtr}
+    (hParent : (block.get! ctx.raw).parent = some region)
+    (hFirstBlock : (region.get! ctx.raw).firstBlock = some block) :
+    (block.get! ctx.raw).firstUse = none := by
+  obtain ⟨_, -, hCheck⟩ := Except.bind_eq_ok.mp (ctxVerified.blockChecks blockIn)
+  have hParent' : (block.get ctx.raw blockIn).parent = some region := by grind
+  unfold BlockPtr.verifyNoEntryBlockPredecessors at hCheck
```
math-fehr, line 411: Same here with unfold

math-fehr (approved):
Just style comments, but otherwise all good to me!

Veir/Verifier.lean
tobiasgrosser: You recalled right: the two lemmas just above this one already use `simp only [WfIRContext.Verified, WfIRContext.verify]`, so these were the odd ones out. Switched in b78576a7c, which also drops the `dsimp only` that was only there to clean up after `unfold`.

Veir/Verifier.lean
tobiasgrosser: Done in the same commit. Here the definition folds into the rewrite that follows it, so the two `simp only` calls on this hypothesis become one.

Veir/Verifier.lean
tobiasgrosser: Agreed, and it turned out to be hiding something. In d20c58fed the helper yields the two checks as a conjunction instead, and each one gets a theorem of its own, so no statement mentions `do` or a bind. Writing the right-hand side as `.ok ()` rather than `.ok ⟨⟩` also pins the universe, which is what the `PUnit.{1}` ascription was there for. Both new theorems have to stay private, since `verifyTerminator` and `verifyNoEntryBlockPredecessors` are defined above the `public section` and so cannot appear in an exported statement.

tobiasgrosser:
I addressed the style comments.

PR #1519, opened 2026-09-21
feat(nix): Provide packages as executables

### LINKED ISSUE #1514: Provided Nix packages use local source repository
Currently, running a command such as `nix run 'github:opencompl/veir/<pinned-commit>#veir-interpret'` does not work as intended.
Because the package definition just calls out to `lake exec`, this only succeeds when in a local VeIR checkout. Then, it builds the package from the local checkout instead of `<pinned-commit>`

When not in a VeIR checkout:
```
error: [root]: no configuration file with a supported extension:
<cool-project-which-consumes-veir>/lakefile.lean
<cool-project-which-consumes-veir>/lakefile.toml
```

EDIT: After reading the README a bit closer, I saw this was intended 😅. MB!

naveen-seth:
Fixes #1514.

Currently, the flake-provided executables are convenience wrappers around `lake exec` or `make`. This makes it impossible to consume them in other projects / use them from outside a local VeIR checkout.

This changes the flake to provide executables which can be consumed in other projects such as our benchmark/progress-tracking repositories. This also improves reproducibility a tiny bit, since the executables are now built in the isolated Nix environment.

This flake-provided apps which wrapped Makefile targets have been removed and this use-case now requires an invocation like  `nix develop -c make tests`.


axelcool1234:
As mentioned before, this would mean relying on Lean4Nix which isn't always up to date. Definitely keep this PR around though in case anyone wants to take it upon themselves with helping out Lean4Nix. The non Nix way of doing things in the flake will have to do for now unfortunately 😔

tobiasgrosser:
What is the conclusion? Do we merge or close the PR. Let's not keep PRs lingering. If this is not something we can merge in code, we can think of documenting choices somewhere else, if needed.

naveen-seth:
We can close this! The conclusion was that the lake4-nix flake used here often breaks with the newest lean release and that this would only be possible if someone helps fix the lake4-nix flake regularly.
I'll try myself at fixing it for the next release; maybe we could revisit this in the future.

PR #1520, opened 2026-09-21
fix(PBV): solve broken sign extension case

luigirinaldi:
Previously, `signExtend`ing by the same width as the `blastWidth` would cause a failure, since the `Push.setWidth_setWidth` rule would stop the `Push.signExtend_...` rule from being applied. 

In this patch, the `setWidth` simplification rules are given less priority than other theorems, and the `signExtend` push rule is modified to have the `blastWidth` be explicit, fixing the failing case.


tobiasgrosser:
@naveen-seth, can you do a first review?

naveen-seth (approved):
Nice! LGTM.

PR #1522, opened 2026-09-22
feat(dataflow): Constant domain now holds RuntimeValue

axelcool1234:


regehr (approved):
looks great-- @math-fehr should weigh in as well!

